PR #25, opened 2011-11-12
Wait for STDOUT to drain before exiting running process.

jhamlet:
When writing copious amounts of help information, exiting immediately truncates it. Wait for the STDOUT stream's drain event before exiting, so everything has time to render.


SomeKittens:
Closing, will reopen if this gets updated.

PR #121, opened 2013-01-26
No error message when an unknown option and two args are given

nitoyon:
When an option `-u` is not defined:

  * `node foo.js -u` -> unknown option error (expected)
  * `node foo.js -u arg1` -> unknown option error (expected)
  * `node foo.js -u arg1 arg2` -> NO ERROR (unexpected)

Another example: `mocha --reporterrrr spec test` (should be `mocha --reporter spec test`) shows no error.

This bug is caused by `Command.parseArgs()` assuming `arg2` to be a command name.

### DIFF
--- a/index.js
+++ b/index.js
@@ -461,21 +461,26 @@ Command.prototype.normalize = function(args){
  */
 
 Command.prototype.parseArgs = function(args, unknown){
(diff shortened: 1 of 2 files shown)


SomeKittens:

PR #122, opened 2013-02-01
Adds an option's default value to the help output

buerkle:
If a default value is supplied to the option, this patch appends a '(Default: value)' message to the end of the help for the option. For example,
option('-f', 'Number of foos', 100)

becomes

-f   Number of foos (Default: 100)

### DIFF
--- a/index.js
+++ b/index.js
@@ -384,7 +384,10 @@ Command.prototype.option = function(flags, description, fn, defaultValue) {
     // when --no-* we make sure default is true
     if (false == option.bool) defaultValue = true;
     // preassign only if we have a default
-    if (undefined !== defaultValue) self[name] = defaultValue;
+    if (undefined !== defaultValue) {
+      self[name] = defaultValue;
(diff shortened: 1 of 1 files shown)


ohjames:
This would be really nice, any reason the issue is open for 2 years without any notice?

abetomo:
@buerkle Could you please resolve the conflict?

abetomo:
@vanesyan I think that it is a useful fix, but what do you think?

vanesyan:
Yes, I think it's good idea to include default value in help message. But maybe we should write `default` all characters in low-case, IMO, otherwise it looks good to me. Need to resolve conflict and lets merge it.

abetomo:
@vanesyan Thank you.

@buerkle Could you please fix it?
If I do not receive a response for a while, I will try to fix it by myself.

buerkle:
Fixed conflicts and change Default to default

abetomo:
@buerkle Thank you.

PR #134, opened 2013-03-06
fixes #61 Adds check is a number check on param

### LINKED ISSUE #61: Accept negative number values
I noticed that negative values are seen as possible options because of the minus sign. What I'm trying to do:

`node something.js --value -10`

Please note that `--value` is defined as `--value <n>` so it should accept it, I think.

davidcl64:
Adds/Fixes support for negative numbers on the cmd line

Note: while this does fix the issue, a better approach may be as @dresende commented in issue #61 - if the parameter is provided and expects a value, automatically slurp up the next arg in the chain.

### DIFF
--- a/index.js
+++ b/index.js
@@ -434,7 +434,7 @@ Command.prototype.normalize = function(args){
 
   for (var i = 0, len = args.length; i < len; ++i) {
     arg = args[i];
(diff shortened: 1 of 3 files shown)


SomeKittens:

PR #144, opened 2013-04-26
Update help

jimthedev:
Updated example to use rawArgs length.


zhiyelee:
I also prefer `program.args.length`, but thx all the same.

PR #167, opened 2013-09-12
Fix for passing ssh private key to argument

lancejpollard:
I have a script that gets run like this:

```
node upload-key < path/to/key
```

I then internally convert that blob to an argument for commander, which ends up being like this:

``` js
// compiled process.argv
[ 'node', 'upload-key', '--key', '-----BEGIN RSA PRIVATE KEY-----\nMx1....' ]
```

It seems that just adding a check to see if there are more than two `--` dashes will do the trick (it works for me locally). Thoughts?

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,4 +1,3 @@
-
 /**
  * Module dependencies.
  */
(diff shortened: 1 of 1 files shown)


abetomo:
@lancejpollard Would you please resolve the conflict?

abetomo:
An example of a case where the value begins with a hyphen.
We plan to make modifications corresponding to all cases starting with hyphens.

#61

PR #172, opened 2013-09-29
Clean up help screen

srquinn21:
Quick little fix to get the option and command descriptions starting on the same column. Satisfies my anal retentiveness.


killswitch:
:+1: for anal retentiveness.

abetomo:
@srquinn21 
Sorry for the delay in confirmation.
Since the difference from the current code got bigger, I close it once.
Even now, if you need it, we will sorry for the inconvenience but please do PR once again.

PR #187, opened 2013-11-22
fix up errant output & escape codes in make test

dbushong:
* many versions of echo don't support "\033", but printf(1) does
* silence the help output in test.options.commands.js (dangerous?)

### DIFF
--- a/test/run
+++ b/test/run
@@ -5,10 +5,10 @@ export NODE_ENV=test
 echo
 for file in $@; do
   printf "\033[90m   ${file#test/}\033[0m "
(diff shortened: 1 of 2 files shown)


zhiyelee:
Close for no response, pls feel free to reopen this if u still take interest in this PR

dbushong:
Sorry, didn't see messages.  If you're still interested, here's a thread about how this won't work, e.g., on MacOS: http://superuser.com/questions/33914/why-doesnt-echo-support-e-escape-when-using-the-e-argument-in-macosx

PR #202, opened 2014-04-03
Fix for issue #37 : --help should not display first.

nathanpeck:
Appending the --help option instead of prepending it.


subchen:
Yes, it is good idea.

SomeKittens:
Closing, as this has sat here for a while and has conflicts.  Will reopen if conflicts are fixed.

zhiyelee:
Also I think it's a break change, isn't it?

Issue #220, opened 2014-05-14
npm: cannot find module with git-style subcommand


iamvdo:
I created a CLI tool based on Commander. I used git-style subcommand. Some people have problem when installing it globally: not all binary aren't found.

Did I missed something? Does all the binary have to appear in package.json under the `bin` key? For now, I only have the main one.

Thank you. :)

PR #224, opened 2014-06-04
Add fallback to local git-style subcommand

mauvm:
I ran into the same problem as @iamvdo (#220) and figured a local fallback would suffice.

It simply checks if the git-style subcommand binary exists (global), and if not, tracks the symlink of the original file and tries the binary there (local fallback).


mauvm:
@muloka Would you mind elaborating on that?

PR #239, opened 2014-08-08
Added option to by pass process exit on unknown options.

diogogmt:
Commander should not always exit the process in case of an unknown option.
I'll give an example:
I have a grunt app that uses commander to parse some arguments since I like it better then the default grunt arg parsing, however, some arguments are targeted to grunt directly not my app and since commander doesn't allow unknown options to be by passed it prevents me to setting arguments targeted at grunt.

This pull requests adds the option to not exit the process in the case of encountering an unknown argument.
Tests are included.

### DIFF
--- a/index.js
+++ b/index.js
@@ -359,7 +359,9 @@ Command.prototype.option = function(flags, description, fn, defaultValue){
  * @api public
  */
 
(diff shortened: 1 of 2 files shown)


zhiyelee:
Close #310

Issue #226, opened 2014-06-09
Nested commands?


wayneashleyberry:
Would it be possible to, or is it currently possible, to have nested commands?

```js
program
  .command('user')
    .command('add [name]')
      .description('Adds a new user')
      .action(function(name) {
        //
      });
```

The reason I'm asking is because it seems commands with spaces in them don't work :/

ie.

```js
program.command('add user [name]');
```

wayneashleyberry:
Closed due to inactivity.

stewones:
+1

Issue #63, opened 2012-06-06
Support more than one level of subcommands


balupton:
### Right now we can do this:

```
program = require('commander')

program
    .version('0.0.1')

program
    .command('plugin-create <name>')
    .description('create plugin')

program
    .command('plugin-delete <name>')
    .description('delete plugin')
```

which results with:

```

  Usage: woodchuck.js [options] [command]

  Commands:

    plugin-create <name>
    create plugin

    plugin-delete <name>
    delete plugin

  Options:

    -h, --help     output usage information
    -V, --version  output the version number
```
### But we can't do this:

```
program = require('commander')

program
    .version('0.0.1')

program
    .command('plugin create <name>')
    .description('create plugin')

program
    .command('plugin delete <name>')
    .description('delete plugin')
```

which results with:

```
  Usage: woodchuck.js [options] [command]

  Commands:

    plugin <name>
    create plugin

    plugin <name>
    delete plugin

  Options:

    -h, --help     output usage information
    -V, --version  output the version number
```

PR #245, opened 2014-08-27
Completed nested command support.

nowells:
Nested command support was almost completely available, however, a few things
seemed to be missing to me.
- `this` context on `action` event listeners was not bound properly.
- We output `help` too early, the deepest referenced command should handle help.
- We did not have `action` handling logic to delegate to subcommands if they existed.

This should address #226, and #1. Pretty much all of the support was already there, and it seemed useful to just connect the wires. I know that #63 referenced using the git subcommand model, which I love and use frequently, but there are several use cases (creating a command framework that loads external "plugin" module commands into the parent executable) that would be more easily expressed with this simple fix to connect the wires for sub-commands.

### DIFF
--- a/index.js
+++ b/index.js
@@ -215,37 +215,42 @@ Command.prototype.action = function(fn) {
     args = args || [];
     unknown = unknown || [];
 
-    var parsed = self.parseOptions(unknown);
+    var parsed = self.parseOptions(args.concat(unknown));
 
-    // Output help if necessary
-    outputHelpIfNecessary(self, parsed.unknown);
-
-    // If there are still any unknown options, then we simply
-    // die, unless someone asked for help, in which case we give it
-    // to them, and then we die.
-    if (parsed.unknown.length > 0) {
-      self.unknownOption(parsed.unknown[0]);
+    if (parsed.args.length && self.listeners(parsed.args[0]).length) {
+        for (var i = 0; i < parsed.args.length; i++) {
+            args.shift();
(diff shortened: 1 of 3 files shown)


ohjames:
My first day with node-commander and I've already hit issues solved by this PR. Would be so good if someone could look at this but there seems to be many neglected PRs :(

SomeKittens:
@ohjames We can't merge this - there's a bunch of conflicts...

jakiestfu:
@ohjames bump

ohjames:
Don't have the time to help out with this, found another library that supports it.

jakiestfu:
@ohjames Which library?

ohjames:
@jakiestfu it's been added to [yargs](https://github.com/bcoe/yargs) with the `next` flag and will be officially released within a few days.

apla:
@jakiestfu also just released https://github.com/apla/commop

ohjames:
@apla thanks for the update, I'm your first star! Hoping you've finally made the holy grail of node.js cmd line argument parsers.

boutell:
Did this feature (nested commands) ever happen?

ohjames:
@boutell nope I've been using yargs for the past few years instead. It's pretty good.

nowells:
I also switched to `yargs` a while ago (when this support was not added) and have been very happy with it. I will close this as it will likely no longer be a valid fix.

Issue #206, opened 2014-04-10
 if (this._execs[name])  when name is a property of Array tries to run subcommand --with fix


startswithaj:
https://github.com/visionmedia/commander.js/blob/cb446166aad48033316f6b975b0bf1975be94696/index.js#L374
```
 if (this._execs[name]) return this.executeSubCommand(argv, args, parsed.unknown);
```

This code looks for a command in the ```_execs``` array and if found runs this.executeSubCommand.

If ```name``` is equal to a property of array such as 'join' it returns true.

```if this._execs.indexOf(name) > -1```

PR #249, opened 2014-09-02
Fix #248

### LINKED ISSUE #248: Using 'filter' (or other native array methods) as a command causing error
Came across an error today when using `filter` as a command input.

Simple reproduction:

```
com.command('seed <tpl>')
    .action(function(tpl){});
```

Just running the command `<script> seed filter` will cause a module not found error, no special coding required.

```
Error: Cannot find module 'c:\Users\...\bin\cmd-filter'
    at Function.Module._resolveFilename (module.js:338:15)
    at Function.Module._load (module.js:280:25)
    at Function.Module.runMain (module.js:497:10)
    at startup (node.js:119:16)
    at node.js:906:3
```

It is caused by line 380 of commander's `index.js` - Even if the `this._execs` array is empty, calling `this._execs['filter']` will return the native function, thus causing the argument to be true and creating this error.

Best solution in my opinion is double checking the value to make sure it is not a function. If you would like, I can throw in a PR.

Thank you for your time.

jonathanrdelgado:
Fixes #248

Thanks again!

### DIFF
--- a/index.js
+++ b/index.js
@@ -377,7 +377,7 @@ Command.prototype.parse = function(argv){
 
   // executable sub-commands
   var name = result.args[0];
-  if (typeof this._execs[name] != "function") {
+  if (this._execs[name] && typeof this._execs[name] != "function") {
     return this.executeSubCommand(argv, args, parsed.unknown);
   }
 


startswithaj:
resolves #206

PR #260, opened 2014-09-10
fixes exit code from sub-commands not propagating

pirelenito:
if a sub-command was failing, we weren't getting a code 1.


yiminghe:
+1, need this to notify other caller, such as precommit-hook

targos:
+1

For now I use the following workaround :
```js
if (program.runningCommand) {
    program.runningCommand.on('close', process.exit.bind(process));
}
```

zhiyelee:
Close #332

PR #262, opened 2014-09-19
add opts() for getting key-value pairs

tonylukasavage:
A common usage pattern I have when using commander is to process the CLI in a bin file and defer all actual program-specific execution to a separate module. I often pass all options gathered from commander straight my module, like this:

``` js
#!/usr/bin/env node

var program = require('commander'),
  mymodule = require('mymodule');

// setup the program object

mymodule(program);
```

While this works, it would be nice if the object passed through included _only_ the options I specified when setting up the program object. 

This PR add the `opts()` function to the program object (via `Command.prototype`) so as to return the options, with special handling for the version, as a simple object of key-value pairs. No parsing Option objects, no additional properties associated with the program obejcts, I can then do the following in my above example to get the desired behavior.

``` js
#!/usr/bin/env node

var program = require('commander'),
  mymodule = require('mymodule');

// setup the program object

mymodule(program.opts());
```

The included test case also demonstrates this fairly simply. 

**NOTE**: I am entirely open to changing the name of the API. `opts()` was the first thing to occur to me that wasn't already claimed on the `Command` object. If someone has a better alternative I'd be happy to update the code and testing.

### DIFF
--- a/index.js
+++ b/index.js
@@ -602,6 +602,23 @@ Command.prototype.parseOptions = function(argv){
 };
 
 /**
+ * Return an object containing options as key-value pairs
+ *
+ * @return {Object}
+ * @api public
+ */
+Command.prototype.opts = function() {
+  var result = {}
+    , len = this.options.length;
+
+  for (var i = 0 ; i < len; i++) {
+    var key = this.options[i].name();
+    result[key] = key === 'version' ? this._version : this[key];
+  }
+  return result;
+};
+
+/**
  * Argument `name` is missing.
  *
  * @param {String} name
--- a/test/test.options.func.js
+++ b/test/test.options.func.js
@@ -1,0 +1,21 @@
+var program = require('../')
+  , should = require('should');
+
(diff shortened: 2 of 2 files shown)


FokkeZB:
@tonylukasavage was it a deliberate decision for `opts()` not to apply the defaults for options?

```
program
  .version(pkg.version, '-v, --version')
  .description(pkg.description)
  .option('-d, --project-dir [path]', 'the directory containing the project [.]', process.cwd());

console.log(program.projectDir);
console.log(program.opts());
```

The first log has the default, the second one not.

FokkeZB:
Same result (probably copy pasted it wrong):

```
program
  .version(pkg.version, '-v, --version')
  .description(pkg.description)
  .option('-d, --project-dir [path]', 'the directory containing the project [.]', process.cwd());

program.parse(process.argv);

console.log(program.projectDir); 
console.log(program.opts()); 
```

Issue #290, opened 2014-11-18
add test suites for the git-style executive subcommand feature


zhiyelee:
Discussion can be found here  #94  #286

PR #305, opened 2014-12-12
add executable arguments to spawn

DigitalIO:
A slight modification to inject node arguments into sub-commands. Specifically, I made it to allow the --harmony flag to perpetuate to sub-commands.

I'm not quite sure how to write a test for this as no tests exist to test the function as is. See: [#290](https://github.com/tj/commander.js/issues/290)


zhiyelee:
This PR is involved in #387

Issue #306, opened 2014-12-13
Uncaught Error in git-style sub-command


zhiyelee:
run the `examples/pm` get below error:

```
✍ examples/pm search
module.js:333
    throw err;
          ^
Error: Cannot find module '/Users/lizhiye/dev/or/commander.js/examples/pm-search'
    at Function.Module._resolveFilename (module.js:331:15)
    at Function.Module._load (module.js:273:25)
    at Function.Module.runMain (module.js:490:10)
    at startup (node.js:124:16)
    at node.js:807:3
```
The error handler code do not work in https://github.com/tj/commander.js/blob/master/index.js#L459-L465

Issue #196, opened 2014-03-18
the sub-command only supported in the same directory 


eggcaker:
for the examples, the `pm` with sub command pm-install , if it can put the sub commands pm-install in anywhere in the $PATH, then the pm still can find  the sub command. and run it  that would be better to write sub command like git way.

PR #307, opened 2014-12-14
solve #306, and bin in $PATH make sense now

zhiyelee:
solve #306, and bin in $PATH make sense now  #196
Almost revert #173 which lead to the bug in #306

Also remove deprecated `customFds` option in spawn

### DIFF
--- a/index.js
+++ b/index.js
@@ -463,7 +463,12 @@ Command.prototype.executeSubCommand = function(argv, args, unknown) {
 
   // prefer local `./<bin>` to bin in the $PATH
   var local = path.join(dir, bin);
-  if (fs.existsSync(local)) bin = local;
+  try {
+    // for versions before node v0.8 when there weren't `fs.existsSync`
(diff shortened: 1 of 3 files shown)


SomeKittens:
Looks like this fixes more than one edge case.  Can we get the tests passing?

index.js
```diff
@@ -460,13 +461,26 @@ Command.prototype.executeSubCommand = function(argv, args, unknown) {
...
+    }
+  } catch (e) {}
 
   // run it
   args = args.slice(1);
-  args.unshift(local);
-  var proc = spawn('node', args, { stdio: 'inherit', customFds: [0, 1, 2] });
+
+  var proc;
+  if (process.platform !== 'win32') {
```
zhiyelee: After the commit, `bin` file must have shebang itself

zhiyelee:
Closed for too many commits. I'll  open another PR

PR #327, opened 2015-01-10
Fix #306, and bin-file in $PATH make sense

### LINKED ISSUE #306: Uncaught Error in git-style sub-command

zhiyelee:
Almost revert #173 which lead to the bug in #306

Also remove deprecated customFds option in spawn

### DIFF
--- a/index.js
+++ b/index.js
@@ -8,6 +8,7 @@ var spawn = require('child_process').spawn;
 var path = require('path');
 var dirname = path.dirname;
 var basename = path.basename;
+var fs = require('fs');
 
 /**
  * Expose the root command.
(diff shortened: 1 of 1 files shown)


zhiyelee:
@DigitalIO This PR is incompatible with #305 .

Indeed we can put the  `--harmony` in the shebang

file: `~/bin/pm-list`
```js 
#!/usr/bin/env node --harmony
"use strict";

let a = 5
console.log(5)
```

which works well with current PR:

```bash
✍ examples/pm list
5
```

DigitalIO:
Adding ```--harmony``` to the shebang doesn't work in all environments, unfortunately. [Some discussion.](http://stackoverflow.com/questions/3306518/cannot-pass-an-argument-to-python-with-usr-bin-env-python) I've tried the example script on Debian (Stable) and Windows ("Git Bash") and it fails on both for me.

I have been initiating the scripts with node directly

```> node --harmony examples/pm list```

Not the best solution, but the best compromise I have found thus far.

**Comment by zhiyelee** [comment]
@DigitalIO windows will use this branch: https://github.com/tj/commander.js/pull/327/files#diff-168726dbe96b3ce427e7fedce31bb0bcR481
Your patch can be applied here.

Do u mean Debian don't support?

```js
#!/usr/bin/env node --harmony
"use strict";

let a = 5
console.log(5)
```

DigitalIO:
Correct. env attempt to match "node --harmony" as a function, instead of "node" with the "--harmony" flag. It results in this error: 

```shell
/usr/bin/env: node --harmony: No such file or directory
```

zhiyelee:
@DigitalIO  Seems there is a bug in Debian on the `/usr/bin/env: node`.  https://github.com/joyent/node/issues/3911

PR #336, opened 2015-01-30
Support for RegEx options

palanik:
Piggy backing on coercion to support options with regular expressions.

This allows pre-defined set of values for options.

Taking the Pizza example, the options for size can be limited to /^(large|medium|small)$/i.

HTH.


palanik:
Something messed up. 
Closing this.

I'll send a clean PR.

PR #339, opened 2015-02-03
Add docs badge to README

rrrene:
Hi there,

I want to propose to add this badge to the README to show off inline-documentation: [Inline docs](http://inch-ci.org/github/tj/commander.js)

The badge shows an evaluation by [InchJS](http://trivelop.de/inchjs), a project that tries to raise the visibility of inline-docs in Open Source to encourage aspiring programmers to document their code. So far over 500 Ruby projects **including [Commander](https://github.com/tj/commander)** are sporting these badges to raise awareness for the importance of inline-docs and to show potential contributors that they can expect a certain level of code documentation when they dive into your code.

I would really like to do the same for the **JavaScript** community and although this is "only" a passion project, I really would like to hear your thoughts, critique and suggestions. Your status page is https://inch-ci.org/github/tj/commander.js

What do you think?

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -3,6 +3,7 @@
  [![Build Status](https://api.travis-ci.org/tj/commander.js.svg)](http://travis-ci.org/tj/commander.js)
 [![NPM Version](http://img.shields.io/npm/v/commander.svg?style=flat)](https://www.npmjs.org/package/commander)
 [![NPM Downloads](https://img.shields.io/npm/dm/commander.svg?style=flat)](https://www.npmjs.org/package/commander)
+[![Inline docs](http://inch-ci.org/github/tj/commander.js.svg?branch=master)](http://inch-ci.org/github/tj/commander.js)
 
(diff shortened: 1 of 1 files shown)


rusikf:
Looks nice, but I am not really sure, that this badge is very important for this project.
@tj , what you think?

rrrene:
Thanks for your reply! On the chance that I lose your interest for good, let me elaborate: 

This badge is not really about showing off how good your documentation is or getting to 100%, which is  why there are *no numbers* on it. And it certainly is not about bragging "Look at my perfect docs!" 

The idea is to motivate aspiring Node developers to dive into Open Source projects and read the code. 
It is about *engagement*. And for me, while testing and code coverage are important, inline-docs are the humanly engaging factor in Open Source. This project is about making people less adverse to jumping into the code and see whats happening, because they are not left alone while doing so. I know that, because I put off reading other people's code way too long in my life.

And let me be clear: `commander.js` is seriously good in terms of inline documentation. But even if it wasn't: Most times, any amount of inline-docs is better than none. This badge tries to emphasize that.

Issue #344, opened 2015-02-09
Support explicit boolean flag


Globegitter:
I have a cli where I want to be able to set flags such as `--docker=true` as well as `--docker=false` and have the true/false automatically been interpreted as boolean.

```js
program
  .command('up')
  .option('-D, --docker [boolean]', 'Run Sails server and dependencies within docker containers.', config.docker || false)
```

is the code I have and I also have a config file where you can set `{ "docker": true}`, but then still overwrite that via setting the flag `--docker=false`.

To support this I currently have to check `if (options.docker && options.docker !== 'false')` but think just checking for `if (options.docker)` keeps it simpler especially since I want to support this for multiple flags throughout the whole app.

Is there any way to do this, or a possibility to implement boolean interpretation?

SomeKittens:
You will have to implement the interpretation yourself.

I'm not sure about the internals of your CLI but it usually makes more sense to have the absence of a flag be `false`

shadowspawn:
Commander supports negated options, so rather than:

```sh
launch up
launch up --docker=true
launch up --docker=false
```

the calls could look like:

```sh
launch up
launch up --docker
launch up --no-docker
```

PR #345, opened 2015-02-09
Add support for optional boolean flag

Globegitter:
This seems to fix the use-case described in https://github.com/tj/commander.js/issues/344 and should not really break any existing code, since it is based on the existence of `boolean` in the flags.
Now that I am saying that, checking for `<boolean>` and `[boolean]` might be better.

Also don't know if that ever was the intended use, but it seems quite intuitive to me and would be really useful.

### DIFF
--- a/test/test.options.bool.optional.js
+++ b/test/test.options.bool.optional.js
@@ -1,0 +1,21 @@
+/**
+ * Module dependencies.
+ */
+
+var program = require('../')
+  , should = require('should');
+
+//can be loaded from a config file
+var config = {
+  pepper: true,
+  cheese: false
+}
+
+program
+  .version('0.0.1')
+  .option('-p, --pepper [boolean]', 'add pepper', config.pepper || false)
+  .option('-c, --cheese [boolean]', 'remove cheese', config.cheese || false);
+
+program.parse(['node', 'test']);
+program.pepper.should.be.true;
+program.cheese.should.be.false;
--- a/index.js
+++ b/index.js
@@ -388,7 +388,7 @@ Command.prototype.option = function(flags, description, fn, defaultValue) {
         self[name] = option.bool
           ? defaultValue || true
           : false;
-      } else {
+      } else if (option.flags.indexOf('boolean') === -1) {
         self[name] = val;
       }
     } else if (null !== val) {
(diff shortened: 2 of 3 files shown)


SomeKittens:
This is what I get for answering issues before reading them all.  Can you provide a use case where explicit `true/false` would be needed over lack-of-flag implicitly meaning false?

Globegitter:
The use-case I need it for is, I use https://github.com/twokul/yam so users are able to permanently set cli flags within a `.sane-cli` file.

So that file can contain all the potential flags the cli supports such as `docker: true`, or `skipEmber: true`, etc.

But now I have notices myself that I 80-90% use these defaults, but for some test-cases I just want to be able to quickly change these defaults.

Which, as far as I am aware would only be possible by setting e.g. `sane up --docker=false --skip-ember=false`. But to support that for each flag (quite a few now), I would either have to manually convert the string to a boolean for each flag or check for the strings themselves, which both seems like some unnecessary work, especially because commander already gets them as a boolean.

Let me know if that makes sense and seems like a legitimate use-case.

SomeKittens:
Makes great sense!  Thanks so much.  Add a test and I'll merge.

Globegitter:
Good call on the tests. Added them and fixed the implementation - I will also test again in my application asap.

SomeKittens:
Thanks again!

zhiyelee:
This PR may cause some bugs, `[`/`<`  are reserved for `optional/required arguments`. I do not in favor of this way. Maybe we can discuss this feature more, and I'll revert this commit for the new release `v2.7.0`. @SomeKittens @Globegitter

test/test.options.bool.optional.js
```diff
@@ -0,0 +1,21 @@
...
+  pepper: true,
+  cheese: false
+}
+
+program
+  .version('0.0.1')
+  .option('-p, --pepper [boolean]', 'add pepper', config.pepper || false)
+  .option('-c, --cheese [boolean]', 'remove cheese', config.cheese || false);
+
+program.parse(['node', 'test']);
```
zhiyelee: `program.parse(['node', 'test', '-c']);`
will throw an error here

Globegitter:
@zhiyelee Yeah I am very open to any other way of implementing this. Do you have anything in mind how this could be implemented differently?

SimplGy:
Interested party. Related? https://github.com/tj/commander.js/issues/399

celicooo:
Is this working?? My [boolean] flags are strings...

Issue #346, opened 2015-02-10
error when arg has same name as option


tonylukasavage:
- commander.js 2.6.0

It appears that when you give commander an arg that has the same name as an option, it wrongly sees that arg as an option. This only seems to occur when the arg name that matches an option name is the first arg. Take the following test case:

``` js
var program = require('commander');
program.option('--registry <registry>','desc');
program.parse(process.argv);
console.log('opts: %s\nargs: %s', JSON.stringify(program.opts()), program.args);
```

Now here are some outputs from it, notice that only when the first arg matches the option name does the error in the opts field occur:

```
≫ node com.js registry
opts: {"registry":[]}
args: 

≫ node com.js foo registry
opts: {}
args: foo,registry

≫ node com.js foo registry --registry bar
opts: {"registry":"bar"}
args: foo,registry

≫ node com.js registry --registry bar
opts: {"registry":[]}
args: 
```

The error cases, if not apparent, are when the opts equal `opts: {"registry":[]}`. Because this occurs only with the first argument, I wonder if this has to do somehow with an error in subcommand processing, even though I specified no commands. Just for a bit more detail, this is the full output of the `program` object in the last error case:

``` js
{ commands: [],
  options: 
   [ { flags: '--registry <registry>',
       required: -12,
       optional: 0,
       bool: true,
       long: '--registry',
       description: 'desc' } ],
  _execs: [],
  _allowUnknownOption: false,
  _args: [],
  _name: 'com',
  Command: [Function: Command],
  Option: [Function: Option],
  _events: { registry: [Function] },
  rawArgs: 
   [ 'node',
     '/Users/tlukasavage/development/appc-cli/com.js',
     'registry',
     '--registry',
     'bar' ],
  registry: [],
  args: [] }
```

tonylukasavage:
Upon further inspection, this seems to be an error in the logic in [parseArgs()](https://github.com/tj/commander.js/blob/master/index.js#L560). It strips the first argument off the list of arguments if it has a listener. This I believe is in the case of actions. Unfortunately, the listeners for both actions and options share the same space, and can end up having the same name, as in my case above. I'm not sure, at this moment anyway, what the least invasive way to fix this would be. I'm investigating, but would love to hear if there's a simple way to avoid this.

tomgco:
Can we re-open this as it was reverted because it introduced a bug?

shadowspawn:
[...]

PR #347, opened 2015-02-10
Issue #346, fix collisions when option and first arg have same name

tonylukasavage:
This is detailed in full in #346.

This PR has a test case demonstrates the error in the current master, and a less-than-elegant, but non-invasive way to avoid the error of conflicting action and option listeners without negatively impacting either. I am open to better solutions, but other than establishing a separate emitter for both options and actions, I don't know a better way than a naming convention (`action_` prefix) to keep them from colliding, in my brief inspection anyway.

### DIFF
--- a/index.js
+++ b/index.js
@@ -287,8 +287,9 @@ Command.prototype.action = function(fn) {
 
     fn.apply(self, args);
   };
-  this.parent.on(this._name, listener);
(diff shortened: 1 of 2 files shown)


tonylukasavage:
bump.

SomeKittens:
Works for me.

PR #349, opened 2015-02-11
fix #335

### LINKED ISSUE #335: Subcommands not found when installed globally (npm) on Unix-type environments
I have implemented a command line interface using Commander and the "subcommand" feature.  I have a main called xxxx.js, and a series of commands in files named like xxxx-command.js.  This command line interface is intended to be used via npm (and typically installed globally)

The sub-command launching works fine under Windows, but fails in Unix-type environments.  The reason it fails is when installed globally via npm (using install -g or just npm link), which seems like a reasonable use case for a CLI with subcommands, the process.argv received by the base process (xxxx.js) is NOT that file itself, but rather the symlink that launched it, which is /usr/local/bin/xxxx.

The various machinations in executeSubCommand are of course never going to find the subcommand file (since it's not in /usr/local/bin - the only thing there is the symlink to the "main" command that npm created).  Thus it fails to load the module, with a fairly unhelpful generic module load stack trace from Node.

As far as I can tell, the idea is that executeSubCommand should be looking in the same directory where the actual main command file lives.  I am working around this for now by doing:

    process.argv[1] = __filename;

right before I call the Commander parse.

zhiyelee:
fix Subcommands not found when installed globally #335


zhiyelee:
Test cases for this fix will be added later.

zhiyelee:
I want to stop supporting node versions <v0.8 in next version `v2.7.0` , how do u think about this? @SomeKittens

SomeKittens:
Not sure - couldn't find usages stats on how many are still on `0.8`.

SomeKittens:
Talked to some folks on #nodejs and #npm (freenode) - there's no good way to tell usage, but we should keep `0.8` support for now - at least until 0.12 settles down.

zhiyelee:
Yep, IMO versions below `v0.8`(not include) are too old. https://github.com/zhiyelee/commander.js/commit/9c42834db2bf184dd749ab9527b69e762242d515

zhiyelee:
@SomeKittens Dropped versions `< v0.8`  needs a Major version, maybe we can drop it from `v3.0`. I'll revert the commit and merge this PR.

SomeKittens:
@zhiyelee sounds good to me.

PR #353, opened 2015-02-24
Add support for camelCase on opts()

nkzawa:
Fix `opts()` fails to return an option something like `--foo-bar`.

### DIFF
--- a/index.js
+++ b/index.js
@@ -688,7 +688,7 @@ Command.prototype.opts = function() {
     , len = this.options.length;
 
   for (var i = 0 ; i < len; i++) {
-    var key = this.options[i].name();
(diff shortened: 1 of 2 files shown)


SomeKittens:
I'm sorry, the coffee has not kicked in yet.  Can you provide a case where this would fail in `master` that does not fail here?

nkzawa:
I already added a test case which fails on master.

If you use an option which has `-` in it like `--foo-bar` then `opts()` can't parse that correctly.

```js
program
  .option('--foo-bar')
  .parse(process.argv);

program.opts() // { 'foo-bar': undefined }
```

SomeKittens:
Ah, thank you - looks good!

PR #382, opened 2015-03-26
Given git-style sub-commands, display help if none called

claylo:
When one types `git` and hits enter, top-level git help output appears.

Presently, `./examples/pm` does not follow this behavior; the command simply outputs nothing.

This patch fixes this by injecting `--help` into argv. This approach to the problem causes the `outputHelpIfNecessary` method to behave as expected in existing test cases, yet displays help output in use case described.


SomeKittens:
Very nice, thanks so much!

PR #383, opened 2015-03-26
Allow for commands to be hidden from help menu

tonylukasavage:
This PR allows you to specify a command for commander.js, but prevent it from showing up in the help listing. My particular use case is that I'm writing a CLI which is going to be leveraged by a GUI. There are specific commands (and options) that are only going to be relevant when invoked by the GUI's special use cases. The commands are not relevant to the straight-up CLI user, and therefore I don't want to display them to the user when they call for `help`.

I'm not married to the API here, which simply allows for a 3rd arg to be passed to the `command()` function. I think it works, it's functional, and it doesn't interfere with any existing code, but I'm open to suggestions. Particularly considering that if this PR is accepted I'd like to do the same for options. Given the complexity of the `option()` API compared to the `command()` one, though,  I figured `command()` was a simpler place to start.

The change and intended usage can be found in the test included with this PR.

### DIFF
--- a/index.js
+++ b/index.js
@@ -156,7 +156,8 @@ Command.prototype.__proto__ = EventEmitter.prototype;
  * @api public
  */
 
-Command.prototype.command = function(name, desc) {
+Command.prototype.command = function(name, desc, opts) {
+  opts = opts || {};
   var args = name.split(/ +/);
   var cmd = new Command(args.shift());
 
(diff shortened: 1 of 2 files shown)


jhaynie:
+1 :+1:

SomeKittens:
Setting this only on `.command` for now makes sense for now, given your use case.  Thanks so much for the detailed PR + tests!

PR #388, opened 2015-04-05
Add .npmignore to exclude unnecessary development files from npm package

panuhorsmalahti:
I copied .gitignore as .npmignore and added a few development files and folders to be excluded from the npm package as they're unnecessary. Also removed node_modules as it's unnecessary as per npm's documentation.

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -1,0 +1,7 @@
+.DS_Store
+*.sock
+.idea
+
+# Development files
+/test/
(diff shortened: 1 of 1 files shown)


panuhorsmalahti:
Ah, the package.json file already includes a files property.

SomeKittens:
Thanks for taking the time to help us out anyway!

PR #389, opened 2015-04-06
Add support for coffeescript subcommands using explicit .coffee extension

carlosforero:
I added support for coffeescript subcommands using explicit `.coffee` extension.

The test run only if the `coffee` command is available. I renamed the variable `bin` to `symbolBin` in `test.command.executableSubcommand`, because this shadowed the `bin` variable in the coffee subcommand test, since this test is executed in a deep callback.

### DIFF
--- a/index.js
+++ b/index.js
@@ -507,11 +507,16 @@ Command.prototype.executeSubCommand = function(argv, args, unknown) {
   // prefer local `./<bin>` to bin in the $PATH
   var localBin = path.join(baseDir, bin);
 
-  // whether bin file is a js script with explicit `.js` extension
+  // whether bin file is a js or coffee script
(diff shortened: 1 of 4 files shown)


abetomo:
@carlosforero I think that the opportunity to use coffeescript is less than before.
Is this function still necessary now?

carlosforero:
@abetomo 

I think coffeescript has been forgotten these days, no offence intended.

So, this pull request can be discarded.

abetomo:
@carlosforero 
Thank you for your reply.
I am sorry that the reaction slowed down.

abetomo:
I will close, but I'd appreciate it if you can open PR if there is something again.

carlosforero:
@abetomo sure!

Issue #255, opened 2014-09-06
Extra "padding" argument: what is it for?


rrthomas:
In the code for the `action` method of sub-commands is the following code:

```
// Always append ourselves to the end of the arguments,
// to make sure we match the number of arguments the user
// expects
if (self._args.length) {
  args[self._args.length] = self;
} else {
  args.push(self);
}
```

What is it for? Why would the user expect an extra argument at this point? It means that detecting a missing argument is a bit tricky, as it can be either an object (if only one argument is missing at the end of a list) or `undefined` (if more than one is missing). Normally one only has to check for `undefined`.

rrthomas:
I looked into this history of this code, and it was added for the benefit of subcommands, so that the `action` routine never receives undefined arguments. But it seems to me that it is perfectly reasonable for an argument to be undefined if the corresponding command-line argument was not supplied.

rrthomas:
I offer a PR that removes this code, which avoids some checks for undefined arguments in `action` routines at the expense of making it much harder to check whether there were missing or superfluous arguments.

The case of dealing with undefined arguments in `action` routines will be better addressed by strict argument spec parsing; on the other hand, this change makes it easier to implement automatic strict checking, and possible to implement such checking manually in the mean time.

shadowspawn:
This issue has not had any activity in over six months. It isn't likely to get acted on due to this report.

Feel free to open a new issue if it comes up again, with new information and renewed interest.
 
Thank you  for your contributions.

Issue #259, opened 2014-09-09
Allow rejection of wrong number of arguments


rrthomas:
At present, it's really hard to tell whether your command was passed the number of arguments it asked for. I gave up and patched commander, and still had to check in two places (one for too many, one for zero; too few and greater than zero are checked automatically by the argument parser).

A solution would be to insist that the arguments have to match the syntax (i.e. optional arguments may be omitted, required arguments must be present, additional arguments are not allowed), and if zero arguments is not a valid number, treat that case as if "--help" were given (but preferably exit with error).

I have a patch for this which I can make into a pull request.

tecfu:
+1

tecfu:
At a minimum, it is a lot of extra, superfulous code that goes into validating the number of arguments passed to each command individually. 

Granted, I don't know the POSIX spec, but being able to generate a list of commands that enforce the number and type of arguments passed would be ideal.

What is the difference between the <value> and [value] notation? If there is, can we reference this in the README? If there is no spec for this, can we consider a notation that enforces argument type? 

i.e.
.option('-C, --action [alphanumeric] --qty <numeric> --price <numeric>', 'require all these args or return an error', callback)

shadowspawn:
I think the PR referred to here has been merged so this particular report is now stale.

This issue has not had much activity in four years. It isn't likely to get acted on due to this report.

Feel free to open a new issue if it comes up again, with new information and renewed interest.
 
Thank you  for your contributions.

PR #393, opened 2015-04-08
Issue #255: don't append padding arguments to the end of the list

rrthomas:
See the discussion of issue #255 for the situation.

In brief, the removed code tried to reduce potential confusion, but in fact added to it.

### DIFF
--- a/index.js
+++ b/index.js
@@ -289,15 +289,6 @@ Command.prototype.action = function(fn) {
       }
     });
 
-    // Always append ourselves to the end of the arguments,
-    // to make sure we match the number of arguments the user
-    // expects
-    if (self._args.length) {
-      args[self._args.length] = self;
-    } else {
-      args.push(self);
-    }
-
     fn.apply(self, args);
   };
   var parent = this.parent || this;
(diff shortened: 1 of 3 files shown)


SomeKittens:
Hmm.  This seems like a major change (necessitating going to version 3.x.x).  People will be depending on this behavior and I don't want to break it.  I'm with Torvalds on this one (Don't break user space).

rrthomas:
I agree this could be an upsetting change. How else could one detect missing or superfluous arguments? I suppose the module could still arrange internally to detect them (e.g. some kludge like having a separate "real" count), so I could submit a patch for #259 that does not depend on this PR.

If there are (ever?) any 3.x.x plans, I would suggest this would be a good simplifying, if breaking change.

shadowspawn:
I don't think we would consider this change without better understanding what it was trying to achieve, which wasn't discovered in the linked issue. The behaviour is a bit weird so something we might revisit one day.

This PR has not had any activity in over six months. I suggest we close this and invite future readers to open a new PR if wanted.

shadowspawn:

Issue #396, opened 2015-04-16
Regression in multiline descriptions


simast:
Any chance [this feature](https://github.com/tj/commander.js/commit/6134fe44992a301485b087cc7d74a2634311e99d) can be reverted?

Previously, using the newline and tab characters in descriptions allowed us to provide an extended description/help of a command with a lot of space available immediately after the command syntax line. However after this change - all multiline description text lines are crammed and aligned in a column right after the command syntax text..

Example of previous functionality:
```
    -d, --date <YYYY-MM-DD>      select date
    -t, --time <HH:MM>           select time

        Time can also be specified using special values:

        "dawn" - From night to sunrise.
        "sunrise" - Time around sunrise.
        "morning" - From sunrise to noon.
        ...
```

.. and after the change:
```
    -d, --date <YYYY-MM-DD>      select date
    -t, --time <HH:MM>           select time

                                 Time can also be specified using special values
:

                                 "dawn" - From night to sunrise.
                                 "sunrise" - Time around sunrise.
                                 "morning" - From sunrise to noon.
                                 ...
```

zhiyelee:
Sorry, seems that I't a break change.  I think it should be revert, what's your opinion  @SomeKittens ?

SomeKittens:
Yes, this is a bug, should be reverted.  Sorry about that.

PR #208, opened 2014-04-10
multiline descriptions support

zxqfox:
fixes #207

PR #184, opened 2013-11-18
Implement the wrapping of command and option descriptions

socsieng:
Implemented as suggested in issue #123

Formatting tweak to make descriptions wrap and align vertically with the command/option description blocks

Output prior:

```
  Usage: test [options] [command]

  Commands:

    long                   This is a long command description. The purpose of this is to demonstrate how command descriptions will wrap over multiple lines
    short                  This is a short command description

  Options:

    -h, --help   output usage information
    -l, --long   This is a long option description. The purpose of this is to demonstrate how option descriptions will wrap over multiple lines
    -s, --short  This is a short option description
```

Output after:

```
  Usage: test [options] [command]

  Commands:

    long                   This is a long command
                           description. The purpose of this
                           is to demonstrate how command
                           descriptions will wrap over
                           multiple lines
    short                  This is a short command
                           description

  Options:

    -h, --help   output usage information
    -l, --long   This is a long option description. The
                 purpose of this is to demonstrate how
                 option descriptions will wrap over multiple
                 lines
    -s, --short  This is a short option description

```

PR #397, opened 2015-04-17
Revert "multiline descriptions support"

zhiyelee:
Reverts tj/commander.js#208

Refer: Regression in multiline descriptions #396


zhiyelee:
cc @zxqfox , I think #184 is much better in compatible with old version.

zxqfox:
I'm agreed with @tj, but ok.

PR #405, opened 2015-05-20
Parse arguments even without 'action'.

IvanGoncharov:
In case if no arguments provided there was no error about missing argument.
Also arguments was validated if no action specified.

[Review on Reviewable](https://reviewable.io/reviews/tj/commander.js/405)

### DIFF
--- a/index.js
+++ b/index.js
@@ -257,53 +257,13 @@ Command.prototype.parseExpectedArgs = function(args) {
 
 Command.prototype.action = function(fn) {
   var self = this;
-  var listener = function(args, unknown) {
-    // Parse any so-far unknown options
-    args = args || [];
-    unknown = unknown || [];
-
-    var parsed = self.parseOptions(unknown);
-
-    // Output help if necessary
-    outputHelpIfNecessary(self, parsed.unknown);
-
-    // If there are still any unknown options, then we simply
-    // die, unless someone asked for help, in which case we give it
-    // to them, and then we die.
-    if (parsed.unknown.length > 0) {
-      self.unknownOption(parsed.unknown[0]);
-    }
-
-    // Leftover arguments need to be pushed back. Fixes issue #56
-    if (parsed.args.length) args = parsed.args.concat(args);
-
-    self._args.forEach(function(arg, i) {
-      if (arg.required && null == args[i]) {
-        self.missingArgument(arg.name);
-      } else if (arg.variadic) {
-        if (i !== self._args.length - 1) {
-          self.variadicArgNotLast(arg.name);
-        }
-
-        args[i] = args.splice(i);
-      }
-    });
-
-    // Always append ourselves to the end of the arguments,
-    // to make sure we match the number of arguments the user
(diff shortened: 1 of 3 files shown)


index.js
```diff
@@ -602,26 +560,57 @@ Command.prototype.normalize = function(args) {
  */
 
 Command.prototype.parseArgs = function(args, unknown) {
-  var name;
+  var self = this;
 
-  if (args.length) {
```
IvanGoncharov: No check for required args, if no arguments provided.

index.js
```diff
@@ -257,53 +257,13 @@ Command.prototype.parseExpectedArgs = function(args) {
 
 Command.prototype.action = function(fn) {
   var self = this;
-  var listener = function(args, unknown) {
```
IvanGoncharov: Check argument only if 'action' registered.

test/test.variadic.args.js
```diff
@@ -26,7 +26,10 @@ program.args.should.have.lengthOf(3);
 program.args[0].should.eql('arg0');
 program.args[1].should.eql(['arg1', 'arg2', 'arg3']);
 
-program
+//Create fresh instance
+var program2 = new program.Command();
+
+program2
```
IvanGoncharov: same sub command was registered twice.
Previously it work because events was overridden.
But inside `commands` array it was two instances with same name.
This lead to problem with help messages and should be handled in separate commit.
IMHO, exception should be thrown then duplicated commands registered.

IvanGoncharov:
@zhiyelee Did you have a chance to look at my changes?
It pretty simple and safe, because most of changes is code move from one function into another.
If required I can try split them into several commits.

SomeKittens:
@IvanGoncharov it'd help if you could summarize the why (not what) of this PR

IvanGoncharov:
Main reason is following scenario:
```javascript
#!/usr/bin/env node
var program = require('./index.js');

program
  .arguments('<ARG>')
  .parse(process.argv);
```
Before changes:
>$ ./test.js
>$

After my changes:
>$ ./test.js 
>
>  error: missing required argument `ARG'

IvanGoncharov:
It could be workaround simply be removing [this](https://github.com/tj/commander.js/blob/master/index.js#L607) check.
But I think delaying args parsing until event handler unreasonably complicate code.
It also force register events for aliases and create complicated wrapper for listener.

SomeKittens:
This is a pretty big change - not sure if I want to make it at this point.  @zhiyelee ?

IvanGoncharov:
When I debug these issue, I spent quite some time just to understand where args parsing is happening.
I never would have guessed, it resides inside listener wrapper.
The simpler and clean code is, the more people would contribute.
That is my two cents.

To simplify review, I can break it into several commits.

zhiyelee:
@SomeKittens Really big change, I will review it later this weekend.

tvervest:
Any news on this PR? I'd love to see built-in required arguments :+1:

PR #408, opened 2015-05-25
Fix #407 of commander.js

### LINKED ISSUE #407: Conflict with collections module when using "all" parameter
Hi.

I encounter the following problem:

```
$ ./supersoftware reset all
all
execvp(): No such file or directory

  supersoftware-all(1) does not exist, try --help
```

This is due to a conflict with an external module [collections.js](https://github.com/montagejs/collections):

```js
#!/usr/bin/env node
"use strict";
require('collections/set'); // <----- Here is the collision

program
  .version('0.0.1')
  .usage('<command> [options]')
;

program
  .command('reset <what>')
  .description('Reset some parameter')
  .action(function (what) {
    console.log(what);
  });

program.parse(process.argv);
```

Indeed, `collections` define an `all` method on Arrays, which conflicts with the `all` parameter I use in my command.

A turnaround is to lazy initialize `collections` so Commander is not polluted:

```js
#!/usr/bin/env node
"use strict";

program
  .version('0.0.1')
  .usage('<command> [options]')
;

program
  .command('reset <what>')
  .description('Reset some parameter')
  .action(function (what) {
    console.log(what);
    setTimeout(function() {
      require('collections/set');
    }, 10);
  });

program.parse(process.argv);
```

But more deeply, the problem comes from Commander code to me, since the bug is [here](https://github.com/tj/commander.js/blob/master/index.js#L86):

```
function Command(name) {
  this.commands = [];
  this.options = [];
  this._execs = []; // <---- this
  this._args = [];
  this._name = name;
}
```

And then [here](https://github.com/tj/commander.js/blob/master/index.js#L167):
```
if (desc) this._execs[cmd._name] = true;
```

Why using `Array` as an object? Replacing `[]` by `{}` completely solves the problem.

c-geek:


zhiyelee:
Thank you @c-geek

PR #411, opened 2015-06-15
killing the subcommand if the main command is terminated

thomashoffmann1979:
The sub command keeps running if the main command was killed. I added listeners to that events to terminate the sub command.


index.js
```diff
@@ -1100,4 +1109,3 @@ function exists(file) {
     return false;
   }
 }
-
```
SomeKittens: Keep the newline here

SomeKittens:
Tests would be great, :+1: otherwise

zhiyelee:
Should it be an option or default?

index.js
thomashoffmann1979: Ok, my atom editor kills extra lines. I add the line again.

thomashoffmann1979:
I think this should be the default behavior. If you try to store the process pid, you will get only the pid of the main process. 

```
mycommand service  1> /tmp/log.txt 2>&1 &
```
Returns something like:
```
[1] 55274
```
`ps aux | grep mycommand` shows:
```
me  55275   0,0  0,3  3073148  50068 s000  S    10:05am   0:00.40 node ./bin/mycommand-service
me  55274   0,0  0,3  3069368  46176 s000  S    10:05am   0:00.32 node ./bin/mycommand service
```
 
If you try to kill that pid (55274), the sub command (55275) keeps running.

index.js
```diff
@@ -534,7 +534,16 @@ Command.prototype.executeSubCommand = function(argv, args, unknown) {
...
   }
 
-  proc.on('close', process.exit.bind(process));
+  //killing the child process if the main process is terminated
+  var processClosing = function(){
+    proc.kill('SIGHUP');
+  }
+  process.on('exit',processClosing );
+  process.on('SIGINT', processClosing );
+  process.on('SIGHUB', processClosing );
```
zhiyelee: `SIGHUP`?  Have you test all this listener? @thomashoffmann1979

zhiyelee:
Hi @thomashoffmann1979 , we can merge this if you add a test. Thank you.

index.js
thomashoffmann1979: SIGHUP seem to be right see https://en.wikipedia.org/?title=SIGHUP.  I will add a test for it.

zhiyelee: Yea. To double check, I mean here we should use `SIGHUP` instead of `SIGHUB`.

thomashoffmann1979: Oh, you are right! I will fix it.

thomashoffmann1979: I fixed it. Now the exit signal of the main process is used for the sub command, too.

SomeKittens:
This should be default in my opinion - seems expected behavior to kill the whole process.

DLMR:
Is the only thing that prevents this from being merged that it has conflicts? 

Would really like this functionality included in Commander.js.

zhiyelee:
@thomashoffmann1979 Can you solve the conflict in this PR.  I will merge it.

DLMR:
Seems like the build fails for Node.js v0.6 and v0.8. Is this still something that should be supported and have anyone looked into what the cause is?

SomeKittens:
@DLMR IIRC, we're still supporting v0.8.  Dropping it would be a major version change.  See #431

vanesyan:
closing in favor of #632

PR #415, opened 2015-06-21
add default executable (subcommand) option

Qix-:
This PR introduces an option for the `.command()` call that allows git-style subcommands to default to a specific command in the event one isn't provided on the command line.

The use case that brought about this PR was a build system I'm developing needing to run the subcommand `build` automatically when nothing was specified on the command line. I'm sure there is a hack that could have done this with `.action()` but alas I wanted it to work with the executable functionality (i.e. having another script to do run the command).

I'm sure I'm not the only one who can find a use case for this.

Test is included (based off of the pm-subcommand test) and shouldn't interfere with the existing test.

Also added some documentation (including documentation for the previously undocumented `opts.noHelp` option).

I tried to keep the implementation as minimal as possible. Let me know if I need to change anything.

### DIFF
--- a/test/fixtures/pm-default
+++ b/test/fixtures/pm-default
@@ -1,0 +1,2 @@
+#!/usr/bin/env node
+console.log('default');
--- a/Readme.md
+++ b/Readme.md
@@ -167,12 +167,14 @@ program
   .version('0.0.1')
   .command('install [name]', 'install one or more packages')
   .command('search [query]', 'search with optional query')
-  .command('list', 'list packages installed')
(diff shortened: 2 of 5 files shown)


noamokman:
:+1:

test/fixtures/pm-default
```diff
@@ -0,0 +1,2 @@
+#!/usr/bin/env node
+console.log('default');
```
SomeKittens: What's this for?

SomeKittens:
LGTM

PR #420, opened 2015-07-13
add optional callback parameter in parse method

jonatasnona:
This patch allows execute a callback after the parse
step by passing an optional 'function' as the
second parameter of the parse method.

This feature is useful when we working with async tasks.

### DIFF
--- a/examples/callback
+++ b/examples/callback
@@ -1,0 +1,24 @@
+#!/usr/bin/env node
+
+/**
+ * Module dependencies.
+ */
+
+var program = require('../');
+
+function list(val) {
+  return val.split(',').map(Number);
+}
+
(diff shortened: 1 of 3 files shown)


SomeKittens:
Parsing isn't for situations where you'd need async tasks but rather where you'd want to convert types and the like.

abetomo:
@vanesyan I think that SomeKittens says, but what do you think?

vanesyan:
@SomeKittens is right. Plus, IMHO it's unnecessary, we really can pass the result after parsing as always:
```js
const result = program.parse(process.argv)

func(result)
```

vanesyan:
Let's close this one.

PR #425, opened 2015-07-30
Move 'description' to the top

k5e:
This moves the program's description in the help display up from between commands and options to the top.


zhiyelee:
No, thank you.

PR #429, opened 2015-08-24
Do not parse an arg with spaces as an option.

chasingmaxwell:
This allows for correct parsing in cases like this:

`command "-y this should not be parsed as an option"`


chasingmaxwell:
Apparently it's not going to be that easy to solve. :(

zhiyelee:
yep, thanks, i will close this. You can reopen this after figuring out a solution.

PR #433, opened 2015-09-10
Default value not set on function options

malthejorgensen:
When a function is passed as the third argument to `.option` the default
value from the fourth argument isn't set when the flag/option isn't
given on the commmand line.

E.g. in the README 

```
.option('-v, --verbose', 'A value that can be increased', increaseVerbosity, 0)
```

if no `-v` or `-verbose` flag is passed then `program.verbose` will be `undefined` rather than `0`.
I feel that this is unexpected and not intended behavior.

This commit fixes the problem (as far as I can figure) and adds tests to
ensure the functionality works.

### DIFF
--- a/index.js
+++ b/index.js
@@ -376,6 +376,10 @@ Command.prototype.option = function(flags, description, fn, defaultValue) {
       defaultValue = fn;
       fn = null;
     }
+  } else if (undefined !== defaultValue) {
+    // If the 3rd arg is a function and a default value is given preassign it so
+    // that the default value is available even if the option isn't passed
+    self[name] = defaultValue;
   }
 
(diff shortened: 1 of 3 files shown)


SomeKittens:
IIRC, in this situation, isn't the proper thing for `increaseVerbosity` to return zero when nothing's passed in?

malthejorgensen:
@SomeKittens Could be, but in that case I think it should be reflected in the readme. Also I feel the behaviour implemented here is less surprising than having to make a special-case in `increaseVerbosity`.

Thoughts?

shadowspawn:
I am keen to try a big rewrite of options + functions in README. I think I have seen code ignoring default value for boolean flags, so might be able to cover that. See #943 for big triage item.

PR #436, opened 2015-09-15
Add support for exit codes.  

### LINKED ISSUE #188: optional exit value for Help. (feature request)
It would be extremely convenient for help to take an optional return value so that the program could easily exit with an error code. This is useful in cases where a combination of options are required together and you would want to report to the user the usage but also to the system not to continue with anything else that may be in the execution chain.

### LINKED ISSUE #435: process.exit(64)
When you call exit, the return code probably should not be `0` (success).  Traditionally, option parsing errors have resulted in error code 64, as documented in many systems' `sysexits.h` header file, which on my system says:

```
#define EX_USAGE    64  /* command line usage error */
```

I would be happy to submit a PR if you're open to this change.

hildjj:
Fixes #435.  
Fixes #188.

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -95,7 +95,7 @@ program
   .option('-s --size <size>', 'Pizza size', /^(large|medium|small)$/i, 'medium')
   .option('-d --drink [drink]', 'Drink', /^(coke|pepsi|izze)$/i)
   .parse(process.argv);
-  
+
 console.log(' size: %j', program.size);
 console.log(' drink: %j', program.drink);
 ```
@@ -298,6 +298,10 @@ function make_red(txt) {
   Output help information and exit immediately.
   Optional callback cb allows post-processing of help text before it is displayed.
 
+## .exitCode(code)
+
+  Set the exit code for option parsing errors.  Defaults to `64`.
+
 ## Examples
 
 ```js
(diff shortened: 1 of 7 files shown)


SomeKittens:
Broken tests, can you fix those up?

zhiyelee:
@SomeKittens  I'm not sure whether we need that

zhiyelee:
Also that would be break change to change default code from 1 to 64.

hildjj:
Sorry, thought I had gotten all of the tests fixed up.  I think there's a legitimate discussion to be had about what exit code to use when the user *asks* for help with `--help` vs. when they supply invalid arguments.  I would be fine with `--help` returning 0, if you prefer.

hildjj:
Oh, and I'm also happy to change the defaults so that the interface doesn't change, and only the people that want 64-exits get them.

zhiyelee:
@hildjj Thank you!  I'm not sure `64` is convention for such error. I have checked `wget` and `git` with unknown option, they give `2` and `129` respectively. 

I am open to allow user set custom exit code.

SomeKittens:
@zhiyelee good checking - I'm ok with user-set as well.

vanesyan:
@zhiyelee is it still necessary?

shadowspawn:
I have added this pull request to a triage collection of related issues for future reference.

PR #442, opened 2015-09-29
Fix -- handling, don't throw away subsequent -- .

geophree:
* Add test for -- in normalize


zhiyelee:
That is designed. `--` is used in bash to  signify the end of command options
http://unix.stackexchange.com/questions/11376/what-does-double-dash-mean-also-known-as-bare-double-dash

geophree:
Yes, but you're throwing away all the `--` in the string.  You should be retaining all the ones after the first.

PR #444, opened 2015-09-30
Explicitly return from error methods

### LINKED ISSUE #443: Explicitly return when an error occurs
Methods that rely on error methods (such as `unknownOption`, `missingArgument`, etc...) use `process.exit()` to stop execution and exit the program. This is quite inconvenient if you want to use Commander in a context other than a shell.

If a client extends the Command prototype to avoid having these error methods call `process.exit()`, command execution continues and commander blows up.

Personally I am using commander to parse arguments for my Slack bot, extending Command to prevent `process.exit()`s but even still if you input an incorrect command the program will crash.

plemarquand:
Instead of relying on process.exit() to terminate execution, also explicitly
return from methods when errors are encountered. This is more flexible for
clients who which to override the methods that use process.exit() in order
to use commander in other contexts than the shell

Fixes #443

### DIFF
--- a/index.js
+++ b/index.js
@@ -264,6 +264,9 @@ Command.prototype.action = function(fn) {
     unknown = unknown || [];
 
     var parsed = self.parseOptions(unknown);
+    if(null == parsed) {
+      return;
+    }
 
     // Output help if necessary
     outputHelpIfNecessary(self, parsed.unknown);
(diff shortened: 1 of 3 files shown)


zhiyelee:
This will break the old manner?

plemarquand:
@zhiyelee No, this only comes in to play when you override `process.exit`, or the commander methods that use it. Commander relies on process.exit to halt program flow. These returns mimic that if process.exit doesn't do it. 

By doing so now you can use commander in contexts other than a shell.

plemarquand:
@SomeKittens @zhiyelee Any thoughts? The patch has some added tests.

SomeKittens:
@zhiyelee what will break?  This looks good to me.

zhiyelee:
@SomeKittens Nothing break. But I'm not a big fan of this since to use this, user must override `process.exit` first. Not very graceful, but I have no better solution.
Besides I think Commander is designed to use in command line instead as a class can be extended.

plemarquand:
@zhiyelee I agree its not the most graceful approach when it comes to overriding `process.exit`. A better one long term would be centralizing all the `process.exit`s into a single method that can then be overridden, and passing to that method the exit code as well as error message. The default approach, as it is now, would be to print that message and exit. However then if you wanted to do something else like forward the message to a remote client, you could do so.

Even if we do that, we'll still need to implement these changes where program flow is interrupted by returns instead of relying on `process.exit` to do it for us.

plemarquand:
I should also mention this patch goes most of the way towards fixing the second Con on this list: http://trentm.com/2014/01/a-critique-of-commander-for-nodejs.html

inikulin:
In addition to the reason of this PR (testing purposes) we have a requirement to apply custom styling for the error messages to make them consistent across our app (e.g. we are using bold red `ERROR` prefix). I think the best solution for the both issues will be Command's `error` event. If it hasn't listeners it will behave as it does currently. If it has listener, when process will not exit and message text is passed as event argument. I can issue PR if you are OK with this behavior.

kevinoid:
I'd also like to see this (and preferably a way to pass an exit function to the `Command` constructor).  I'm currently overriding `process.exit` to throw an exception, which also works, but it would be nice to avoid patching `process.exit` altogether.

shadowspawn:
I have added this pull request to a triage collection of exit related issues for future reference.

PR #457, opened 2015-10-20
stick to module.exports for clarity

tejasmanohar:
- no point using `exports.` as Node ALWAYS uses `module.exports`
- if `exports.x` and `module.exports.x` are stated, which is used? no one knows... that's why better to stick to one, just cleaner... avoid clutter, too :)

### DIFF
--- a/index.js
+++ b/index.js
@@ -14,19 +14,19 @@ var fs = require('fs');
  * Expose the root command.
  */
 
-exports = module.exports = new Command();
+module.exports = new Command();
 
 /**
  * Expose `Command`.
  */
 
-exports.Command = Command;
+module.exports.Command = Command;
 
 /**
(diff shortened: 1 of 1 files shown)


zhiyelee:
IMO, `exports.xxx = ` is equal to `module.exports.xxx = ` 
But need to use `module.exports = ` instead of  `exports = `

SomeKittens:
>  no one knows...

[there's very much a difference](http://stackoverflow.com/a/7142924/1216976)

> Setting module.exports allows the database_module function to be called like a function when required. Simply setting exports wouldn't allow the function to be exported because node exports the object module.exports references.

tejasmanohar:
@SomeKittens oh neat i didn't know htat. but yeah that gives another reason to use module.exports i'd say

PR #458, opened 2015-10-20
prefer ES6 Function#bind to var self = this

tejasmanohar:


SomeKittens:
Not bad but this is just a syntax change that doesn't add anything.  It'd help if you explained why this format is better over the current way (which is pretty standard for ES5 environments).

tejasmanohar:
@SomeKittens removes unnecessary variable declaration- very very minor change. no drawbacks afaik. this feature exists in es5 and is slightly cleaner than people looking at why are they doing `var self = this` so why not use it :)

zhiyelee:
I prefer the `self`.

Issue #230, opened 2014-07-13
Required arguments for options and commands are not enforced


foxx:
It would be pretty awesome to have support for required arguments. Although there is support for required argument values, there doesn't appear to be anything for ensuring the argument has been specified.

For example;

```
    -e, --entry <name>          Add module to entry point
```

I'd like to make it so an error is raised if `-e` has not been specified like so;

```
  error: option `-e, --entry <name>' missing
```

NameFILIP:
I see a lot of open pull-requests.. Will this be merged if I spend time implementing it?

SomeKittens:
@NameFILIP Hopefully.  If you want to ensure your PR is merged, make sure it's wanted (already shown in this case), and has proper tests & documentation.

czardoz:
Is anyone actually working towards this?

marcelofarias:
My 2 cents: https://github.com/tj/commander.js/pull/462

foxx:
Just come up against this again, seems to be quite a few PRs, are any due to be merged by core dev?

foxx:
Any word on this @tj?

Petah:
![Any word yet?](https://i.ytimg.com/vi/bDQpF92l1vg/mqdefault.jpg)

amrynsky:
just came across this issue and didn't really get why it's still not implemented. I saw different reasons for not doing it, but the main problem that current implementation is not really consistent:
- In some cases, validation is performed 
`program.command('command <option>')`
in some not
`program.option('--option <option>', 'option')`

what the reason to define required option and don't perform any validation? Of course, it's possible to implement custom validation, but then you need to align error messages with standard validation messages like '  `error: missing required argument `option''`

ratancs:
Any update on this?

Strandedpirate:
I immediately regret using this library.

schovi:
Any update? :)

amir-arad:
+1

cyrfer:
+1

teazean:
+1
really want this feature

joereynolds:
Just came across this thread and I'm wondering the same thing. How is there not support for required arguments, and why the reluctance? :(

danihwsr:
+1

Tiim:
+1

chumster:
+1

xCykrix:
The fact that this **still** isn't a thing almost hurts. If you want to be a command-line library this should be one of the core features.

[...]

PR #462, opened 2015-10-29
Automatically check for missing required options

marcelofarias:


travisperson:
Why not have a `required` function, which acts the same as `option` but sets a flag?

marcelofarias:
Isn't it what's already implied by <> like in "-r <required>"? I mean, does it make sense to mark an option as <required> and allow the user to not supply it at all?

travisperson:
> I mean, does it make sense to mark an option as <required> and allow the user to not supply it at all?

I think so. The angel brackets say that if you use this flag, you must also provide a value. There might also be times when I want to force the user to provide an argument, but I wish to have it named.

So I can see two types of flag style arguments for required.

1) A flag that if specified, has a required value (angle brackets)
2) A flag that must be supplied (non optional flag)

rottmann:
:+1: 
Or name the function `requiredOption` (no need to change or touch the option() args), so it is obvious that an option is required.

snow01:
I have fixed this in different manner. No additional function or flag is required. See here - https://github.com/tj/commander.js/pull/503

brunapereira:
@snow01 Your fix just makes sure that if I have a flag `-C` without a value, it will throw an error. I think the main idea of the functionality is throw an error when the flag itself is not specified.
Nowdays: `./command -C => throws an error`
What is being required: `./command => throws an error. Where is the flag -C?`

So, will the feature be merged? 😄

nickbarry:
I agree something like this would be helpful. In the meantime, I wrote a utility function that automatically checks for required params:

```javascript
cli
  .version('0.0.1')
  .option('-y, --year <y>', 'Year for reports')
  .option('-d, --directory <path>', 'Directory with the report subdirectories')
  .option('-u, --upload', 'Do upload files for matched companies')
  .parse(process.argv);
assertRequiredArgs(cli);

function assertRequiredArgs(cli) {
  const ommittedRequiredParameters = [];
  cli.options.forEach((option) => {
    if (option.required && !cli.hasOwnProperty(name)) {
      const name = option.long.slice(2);
      ommittedRequiredParameters.push(name);
    }
  });

  if (ommittedRequiredParameters.length !== 0) {
    throw new Error(`The following required parameters were ommitted: ${ommittedRequiredParameters.join(', ')}`);
  }
}
```

shadowspawn:
I think this Pull Request allows treating all options with required values as mandatory options that must be specified? I consider those two independent things, as noted by some of the other comments.

See https://github.com/tj/commander.js/issues/230#issuecomment-478181355 for a long description.


This Pull Request has not had any activity in over six months.

PR #464, opened 2015-10-30
Make return of the `parse` api consistent.

lsm:


vanesyan:
It's unnecessary here, because of that fact, after using `parse` method, both `command` and `option` has no effect.

abetomo:
@lsm Could you tell me the assumed use case?

lsm:
@abetomo It's been almost 2 years.  Let me double check if this is still necessary.

vanesyan:
Any updates on the thread? Or we can safely close the pr?

lsm:
I was using commander in a lib instead of using it directly in app.  My use case was I need to modify the returned result after calling `parse`.  So it's nice to have the `parse` function always returns the `result` (instance of commander) regardless the input (having subcommands or not).  I guess no one is using it that way so maybe it's not necessary.  But, I do personally prefer functions to have consistent returns which I believe causes less `surprises`.

abetomo:
@lsm thank you for your answer.

@vanesyan What do you think?

vanesyan:
I think api is consistent with what we have today. @ism you can just path commander instance after parse it.
```js
...
commander.parse()

func(commander)
```

vanesyan:
I'll close this one, nevertheless thanks for the prompt :)

PR #478, opened 2015-12-16
fix lint issues

fscherwi:


SomeKittens:
This pretty drastically changes the code style of the project.  Thanks for taking the time to create this but I don't think we'll be merging it.

PR #480, opened 2015-12-21
Update README.md

pmbenjamin:
Update README.md to make it beginner friendly and elaborate on the difference between angled and square brackets.


nVitius:
This is good to know. It would be helpful to have in the README.

SomeKittens:
Nice catch, thanks!

Issue #422, opened 2015-07-23
program --help is broken


jadbox:
I'm getting this error when trying to get help for any program on it's base command. For example: "./myapp --help" doesn't work but "./myapp cmd --help" does work.

Error:
../node_modules/commander/index.js:945
    , cmd.description()
          ^
TypeError: Property 'description' of object #<Command> is not a function

jadbox:
I was able to patch the problem by changing this in the source...
Line 936   
 return [
      cmd._name
        + (cmd._alias
          ? '|' + cmd._alias
          : '')
        + (cmd.options.length
          ? ' [options]'
          : '')
        + ' ' + args
    , cmd._description  // WAS cmd.description()
    ];

jadbox:
Any updates on this?

SomeKittens:
This is odd.  Open a PR?

jadbox:
Odd indeed. I don't think I'll have time to open a PR for this one... I've since needed to deviate a bit from head.

jadbox:
node --version
v0.10.38

ignacioiglesias:
@jadbox did your command take a `--description`?

SomeKittens:
Fixed in above PR, thanks @ignacioiglesias !

PR #484, opened 2016-01-05
Fixes TypeError when a command has an option called `--description`

ignacioiglesias:
Using the `description()` method instead of the private property causes a `TypeError` when a command has an option called "description".

Since in the rest of the code the private property is used, I changed this line to use it too.

(this fix was proposed by @jadbox in #422 -- thx!)


SomeKittens:
LGTM, thanks!

PR #486, opened 2016-01-08
colors, l10n, args details

cravler:


vanesyan:
Sorry, we trying to keep commander out of such stylistic things. Plus, the second goal of commander is to keep it zero dependencies package.

PR #491, opened 2016-01-25
Fail when using same command and alias

### LINKED ISSUE #490: using same command and alias runs action twice
Failure program:

``` javascript
'use strict';

commander.version('0.0.0').description('fun');

commander.command('fail')
         .alias('fail')
         .description('fail twice')
         .action(fail);

commander.parse(process.argv);

function fail() {
  console.log('Should print this once');
}
```

Running `node <program-name> fail` produces:

```
Should print this once
Should print this once
```

The problem is that the alias and command are the same thing (`'fail'`). Arguably this is bad program, but running `fail()` twice is very much unexpected. Throwing an exception if the alias is the same as the command may be more appropriate.

ajfranzoia:
Fixes #490. Prevents running a command twice when -accidentally- using same command alias and name.

### DIFF
--- a/index.js
+++ b/index.js
@@ -880,6 +880,8 @@ Command.prototype.alias = function(alias) {
 
   if (arguments.length === 0) return command._alias;
 
+  if (alias === command._name) throw new Error('Command alias can\'t be the same as its name');
+
   command._alias = alias;
   return this;
 };
--- a/test/test.command.failOnSameAlias.js
+++ b/test/test.command.failOnSameAlias.js
@@ -1,0 +1,14 @@
+var program = require('../')
+  , should = require('should');
+
+var error;
+
+try {
+  program
+    .command('fail')
+    .alias('fail');
+} catch (e) {
+  error = e;
+}
+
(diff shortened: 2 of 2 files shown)


deian:
LGTM. Thanks for doing this. The only thing I would point out is a style difference: it seems like the rest of the file doesn't use curly braces for a single-statement `if`. (Though I'm in no position other than to informally comment.)

ajfranzoia:
@SomeKittens Let me know if there's anything that needs modification (same with #492 and #494)

hellowin:
@ajfranzoia I think what @deian said is, you just need to change
```
if (alias === this._name) {
  throw new Error('Command alias can\'t be the same as its name');
}
```
to
```
if (alias === this._name) throw new Error('Command alias can\'t be the same as its name');
```

abetomo:
@ajfranzoia
Could you please resolve the conflict?

ajfranzoia:
@abetomo done!

abetomo:
Thank you!

PR #494, opened 2016-01-26
Prefix events to prevent conflicts between commands and options

### LINKED ISSUE #471: 'version' is interpreted as '--version'
According to the docs, only "my-cmd --version" should output the version number, but I have found that commander also picks up the string 'version' anywhere in the arguments.

As an example of where this causes problems, you can have a look at a little utility I wrote called [pick_json](https://github.com/fatso83/pick_json/blob/master/pick_json.js#L25). It is used it to extract fields from json on stdin.

```
$ echo '{ "name" : "myapp", "version" : "3.4.5" }' > appinfo.json

$ pick_json name < appinfo.json
myapp

$ pick_json version < appinfo.json  #WTF moment
1.1.2

$ pick_json --version
1.1.2
```

As you can see, instead of picking the version field from the json, it shows the version number of `pick_json`. This is not inline with the docs.

ajfranzoia:
Prefix internal command events with `'command:'` and options events with `'options:'` to prevent conflicts between equally named commands and options (both user and plugin options, like `--version`).
This fixes #471 as well.

### DIFF
--- a/index.js
+++ b/index.js
@@ -302,8 +302,8 @@ Command.prototype.action = function(fn) {
   };
   var parent = this.parent || this;
   var name = parent === this ? '*' : this._name;
-  parent.on(name, listener);
-  if (this._alias) parent.on(this._alias, listener);
+  parent.on('command:' + name, listener);
+  if (this._alias) parent.on('command:' + this._alias, listener);
   return this;
 };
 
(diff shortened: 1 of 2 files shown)


vanesyan:
LGTM

yausername:
Shouldn't this be added to the README? It is not documented anywhere.
I was wondering why this isn't working

```
program.on('verbose', function () {
  console.log('reached here')
})
```

vanesyan:
Yes, I actually should be documented. Feel free to send a pr :+1:

PR #497, opened 2016-01-31
Add 'Other modules' section to Readme.md

justinhelmer:
- add commander.js-errors to the list of other modules


justinhelmer:
Not sure if there is any interest in this, but thought I would offer.

PR #498, opened 2016-01-31
Add 'Other modules' section to Readme.md

justinhelmer:
- add [commander.js-error](https://github.com/justinhelmer/commander.js-error) to the list of other modules


justinhelmer:
Not sure if there is any interest in this, but thought I would offer.

vanesyan:
It seems, we trying to keep our readme file clear and free of such extensions. Sorry :)

PR #501, opened 2016-02-04
test node.js stable

revington:
Added `node` to the target list in order to test commander with stable (actually node 5.x).
nvm uses `node` as a pointer to stable release.


SomeKittens:
I like this idea (we should be testing on all major versions at least), though wouldn't it be smart to point to `5` instead of `node`?

revington:
The idea is to always point to `stable` no matter which version `stable`. If 5 become LTS and 6 become stable then `node` will point to 6 and we can add 5 as LTS

SomeKittens:
Alright, makes sense.  In either case we'll need to update the tests when there's a new major version but that's not too often.  Merged!

PR #502, opened 2016-02-11
Fix Sub-Subcommands

bkendall:
When commands were nested, I was finding that it was trying to run a file that didn't exist. The logic here _found_ the correct path and file, but then used an incorrect variable later in the code.

I added a test for this since I didn't see one already in place.

### DIFF
--- a/index.js
+++ b/index.js
@@ -525,7 +525,7 @@ Command.prototype.executeSubCommand = function(argv, args, unknown) {
   var proc;
   if (process.platform !== 'win32') {
     if (isExplicitJS) {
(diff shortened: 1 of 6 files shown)


rsandor:
:+1:

zhiyelee:
LGTM

SomeKittens:
Oh hay guise!

Yup, looks great.

KyleAMathews:
Oh nice. This might the issue some of my users are running into? https://github.com/gatsbyjs/gatsby/issues/80

Issue #330, opened 2015-01-12
incomplete JSDoc comments


DarkPark:
in some places there are some problems preventing IDE from validating code successfully
for example `Command.prototype.option`
this function has two last arguments as optional but JSDoc comments don't show this

```
@param {Function|Mixed} fn or default
```

but should be

```
@param {Function|Mixed} [fn] or default
```

or even better

```
@param {Function|*} [fn] or default
```

zxqfox:
Probably a big part of mistakes like this can be found automatically with jscs-jsdoc.

vanesyan:
Merged

PR #504, opened 2016-02-17
#330 make .option defaultValue and fn and .version flag optional

twk-b:
#330 make option defaultValue and fn and .version flag optional according to JSDoc's to resolve  IDE reporting wrong argument count number errors.

https://github.com/tj/commander.js/issues/330


GM-Alex:
Please merge it for the next version, should not affect the program logic and is a huge improvement if you are using an IDE.

PR #510, opened 2016-03-17
The unknownOption method is now called with the array of unknown options...

nzhenry:
…as the argument, instead of just the first item in the array.

This makes it more flexible for anyone who wants to override the unknownOption method. They will be able to see the full list of unknown options.

### DIFF
--- a/index.js
+++ b/index.js
@@ -271,7 +271,7 @@ Command.prototype.action = function(fn) {
     // die, unless someone asked for help, in which case we give it
     // to them, and then we die.
     if (parsed.unknown.length > 0) {
-      self.unknownOption(parsed.unknown[0]);
+      self.unknownOption(parsed.unknown);
     }
 
     // Leftover arguments need to be pushed back. Fixes issue #56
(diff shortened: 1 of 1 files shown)


index.js
```diff
@@ -781,10 +781,10 @@ Command.prototype.optionMissingArgument = function(option, flag) {
  * @api private
  */
 
-Command.prototype.unknownOption = function(flag) {
+Command.prototype.unknownOption = function(unknown) {
```
vanesyan: So, if the method now accepts array of unknown options, that it should be named as `unknownOptions` etc.

index.js
```diff
@@ -781,10 +781,10 @@ Command.prototype.optionMissingArgument = function(option, flag) {
...
   if (this._allowUnknownOption) return;
   console.error();
-  console.error("  error: unknown option `%s'", flag);
+  console.error("  error: unknown option `%s'", unknown[0]);
```
vanesyan: We can inform about all unknown flags, instead of the first one.

index.js
vanesyan: So, if we change the function name, it'll be considered as breaking change, and should we deferred until the next major release.

shadowspawn:
The `unknownOption` method is listed as private, so not appropriate to override.

PR #513, opened 2016-03-25
Remove dependancy on graceful read link

ShaneQful:
Remove the dependency on [graceful-readlink](https://github.com/zhiyelee/graceful-readlink) to make sure the removal of such a small dependency can't break commander.js.

### DIFF
--- a/index.js
+++ b/index.js
@@ -4,7 +4,6 @@
 
 var EventEmitter = require('events').EventEmitter;
 var spawn = require('child_process').spawn;
-var readlink = require('graceful-readlink').readlinkSync;
 var path = require('path');
 var dirname = path.dirname;
 var basename = path.basename;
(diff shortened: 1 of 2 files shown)


zhiyelee:
thanks, but that's not a problem to me.

zhiyelee:
https://github.com/sindresorhus/ama/issues/10#issuecomment-117766328

ShaneQful:
I can certainly see some of the points made in the comment but commander.js doesn't rely on loads of little modules. It's commander.js's only dependency, surely it's worth having one if statement to have no non dev dependencies. 

I also realize that this is your module we are discussing and you've been the main contributor to commander.js since @tj stopped committing two years ago so commander.js wouldn't be in great shape if you decided to stop contributing or pull your modules like what happened with Azer a few days ago.

Issue #404, opened 2015-05-16
argument 'once' (and probably others) cause naming collision


hotaru355:
Just want to give some feedback here to help others avoid a little pitfall: you cannot use `--once` as an argument, probably because the name collides with some internals of commander. Interestingly, the `once` argument will revert the expected behavior: `myprog --once` will set `commander.once=undefined`, so a falsy value, while running `myprog` will set it to a function as expected. Maybe you guys can fix the documentation to mention these reserved arguments, so others can save their time. Thanks!

tonylukasavage:
There's a handful of these since when you create and parse options, they are placed directly on the `program` object, which an [EventEmitter](https://nodejs.org/api/events.html). This means you'd also collide with anything on the `Command` object/prototype contained with commander.js. Frankly, there's lots of them and you'd have to check the source to know exactly which ones, though some of the ones that are likely to cause collisions are `action`, `option`, `command`, `opts`, and JS internals like `apply` and `call`.

A while ago I submitted #262 that added the `opts()` function to return the set of options defined on the `program` object in order to return options cleanly and reduce the number of possible collisions. This was only a first step though. The change that would reduce the collisions for good would be then no longer writing the options back onto the `program` object after parsing. This, IMO, is unlikely to happen since it would be a major breaking change, fundamentally changing how commander is used.

As far as the docs, I'm sure if someone submits an accurate PR I'm sure it would be considered for a **caveats** section or something similar.

tonylukasavage:
So I hacked together a script that will list every possible collision (I think). Note that these are _possible_ collisions. For example, `Function.name` is never used in commander.js, so while `name` is on the list, it won't be a problem. If `Function.name` _is_ ever used, it would then become a real conflict, not just a possible one.

Got the `getAllPropertyNames` function on this [SO post](http://stackoverflow.com/questions/8024149/is-it-possible-to-get-the-non-enumerable-inherited-property-names-of-an-object)

```js
function getAllPropertyNames( obj ) {
    var props = [];
    do {
        props= props.concat(Object.getOwnPropertyNames( obj ));
    } while ( obj = Object.getPrototypeOf( obj ) );
    return props;
}

function unique(val, index, array) { 
    return array.indexOf(val) === index;
}

var program = require('commander');
program.parse(['node','test']);
getAllPropertyNames(program.Command).concat(getAllPropertyNames(program)).sort().filter(unique);
```

And right now, you'll come out with a list like this:

```js
[ 'Command',
  'Option',
[...]

PR #515, opened 2016-03-25
Data overload

ericuldall:
New features allow the following:

```
#!/usr/bin/env node

program
  .version('0.0.1')
  .description('Test')
  .usage('[options] string')
  .option('-t, --type [value]', 'Type')
  .option('-f, --format', 'Format')
  .parse(process.argv);

var type = program.get('type'); //if exists return type, else returns false
var options = program.get(); //return all options, or empty object

var has = program.has('type', function(type){
    console.log(type); //will return if type exists
});
console.log(has); //boolean

var has_all = program.has_all(['type', 'format'], function(type, format){
    console.log(type, format); //will return if both exist
});
console.log(has_all); //boolean

}
```


zhiyelee:
hold on this.  New added `get` `has` will overwrite custom command with same name. https://github.com/tj/commander.js/issues/404

ericuldall:
So commands will have to be moved out of the objects main scope as well, for this to be viable?

ericuldall:
UPDATE: I'm not seeing any conflict with custom commands. Can you provide an example of this case?

ericuldall:
I've tested this with the following scenario:

cmd: mycmd
subcmd: mycmd-get
```
mycmd get -y
```
works as expected. Am I missing something?

zhiyelee:
oh, sorry, it's about options. 
> Just want to give some feedback here to help others avoid a little pitfall: you cannot use --once as an argument, probably because the name collides with some internals of commander. Interestingly, the once argument will revert the expected behavior: myprog --once will set commander.once=undefined, so a falsy value, while running myprog will set it to a function as expected. 

I don't want to introduce more these cases until we fix this issue.

ericuldall:
I think my pull request fixes that issue.

shadowspawn:
I am still investigating options. For now, I'll note that `get()` and `has()` are similar in name and purpose to the new `Map`object, and should probably behave in similar way.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map

shadowspawn:
I am closing this in favour of #951, which builds on work in this Pull Request and aims for backwards compatibility.
 
shadowspawn:
I have opened a new Pull Request which allows storing option values separately rather than as command properties (access using .opts()), and passes the options (rather than the command) to the action handler. Opt-in using

```
program.configureCommand({ modern: true });
```

See #1102

PR #519, opened 2016-04-04
added keyword commander for better search result on npm

ycmjason:
Well, just it irritates me when I don't see commander to be the first result when I search "commander" on npm. hope this helps?


zhiyelee:
thanks!

PR #525, opened 2016-04-20
Allow literal '--' to be passed-through as an argument

JustATrick:


zhiyelee:
thanks, but we want to stick to the old manner.

fyi.
>  a double dash (--) is used in bash built-in commands and many other commands to signify the end of command options, after which only positional parameters are accepted.

http://unix.stackexchange.com/a/11382/87921

JustATrick:
Thanks for your attention @zhiyelee, but I think maybe I didn't explain my change well enough.

I want exactly the behaviour as explained in the stackoverflow link. I noticed that the current code has a problem when I give a set of arguments where double-dash appears more than once. I expected the first double-dash to mark the start of positional parameters, and then all subsequent double-dashes to be treated as positional parameters. However, the current code removes the subsequent double-dashes from the positional parameters.

My change ensures that, once the first double-dash is detected, any other double-dashes are passed along as positional parameters. It still allows the first double-dash to mark the end of command options.

zhiyelee:
@JustATrick  may bad, you are correct.  Could you double check the travis tests? thanks

JustATrick:
I don't think the travis test failures are related to this change, as they're failing in the same way on master. See #550 for a possible solution.

PR #535, opened 2016-05-14
flexible filename extensions

### LINKED ISSUE #534: commander expects only .js files
In various places, hardcoded references are made to the `.js` extension.
Perhaps better to use `path.extname(argv[0])` to be able to deal with eg `.coffee` and others

hans-d:
fixes #534


hans-d:
too much hassle to implement the proper bin to execute for files <> .js

PR #536, opened 2016-05-16
added .js extension to the command file path if it doesn't have already

ibrahimduran:
While using sub-sub commands, at the second one it was looking for the command file without .js extension and throwing an ENOENT error.


abetomo:
@ibrahimduran Could you fix it because the test failed?

ibrahimduran:
@abetomo Hey, I've just re-debugged multi-level sub commands on master branch and it's working nicely. This PR is not required anymore.

PR #549, opened 2016-06-30
Force version of npm that understands caret (^) dependencies

JustATrick:
Try to fix travis build for older versions of node

Cribbed from travis-ci/travis-ci#2076


JustATrick:
Hmm, doesn't seem to work for Node 0.6. I'll keep reading.

PR #550, opened 2016-07-01
Try to fix travis build for node versions 0.6 and 0.8

JustATrick:
Limits version of should to be <9.0.0. Version 9.0.0 and above
use caret (^) dependencies, which are not understood by the versions
of npm available on old versions of node

### DIFF
--- a/package.json
+++ b/package.json
@@ -15,7 +15,7 @@
     "url": "https://github.com/tj/commander.js.git"
   },
   "devDependencies": {
-    "should": ">= 0.0.1",
+    "should": ">= 0.0.1 <9.0.0",
     "sinon": ">=1.17.1"
   },
   "scripts": {


JustATrick:
On 29th May a new version (v9.0.0) of should was released. It specifies its dependencies using caret (^) syntax, which are not understood by the versions of npm that are installed by default on node versions 0.6 and 0.8, and so all subsequent travis-ci builds have failed.

I first tried to install newer versions of npm as suggested at travis-ci/travis-ci#2076. That worked for node 0.8 but not 0.6. On 0.6 a sufficiently new version of npm fails to install.

Instead, this 'fixes' the problem by preventing the use of should version 9.0.0 or higher. I'm not sure if this is the best way forward, but wanted to document what I'd found.

zhiyelee:
Sounds good. I think we need to remove v0.6 and v0.8 in next major version.

zhiyelee:
thanks @JustATrick

PR #556, opened 2016-08-02
use current node

lzilioli:
I'm running an app as that I wrote with commander.js (using the sub command feature)

I'm using nvm and running a subcommnd as a LaunchAgent through osx' launchctrl. 

Under this configuration, I had an issue where commander ended up using the node on my system path instead of the node that my main binary script was invoked with, and that version of node couldn't find my globally installed package.

This commit fixes it by using the first argument that was passed to the node process, which is node's process.execPath https://nodejs.org/docs/latest/api/process.html#process_process_execpath
https://nodejs.org/docs/latest/api/process.html#process_process_argv

This resolves my issue

### DIFF
--- a/index.js
+++ b/index.js
@@ -542,7 +542,7 @@ Command.prototype.executeSubCommand = function(argv, args, unknown) {
       // add executable arguments to spawn
       args = (process.execArgv || []).concat(args);
 
-      proc = spawn('node', args, { stdio: 'inherit', customFds: [0, 1, 2] });
+      proc = spawn(process.argv[0], args, { stdio: 'inherit', customFds: [0, 1, 2] });
     } else {
       proc = spawn(bin, args, { stdio: 'inherit', customFds: [0, 1, 2] });
     }


lzilioli:
@tj can you consider merging this into the package?

lzilioli:
@tj any thoughts regarding this fix?

tj:
oops sorry, maybe we need more maintainers here, I don't touch node stuff anymore

zhiyelee:
LGTM, thanks @lzilioli 
Could you add some tests?

abetomo:
@lzilioli I want to merge. Could you add a test?

lzilioli:
@abetomo @zhiyelee its not exactly clear to me how I would write a test for this change, wouldn't I need to hook into the spawning process?

vanesyan:
I have no idea how to write tests to cover the changes. friendly cc @zhiyelee @abetomo

abetomo:
Let's merge

Issue #553, opened 2016-07-04
Command log + option log


Unitech:
When declaring a command with a defined name and declaring option with the same name, it does not works:

``` javascript
var program = require('commander');

program.version('1.0.0')
  .option('--log [path]', 'log path')

program.command('logs')
  .alias('log')
  .action(function() {
    console.log(program.log);
  });

program.parse(process.argv);
```

``` bash
$ node test.js log --log 'test'
```

It throws:

```
/home/unitech/keymetrics/pm2/node_modules/commander/index.js:298
      args.push(self);
           ^

TypeError: args.push is not a function
    at Command.listener (/home/unitech/keymetrics/pm2/node_modules/commander/index.js:298:12)
    at emitOne (events.js:82:20)
    at Command.emit (events.js:169:7)
    at Command.parseOptions (/home/unitech/keymetrics/pm2/node_modules/commander/index.js:699:14)
    at Command.parse (/home/unitech/keymetrics/pm2/node_modules/commander/index.js:455:21)
    at Object.<anonymous> (/home/unitech/keymetrics/pm2/test.js:13:9)
    at Module._compile (module.js:409:26)
    at Object.Module._extensions..js (module.js:416:10)
    at Module.load (module.js:343:32)
    at Function.Module._load (module.js:300:12)
```

zhiyelee:
I will check this @Unitech 
Btw, you can send a PR if you find a solution. Thanks!

shadowspawn:
I was not able to reproduce this issue.

Feel free to open a new issue if it comes up again, with new information and renewed interest.
 
Thank you  for your contributions.

PR #557, opened 2016-08-04
Add a prefix to the event name of OptionClass

rike422:
Fix this issue
https://github.com/tj/commander.js/issues/553


rike422:
this pull-request is duplicates of #494

Issue #540, opened 2016-05-18
Max length of help output?


endquote:
I have a tool with a fair number of arguments and subcommands, and the help is getting cut off.

![screenshot 2016-05-18 11 19 11](https://cloud.githubusercontent.com/assets/330010/15370150/662ceb60-1cea-11e6-88fa-6ac134755bc2.png)

https://github.com/endquote/laundry

zhiyelee:
in node v6? https://github.com/tj/commander.js/issues/530

endquote:
6.1.0...

derek-ellison:
A good work around for this, although time consuming, is to build your own help menu using the program.on('--help') functionality. It's time consuming as I said, but you do have a lot more flexibility and you avoid this bug with the newer versions of Node.

```
program.on('--help', function() {
      console.log('--stuff-here       things');
});
```

creeperyang:
Cut off +1, `node v6.2` and `command v2.9.0`.

<img width="601" alt="2016-06-14 4 04 12" src="https://cloud.githubusercontent.com/assets/8046480/16035432/af5a056a-3249-11e6-9a75-0114934252be.png">

GCheung55:
Cut off here as well. `node v6.1.0` and `commander v2.9.0`.

tyge68:
I have the same issue with node 6.2.0 and commander 2.9.0

tyge68:
It appear that the cut off is due to the program.exit(0) which doesn't give time for the stdout.write operation to complete. Similar issue have been reported. So most likely the implementation should avoid to force exit, instead it should end in a normal flow without this call.

shadowspawn:
The comments in #540 suggest this was fixed in node 6.6.0.

This issue has not had any activity in over six months. It isn't likely to get acted on due to this report.

Feel free to open a new issue if it comes up again, with new information and renewed interest.
 
Thank you  for your contributions.

shadowspawn:
Future readers: this may still be an issue on Windows TTY which are asynchronous according to:
- https://nodejs.org/api/process.html#process_a_note_on_process_i_o

Please open a new issue if you are seeing truncated help.

PR #572, opened 2016-09-18
#540 fix with sync write in stdout

tyge68:
@tj that should help to resolve the issue #540 can you please confirm and release it ? that would be really helpful.

### DIFF
--- a/index.js
+++ b/index.js
@@ -1039,7 +1039,8 @@ Command.prototype.outputHelp = function(cb) {
       return passthru;
     }
   }
-  process.stdout.write(cb(this.helpInformation()));
+  fs.writeSync(1, cb(this.helpInformation()));
+  fs.fsyncSync(1);
   this.emit('--help');
 };
 


tyge68:
@tj forget it I was still running with 6.2.0 on mac os, updating to 6.6.0 resolved the issue #540 as it was suggested (6.2.1  contained the fixed I supposed). So I will close this one.

tmarciszewski:
Is there a chance to introduce this change for previous versions of Node?

tyge68:
Sure it can be used for older version of Node. Do you want me to reopen the
pull request ?

tmarciszewski:
I'm not sure what's commander's approach to supporting older versions of node. I was wondering if it should be merged for older versions of node to work without problems.
Our project will probably upgrade from 6.0.0 to 6.6.0 in short time so it's not critically important for us, but others could run into this problem.

zhiyelee:
Could you also edit the travis.yml and run the tests?

tyge68:
The test will not help because it does write in the stdout and directly
exit. So it would exit the test as well abruptly. It can only work with a
manual test from what I could see.

tyge68:
Actually a test is possible , it would be simply about creating a command with many options so that the result help text is large enough to fill the the buffer (expecially on Mac OS this happens quickly). Then the test would run that command as an external process (child_process), and read the output of that external process, then compare it to expected value, similar to the tests already existing.

shadowspawn:

PR #503, opened 2016-02-11
For a required option, when even option flag is not given throw error.

snow01:
For a required option, when even option flag is not given throw error. 
Fixes tests accordingly.

PR #573, opened 2016-09-20
Revert "For a required option, when even option flag is not given throw error."

zhiyelee:
Reverts tj/commander.js#503


faceleg:
Why was this rolled back?

Kurt-von-Laven:
I would guess because it breaks existing dependencies. Is there a plan to replace this feature in a more backwards-compatible manner?

PR #583, opened 2016-10-13
implement ability to use negative numbers as option values 

iamjochem:
this is based on an earlier fix by @qthru: 

> relevant commit: https://github.com/qthru/commander.js/commit/db2c69e353cd80a15f385184572061ad126db43f 
> relevant PR: https://github.com/tj/commander.js/pull/134 

@SomeKittens closed PR#134 because the it was no longer up to date with origin master, this PR is (for now) update with master - hoping someone can/will now merge this.


shadowspawn:
I have not worked through the open issues and pull requests around negative numbers as options, so not ready to review this yet, but keeping open for now as looks interesting.

However, I note there is a `parseBool` which has been added and not used.

iamjochem:
hi there, this was from a long time ago, not sure why the `parseBool` is in there, I think it was an extra idea I was playing with and the function def accidently got included. it's not doing any harm (and I think the idea is fairly self-explanatory ... it can be stripped if you want.

shadowspawn:
(Thanks for comment. I am doing triage on some old issues and pull requests, so not expecting everyone to remember the details! 😄  )

shadowspawn:
I am not able to reproduce the issue this Pull Request is fixing. The current behaviour is when an option has a required value, the following argument is used whether or not it starts with a dash. I tested both short and long options in a small program. (I have not looked for a commit that changed the behaviour.)

Feel free to add new information if you are still seeing the issue.
 
shadowspawn:
Update: I think negative numbers are still a problem for options which have an _optional_ value, i.e. which are specified with `[value]` rather than `<value>`. However, this Pull Request added unit tests for the `<value>` case which I believe is already working.

I see some confusion in comments in other issues over whether "required" means the option is required or the value is required. In the first instance I would like to change the README for options for make it clearer that an option with a required value is the normal case. Then if negative numbers are still an issue, might come back to reconsider the merits of this Pull Request!

iamjochem:
I confused `[value]` and `<value>` between the code I originally had the problem with and the unit-test ... I am pretty sure this PR does fix things for the `[value]` case (I definitely deployed my fork to get round the problem I was having) ... it's the unit-test that is testing for the wrong case ... I just did an adhoc test with the latest version and get an error when using `[value]`  (`error: unknown option `-1'`).

PR #595, opened 2016-12-07
Resolve symlinks recursively; Fix #588

### LINKED ISSUE #588: sub-command can not found when `ln` link file again
If we make a command `c-test`,and `c-test` has a sub-command `c-test init`.after we use `npm/yarn` install `c-test` command, we  excute`ln` which c-test` test` and excute command `test init`, we will get  a `test-init` not found error.

h2so5:
Fix sub-command not working when a program is executed via nested symlinks.
This problem could occur when you install a global command with `yarn global add`.


shadowspawn:
This has been fixed by #935.

Issue #395, opened 2015-04-15
short switches don't take numeric arguments


Sequoia:
in minimist (or `tail` for that matter), `-n10` sets `n` (or the switch it aliases) to `10`.

A similar call with `commander` sets `n` (or the switch it aliases) to `-1`. This seems to mean a minimist to commander rewrite forces a CLI change, i.e. a cronjob that previously called `foo -n10` will stop functioning properly if `foo` is upgraded from the minimist version to the commander version.

Is there a workaround? Am I mistaken?

TheRoSS:
The same issue.
commander.option("-c, --count <positiveInteger>", "...")
with call:
node myCoolScript -c2
gives:
commander.count = -2

TheRoSS:
This is a bug in the Command.prototype.normalize method
It is him who splits -c2345 option with parameter to set of no parameter options -c -2 -3 -4 -5
Then parser behaves correctly and takes "correct" parameter for option "c"

shadowspawn:
This issue will be partially  resolved when v3.0.0 is released. Available now as a prerelease. See #1001

Support for passing option directly with short flags is added for recognised options (but not for options for subcommands with executable file as handler.)

shadowspawn:
Support for directly appending values to short options has shipped in v3 (thanks @TheRoSS): https://github.com/tj/commander.js/releases/tag/v3.0.0

As per previous comment, currently only working for action handler based subcommands.

PR #599, opened 2016-12-28
#395  fix, short options do not understand adjacent values

TheRoSS:
Example of the bug:

commander.option("-c <value>")
console.log("C:", commander.C)

$ node program -c234
C: -2

### DIFF
--- a/index.js
+++ b/index.js
@@ -452,7 +452,8 @@ Command.prototype.parse = function(argv) {
   }
 
   // process argv
-  var parsed = this.parseOptions(this.normalize(argv.slice(2)));
+  var normalized = this.normalize(argv.slice(2));
+  var parsed = this.parseOptions(normalized);
   var args = this.args = parsed.args;
 
   var result = this.parseArgs(this.args, parsed.unknown);
@@ -579,7 +580,9 @@ Command.prototype.normalize = function(args) {
   var ret = []
     , arg
     , lastOpt
-    , index;
+    , index
+    ,short
+    ,opt;
 
   for (var i = 0, len = args.length; i < len; ++i) {
     arg = args[i];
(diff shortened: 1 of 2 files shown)


shadowspawn:
It looks like this pull request works with `-a234 => a=234` for an option taking a value, and `-xyz => -x -y -z`  if x is a boolean flag, but leaves some combination cases like `-axz 3` or `-xa234`. 

It would be nice if there was a standard we could refer to for short flag handling. Should we only consider numbers?

Keeping pull request open for now.

shadowspawn:
Link to issue: #395

shadowspawn:
Found a couple of posix references which align with suggested pull request:

1) https://www.gnu.org/software/libc/manual/html_node/Argument-Syntax.html

> Multiple options may follow a hyphen delimiter in a single token if the options do not take arguments. Thus, ‘-abc’ is equivalent to ‘-a -b -c’.

2) https://www.gnu.org/software/libc/manual/html_node/Example-of-Getopt.html#Example-of-Getopt

```bash
% testopt -a -b
aflag = 1, bflag = 1, cvalue = (null)

% testopt -ab
aflag = 1, bflag = 1, cvalue = (null)

% testopt -c foo
aflag = 0, bflag = 0, cvalue = foo

% testopt -cfoo
aflag = 0, bflag = 0, cvalue = foo
```

TheRoSS:
Good day. I am rather busy just now. But can participate in this issue in the end of May

index.js
```diff
@@ -579,7 +580,9 @@ Command.prototype.normalize = function(args) {
   var ret = []
     , arg
     , lastOpt
-    , index;
+    , index
+    ,short
```
shadowspawn:
```suggestion
    , short
```

index.js
```diff
@@ -579,7 +580,9 @@ Command.prototype.normalize = function(args) {
...
+    ,opt;
```
shadowspawn:
```suggestion
    , opt;
```

shadowspawn:
Merging into `release/v3.0.0` branch

shadowspawn:
(I deleted some requested style changes that turned out to be stale, format on master had changed.)

shadowspawn:
I made some minor changes, and already had one approval. Merged into v3 and will be included in that release. Closing to make it clear that should not be merged into master.

shadowspawn:
Update: I am working my way through the parsing process looking into another bug. I suspect this fix only works for action based commands.

shadowspawn:

shadowspawn:

PR #601, opened 2017-01-07
Adding support for required arguments.

jimwang:
This checks to make sure that a required argument is specified with either a default or a non-null value passed in.

Updated unit tests and added two new unit tests to test conditions.


shadowspawn:
I think this Pull Request makes all options with required values be mandatory options that must be specified? I consider those two independent things, and this is not the desired behaviour.

See https://github.com/tj/commander.js/issues/230#issuecomment-478181355 for a long description.


This Pull Request has not had any activity in over six months.

PR #612, opened 2017-03-21
Pass un-normalized unknown options to sub-commands

joshbetz:
I want to access the un-normalized unknown options (via `process.argv`) in a sub-command. This should only effect commands that use rawArgs or process.argv directly as the argument are still normalized within the subcommand.


shadowspawn:
Without examples, not clear what this is intended to achieve.

Issue #461, opened 2015-10-22
Default subcommand with options


oscarotero:
Hi, I want to use a subcommand as default, and this subcommands has options. Something like this:

``` js
//app.js

program
    .command('run', 'Execute stuff', {isDefault: true})
    .command('list', 'List stuff')
    .parse(process.argv);
```

``` js
// app-run.js

program
    .option('-c, --config <file>', 'use a config file')
    .parse(process.argv);
```

This command works fine:

```
app run -c config.json
```

And this as well (executes the default subcommand):

```
app
```

but on execute the default subcommand with options: 

```
app -c config.json
```

I have the followin error:

```
  error: unknown option `-c'
```

natew:
Big +1 on this

My workaround is not to use isDefault and just coerce argv, for now:

```js
// make `flint run` default command
var flintIndex = getFlintIndex()

 // find where index of flint is
function getFlintIndex() {
  let index = 0
  for (let arg of process.argv) {
    if (arg.indexOf('flint') > 0)
      return index

    index++
  }
}

// make sure flags are still passed to `flint run`
var firstFlag = process.argv[flintIndex + 1]

if (flintIndex === process.argv.length - 1 || (firstFlag && firstFlag[0] === '-')) {
  process.flintArgs = [].concat(process.argv);
  process.flintArgs.splice(flintIndex + 1, 0, 'run');
}
```

SomeKittens:
Not sure how we'd accomplish this - you want us to skip all params if there's a default subcommand?

natew:
No, it's just to pass those options down to the subcommand.

So `flint run --cached` works, and if `run` is default, then `flint --cached` would work the same.

SomeKittens:
My issue is what happens if the root program (`app` in this case) also has a `-c` option.

natew:
Ah, so could we have an option `isDefault: true, passOptions: true`

alexcanessa:
Any updates on this one?

shadowspawn:
Took me a while to get on board with this one, but I now agree this is the correct behaviour and interested in making it happen.

The PR in flight is #614, and an easy code change. :-)

shadowspawn:
There are improvements on the develop branch for the next release (#614 and #1047).

shadowspawn:
v4.0.0-0 prerelease published: #1067

shadowspawn:
v4.0.0 has been released which hopefully resolves this issue

PR #614, opened 2017-04-04
Inherit arguments to default subcommand

Justinidlerz:
You can use the option on the subcommand.Like so:  
### pm.js
```javascript
program
  .command('default', 'default command', { noHelp: true, isDefault: true });

program
  .command('server')
  .option('-p --port <port>', 'port', /^\d+$/, 3000)
   .action(port => {
     // start a server with the option port
   })
  .parse(process.argv);
```
### pm-default.js
```javascript
#!/usr/bin/env node

program
    .option('-t --test <number>', 'Test inherit subcommand option', /^\d+$/, 500)
    .parse(process.argv);

console.log('default');
console.log(program.test);
```

## RUN
```command
./pm.js -t 800 // echo default 800
```

### DIFF
--- a/examples/pm
+++ b/examples/pm
@@ -8,6 +8,7 @@ program
   .command('search [query]', 'search with optional query').alias('s')
   .command('list', 'list packages installed')
   .command('publish', 'publish the package').alias('p')
+  .command('default', 'default command', { noHelp: true, isDefault: true })
   .parse(process.argv);
 
 // here .command() is invoked with a description,
@@ -25,3 +26,4 @@ program
 //   ./examples/pm install -h
 //   ./examples/pm install foo bar baz
 //   ./examples/pm install foo bar baz --force
+//   ./examples/pm -t 800
--- a/examples/pm-default.js
+++ b/examples/pm-default.js
@@ -1,5 +1,5 @@
 #!/usr/bin/env node
-var program = require('../../');
+var program = require('..');
 
 program
     .option('-t --test <number>', 'Test inherit subcommand option', /^\d+$/, 500)
--- a/index.js
+++ b/index.js
@@ -634,7 +634,8 @@ Command.prototype.parseArgs = function(args, unknown) {
 
     // If there were no args and we have unknown options,
     // then they are extraneous and we need to error.
-    if (unknown.length > 0) {
+    // Extend unknown args to defaultExecutable
+    if (unknown.length > 0 && !this.defaultExecutable) {
       this.unknownOption(unknown[0]);
     }
   }
--- a/index.js
+++ b/index.js
@@ -634,7 +634,7 @@ Command.prototype.parseArgs = function(args, unknown) {
 
     // If there were no args and we have unknown options,
     // then they are extraneous and we need to error.
-    // Extend unknown args to defaultExecutable
+    // Inherit unknown args to 'defaultExecutable'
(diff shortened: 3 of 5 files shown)


Justinidlerz:
Tests unhappy when node version 0.6 or 0.8, 
but my update is support this version, 
update package engines: { node: 0.10 } ?

shadowspawn:
I was bit concerned whether this was a good idea, but I see it requested in #461, and I checked `yargs` and it allows options for default command, so will be considered further.

shadowspawn:
I have a concern that unrecognised commands will get passed to default hander as arguments. For example with commands `server` and `default`:

```bash
# oops, server typo calls default command with argument servers
./pm.js servers -p 90
```

Basically the fallback error handling shifts from the outer program to the default command. Heavier weight change than might first appear.

shadowspawn:
I still have mixed feelings about this one. I am leaving it open but not considering it for v3, so no action likely for a while.

shadowspawn:
I was concerned about what I thought was changing behaviour to call default subcommand when an option was specified, but was looking at that code recently and saw that this is exactly what happens if an unrecognised argument is included on the command line such as when the default command takes an argument.

Happier that this change is consistent with existing behaviour and will be taking another look after v3.

shadowspawn:
@Justinidlerz 
Are you interested in reworking this, or should I take it over?

shadowspawn:
Stripped down `pm.js`

```js

program
  .command('default', 'default command', { noHelp: true, isDefault: true })
  .parse(process.argv);
```

`pm-default.js`

```js
#!/usr/bin/env node
const util = require('util')

console.log('Called default subcommand with')
console.log(process.argv.slice(2));

program
  .arguments('<file>')
    .option('-a, --all')
    .action((file) => {
      console.log(`File is ${file}`);
    })
    .parse(process.argv);

console.log(`option all is ${program.all}`);
```

Behaviour before fix:

```bash
$ node pm.js
Called default subcommand with
[]
option all is undefined
$ node pm.js x
Called default subcommand with
[ 'x', '[object Object]' ]
File is x
option all is undefined
$ node pm.js --all
error: unknown option '--all'
$ node pm.js x --all
Called default subcommand with
[ 'x', '[object Object]', '--all' ]
File is x
option all is true
```

After fix:


```bash
$ node pm.js
Called default subcommand with
[]
option all is undefined
$ node pm.js x
Called default subcommand with
[ 'x', '[object Object]' ]
File is x
option all is undefined
$ node pm.js --all
Called default subcommand with
[ '--all' ]
option all is true
$ node pm.js x --all
Called default subcommand with
[ 'x', '[object Object]', '--all' ]
File is x
option all is true
```

(I am annoyed by those `[object Object` in the args!)

Justinidlerz:
@shadowspawn Sorry, it's a loooooong time PR, I forgot the contents. Let me see again it to recall.

shadowspawn:
I'm going to start work on this for 4.0.0 in a new PR. I'll add you (@Justinidlerz) as co-author on the commit.

shadowspawn:
Closing in favour of #1047 which finishes this work.

shadowspawn:
#1047 has been merged to develop and will be released in v4.0.0. 

shadowspawn:
v4.0.0 has been released. 

PR #619, opened 2017-04-09
On SemVer you do not start from 0.0.1

Nepomuceno:
You should not start one application with a fix but with an implementation of a new feature. this is why I think that this approach it is better than 0.0.1

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -28,7 +28,7 @@
 var program = require('commander');
 
 program
-  .version('0.0.1')
+  .version('0.1.0')
   .option('-p, --peppers', 'Add peppers')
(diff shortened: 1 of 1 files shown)


scott-joe:
@Nepomuceno makes total sense. Never thought of it that way. Thanks

iowillhoit:
Good call, here is the same thing suggested in the Semver 2.0.0 Spec: [How should I deal with revisions in the 0.y.z initial development phase?](http://semver.org/#how-should-i-deal-with-revisions-in-the-0yz-initial-development-phase)

zhiyelee:
Thanks, make sense

PR #621, opened 2017-04-11
Added color output of main command

Piterden:
image


vanesyan:
Imho, a bit unnecessary, plus not all tests are passed. Feel free to open a new pull request, if this one was closed incorrectly

Piterden:
Maybe you are right, it's not such necessary, but it looks much clearer for interaction.

I suppose, test errors was causing by strings of regular expression.

I will make another PR, thanks, when I find some time...

PR #632, opened 2017-05-25
feature: support for signals to subcommands

ThomasSimoens:
- Supported signals: 'SIGUSR1', 'SIGUSR2', 'SIGTERM', 'SIGINT' & 'SIGHUP'
- Also included a fix for the builds in node 0.6 & 0.8

A merge is not required immediately, maybe we can review how the tests are written? I would like to avoid the timeouts, but haven't really found a good solution for it.


call-a3:
@abetomo any guidelines on what we still can/must do to get this into master?

We're maintaining our own fork for now at Skedify because we need this feature, but we'd like to be able to keep tracking the "official" version without having to reapply our patch each time a new version comes out...

abetomo:
@ThomasSimoens @call-a3 
Thank you PR!
Sorry for being late.

@vanesyan
If there is no problem merge please.

vanesyan:
What I really dislike about pr, that it use timer with magic number of time out. Are there any other ways to handle it?

abetomo:
(Note: Similar fix #411)

vanesyan:
Sorry, my bad, miss understood the pr implementation, timers are in tests

ThomasSimoens:
Thanks guys, greatly appreciate it!

@ValentinTrinque I hated the timeout approach as well, but couldn't find another way. Don't know if there's an event based approach possible for these kind of things.

ValentinTrinque:
(( @ThomasSimoens, I guess you wanted to mention @vanesyan instead ;) ))

PR #639, opened 2017-06-19
Update version to be tested

abetomo:
- Do not test older versions
- Add new version

https://github.com/nodejs/LTS/


vanesyan:
Maybe we can also drop support for 0.10, 0.12 and I.js, as there're seem also deprecated.

vanesyan:
Cool

PR #641, opened 2017-06-19
version bump v2.9.1

abetomo:


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "commander",
-  "version": "2.9.0",
+  "version": "2.9.1",
```
vanesyan: Shouldn't we increase minor version instead of patch, because of "Allow literal '--' to be passed-through as an argument" feature?

abetomo: I will rebuild it at 2.10.0!

PR #646, opened 2017-06-24
feat(typings): add typings to support TypeScript

alan-agius4:
Add typings to support TypeScript

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,7 +23,7 @@
     "index.js"
   ],
   "dependencies": {
-    "@types/node": "^8.0.2"
+    "@types/node": "^7.0.39"
   },
   "devDependencies": {
     "should": "^11.2.1",
--- a/package.json
+++ b/package.json
@@ -16,10 +16,12 @@
   },
   "devDependencies": {
     "should": ">= 0.0.1 <9.0.0",
-    "sinon": ">=1.17.1"
+    "sinon": ">=1.17.1",
+    "typescript": "^2.3.4"
   },
   "scripts": {
-    "test": "make test"
+    "test": "make test && npm run test-typings",
+    "test-typings": "node_modules/typescript/bin/tsc -p tsconfig.json"
   },
   "main": "index",
   "engines": {
@@ -30,5 +32,6 @@
   ],
   "dependencies": {
     "graceful-readlink": ">= 1.0.0"
-  }
+  },
+  "typings": "typings/index.d.ts"
 }
--- a/package.json
+++ b/package.json
@@ -31,6 +31,7 @@
     "index.js"
   ],
   "dependencies": {
+    "@types/node": "^8.0.2",
     "graceful-readlink": ">= 1.0.0"
   },
   "typings": "typings/index.d.ts"
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -1,0 +1,20 @@
+{
+    "compilerOptions": {
+        "module": "commonjs",
+        "lib": [
+            "es6"
+        ],
+        "noImplicitAny": true,
+        "noImplicitThis": true,
+        "strictNullChecks": false,
+        "types": [
+            "node"
+        ],
+        "noEmit": true,
+        "forceConsistentCasingInFileNames": true
+    },
+    "files": [
+        "typings/index.d.ts",
+        "typings/commander-tests.ts"
+    ]
+}
--- a/typings/commander-tests.ts
+++ b/typings/commander-tests.ts
@@ -1,0 +1,96 @@
+import * as program from './index';
+
+interface ExtendedOptions extends program.CommandOptions {
+    isNew: any;
+}
+
+const commandInstance = new program.Command('-f');
+const optionsInstance = new program.Option('-f');
+
+program
+    .version('0.0.1')
+    .option('-p, --peppers', 'Add peppers')
+    .option('-P, --pineapple', 'Add pineapple')
+    .option('-b, --bbq', 'Add bbq sauce')
+    .option('-c, --cheese [type]', 'Add the specified type of cheese [marble]', 'marble')
+    .parse(process.argv);
+
+console.log('you ordered a pizza with:');
+if (program['peppers']) console.log('  - peppers');
+if (program['pineapple']) console.log('  - pineapple');
+if (program['bbq']) console.log('  - bbq');
+console.log('  - %s cheese', program['cheese']);
+
+function range(val: string) {
+    return val.split('..').map(Number);
+}
+
+function list(val: string) {
+    return val.split(',');
+}
+
+function collect(val: string, memo: string[]) {
+    memo.push(val);
+    return memo;
+}
+
+function increaseVerbosity(v: any, total: number) {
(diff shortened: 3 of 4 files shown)


package.json
```diff
@@ -16,10 +16,12 @@
   },
   "devDependencies": {
     "should": ">= 0.0.1 <9.0.0",
-    "sinon": ">=1.17.1"
+    "sinon": ">=1.17.1",
+    "typescript": "^2.3.4"
```
michael-yx-wu: unlikely to make a different to this PR, but 2.4.1 is available now 😄

alan-agius4:
With the carot it will take 2.4.1 anyways

vanesyan:
Can you resolve the pr on top of master, please :)

alan-agius4:
What do you mean by 'resolve' please? I am current away and ill be able to do it on Friday

vanesyan:
Sorry, mean rebase, mistyped :)

alan-agius4:
Ah okay, will do that on friday :)

alan-agius4:
@vanesyan rebase done :)

package.json
```diff
@@ -28,8 +23,12 @@
     "index.js"
   ],
   "dependencies": {
-    "@types/node": "^8.0.2",
-    "graceful-readlink": ">= 1.0.0"
+    "@types/node": "^8.0.2"
```
vanesyan: Shouldn't this go in peerDependencies?

alan-agius4:
Can go, if you prefer it that way. Though its not neccessary.

Not a bog fan of peers unless it's neccessery.

I'll do it as you prefer though. Just let me know :)

alan-agius4:
The simple reason why its dependency is that it that node is requeried by
commander.js because of the 'eventEmitter'

vanesyan:
Just worry about people that don't need typing for commander, they'll get extra deps.

alan-agius4:
Thats trie, but its a very small deps.

Though i can change it if you wish

vanesyan:
I'm not too familiar with TypeScript, but what do other publishers do?

alan-agius4:
Usefully they are set as dependencies even with the typyings packages
themselves.


https://www.npmjs.com/package/@types/commander

vanesyan:
Sorry me, if it's a bit stupid question, but what do we want to include typings within package, when typings are already released in DefinitelyTyped? :)

alan-agius4:
Having typyings in DefinitlyTyped is a 'last' resort for the community to
have typings of a library that the owner doesn't want to have within his
library. TypeScript recommand having them within the library and not within
DefinitlyTyped.

vanesyan:
Sorry me, but I haven't found such suggestion by TypeScript. Furthermore as I understand they heavily rely on `@types` npm namespace, which from my understanding is referring to DefinitelyTyped.

alan-agius4:
https://github.com/Microsoft/TypeScript-Handbook/blob/master/pages/declaration%20files/Publishing.md

alan-agius4:
@types is indeed for DefinitlyTyped which is used only for non typescript
libaries, as typically typescript libraries always have typing inside the
library itself and now andays even js libraries have typings in them such
as 'moment'

Reason being is for maintability and no need for addition packages.

TypeScript doesn't realy on DefinitlyTyped at all, its the consumer that
realies on them if he needs non TypeScript ready libraries.

vanesyan:
Ok, sorry me again, but previously I was able to use Node.js api out of the box when used typescript last time in the past. But sometimes not so far, I tried it again and now only way to use node.js api was to install 'em from DefinitelyTypes.

michael-yx-wu:
Yes that's because node does not ship with typings by default. Despite adding an extra dependency that's unnecessary for JavaScript projects having a peer dependency of node typings 8+ will give typescript projects targeting node <8 build errors because of unsatisfied peer dependencies. 

Commander seems to work with node 7+ (possibly even earlier versions). I think we should use the typings of the minimum version of node that commander is compatible with and set as a peer dependency.

michael-yx-wu:
Including node typings as a peer dependency is fine since it's a declaration that commander works with a minimum version of node, but you'll run into problems with:

0. projects not using typescript (npm / yarn will warn that the peer dependency is not met and will be annoying)
0. typescript projects relying on an older version of node: I believe commander works with at least node 7+ and possibly earlier versions?

I think we should keep it as an actual dependency to avoid `1.`. We should also use the actual minimum version of node typings that commander is compatible with although `2.` is not an immediate concern unless we change the typings dependency to a peer dependency.

alan-agius4:
So, what do you guys want me to do?

michael-yx-wu:
My preference is to keep it as a dependency but use the version that commander is minimally compatible with.

vanesyan:
Thanks :+1:

Marak:
Hrmm. 

By adding `@types/node` dependency, `commander` project is no longer zero-dependencies.

The total size of the installed code-base seems to be increased by over 10x.

Project was previously `57kb` now is `503kb`. That seems wrong to increase package size by 10x in order to support one specific framework ( TypeScript ).

Not sure what the project maintainers are thinking here. Please review this again @michael-yx-wu.

xavdid:
There's a couple of gotchas happening here:

1. the `package.json` files array doesn't include the `index.d.ts`, so the typings aren't shipped to npm. TS complains when you try to import commander, since the types aren't installed. That's an easy fix. 
2. The more involved piece is the dependencies. The bundle size increase is probably mostly due to the Typescript devDep, which commander doesn't _really_ depend on. The more harmonious way to do this is to include the `index.d.ts` file, but not introduce any dependencies for it. You can include the `/// <reference types="node">` flag to access `EventEmitter` and anything else needed. 
3. Semi-related, after cloning, I couldn't get the type tests to fail. I changed a `.option` line to `.asdf` and tests still passed fine. That could super be a configuration issue on my end, but if they're not actually catching anything, they're adding complexity. It's easy enough to have maintainers edit the `index.d.ts` file whenever there are API changes (usually semver major & minor). 

So! That said, if I were to make a PR now, I'd remove deps, add a reference to the index, and edit files. That should get the bundle back down, because shipped files would go up by exactly 1 and we'd be back to the deps we had before. 

How does that sound?

alan-agius4:
@xavdid the bundle size is still the same considering the fact that TypeScript is only for devDeps, to be honest though, finding solutions to keep the devDeps down by removing TS, for me it's not a good idea as like this you will remove the tests will keeps the sanity of the definition file.

Why is it such a problem to have an extra devDeps which is actual used? Just for an example would you consider removing `sinon`?

xavdid:
I don't care about the bundle size, but other commenters seem to take issue with it. Since the API for a big project like this changes so infrequently, it's pretty easy to keep the typings up-to-date. I don't see a ton of value in the tests, but it's not a big deal. 

That's a good point about the tests, I'm still pretty new with commander. 

Nevertheless, the `files` issue remains.

alan-agius4:
The files issue is fixed
https://github.com/tj/commander.js/pull/725#pullrequestreview-79401278

PR #650, opened 2017-06-26
Removing dependency on graceful-readlink

TheToolbox:
It's almost exactly the same code as written in graceful-readlink, but replaced with a quick one-liner. Hopefully it might increase attractiveness to those that are dependency-averse. Unlikely to be any regressions due to the tiny nature of the change, and tests passing from my point of view.


vanesyan:
Cool, that was the plan.

PR #652, opened 2017-06-28
Fix help section order and padding

### LINKED ISSUE #500: Display description and options before subcommands
See:
https://github.com/tj/commander.js/blob/master/index.js#L1004-L1008

The global description and options are shown _after_ the subcommands list.

ValentinTrinque:
Fixes #500

### DIFF
--- a/examples/pm
+++ b/examples/pm
@@ -4,6 +4,7 @@ var program = require('..');
 
 program
   .version('0.0.1')
+  .description('Fake package manager')
   .command('install [name]', 'install one or more packages').alias('i')
   .command('search [query]', 'search with optional query').alias('s')
   .command('list', 'list packages installed')
--- a/index.js
+++ b/index.js
@@ -1022,9 +1022,9 @@ Command.prototype.helpInformation = function() {
   ];
 
   return usage
-    .concat(cmds)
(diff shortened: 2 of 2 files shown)


ValentinTrinque:
Ping @abetomo

abetomo:
@ValentinTrinque Thank you!

ValentinTrinque:
Thank you for your reactivity :)

abetomo:
@vanesyan Please review.

vanesyan:
Shouldn't commands come first and options after em' (imho, I don't really convenient way)? Otherwise looks good to me

ValentinTrinque:
I've used Git man pages as an example for the ordering. Also, in my opinion, the options applies directly to the executable such as verbosity, or any option to _configure_ the execution. Then, the commands have arguments which can be expressed as plain word `list`, `delete` or as _option-like_ style `-l`, `-d` respectively.

Some may say that is a matter of style / taste, but it _makes sense_ to me. Also, since git sees it this way as well, it should make us confident about this changes.

vanesyan:
Makes sens

vanesyan:
LGTM that way

PR #654, opened 2017-06-29
Pass unknown also as a second argument to the command:* event

jseminck:
Thanks a lot for the amazing library.

I was browsing through the code to find a way to get access to the `unknown` options array that is currently only exposed through the `command:${commandName}` event. But this event also calls `.shift()` on the arguments array, which is not something that we'd want.

Therefore I was wondering, if there is a specific reason that the `unknown` array is not passed as a second argument to the `command:*` event?

I added it there, but had to fix one test as this was passing unknown options and without specifying that unknown options are allowed, the test would crash now.

I also added a test case to actually test that the `unknown` array is being passed as the second argument in both `command:[*|commandName]` events.

### DIFF
--- a/index.js
+++ b/index.js
@@ -626,7 +626,7 @@ Command.prototype.parseArgs = function(args, unknown) {
     if (this.listeners('command:' + name).length) {
       this.emit('command:' + args.shift(), args, unknown);
     } else {
-      this.emit('command:*', args);
+      this.emit('command:*', args, unknown);
     }
   } else {
     outputHelpIfNecessary(this, unknown);
@@ -1127,4 +1127,3 @@ function exists(file) {
     return false;
   }
 }
-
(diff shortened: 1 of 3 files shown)


GreenGremlin:
I was just wondering the same thing. It would be nice to be able to access unknown options from the action as well. This could be done by pushing `unknown` to args before [calling `fn.apply`](https://github.com/jseminck/commander.js/blob/e8472684d3c52417cf560ee5e1eed63cf4dd5252/index.js#L300).

Something like this

``` javascript
Command.prototype.action = function(fn) {
  var self = this;
  var listener = function(args, unknown) {
    ...

    if (self._args.length) {
      args[self._args.length] = self;
    } else {
      args.push(self);
    }
    args.push(unknown); // Adding this line would give the action callback access to the unknown options array as well.

    fn.apply(self, args);
  };
  ...
};
```

Or maybe even expose the unknown options on the commander instance so a user could access them via `program.unknownOptions`.

shadowspawn:
@GreenGremlin I think a method on the commander object would be tidier than pushing arguments into callback.

Issue #645, opened 2017-06-22
-h help flag collides with user defined -h


nickgarvey:
Configuration:
User defines their own flag that uses -h

Expected behavior:
Running tool --help doesn't include -h as short hand for --help

Actual behavior:
-h is included for both --help and the custom flag

e.g.
```
ngarvey@challenger:~/projects/screeps-docker$ docker run tester cli --help

  Usage: cli [options]

  Connect to the CLI interface of the main process.

  Options:

    -h, --help         output usage information
    -p, --port <port>  Default is 21026
    -h, --host <host>  Default is localhost
```

vanesyan:
Duplicate of #400

vanesyan:
My bad, misunderstood

shadowspawn:
If someone wants to have a go at this, there is some code in #658 to reference which is doing the right things, but I had some style concerns. (Optionally could also cope with user defined `--help` although since much less likely than `-h`, it is fine to leave that out of this fix.)

shadowspawn:
The masked help options are no longer displayed from Commander v5.1.

https://github.com/tj/commander.js/releases/tag/v5.1.0

Thank you  for your contributions.

PR #658, opened 2017-07-10
fix help option text for user-defined -h short flags

TDress:
This fixes issue [#645 ](https://github.com/tj/commander.js/issues/645)
if there is a user-defined -h option, the text displayed when running --help should not list a -h short flag for both the help and user-defined options.

### DIFF
--- a/index.js
+++ b/index.js
@@ -932,20 +932,35 @@ Command.prototype.largestOptionLength = function() {
 };
 
 /**
+ * Return `Option` object for help option.  
+ *
+ * @return {Option}
+ * @api private
+ */
+Command.prototype.optionHelp = function() { 
+  // check if there is an overriding user-defined `-h` short flag
+  var shortHOptions = this.options.filter(function(option) { 
+    return option.short === '-h';
+  });
+  var helpFlags = shortHOptions.length ? '--help' : '-h, --help';
+  
+  return new Option(helpFlags, 'output usage information');
+};
+
+/**
  * Return help for options.
  *
  * @return {String}
  * @api private
  */
 
-Command.prototype.optionHelp = function() {
+Command.prototype.optionHelpText = function() {
   var width = this.largestOptionLength();
 
-  // Append the help information
-  return this.options.map(function(option) {
-      return pad(option.flags, width) + '  ' + option.description;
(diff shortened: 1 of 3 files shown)


index.js
```diff
@@ -932,20 +932,35 @@ Command.prototype.largestOptionLength = function() {
 };
 
 /**
+ * Return `Option` object for help option.  
+ *
+ * @return {Option}
+ * @api private
+ */
+Command.prototype.optionHelp = function() { 
+	// check if there is an overriding user-defined `-h` short flag
```
abetomo: Could you modify the tab width to match other codes?

TDress: my bad!
See amended commit.

vanesyan:
As far as I understand the pr, it introduce only overrides of short option for help (e.g. `-h`), but `--help` is still non overredable?

TDress:
@vanesyan, correct.   On master, before PR, a custom short option -h overrides the help short option, but the help information text was still showing -h as an option for the help command.  The PR doesn't touch the functionality of the options, and --help is still not overridable.

index.js
abetomo: 👍

vanesyan:
BTW, I think it's breaking change, and we should come back here, when the next major release happen.

TDress:
Thanks for taking a look.  Let me know if I can make any changes to fix
conflicts with new changes that you have.

vanesyan:
Yes, I will addressing this issue when we come to 3.0. Thank you :)

HRK44:
How is this a breaking change?
It's just modifies an output to remove '-h' from the built-in help option when it is already used by an user-defined option.

shadowspawn:
I had a look into this and don't think it is much risk of it being a breaking change @vanesyan since it just affects the display of the built-in help line in the built-in help and does not change the runtime behaviour otherwise. Add a comment if you have a usage case in mind which would break. Almost every change in a widely used program may break someone @HRK44, but I am trying not to be **too** paranoid! 😨

TDress:
Thanks for reaching out, but I won't be able to look at it in the near
future

shadowspawn:
Thanks @TDress 
I have explicitly referenced this pull request in #645 and closing this Pull Request for now.

PR #661, opened 2017-07-18
[WIP] Modernize codebase to ES6

vanesyan:


vanesyan:
@abetomo @zhiyelee  Hey, I considering updating testing infra and I'd like to decide on testing framework as well, should we update sinon to the most current version, or should we switch to something like jest, ava, etc.?

abetomo:
I would like to cast a vote on the plan to switch to jest.

abetomo:
@vanesyan Could you tell us about your future development schedule?

vanesyan:
Currently the branch contains only convertion of the codebase to es6 and separation of classes per file, everything working pretty well on Node.js 8+. The first thing we should recognize which minimal version of Node.js we support.

vanesyan:
Also, the plan was to incrementally improve the codebase. Currently commander consists of set of hacks to support different cases IMHO, we should minimize such a things and canonize most of parts of commander.

vanesyan:
Another one part of the plan was to implement critical futures people request for a long time, for instance long awaited future like control the flow of processing of required options, arguments.

vanesyan:
Probably, we shouldn't merge this pr into master, rather we should create `dev` branch and merge all commits the we want to be landed into v3 there.

abetomo:
@vanesyan Thank you for your reply.

I also have some suggestions for improvement.
Probably most of it will be improved with this fix.
I think that this code will eventually remain, so it is highly likely that you will be wasted even if you modify the existing code.

So I would like to do this with this new branch if I am going to fix it.
Could you please let me make it possible for multiple people to develop it?

vanesyan:
Yes, sure, do what ever you'd like to with it.

sangdth:
Any update about this? I'm looking forward.

shadowspawn:
Closing this WIP, see issue #666 instead.

Issue #663, opened 2017-07-19
ver@2.11.0 log help msg in wrong format


donaldshen:
should be:
```
  Usage: vue2-init [options]

  Options:

    -h, --help  output usage information

  Examples:
```
instead:
```
  Usage: vue2-init [options]


  Options:

    -h, --help  output usage information
  Examples:
```
btw, ver@2.9.0 is fine

vanesyan:
You mean about linespace?

drinchev:
That's because of https://github.com/tj/commander.js/pull/652/files .

shadowspawn:
This issue has not had any activity in over six months. It isn't likely to get acted on due to this report.

Feel free to open a new issue if it comes up again, with new information and renewed interest.
 
Thank you  for your contributions.

PR #664, opened 2017-07-20
Add empty element of array to add linefeed

abetomo:
Relation: #663 (#652)


vanesyan:
Isn't it better to make changes with `commandHelp` function directly?

abetomo:
I was at a loss as to whether to implement it in commandHelp.
Since this modification was formatting of display, I felt it was overkill in commandHelp.
So I made a fix like this, should I do it in commandHelp?

Issue #648, opened 2017-06-25
Init value of option '--name' is a [Function]


bySabi:
When a option `--name` is not set on CLI, initial value is a [Function]

Example:
```js
var Command = require('commander').Command;
var program = new Command('my-cli');

program
  .version('0.0.1')
  .option('--name <name>', 'Set name')
  .option('--name-prefix <name>', 'Set prefix name')
 .parse(process.argv);
  })
...

console.log('name: ' , program.name)
console.log('namePrefix: ', program.namePrefix)
```

### The ouput of  `$ my-cli --name-prefix=test`  is:
```
name: [Function]
namePrefix: test
```
> Here `name` should be `undefined`

### The ouput of  `$ my-cli --name=test`  is:
```
name: test
namePrefix: undefined
```
> Here is ok!

IMHO a widely use option like '--name' that collide with `[Command.name]` should have a dedicated check and initialization branch in code.

bySabi:
I create a workaround module for this problem: [safe-commander](https://www.npmjs.com/package/safe-commander)

elliot-nelson:
There are a couple flags like this.  `--parent`, for example.

bySabi:
@elliot-nelson you are right, too many options collide. That's for I could not wait for commander 3.x.x release and create [safe-commander.js](https://www.npmjs.com/package/safe-commander). 

Any feedback is welcome

shadowspawn:
Closing this as covered by #183 and #933

shadowspawn:
I have opened a Pull Request which allows storing option values separately rather than as command properties (access using `.opts()`), and passes the options (rather than the command) to the action handler. Opt-in using

```
program.configureCommand({ modern: true });
```

See #1102

PR #673, opened 2017-08-03
Fix options key name conflict with commander object's' key

brn:
Fix options key name confliction #404, #648 to use `Object.create(null)` and commander.opts() return object that has only options key value.
`Object.create(null)` is create object that hasn't any prototype, so it can hold any key without confiliction.

But, I'm warried about that it will become problem.
`Object.create(null)` hasn't any prototype chain, so if user extend Object.prototype and expect opts() result object is extended too, it will become problem.
What do you think of that?

### DIFF
--- a/index.js
+++ b/index.js
@@ -85,6 +85,7 @@ function Command(name) {
   this._allowUnknownOption = false;
   this._args = [];
   this._name = name || '';
+  this._optionKeyValueHolder = Object.create(null);
 }
 
 /**
@@ -357,6 +358,7 @@ Command.prototype.action = function(fn) {
 
 Command.prototype.option = function(flags, description, fn, defaultValue) {
   var self = this
+    , optionKeyValueHolder = this._optionKeyValueHolder
     , option = new Option(flags, description)
     , oname = option.name()
     , name = camelcase(oname);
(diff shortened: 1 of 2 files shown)


brn:
any update?

vanesyan:
Potentially breaking change, preserve until v3.0

shadowspawn:
Name clashes is one of the pain points with commander!

This pull request addresses just one aspect, with keeping track of the option values separate from the commander object. However, writing the option values onto the commander object is still going to cause clashes in that direction:

i.e. the lines like this:
```
self[name] = optionKeyValueHolder[name] = val;
```

(I am currently wondering about adding a feature-flag to track the values separately, as you are doing here, and not write them onto the commander object at all.)

shadowspawn:
See also #515 which removed the writing to the command object but added more syntax.

shadowspawn:
I am closing this in favour of #951, which builds on work in this Pull Request and aims for backwards compatibility.
 
shadowspawn:
I have opened a new Pull Request which allows storing option values separately rather than as command properties (access using .opts()), and passes the options (rather than the command) to the action handler. Opt-in using

```
program.configureCommand({ modern: true });
```

See #1102

PR #685, opened 2017-08-24
add attributeName() method to Option objects

yanick:
Mostly because it should be possible to
query what will be the `program` attribute name
for a given option. As it was, the `camelcase` function
was internal to the Option class and not accessible from
the outside.

### DIFF
--- a/index.js
+++ b/index.js
@@ -60,6 +60,18 @@ Option.prototype.name = function() {
 };
 
 /**
+ * Return option name, in a camelcase format that can be used
+ * as a object attribute key.
+ *
+ * @return {String}
+ * @api private
+ */
+
+Option.prototype.attributeName = function() {
+  return camelcase( this.name() );
+};
+
+/**
  * Check if `arg` matches the short or long flag.
  *
  * @param {String} arg
@@ -362,7 +374,7 @@ Command.prototype.option = function(flags, description, fn, defaultValue) {
   var self = this
     , option = new Option(flags, description)
     , oname = option.name()
-    , name = camelcase(oname);
+    , name = option.attributeName();
 
   // default as 3rd arg
   if (typeof fn != 'function') {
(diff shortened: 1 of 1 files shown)


vanesyan:
It's breaking change, and should be landed in v3.0, if landed. We've already accidentally broken commander compatibility in minor release once, and it brings us a bit of headache, so I'd like to preserve any potentially breaking changes.

yanick:
@vanesyan I'm curious: in which way is it breaking back-compatibility? As far as I know, the attribute names will resolve to be the same, I've just made them accessible from the outside. But then, I might be blind to a tiny, thorny detail. It happens. :-)

vanesyan:
Oh, sorry, now I see that there're no breaking changes, but I'm still not sure why this change should be landed?

yanick:
@vanesyan Heh, it's okay. I was totally expecting to be pointed at a weird side-effect I had not considered. :-)

As for the why, it's simply to allow to make the connection between the `Option` objects and their instance values in `program`.

To use the canonical example, for `--bbq-sauce`, the `name()` of the `Option` object is `bqq-sauce`. For a human, it's not hard to make the leap to `bbqSauce`, but there  is really no reason to have people reimplement (probably in a wrong way) the `camelcase` function if they want their code to figure out which `Options` a program value belongs to.

vanesyan:
But it's private method, and people should not relies on it, as privates are always maybe changed without any major bumps in breaking compatible way.

yanick:
Let me flip the question around. I have a `program` with an attribute named `bbqSauce`. I have to find back which `Option` object generated it (because I want to know if its value is different than the default, or what is the option's description, etc).  With the current public methods, how can I achieve that?

I think that currently I (programmatically) can't, because the way the option name is mapped to `bbqSauce` is internal and hidden.

abetomo:
@vanesyan Could you merge if there is no particular problem?

sweth:
Ping!  Any chance of this getting merged?  I personally would find it useful for being able to have a debug mode for a tool be able to programatically go through all of the options and list their current values, without having to hard code each of the option's var's names in addition to the option names themselves.

vanesyan:
@abetomo hey, I'm no longer interesting in maintaining the project, you can do whatever you thing is right, without me requiring to approve the change, thx :)

vanesyan:
Merging this into the master for now. As soon as new release occurs the change will be landed :)

PR #704, opened 2017-09-28
feat(typings): update to match js api

alan-agius4:


typings/index.d.ts
```diff
@@ -246,12 +245,19 @@ declare class Command extends NodeJS.EventEmitter {
...
     /**
-     * Get or set the name of the command.
+     * Set the name of the command.
      *
      * @param {string} str
-     * @return {string|Command}
+     * @return {Command}
      */
-    name(str?: string): string | Command;
+    name(str: string): Command;
```
vanesyan: It's wrong, previous was correct, name also can return string (self)

typings/index.d.ts
```diff
@@ -246,25 +246,22 @@ declare class Command extends NodeJS.EventEmitter {
...
+    name(str?: string): string | Command;
 
     /**
      * Output help information for this command.
      *
-     * @param {(str: string) => string} [cb] Callback method
+     * @param {() => void} [cb] Callback method
      */
-    outputHelp(cb?: (str: string) => string): void;
+    outputHelp(cb?: () => void): void;
```
vanesyan: Callback takes a string as the first parameter.

alan-agius4: fixed

PR #713, opened 2017-11-13
typedefs: Command and Option types added to commander namespace

sveinburne:
**Before** commit, the constructor is resolved but not the type<sup>1</sup> 

image

**After** commit, the type `Command` is resolved

image

The goal of this typedef enhancement is to allow the library consumer to use `Command` and `Option` types which he cannot yet use, in peculiar when receiving `action` callback arguments.
The way I do this is by creating a `local` namespace which contains class definitions and to cast those definition to a type (i.e. interface) in the `commander` namespace. By doing so, the types are accessible to library consumer.

<hr/>
<sup>1</sup> with typescript 2.5.3 and <code>CommonJS</code> modules


ball6847:
How long it's gonna take before releasing the next version ?

I'm facing this problem and looking forward to use this. 

Thanks

PR #720, opened 2017-11-23
Moved @types/node to devDependencies

### LINKED ISSUE #719: Move @types/node to devDependencies
Please move `@types/node` to `devDependencies` section instead of `dependencies` https://github.com/tj/commander.js/blob/master/package.json#L26.

I guess that nobody want to have this types in their web-based project (for example `tslint` have `commander` as deps) (`@types/node` may be installed as cross-deps)

timocov:
Fixes #719


abetomo:
It is fixed by #718

PR #622, opened 2017-04-22
sorted by name in commandHelp

shuogawa:
Please see examples (AdonisJs command line)

Before

After

PR #724, opened 2017-11-26
In help display sort commands alphabetically

### LINKED ISSUE #625: Feat Request: Sort help commands alphabetically
Either by default or through an option, the help text commands should be sorted alphabetically.

I can submit a PR if adding this line around [here](https://github.com/tj/commander.js/blob/master/index.js#L968) is sufficient. I'm not aware of any other usecases where this may cause a problem unless people are using weird characters in their command names. Works on my end anyway.

```
commands.sort(function (a, b) {
    return a[0] > b[0];
});
```

tedivm:
resolves #625

### DIFF
--- a/index.js
+++ b/index.js
@@ -981,6 +981,8 @@ Command.prototype.commandHelp = function() {
 
   var commands = this.commands.filter(function(cmd) {
     return !cmd._noHelp;
+  }).sort(function(a, b) {
+    return a._name.localeCompare(b._name)
   }).map(function(cmd) {
     var args = cmd._args.map(function(arg) {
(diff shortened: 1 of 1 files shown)


tedivm:
Unlike #622, this uses `localeCompare` for sort. That should make it more robust against special characters, and also makes it ignore case for sorting.

mojavelinux:
:+1:

shadowspawn:
I am concerned this change will undo custom ordering that people have done over the years and so we might need to allow an opt-out (although in general I don't want options for every behaviour). Closing this and making some notes in the linked issue for discussion.

PR #725, opened 2017-11-28
fix: typings are not shipped

alan-agius4:


alan-agius4:
Sry closed the PR by mistake

alan-agius4:
This is urgent can we have it merged and released asap pls?

Issue #727, opened 2017-12-04
Passing empty string for option on command is set to undefined


sorohan:
For this program:

```
program
  .name('test')
  .command('mycommand')
  .option('-c, --cheese [type]', 'optionally specify the type of cheese')
  .action(function(cmd) {
    val = cmd.cheese;
  });
```

And calling it like this:

```
test mycommand --cheese ""
```

I would expect the value of "val" to be an empty string, but it is actually undefined.

PR #728, opened 2017-12-04
Fixes issue #727: Passing empty string for option on command is set to undefined

sorohan:
This is probably a breaking change, but I would consider it the expected behaviour. It's currently checking if the next argument after an --option is truthy, but that makes it impossible to pass an empty string, which is sometimes needed.

### DIFF
--- a/index.js
+++ b/index.js
@@ -747,7 +747,7 @@ Command.prototype.parseOptions = function(argv) {
       // If the next argument looks like it might be
       // an argument for this option, we pass it on.
       // If it isn't, then it'll simply be ignored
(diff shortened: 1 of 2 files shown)


lipeRomani:
hi, I've had this problem and need this fix.
Some prevision to release this fix?

abetomo:
@sorohan Thank you pull request!
Could you confirm the conflict?

sorohan:
@abetomo I've rebased my changes onto master, and fixed a couple of small lint issues (missing semi colons).

abetomo:
Thank you!

Issue #289, opened 2014-11-16
Subcommand option with dashes


achingbrain:
This might be related to #250 but I've only been able to reproduce the problem reliably with subcommands.

The following test fails:

``` javascript
var program = require('../')
  , should = require('should');

program
  .version('0.0.1')
  .option('-f, --foo <foo>', 'a main command argument');

program
  .command('sub')
  .description('a subcommand')
  .option("-b, --bar <bar>", "a subcommand argument")
  .action(function(options) {
    options.bar.should.equal('-c -d');
  });

program.parse(['node', 'test.js', 'sub', '-b', '-c -d']);
```

This can be triggered in a real app by invoking it like:

``` bash
$ node test.js sub -b '-c -d'
```

If you remove the `-` characters from the option value (and the assertion in the test) it works as expected.

I think the problem is in the [`normalize`](https://github.com/tj/commander.js/blob/master/index.js#L474) method - it [looks up](https://github.com/tj/commander.js/blob/master/index.js#L483) the `Option` object for the previous argument.  It consults `this.options` which seems to be populated with options for the main command (`-f` in the example above) so the subcommand's options list is not checked and the method always returns undefined.

If it found the right `Option`, it'd trigger [this branch](https://github.com/tj/commander.js/blob/master/index.js#L491) which would add the option to the `ret` array but instead it reaches [this one](https://github.com/tj/commander.js/blob/master/index.js#L493) with undesirable results in this case.

curtisr7:
Still seeing this problem on commander@2.9.0. 

A workaround for the problem is to use the long version of the option ie: `--bar='-c -d'`

achingbrain:
In the end my workaround was to switch to [yargs](https://www.npmjs.com/package/yargs).

shadowspawn:
This issue has not had any activity in over six months. It isn't likely to get acted on due to this report.

Feel free to open a new issue if it comes up again, with new information and renewed interest.
 
Thank you  for your contributions.

Issue #508, opened 2016-03-12
git-style subcommands mess up argument order when using boolean flags


mal:
**index.js**

``` js
var command = require('commander')

var cmd = command
  .command('hello', 'description')
  .parse(process.argv)
```

**index-hello.js**

``` js
var command = require('commander')

var cmd = command
  .option('--dry-run', 'prints actions, but takes none')
  .parse(process.argv)

console.log('opts', cmd.opts())
console.log('args', cmd.args)
```

**output**

```
$ node index.js hello 1 2 --dry-run 3 4 5 6
opts { dryRun: true }
args [ '1', '2', '4', '5', '6', '3' ]
```

Note how `3` has been shifted to the end. This really messes with you when working with subcommands that expect a specific argument ordering.

_Found and tested against `commander@2.9.0`_

jmatsushita:
Ran into the same problem just now.

nibblesnbits:
I'm having the same issue.  It may have to do with the fact that even `process.argv` does not contain the flags passed in.

I have a simple cli with this definition in `app`:
```js
#!/usr/bin/env node

var program = require('commander');
var package = require('./package.json');

program
  .version(package.version)
  .option('-f, --force', 'skip confirmation')
  .command('flush <queue>', 'flush the queue of all jobs');
  
program.parse(require('process').argv);
```
And this in `app-flush.js`:
```js
console.log(JSON.stringify(process.argv));
console.log(JSON.stringify(program));
```

But the output of `./app flush -f` is:

```
["C:\\Program Files\\nodejs\\node.exe","C:\\dev\\app\\app-flush"]
{"commands":[],"options":[],"_execs":{},"_allowUnknownOption":false,"_args":[],"_name":""}
```
It seems that nothing is catching that flag.

nukelytide:
I can confirm this issue as well.

stnever:
I can confirm this issue on commander@2.18.0 as well.

shadowspawn:
This was reported again in #962 which has some good reproduction examples and discussion.

shadowspawn:
Possible related: #56 #75

shadowspawn:
Reproduced related issue with action handler:

```
const program = require('commander');

program
  .command('sub <args...>')
  .option('-a')
  .action((args) => {
    console.log(args);
  })

program
  .parse(process.argv);
```

```
$ node badParse.js sub foo bar
[ 'foo', 'bar' ]
$ node badParse.js sub -a foo bar
[...]

PR #734, opened 2017-12-16
Restore arguments order to pass to subcommands

maxlath:
so that option parsing is done by the subcommand, which may have its own options.

Addressed cases of the kind: `cmd subcmd -o arg1 arg2` where `-o` is a boolean option defined on the subcommand, and thus will not be identified by `parseOptions` when called from the parent command, and fall in the `// looks like an option` block, which will alter the arguments order: the subcommand will receive an args array that looks like `[ arg2, -o, arg1 ]`

Possibly related to #692 and #289

### DIFF
--- a/index.js
+++ b/index.js
@@ -485,15 +485,16 @@ Command.prototype.parse = function(argv) {
   }
 
   if (this._execs[name] && typeof this._execs[name] !== 'function') {
-    return this.executeSubCommand(argv, args, parsed.unknown);
+    this.args = args = argv.slice(2)
+    return this.executeSubCommand(argv, args);
   } else if (aliasCommand) {
     // is alias of a subCommand
-    args[0] = aliasCommand._name;
-    return this.executeSubCommand(argv, args, parsed.unknown);
+    this.args = args = [ aliasCommand._name ].concat(argv.slice(3))
+    return this.executeSubCommand(argv, args);
   } else if (this.defaultExecutable) {
     // use the default subcommand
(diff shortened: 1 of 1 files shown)


shadowspawn:
I think we need some better details about expected pattern for git-style commands! I need to do some experimentation to better understand before reviewing this PR. 

(I am surprised things are potentially this broken, which makes me wonder if there are some calling patterns which are ok.)

shadowspawn:
This is being considered for v3 (since it seems a nasty bug!), but has not passed any reviews yet, and won't be merged until approved. I am willing to help with code changes if needed.

shadowspawn:
Another possibly related issue: #508 
And noting that #962 has some analysis.

shadowspawn:
There is a bug but it is unfortunately harder and  more subtle to fix than this. The top level command can have options which are currently removed before calling the sub command. This may not be common, but it would dramatically change the behaviour to pass the original args through to the sub-command.

```
node index.js --debug hello 1 2 --dry-run 3 4 5 6 
```

maxlath:
hi! sorry for the slow reaction, I made a test repo to demonstrate the issue: https://github.com/maxlath/commander-subcommand-argument-order-demo

shadowspawn:
Thanks @maxlath. I think this is the same issue as in #508 and #962 (and is a bug).

shadowspawn:
The parsing got a major shakeup and argument order is now preserved in the published pre-release of Commander v5.0.0 (#1163)

Issue #691, opened 2017-09-07
Proposals for on/off switch


ntkme:
The justification for this has been discussed #108. I have working code but I'd like to open a discussion before opening a PR.

# Proposal 1

``` javascript
program
  .version('0.0.1')
  .option('    --[no-]color', 'optionally turn on or off the color');
```

Proof of concept implementation:

https://github.com/ntkme/commander.js/compare/master...ntkme:optional-switch

### Pros
- Small overhead
- Default value works as expected

### Cons
- No short flag

# Proposal 2

``` javascript
program
  .version('0.0.1')
  .option('-c, --color', 'enable color')
  .option('-C, --no-color', 'disable color')
```

Proof of concept implementation:

https://github.com/ntkme/commander.js/compare/master...ntkme:boolean-switch

### Pros
- Short flag works fine

### Cons
- The proof of concept implementation has to take a look back at all previous options, which means:
  - Weak performance 
  - Default value has to be set on the second option
  - Can be overcome by setting default values only when calling `.parse()` (breaking change, not a good idea)

abetomo:
@ntkme Thank you.
I think that `Proposal 1` is good.

@vanesyan What do you think?

syberkitten:
Hey Guys
It's so missing.... there are so many times I want to use a switch,
and right the way to achieve this is to now use normal arguments adding them a value just to 
make a switch.

Issue #108, opened 2012-12-07
On and off switches are not working


dignifiedquire:
If I try 

``` js
program.command('start')
  .option('--auto-watch', 'Description')
  .option('--no-auto-watch', 'Description')
  .action(function(options) {
    console.log(options.autoWatch);
  });
```

I get the following output:

``` bash
$ node program.js start --auto-watch
false
$ node program.js start --no-auto-watch
false
$ node program.js start
true
```

That should not happen. I need the first one to be `true`. Also if I try to set a default value of `true` that won't work either.

adius:
Any progress on this one?

akdor1154:
Still hitting this in 2017..

usmonster:
I have an open PR that should fix this: #795. Please review and comment!

shadowspawn:
This issue will be resolved when v3.0.0 is released. Available now as a prerelease. See #1001

shadowspawn:
Support for combined `--foo` and `--no-foo` has shipped in v3 (thanks @usmonster): https://github.com/tj/commander.js/releases/tag/v3.0.0

Closing second oldest open issue. Woop!

PR #736, opened 2018-01-03
Do not print default for "--no-"

ntkme:
Consider the following case:

```
  Options:

     --no-color         Disable color. (default: true)
```

**Internally** the "default: true" would make sense because we normalize `--no-color` to `color`, so `color` would have default true.

However, from a user perspective it makes absolutely no sense. Users would interpret it as: `--no-color` has default true, which means `color` would be off by default, which is the opposite to what it is.

Let's assume we fix it by change it to print false, now look at the following case (#108 & #691):

```
  Options:

     --color            Enable color.
     --no-color         Disable color. (default: false)
```

In this case no matter which default it prints, it would be nothing but confusions.

So instead of printing `default: false`, I suggest to not print any default for `--no-`, and this PR would do that.

### DIFF
--- a/index.js
+++ b/index.js
@@ -964,7 +964,7 @@ Command.prototype.optionHelp = function() {
   // Append the help information
   return this.options.map(function(option) {
       return pad(option.flags, width) + '  ' + option.description
-        + (option.defaultValue !== undefined ? ' (default: ' + option.defaultValue + ')' : '');
+        + ((option.bool != false && option.defaultValue !== undefined) ? ' (default: ' + option.defaultValue + ')' : '');
   }).concat([pad('-h, --help', width) + '  ' + 'output usage information'])
     .join('\n');
 };


vanesyan:
Good catch

michael-crawford:
This issue seems to be related to something I'm trying to implement, but I'm not sure. I still can't seem to figure out a way to do what I think should be something simple, so not sure if this is the best place to ask, but as I think it's related, here goes:

Summary: How are we supposed to set a default value for a boolean option?

Two use cases. Default true vs default false. 

- options.monitor should default to false or undefined, but have a way to be set to true
- options.check should default to true, but have a way to be set to false.

Current design as I understand it seems to support what I want as explicit command line options:

- Add "--monitor" to CLI to get options.monitor = true, otherwise it's undefined
- Add "--no-check" to CLI to get options.check = false, otherwise it's true

But I can't see how to make this work with default values, allowing me to set the default behavior via environment variables, but still allow for a command-line override.

- If MONITOR env var exists and is 1 or true, options.monitor should default to true, and "--monitor" should have no effect. Otherwise, options.monitor should default to undefined or false, and "--monitor" would then set it to true.
- If NOCHECK env var exists and is 1 or true, options.check should default to false, and "--check" should override this default to set it to true. Otherwise, options.check should default to true, and "--no-check" should override this and set it to false.

I thought this should work, but it doesn't. If there's not a way to do this, there should be.
```
let MONITOR = (process.env.MONITOR && (process.env.MONITOR == 'true' || process.env.MONITOR == '1')) ? true : false;
let CHECK = (process.env.NOCHECK && (process.env.NOCHECK == 'true' || process.env.NOCHECK == '1')) ? false : true;

program.command('test')
  .option('-m, --monitor', 'Monitor build process', MONITOR)
  .option('-C, --no-check', 'Do not check prerequisites', CHECK)
```

shadowspawn:
Interesting question @michael-crawford and nice detailed explanation of what you are trying to achieve. 

Would you please open a new issue for this? I think your issue is different to what this Pull Request is about and this is not the right place to discuss it. (You could perhaps add to #108, but I think your question is long enough and well constructed enough to go in its own issue.) 

One thing you could add is example output to show what didn't work with what you tried at the bottom, so readers don't need to try the code to find out what goes wrong.

PR #738, opened 2018-01-05
resolves #737 add commandFor method to Command

### LINKED ISSUE #737: Add commandFor method to Command
When writing custom actions for commander.js, I often find the need to look up a (sub)command by name so that I can retrieve information about it, such as help or its options. There's a method to look up options by name (optionFor), but not commands. I propose to add the helper commandFor. It will find a command either by name or alias.

mojavelinux:


vanesyan:
Personally, I'm against of adding a new feature for each user cases. I fills like a more specific problem, rather than common and can be implemented in userland.

mojavelinux:
> Thanks for helping us. We are really appreciate it :)

You're welcome. I'm happy to help. I should be able to stay involved as I've selected commander.js as the interface for my new project ;)

I'll explain the use case, which should help you understand why I think this is actually pretty core. I want to provide a `help <command>` feature like that which is found in many command-oriented programs. Now, of course, we could just enhance commander to provide this as an alternative to `<command> -h`, but for now, it has to be done manually. Here's what I made:

```js
cli
  .command('help [command]', { noHelp: true })
  .action((name, command) => {
    if (name) {
      const helpCommand = cli.commands.find((candidate) =>
        candidate.name() === name || candidate.alias() === name
      )
      // or const helpCommand = cli.commandFor(name)
      if (helpCommand) {
        process.stdout.write(helpCommand.helpInformation())
      } else {
        console.error(`'${name}' is not a valid command in ${cli.name()}. See '${cli.name()} --help' for a list of commands.`)
      }
    } else {
      cli.outputHelp()
    }
  })
```

So it seems natural to have a commandFor for performing operations over commands.

If you think a better way to go here is to add built-in support for "help <command>" as an alternative to "<command> -h", I'm happy to explore that route.

(I would also like to add support for a default command when it's not an executable, which would also require this method to determine if one of the members of argv is a command).

vanesyan:
Oh, I understand your use case. But you can maintain constant list of commands, that are supported by program, and lookup for provided name there :)

mojavelinux:
That's fine. It was just an idea. We've got other ideas to pursue, so I'm happy to focus on those.

PR #740, opened 2018-01-05
resolves #739 add parent command as prefix of subcommand

### LINKED ISSUE #739: Usage for command should include parent commands
The usage for an action-based (sub)command is misleading. It seems to suggest that the command can be used by itself, without the parent.

For example, if we were making the git cli with commander.js, the usage for `add` would look like:

```
Usage: add [options] <pathspec>
```

However, it should show the parent command (the context) as well:

```
Usage: git add [options] <pathspec>
```

I guess if it's an executable it's fine, so this just applies to subcommands that are not executables.

mojavelinux:


index.js
```diff
@@ -1031,9 +1031,12 @@ Command.prototype.helpInformation = function() {
   if (this._alias) {
     cmdName = cmdName + '|' + this._alias;
   }
+  var ancestor = this
+  var prefix = []
+  while ((ancestor = ancestor.parent)) prefix.push(ancestor.name() + ' ')
```
vanesyan: Isn't it better to push just a name of ancestor to array (e.g. `prefix.push(ancestor.name())`) and then...

index.js
```diff
@@ -1031,9 +1031,12 @@ Command.prototype.helpInformation = function() {
...
   var usage = [
     ''
-    ,'  Usage: ' + cmdName + ' ' + this.usage()
+    ,'  Usage: ' + prefix.join('') + cmdName + ' ' + this.usage()
```
vanesyan: ...join here by space (`' '`) instead of empty string?

mojavelinux: The reason I did it this way is because it avoids adding an extra space if the list turns up empty. There are other ways to accomplish that of course, this was just a clever way to do it.

vanesyan:
```js
[].join(' ') // => ''
```
Isn't it?

mojavelinux: Yes, but it's not the spaces between the arguments we're worried about, it's the trailing space. If the array is not empty, we *want* a trailing space. If the array is empty, we don't want a trailing space. That's why I add a trailing space to each entry. It accomplishes this outcome.

Again, it's just one way of getting the spaces correct. I'm happy to change it to something more clear if you don't feel comfortable with it.

index.js
simonbuchan: If this has multiple ancestors, e.g. `git remote add`, this will add them in reverse order, so you get `remote git add`

index.js
simonbuchan: Something clearer might be:

```js
const names = [];
for (const cmd = this; cmd; cmd = cmd.parent) {
  names.unshift(cmd.name());
}
...
   ,'  Usage: ' + names.join(' ') + ' ' + this.usage(),
```

kvmamich:
Hi! When PR will be merged? Need this :)

shadowspawn:
Thanks for questions from @vanesyan and @abetomo, replies from @mojavelinux,  and comments from @simonbuchan.

shadowspawn:
Adding link to #739 so shows up there.

shadowspawn:
Merging into `release/3.0.0` branch.

shadowspawn:
Closing in favour of #980 (which includes these commits)

shadowspawn:

PR #743, opened 2018-01-06
resolves #741 make indentation configurable

### LINKED ISSUE #741: Make help indentation configurable
commander.js indents all lines in the help output. This is atypical. I don't know many commands that output help with this formatting (if you consider node, npm, lerna, ruby, docker, git, grep, find, vim, rpm). I'd like to make my command look more like the typical command. Please make the indentation size configurable, or provide a switch to disable it outright.

Commands which do use indentation include yarn and gem. Yarn happens to have indentation because it's using this package ;)

mojavelinux:


mojavelinux:
I'm still playing with this one a bit. And I need to add a test.

mojavelinux:
Okay, that should give us something to discuss.

Here's what I'm thinking. First, we may want to use numbers instead of strings for indent and spacing. For example, INDENT=2, SPACING=1 (which are the defaults). That will require some rework.

I'm also thinking that perhaps this logic should be moved to the functions `space()` and `indent()`. That's pretty simple for everything except `helpInformation` which requires adding a few extra lines to keep the sections from crashing into each other.

vanesyan:
Generally this is good idea. Yes, I like the second form of control indentation (through numbers) more that the first one.

shadowspawn:
Related issue was closed and OP happy with subsequent improvements: https://github.com/tj/commander.js/issues/741#issuecomment-452632316

mojavelinux:
Agreed. I think we came to a different agreement to solve the same problem.

PR #745, opened 2018-01-07
Fix examples' extensions

buoyantair:
I felt like the example files needed their extensions, and added them. :)


vanesyan:
Hello there! We are appreciate your help, but files as extensions-less deliberately as bin files, are almost always goes without extension and for execution of the file response line started with shebang token (the most top line).

PR #750, opened 2018-01-10
resolves #749 pass extra args as final argument of action callback

### LINKED ISSUE #749: Extra arguments passed to non-variadic command are swallowed
If an action-based command does not have a variadic argument, extra arguments supplied by the user are swallowed. I would expect either for the command to exit with an error, or for those extra arguments to be passed to the callback method so they can be handled appropriately.

The root of the problem is that the program is overwriting the extra entry in the args array.

mojavelinux:


shadowspawn:
This is a possible approach, but my current thinking is I do not like having *two* magic extra parameters for callback (in addition to passing the command object).

There is another open PR #654 where this was also suggested in a comment. I noted there there: I think a method on the commander object would be tidier than pushing arguments into callback.

PR #751, opened 2018-01-11
document using options per command

mojavelinux:
- document that options can be added to a command
- document that command options are validated when the command is used
- document that options are not validated if command has no action

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -64,6 +64,27 @@ if (program.sauce) console.log('  with sauce');
 else console.log(' without sauce');
 ```
 
+## Command-specific options
+
+You can attach options to a command.
+
+```js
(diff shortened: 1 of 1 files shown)


Readme.md
```diff
@@ -64,6 +64,27 @@ if (program.sauce) console.log('  with sauce');
 else console.log(' without sauce');
 ```
 
+## Options for a command
+
+You can attach options to an action-based command.
+
+```ruby
```
abetomo: javascript?

mojavelinux: OMG! I'm having a polyglot moment!

mojavelinux: Fixed!

abetomo: Thank you!

PR #753, opened 2018-01-19
resolves #560 respect custom name for version option

### LINKED ISSUE #560: Add support for -v to get the version
Please add support for `-v` instead of `-V` to get the `--version` to be consistent with Node and npm (and a lot of other CLIs).

Related to #509

mojavelinux:


mojavelinux:
I'm still working on a test. The version function isn't tested at all from what I can tell, so I'll add a version for both the implicit and custom scenarios.

mojavelinux:
Tests added!

mojavelinux:
I made it clear that the long option is required.

mojavelinux:
:+1:

PR #755, opened 2018-01-22
WIP resolves #649 switch to a modern test infrastructure

### LINKED ISSUE #649: Convert test infa to something more reliable and new
commander.js has pretty old test infra based on sinon and should, we can convert it to something more popular. With convertion we may archive the following goals: keep it up to date to today's requirements, to make it easier to make contributions and convert it to es6 (that plan, I has in my mind), and, finally, to make it cross-platform. The current test infra can only be run on *nix systems, but no way to run test on env. that has no bash (or similar shells) and make, like windows.

mojavelinux:
This PR will hopefully kick off the migration to a modern test infrastructure. There are still many tests to rewrite. But we first need to decide whether ava is the right choice for this project. Then we can divide and conquer to finish migrating the tests.

The tests are temporarily stored in the test-ava folder until the old tests are cleared out of the way.

Use the following command to run the tests:

```sh
npm run ava
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -15,6 +15,7 @@
     "url": "https://github.com/tj/commander.js.git"
   },
   "scripts": {
+    "ava": "ava test-ava",
     "test": "make test && npm run test-typings",
     "test-typings": "node_modules/typescript/bin/tsc -p tsconfig.json"
   },
@@ -25,6 +26,7 @@
   ],
   "dependencies": {},
   "devDependencies": {
+    "ava": "^0.24.0",
     "@types/node": "^7.0.48",
     "should": "^11.2.1",
     "sinon": "^2.4.1",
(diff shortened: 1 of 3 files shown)


mojavelinux:
Thoughts?

abetomo:
I think that there is no problem in selecting `ava`.

mojavelinux:
:+1: 

I will continue adding a few more test cases. I'm sure it will raise some questions.

shadowspawn:
(Adding link to #649 for easier clicking.)

CheshireSwift:
I have minor quibbles with ava, but not enough to want to bikeshed about it. I'd say go for it 👍

shadowspawn:
I like Jest. The main advantage for Commander is it is much more widely used than Ava, so will be familiar to more people contributing to Commander. The main personal advantage is that Jest is the testing framework I use at work and at home!

I'm working on the conversion in #1035

Are you still ok with Jest @abetomo ?
https://github.com/tj/commander.js/pull/661#issuecomment-316240282

abetomo:
I think Jest is better.

shadowspawn:
Closing in favour of #1035. Started move to Jest.

Issue #686, opened 2017-08-30
option doesn't work correctly when given four arguments


tk1:
This gist https://gist.github.com/tk1/70fa2ea28b359b90eb1e0181c6cec977 demonstrates a case where the option value is completely wrong.

vanesyan:
Can you, please, test the code in previous versions of commander v2.10 and v2.9 and feedback it here, thx! We need to know was this regression introduced in newer version of commander, or exists in previous as well.

tk1:
Version 2.0.0 and 2.1.0 are correct. The error exists since version 2.2.0

codeclown:
Based on my research, this happens because the default value is passed as a second argument to the coerce function.
```js
program.option('-i, --int <n>', 'pass an int', parseInt, 100)
program.parse(['node', 'test', '-i', '5.5']);
// will invoke
parseInt('5.5', 100)
```
Obviously results in an unexpected result in case of `parseInt`.

Relevant line where coerce function is called.
https://github.com/tj/commander.js/blob/0ae835315a4ff9162c963eb2223c327fd9119996/index.js#L400

I'm willing to contribute a PR but I'm not sure what would be the best course of action to handle this.

edemaine:
For what it's worth, I've found that specifying the default *before* the transform works right:
```js
.option('-d --integer-with-default <n>', 'An integer argument with default value', 22, parseInt)
```
Whereas listing the default after the transform gets it passed in the transform:
```js
.option('-a --append <x>', 'A list argument', (x, list) => list.push(x), [])
```
So perhaps this is all intended behavior.  It's not exactly well-documented, though.

shadowspawn:
Closing as covered by #201 and #943

PR #757, opened 2018-01-23
Fix Coercion for Repeated  Options

icirellik:
This fixes Issue #686 that appears to have been introduced when
repeated options were added PR #198 that causes the previous value or
default value to be passed as a second parameter to the coercion
function. This obviously causes unintended side effects when a user
specifies a function and a default value unaware that a second value
passed, such as with parseInt.

This change makes the assumption that when a default array is specified
that the user has expecting repeated options. This change is backward
compatible with the existing code.

### DIFF
--- a/index.js
+++ b/index.js
@@ -390,6 +390,17 @@ Command.prototype.option = function(flags, description, fn, defaultValue) {
     }
   }
 
+  // Is the default array like, if so its likely repeated options.
+  if (defaultValue != null &&
+    typeof defaultValue != 'function' &&
+    typeof defaultValue.length == 'number' &&
+    defaultValue.length > -1
+  ) {
+    option.repeated = true;
+  } else {
+    option.repeated = false;
+  }
+
   // preassign default value only for --no-*, [optional], or <required>
   if (!option.bool || option.optional || option.required) {
     // when --no-* we make sure default is true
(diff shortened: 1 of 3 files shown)


loicgasser:
any chance to merge that?
It's a pretty common use case.a

abetomo:
@icirellik Thank you PR!
Could you please resolve the conflict?

abetomo:
@icirellik Thank you!

@vanesyan Could you review it?

shadowspawn:
Good investigation, thanks @icirellik . It was bad that the behaviour changed for this case when the repeatable options were added.

However, I do not like trying to guess the intended usage based on the type of the extra parameter. That makes the behaviour a bit fragile. I would prefer to untangle the usage patterns and have it work consistently.

#829 is a step in that direction. Documenting the working pattern.

shadowspawn:
I should clarify that I have not worked through the code and behaviour in detail yet, my comment was based on reading the two Pull Requests and some of the related issues. I might change my mind about the suggested approach here when I understand more, and whether it is possible to "untangle" the usage patterns.

shadowspawn:
I came up with an idea I'll write up as a new proposal to clean up the behaviour, and see if it makes enough sense and seems worth the effort.

Sneak preview:

```
# callback takes one parameter
.option('-i, --integer <n>', 'An integer argument', parseInt)
.option('-i, --integer <n>', 'An integer argument', parseInt, defaultValue)

# callback takes two parameters
 .options('-c, --collect [value]', 'A repeatable value', collect, [])
 .options('-v, --verbose', 'A value that can be increased', increaseVerbosity, 0)
```

shadowspawn:
I am currently leaning towards just documenting the existing behaviour properly. It has been that way for a long time.

shadowspawn:
I am going to have a go at documenting the existing behaviour. If it fails to work out, this Pull Request might be reconsidered!

Thanks for your contributions.

shadowspawn:
The options coverage in the README has been updated and includes new description of custom option processing, including a working parseInt example.

https://github.com/tj/commander.js#custom-option-processing

PR #758, opened 2018-01-23
resolves #754 add linter configuration to project

### LINKED ISSUE #754: Add a linter to the project
I'm trying to contribute to this project and I have no idea what code style I'm supposed to be using. And there's nothing to check that I'm doing it right. A linter would make contributing to this project a lot easier.

I also find the use of commas to be rather eclectic, so it would be nice to either enforce that style or use a more standard style.

My proposal is to use [standardJS](https://standardjs.com/) with eslint. It's extremely popular and effective. I've used it for several months now and I hardly ever having to think about it.

I'm happy to set up this tooling.

mojavelinux:
- add linter configuration based on StandardJS
- customize linter to fit with style of project
- update source to reconcile violations

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -1,0 +1,9 @@
+{
+  "extends": "standard",
+  "rules": {
+    "eqeqeq": "off",
+    "one-var": "off",
+    "semi": ["error", "always", { "omitLastInOneLineBlock": true }],
+    "space-before-function-paren": ["error", "never"],
+  }
+}
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,11 @@
 language: node_js
 sudo: false
 node_js:
-  - "4"
-  - "6"
-  - "8"
-  - "9"
-  - "node"
+- "4"
+- "6"
+- "8"
+- "9"
+before_install: if [[ $(node -v) == v4* && $(npm -v) != 3* ]]; then npm i -g npm@3; fi
+script:
+- npm run lint
+- npm test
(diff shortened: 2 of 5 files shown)


abetomo:
Thank you!
Could you check it because the test failed?

mojavelinux:
The version of npm that comes with node 4 doesn't honor the lock file, so it wasn't hoisting eslint. We should be upgrading npm to at least npm 3 since npm 2 is deprecated.

mojavelinux:
It was also necessary to move the util.inherits call to the top of the file for node 4. I'm not sure why it gets confused about the order, but it seems to have been fixed in node 6.

mojavelinux:
This exercise as proved how useful the linter actually is ;)

mojavelinux:
Looks like we're going to have to add eslint as a dependency as no version of npm seems to agree on where to put it when it's a transitive dependency ;(

mojavelinux:
I also dropped "node" from the list of node versions to test since it's exactly the same as "9".

mojavelinux:
Finally!

abetomo:
Thank you!

vanesyan:
Thanks!

mojavelinux:
@vanesyan My pleasure.

Shall we reenable that eqeqeq rule?

vanesyan:
Yes, it's definitely a good idea, except that current in some parts of code we rely on poor `null` and `undefined` equality (e.g. `thing == null` -> `thing === undefined || thing === null`), we should be careful in such checks.

mojavelinux:
@vanesyan Fortunately, those are already exceptions in StandardJS, so nothing to worry about.

PR #761, opened 2018-01-31
only register the option:version event once

mojavelinux:
- create Option directly to avoid registering the option:version event twice
- defer to the Option instance to resolve the long name of the option for the event


mojavelinux:
Currently, two handlers are being registered for the option:version event. This consolidates it to one.

PR #771, opened 2018-02-27
resolves #766 suggest to users run --help command

### LINKED ISSUE #766: Updating console.error messages suggesting to clients run --help command
I post a possible **enchantment**.  

## Description
One of the core features of commander.js is the ability to show meaningful error messages to the user. The API provides `.on('--help', cb)` listener on commands that allow users to add custom help messages (a description, examples, etc). However, to be able someone gets those help information he needs to provide `--help` argument in the command. For instance, if we have the following command `node index.js rm <file>` we are able to provide an `.on('--help', cb)` listener to `command('rm <file>')`.  

When new clients want to use our CLI tool will find themselves, most of the time, struggling to figure out how to use a command. For instance, if they forget the `<file>` argument in `rm` they'll face the following error message:

```
error: missing required argument `file'
```

Our client now needs to go to our API documentation and search what the `<file>` argument is (a path or name?). For simple commands, such as the above example, this isn't something interesting, but in more sophisticated commands, this actually can add headaches. Of course, the client can always add `--help` argument to that command and get the appropriate information. But most of the times, he will miss it. 

## Proposal Fix
I propose to update those error messages by adding one more line that will suggest our client try the `--help` argument. As such, the above message will be changed to:

```
error: missing required argument `file'
Try '--help' for more information.
```

This is something that most UNIX like commands do (check `rm`, `cp`, `mv`, etc)  and I believe that increases the User Experience. Of course, we can propose other alternatives as well: Print the whole help page below the error message, but in my opinion, this is something that would add more noise.

**A fix like this will not add any break changes to the codebase nor will decrease the test coverage.**

If you agree with this idea, I will immediately open a new pull request.

GeorgeGkas:
Adding one more console.error statements to bellow methods:
- Command.prototype.missingArgument
- Command.prototype.optionMissingArgument
- Command.prototype.unknownOption
- Command.prototype.variadicArgNotLast

This console.error messages propose to user run the same command
with `--help` option to get more info about the command which try
to run. This is something that most UNIX like commands do
(like rm, cp, mv, etc)

### DIFF
--- a/index.js
+++ b/index.js
@@ -787,6 +787,7 @@ Command.prototype.opts = function() {
 Command.prototype.missingArgument = function(name) {
   console.error();
   console.error("  error: missing required argument `%s'", name);
+  console.error("  Try '--help' for more information.");
   console.error();
   process.exit(1);
 };
@@ -806,6 +807,7 @@ Command.prototype.optionMissingArgument = function(option, flag) {
   } else {
     console.error("  error: option `%s' argument missing", option.flags);
   }
+  console.error("  Try '--help' for more information.");
   console.error();
   process.exit(1);
 };
@@ -821,6 +823,7 @@ Command.prototype.unknownOption = function(flag) {
   if (this._allowUnknownOption) return;
   console.error();
   console.error("  error: unknown option `%s'", flag);
+  console.error("  Try '--help' for more information.");
   console.error();
   process.exit(1);
 };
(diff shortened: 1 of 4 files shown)


shadowspawn:
I am closing this PR as has not had recent activity and there are multiple conflicts, and it is a simple change so easier to make it again fresh after discussing desired error format in #766.

I found there is already one error in commander which suggests trying `--help` and on a single line like `git` does. I think this is probably tidier, but have not worked through all four cases that are in this PR to check it will look ok. 

```
      console.error('error: %s(1) does not exist, try --help', bin);
```

PR #774, opened 2018-03-07
added ability to translate help output

cravler:
```js

program.version('0.0.1')
    .command('setup [env]')
    .description('application simple description', {
        env: 'env value'
    })
    .option('-f, --foo', 'enable some foo')
    .action(function(env, options) {});

var translations = {
    'Usage:': 'Применение:',
    'Commands:': 'Команды:',
    'Arguments:': 'Аргументы:',
    'Options:': 'Опции:',
    'output usage information': 'информация об использовании',
    'output the version number': 'показать номер версии',
    'application simple description': 'простое описание приложения',
    'enable some foo': 'включить некоторые foo',
    'env value': 'значение env'
    // ...
};

program.translator = function(str) {
    return translations[str] || str;
};

program.parse(process.argv);
if (!program.args.length) program.help();
```

### DIFF
--- a/test/test.translator.js
+++ b/test/test.translator.js
@@ -1,0 +1,48 @@
+var program = require('../')
+  , should = require('should');
+
+var translations = {
+    'Usage:': 'Применение:',
+    'Commands:': 'Команды:',
+    'Arguments:': 'Аргументы:',
+    'Options:': 'Опции:',
+    'output usage information': 'информация об использовании',
+    'application simple description': 'простое описание приложения',
+    'env value': 'значение env'
+};
+
+program.translator = function(str) {
+    return translations[str] || str;
+};
+
+program
+    .helpInformation()
+    .should.equal([
+        '',
+        '  Применение:  [options]',
+        '',
+        '  Опции:',
+        '',
+        '    -h, --help  информация об использовании',
+        ''
+    ].join('\n'));
+
+program
+    .command('mycommand [env]')
+    .description('application simple description', { env: 'env value' })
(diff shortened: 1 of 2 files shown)


abetomo:
@cravler Thank you!
Could you add a test?

test/test.translator.js
```diff
@@ -0,0 +1,48 @@
+var program = require('../')
+  , should = require('should');
+
+var translations = {
+    'Usage:': 'Применение:',
```
abetomo: Thanks for adding the test!
Could you modify the indent width to 2?
Because the indentation width of other code is 2.

cravler:
When this changes will be merged with master?

vanesyan:
Thanks for the PR. But I'm against of such things, firstly it's looking a bit ugly in part with translation table in you example, IMHO, the second I'm against of all changes that can be done in the userland, it's better to keep core pretty minimal as much as we can.

To make your program behave the same through available API, just put `noHelp: true` to commands and options you'd like to hide from output and put appropriated handler on `--help` event.

vanesyan:
I'm closing it for now. Feel free to re-open pr if you think that it was closed wrongly.

cravler:
Do you think that reapply commander logic with `noHelp: true` is better way then having translations map? I do not think so! For me is better to have translation files for different languages and just require it. 

This functionality is optional and do not add any complexity for commander logic.

shadowspawn:
Anyone following this issue is likely to be interested in #1801

PR #778, opened 2018-03-09
Fix to emit the action even without command

### LINKED ISSUE #729: Action never called when no arguments/options passed
I am using git style sub command and have a command that requires no args/options but needs to use the `.action()` call for an async callback.

````
#!/usr/bin/env node

var program = require('commander')

program
  .action( () => {
    console.log('never called')
  })
````  
The above will never happen. How can I specify that I want this called? Right now I have to make a dummy option.

abetomo:
fix #729

### DIFF
--- a/index.js
+++ b/index.js
@@ -660,6 +660,9 @@ Command.prototype.parseArgs = function(args, unknown) {
     if (unknown.length > 0) {
       this.unknownOption(unknown[0]);
     }
+    if (this._args.filter(a => a.required).length == 0) {
+      this.emit('command:*');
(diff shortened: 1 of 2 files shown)


rostik404:
Great PR! 🔥 
@abetomo @vanesyan, merge it, pls, we are looking forward for this feature. Thanks!

newsomellc:
+1, please merge this.

andrewsosa:
@abetomo @vanesyan This PR has been open for a few months, could it get merged in?

PR #779, opened 2018-03-15
move output information to printer

cravler:
```js
const { Command, Printer } = require('commander');

class CustomPrinter extends Printer {
 
    error(msg, ...args) {
        super.error(this.format(msg), ...args);
    }
 
    outputHelp(program) {
        return this.format(super.outputHelp(program));
    }
 
    format(str) {
        return [
            '\n', str.replace(/^/gm, '  '), '\n'
        ].join('');
    }
 
}

const program = new Command('example', new CustomPrinter());

program.version('0.0.1')
    .command('setup [env]')
    .description('application simple description', {
        env: 'env value'
    })
    .option('-f, --foo', 'enable some foo')
    .action(function(env, options) {});

program.parse(process.argv);
if (!program.args.length) program.help();
```


cravler:
any comments?

cravler:
Is this repository supported?

shadowspawn:
@cravler 
What problem are you solving with this Pull Request? 
Are there open Issues that this will resolve?

Or is this primarily an object separation suggestion, rather than a functional suggestion?

cravler:
This separation adds ability to write custom printers without touching the default. In my case I have `Colorized Printer`

shadowspawn:
Thanks @cravler. This is quite interesting, but changes a lot of code. Offering customisation by inheritance is also more challenging for future maintenance than a callback.

I don't wish to add this at the moment. However we have a variety of requests around help and output. This may get revisited in the future if it can be used to resolve more issues.

PR #788, opened 2018-04-10
Fixed TSC error

### LINKED ISSUE #787: Typescript type definitions incorrect for Command class
When compiling a typescript project the following error is produced

```
node_modules/commander/typings/index.d.ts(26,25): error TS2689: Cannot extend an interface 'NodeJS.EventEmitter'. Did you mean 'implements'?
```

ivankostov:
fixes #787

node_modules/commander/typings/index.d.ts(26,25): error TS2689: Cannot extend an interface 'NodeJS.EventEmitter'. Did you mean 'implements'?

### DIFF
--- a/typings/index.d.ts
+++ b/typings/index.d.ts
@@ -23,7 +23,7 @@ declare namespace local {
     constructor(flags: string, description?: string);
   }
 
-  class Command extends NodeJS.EventEmitter {
+  class Command {
     [key: string]: any;
 
     args: string[];


vanesyan:
Weird error, Command actually extends `EventEmitter`, [on this line](https://github.com/tj/commander.js/blob/master/index.js#L16)

vanesyan:
Are there any errors with commander with TypeScript other then the version (your current)?

ivankostov:
Yes, I have tried it with all versions above 2.11 and with different
typescript versions. And in this case the right keyword would be
"implements" but as far as I see there are no common methods with
EventEmitter

Regards,
Ivan

На ср, 11 апр 2018 г., 12:27 Roman Vanesyan <notifications@github.com>
написа:

ChrisCates:
When will this be merged?
I've tested this on several semvers of Node.js and `tsc`. Just curious what's going on here.

vanesyan:
Sorry for waiting so long. Curios, what will it looks like if we change it a bit and put `implements` instead of `extends`?

ivankostov:
Do not discuss too much, verify, merge, smile :-)

На чт, 7 юни 2018 г., 21:10 Roman Vanesyan <notifications@github.com>
написа:

vanesyan:
But I'd like to find a solution that both is correct and working to satisfy all needs (almost). So currently behavior is that commander actually extends `EventEmitter`, but does this in es5 manner, so TypeScript doesn't recognize it, I'd still like to allow users on TypeScript to use those APIs without headache, so make it visible in `d.ts`.

ChrisCates:
If it passes, tests compile and is usable with the LTS of Typescript and are designed to work for future versions.

Then implements syntax is acceptable.

ChrisCates:
Feel free to reference a pull request using `implements` and I will review.

avendiart:
`EventEmitter` is a class, not an interface, so `implements` is out of place here. The problem is that tsc doesn't know anything about `NodeJS.EventEmitter` if the `@types/node` isn't installed. You could try doing `import { EventEmitter } from 'events'` and actually adding `@types/node` as a dependency.

vanesyan:
@avendiart can you make a pr, please?

shadowspawn:
I tried this change and it caused errors in my own project after using `.on`, the subsequent `.action` had `any` parameters. I do not think it is appropriate to drop the `NodeJS.EventEmitter`.

```js
program
    .command("clone <source> [destination]")
    .on("--help", () => {
      console.log(`blah blah.`);
    })
    .action((source, destination, options) => {
      // ...
    });
```

We do have an undocumented expectation that the types for node are installed separately. I am open to suggestions on improving that situation in TypeScript land, but my current plan is to add the expectation to the README!

PR #791, opened 2018-04-19
updated readme with custom event listeners.

yausername:
adds documentation for #494 as well.


yausername:
@vanesyan any comments? don't you think this would be helpful in the readme?

vanesyan:
Thanks!

PR #794, opened 2018-04-30
fixes tests

usmonster:
- Many tests were not actually invoking assertion functions. Now.they.are()
- Fixed one test that was then failing because of side-effects from
  previous test setups.
- Made small changes in some setups to avoid potential false negatives.

(Also: made a tiny unrelated code change to ensure Options always has
boolean properties; does not change any behavior, but may help with
debugging.)

### DIFF
--- a/index.js
+++ b/index.js
@@ -43,9 +43,9 @@ exports.Option = Option;
 
 function Option(flags, description) {
   this.flags = flags;
-  this.required = ~flags.indexOf('<');
-  this.optional = ~flags.indexOf('[');
-  this.bool = !~flags.indexOf('-no-');
+  this.required = flags.indexOf('<') >= 0;
+  this.optional = flags.indexOf('[') >= 0;
+  this.bool = flags.indexOf('-no-') === -1;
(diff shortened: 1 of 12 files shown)


index.js
```diff
@@ -43,8 +43,8 @@ exports.Option = Option;
 
 function Option(flags, description) {
   this.flags = flags;
-  this.required = ~flags.indexOf('<');
-  this.optional = ~flags.indexOf('[');
+  this.required = flags.indexOf('<') >= 0;
+  this.optional = flags.indexOf('[') >= 0;
   this.bool = !~flags.indexOf('-no-');
```
vanesyan: You can also convert this line as you did with previous one.

usmonster: I can do it if you want, but it will only be for readability/consistency, as the result is boolean as-is.

vanesyan: Yes, for consistency purpose.

usmonster: No problem, it's done. :)

vanesyan:
Thanks

PR #795, opened 2018-05-02
fixes behavior of --no-* options

usmonster:
A bug existed whenever a `--no`-prefixed option was defined along with
its counterpart (e.g. `--status` and `--no-status`), which caused the
corresponding boolean attribute to always be `false` when either option
was specified on the command line.

This was because the definition of a negating option would overwrite the
default value of the attribute, combined with the fact that specifying either
option would emit two events with conflicting logic.

This has been corrected in two ways:
1. During the initialization of a negating option, we now check to see if the
   non-negated counterpart was already defined, in which case we don't
   set/overwrite the default value (which would otherwise be set to `true`);
2. An option's name no longer omits its negation prefix; that is,
   "--no-cheese" is now internally named `"no-cheese"`, which will distinguish
   it from "--cheese". This will allow unique listeners & events to be registered and
   emitted, depending on which option is passed to the command, thus
   avoiding any attribute assignment collision/race.

Additionally, the tests for negated options were updated to more
explicitly demonstrate the expected behavior, and a couple of relevant
examples were fixed to match their intended behavior.

### DIFF
--- a/index.js
+++ b/index.js
@@ -60,9 +60,7 @@ function Option(flags, description) {
  */
 
 Option.prototype.name = function() {
-  return this.long
-    .replace('--', '')
-    .replace('no-', '');
+  return this.long.replace(/^--/, '');
 };
 
 /**
@@ -74,7 +72,7 @@ Option.prototype.name = function() {
  */
 
 Option.prototype.attributeName = function() {
-  return camelcase(this.name());
+  return camelcase(this.name().replace(/^no-/, ''));
 };
 
 /**
@@ -335,14 +333,17 @@ Command.prototype.action = function(fn) {
  *
  * Examples:
  *
- *     // simple boolean defaulting to false
+ *     // simple boolean defaulting to undefined
  *     program.option('-p, --pepper', 'add pepper');
+ *
+ *     program.pepper
+ *     // => undefined
  *
  *     --pepper
  *     program.pepper
- *     // => Boolean
+ *     // => true
  *
- *     // simple boolean defaulting to true
+ *     // simple boolean defaulting to true (unless non-negated option is also defined)
  *     program.option('-C, --no-cheese', 'remove cheese');
  *
  *     program.cheese
@@ -394,8 +395,11 @@ Command.prototype.option = function(flags, description, fn, defaultValue) {
 
   // preassign default value only for --no-*, [optional], or <required>
   if (!option.bool || option.optional || option.required) {
-    // when --no-* we make sure default is true
-    if (!option.bool) defaultValue = true;
+    // when --no-foo we make sure default is true, unless a --foo option is already defined
+    if (!option.bool) {
+      var opts = self.opts();
+      defaultValue = Object.prototype.hasOwnProperty.call(opts, name) ? opts[name] : true;
+    }
     // preassign only if we have a default
     if (defaultValue !== undefined) {
       self[name] = defaultValue;
@@ -426,7 +430,7 @@ Command.prototype.option = function(flags, description, fn, defaultValue) {
       }
     } else if (val !== null) {
       // reassign
-      self[name] = val;
+      self[name] = option.bool ? val : false;
     }
   });
 
--- a/Readme.md
+++ b/Readme.md
@@ -343,13 +343,13 @@ Usage: pizza [options]
 An application for pizzas ordering
 
 Options:
-  -h, --help           output usage information
   -V, --version        output the version number
   -p, --peppers        Add peppers
   -P, --pineapple      Add pineapple
   -b, --bbq            Add bbq sauce
-  -c, --cheese <type>  Add the specified type of cheese [marble]
+  -c, --cheese <type>  Add the specified type of cheese (default: "marble")
   -C, --no-cheese      You do not want any cheese
+  -h, --help           output usage information
 ```
 
 ## Custom help
--- a/examples/pizza
+++ b/examples/pizza
@@ -12,7 +12,7 @@ program
   .option('-p, --peppers', 'Add peppers')
   .option('-P, --pineapple', 'Add pineapple')
   .option('-b, --bbq', 'Add bbq sauce')
-  .option('-c, --cheese <type>', 'Add the specified type of cheese [marble]')
+  .option('-c, --cheese <type>', 'Add the specified type of cheese', 'marble')
   .option('-C, --no-cheese', 'You do not want any cheese')
   .parse(process.argv);
 
@@ -21,9 +21,7 @@ if (program.peppers) console.log('  - peppers');
 if (program.pineapple) console.log('  - pineapple');
 if (program.bbq) console.log('  - bbq');
 
-var cheese = true === program.cheese
-  ? 'marble'
-  : program.cheese || 'no';
+var cheese = !program.cheese ? 'no' : program.cheese;
 
 console.log('  - %s cheese', cheese);
 console.log(program.args);
--- a/package-lock.json
+++ b/package-lock.json
@@ -2276,7 +2276,7 @@
       "dependencies": {
         "minimist": {
           "version": "1.2.0",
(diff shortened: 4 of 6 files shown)


index.js
```diff
@@ -60,9 +60,7 @@ function Option(flags, description) {
  */
 
 Option.prototype.name = function() {
-  return this.long
-    .replace('--', '')
-    .replace('no-', '');
+  return this.long.replace(/^--/, '');
```
usmonster: Note: This may be a breaking change for people who use the (only recently-documented) [Custom Event Listeners](https://github.com/tj/commander.js#custom-event-listeners), though I can also clarify this by adding another illustrative example to the Readme in that section, if required.

usmonster:
Bump–@vanesyan, does this look OK for you?

vanesyan:
Yes, everything is ok for me, but cannot merge this until the next major bump as it's breaking change.

shadowspawn:
Good description of problem and changes, thanks @usmonster 

I am hopeful this will tidy up `--no-` options! I have done a visual inspection of changes. I have not tested runtime yet.

Also good catch that the JSDoc was claiming default value was true rather than undefined, oops.

shadowspawn:
@usmonster 
Thanks for updating PR, and comments in #939.

The change to the custom event listener makes this a breaking change for a major version, so I have been working on some other pull requests first.

Good to know you are still around and interested, makes this one high on my list. :-)

shadowspawn:
(In top two of my PR queue. Hopefully take another look this week.)

shadowspawn:
I have been trying all the permutations by running programs to check behaviour, as the code itself is quite subtle. One request, one comment, one thing I am still  wondering about, one future change. (Sorry, I do some long comments!)

Overall, I think this will make `--no-` useful and usable in combinations, and worth the pain of a breaking change. Thanks! I think the original vision was just for a stand-alone negated boolean, but people quite reasonably want and try to use it in combination with all forms of options. And are disappointed.

1) Request: I think in practice the option with a required value (e.g. `<value>`) should be people's first choice, and the option with optional value (e.g. `[value]`) is less useful and harder to handle. So for the pizza example, I would like `cheese` to be required rather than optional. (But the existing example and output are inconsistent, well spotted!)

2) Musing: I am wondering about `name in opts`. This is the only use of operator `in` and stuck out to me. Would testing `.optionFor` be appropriate and more consistent with existing code?

3) Breaking note. This is a breaking change to the default values of options when no command line arguments are supplied (and other situations, but this is the big one). I think the new behaviour is a reasonable one to explain, simple story: declaring `--no-foo` after `--foo` does not change default value of `foo`.

```js

program
  .option('-c, --cheese <type>', 'Add the specified type of cheese', "marble")
  .option('-C, --no-cheese', 'You do not want any cheese')
  .option('--what <type>')
  .option('--no-what');

program.parse(process.argv);
console.log(`cheese is ${program.cheese}`);
console.log(`what is ${program.what}`);
```

```
# Before:
$ node required.js
cheese is true
what is true

# After:
$ node required.js
cheese is marble
what is undefined
```

4) After: I think we should allow a default value for pure flags. This would allow an externally supplied default which can be overridden on command line. Fixing up `--no-*` is first though and better to deal with one thing at a time, so I am not suggesting you do this in this PR.

```
program
  .option('--what <type>', 'enable on command line', process.env.WHAT)
  .option('--no-what', 'disable on command line');
```

slavafomin:
The existence of the negated option shouldn't make `true` the default value for non-negated one. This is just a bad design and a very non-intuitive behavior.

If we say "user can ask to make pizza without a cheese" it doesn't mean, that if she didn't ask it, we have to always make pizza with cheese. It's a twisted logic.

The much better API would be:
- if option is specified without a value, i.e. `--cheese` it would be resolved as `true`
- If it's specified with `--no` prefix, i.e. `--no-cheese` it should resolve to `false`
- if it's omitted, then it should be `undefined`
- if value is specified, e.g.: `--cheese cheddar`, then it should return a string `cheddar`.

With such approach, the developer don't have to define negated option specifically at all, instead, every  option could be negated. I believe such API would be straightforward, intuitive and easy to work with.

And if we really want to have a default value, then the developer should provide it, we shouldn't make this choice for her, e.g.: `.option('cheese').defaultsTo('mozzarella')` or `.option('cut').defaultsTo(true)`.

slavafomin:
Please, take a look at [my comment here](https://github.com/tj/commander.js/issues/979#issuecomment-504747736).

usmonster:
Thanks for the comment, @slavafomin. I replied in the thread.

Belated thanks, @shadowspawn, for the thorough review and feedback! No need to apologize for long comments, as they are appreciated. :) I'll reply to each below:

1. Totally agree, updated the example!
2. I'm all for consistency, but `.optionFor` doesn't really work in a way that makes it useful in this case. I think `name in opts` is exactly the right tool for the job in this case, so I'll leave it as-is, if you don't mind?
3. Just to clarify, would you like me to document a new example in the readme, or update my commit message to include your example, or is it enough to just add a comment to the pizza example with the new expected output? Personally, I think the changelog/release notes should be enough if you prefix the commit subject with `[BC BREAK]`, but OK for me if you also want something more explicitly documenting the new behavior.
4. I understand what you mean, and I read through #928 to see the use cases, but I'm still not convinced about its utility. Based on the example you gave in the above comment, I think there's already a pretty straightforward way to do that:
```js
const handleWhat = function() {
  // negate `what` if default isn't `true`
  if ('WHAT' in process.env && process.env.WHAT !== 'true') {
    this.what = !this.what;
  }
};

program
  .option('--what', 'enable on command line')
  .option('--no-what', 'disable on command line')
  .on('option:what', handleWhat)
  .on('option:no-what', handleWhat)
  .parse(process.argv);

console.log(`what is ${program.what}`);
```
It can even be done without event listeners by doing the conditional reassignment at some point after the `.parse`. In any case, if enough people want this to be a built-in feature, I'll let someone else make that PR. ;)

Thanks in advance for your feedback, especially on my question in point 3. :)

shadowspawn:
1. Thanks. :-)

2. What about `{}.hasOwnProperty.call(opts, name)`instead then? I am paranoid about (more) conflicts with built-in properties. (I have not tested runtime, suggesting this in theory.)

e.g. https://eslint.org/docs/rules/guard-for-in

3. No changes from you needed thanks. I was thinking out loud and making a note to myself, it isn't just the event listener change that is breaking. I'll update the README section on options and comfortable doing that myself after carefully crafting the current coverage. 

4. I have the ball with that one. I might invite you to review if I get a candidate Pull Request done for v3. 😄

index.js
```diff
@@ -394,8 +395,11 @@ Command.prototype.option = function(flags, description, fn, defaultValue) {
 
   // preassign default value only for --no-*, [optional], or <required>
   if (!option.bool || option.optional || option.required) {
-    // when --no-* we make sure default is true
-    if (!option.bool) defaultValue = true;
+    // when --no-foo we make sure default is true, unless a --foo option is already defined
+    if (!option.bool) {
+      var opts = self.opts();
+      defaultValue = name in opts ? opts[name] : true;
```
shadowspawn: Exploring alternatives to using plain `in`

usmonster: Sounds good! I'll use `Object.prototype.hasOwnProperty.call(opts, name)`.

shadowspawn:
Thanks for switching base branch! I have not been asking people to do that in case it creates problems, but does make it easier for me. :-)

shadowspawn:
Starting merge.

shadowspawn:
Updated example and README with new improved behaviour, and added entries to CHANGELOG

This is one of my favourite changes in v3. 😄 

usmonster:
Thanks for your proactive involvement!

shadowspawn:

shadowspawn:

PR #576, opened 2016-09-21
Make process.exit() injectable.

bolinfest:
This is an initial change to help fix https://github.com/tj/commander.js/issues/575.

Issue #575, opened 2016-09-21
Make calls to process.exit() overridable


bolinfest:
In general, library code should never call `process.exit()`. Instead, clients of the library should decide how they want to handle errors.

In particular, tying `process.exit(0)` to `--help` seems particularly bad. For a CLI parser, this may be a reasonable default, but this behavior should at least be overridable.

I am trying to build a daemon running on a UNIX domain socket that takes requests and uses commander to parse arguments just as it would if it were running in non-daemon mode. Given the way commander works today, passing `--help` has the side-effect of tearing down my daemon.

I have similar issues where commander makes calls to `process.stdout.write()` that are writing to the "wrong" stdout, from my perspective. I think the best way to secure things as a library is to remove all direct references to `process` from `index.js` and make all of its behavior injectable. It can still default to what it does today, but it should be overridable.

bolinfest:
The categorical call to `outputHelpIfNecessary()` (which calls `process.exit(0)`) from `Command.prototype.parseArgs` and `Command.prototype.action` is particularly hard to override from the outside.

cdaringe:
- i agree.  i feel that the issue name here could be changed to something more akin to "don't exit on `.help()` call"
- further, i visited the issues section here today specifically because i run `help()` when something _wrong_ has been entered.  hence, i have strong feelings that the exit code should be **non-zero**, wherein now it is 0.

of course this is opinion.  @tj, are you opposed to de-opinionification `.help()`s behavior?

cdaringe:
here's a [ref](https://github.com/tj/commander.js/blob/9aad0d01baf93d5e509c267da44f6be9695045fd/index.js#L1052) to the offending code

dovidweisz:
Doesn't the outputHelp() method cover these situations?

NearAutomata:
Even though this issue is old I stumbled upon a similar issue where I am having some trouble. I am using `inquirer` and `inquirer-command-prompt` to create an interactive CLI application and wanted to use `commander` to parse options by continuously calling `parse(userInput)` in my CLI prompt.

Unfortunately though, when I run `parse` for a command that is missing a required argument the application writes `error: missing required argument 'cmd'` to the console and then invokes `process.exit` which exits my own application!

@tj I'd like to suggest a change wherein all calls to `process.exit` are replaced by an overridable function that can be freely configured. In my case I would replace it with an empty function in order to display status messages but prevent exiting.

jacksdrobinson:
[...]

PR #800, opened 2018-05-10
Make exit calls overridable by introducing a customizable environment

NearAutomata:
After reading #575 and #576 I combined the latter in a new version of the PR. As described in the issue I am leveraging the capabilities of `commander` to parse an run commandlets in an interactive CLI using `inquirer` and `inquirer-command-prompt`. Unfortunately `commander` is doing `process.exit` calls when it is not supposed to and there was no option to override that in the first place.

Libraries aren't supposed to make calls to `process.exit` because it's way out of their realm of responsibility. The application that uses them is responsible for deciding when to exit thus libraries should throw errors instead.

As `commander` can be considered almost standalone I left the default calls in there to not break any code but introduced the possibility to overwrite the default behavior which allows me to use the library as a helper for parsing and executing commandlets.

### DIFF
--- a/index.js
+++ b/index.js
@@ -323,19 +323,21 @@ Command.prototype.action = function(fn) {
     // Leftover arguments need to be pushed back. Fixes issue #56
     if (parsed.args.length) args = parsed.args.concat(args);
 
+    var hasIssue = false;
     self._args.forEach(function(arg, i) {
       if (arg.required && args[i] == null) {
         self.missingArgument(arg.name);
-        return;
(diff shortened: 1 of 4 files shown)


NearAutomata:
There is currently an issue where the command still gets executed even if it's not supposed to, I will be trying to fix this.

NearAutomata:
I can't seem to get it to work since everytime I change something I notice another part of your code breaking. I am considering to do a different setup for the time being.

PR #804, opened 2018-05-20
Fix types

blair:
Add some TypeScript improvements.

### DIFF
--- a/typings/index.d.ts
+++ b/typings/index.d.ts
@@ -275,7 +275,7 @@ declare namespace local {
      *
      * @param {(str: string) => string} [cb]
      */
-    help(cb?: (str: string) => string): void;
+    help(cb?: (str: string) => string): never;
   }
 
 }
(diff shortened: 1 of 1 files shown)


typings/index.d.ts
```diff
@@ -275,7 +275,7 @@ declare namespace local {
      *
      * @param {(str: string) => string} [cb]
      */
-    help(cb?: (str: string) => string): void;
+    help(cb?: (str: string) => string): never;
```
vanesyan: @blair `never` type is incorrect for help function, it doesn't throw or failing to terminate.
https://blog.mariusschulz.com/2016/11/18/typescript-2-0-the-never-type

blair: @vanesyan According to https://github.com/tj/commander.js/#helpcb the method "Output help information and exit immediately" so it terminates the process.

vanesyan: Sorry, you're correct. `help` terminates process.

vanesyan:
Thanks!

PR #808, opened 2018-05-23
Update Readme.md

tgsoverly:
Added an example with an command option value


vanesyan:
I think, we can unify it under the one example block above. Anyway, options with arguments have already been mentioned in the "Option parsing" paragraph. We also can put link to examples folder, somewhere at the top of README, so people can discover it easier.

PR #820, opened 2018-06-27
Make 'npm test' run on Windows

bzoz:
Uses Node to run the tests. Running `make test` will still work.

### DIFF
--- a/test/run.js
+++ b/test/run.js
@@ -1,0 +1,22 @@
+#!/usr/bin/env node
+
+const { spawnSync } = require('child_process')
+const { readdirSync } = require('fs')
+const { extname, join } = require('path')
+
+process.env.NODE_ENV = 'test';
+
+process.stdout.write('\n')
+readdirSync(__dirname).forEach((file) => {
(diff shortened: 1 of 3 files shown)


test/run.js
```diff
@@ -0,0 +1,22 @@
...
+const { readdirSync } = require('fs')
+const { extname, join } = require('path')
+
+process.env.NODE_ENV = 'test';
+
+process.stdout.write('\n')
+readdirSync(__dirname).forEach((file) => {
+  if (!file.startsWith('test.') || extname(file) !== '.js')
+    return;
+  process.stdout.write(`\x1b[90m   ${file}\x1b[0m `);
```
vanesyan: Does Windows support ANSI colors in terminal?

bzoz: Yes

bzoz: Since Win10 terminal control sequences are supported natively by the console. For previous Windows versions Node (libuv to be more precise) has special emulation layer that translates the codes to winapi calls.

PR #828, opened 2018-07-17
Allow custom usage message to override program name for git-style sub-commands

holmrenser:
This pull request changes the behaviour of the ```usage``` method so that a custom usage message can override the program name.

This is a workaround to correctly display the usage of nested git style sub-commands. 

Example:

foo.js
```js
program
    .command('bar [command]', 'Run bar command')
    .command('qux [command]', 'Run qux command')
````

foo-bar.js
```js
program
    .command('baz [options]')
```

foo-bar-baz.js
```js
program
    .description('This program can be run using git-style sub-commands "foo bar baz"')
    .option('-t, --test', 'Test command')
```

This program can be run using git-style sub-commands as follows: ```foo bar baz```. 
However, the usage information will use the filename and display something like the following:

```
Usage: foo-bar-baz [options]
```

This pull-request allows the user to fully override the usage information, so that the correct usage information for nested git-style sub-commands can at least be correctly added manually. 

I have not yet figured out a way to automatically detect the use of nested git-style sub-commands.

### DIFF
--- a/index.js
+++ b/index.js
@@ -919,7 +919,12 @@ Command.prototype.usage = function(str) {
     return humanReadableArgName(arg);
   });
 
-  var usage = '[options]' +
+  var cmdName = this._name;
+  if (this._alias) {
+    cmdName = cmdName + '|' + this._alias;
+  }
+
+  var usage = cmdName + ' [options]' +
     (this.commands.length ? ' [command]' : '') +
     (this._args.length ? ' ' + args.join(' ') : '');
 
@@ -1105,10 +1110,6 @@ Command.prototype.helpInformation = function() {
     }
   }
 
-  var cmdName = this._name;
-  if (this._alias) {
-    cmdName = cmdName + '|' + this._alias;
-  }
   var usage = [
     '',
     '  Usage: ' + cmdName + ' ' + this.usage(),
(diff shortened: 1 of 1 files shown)


shadowspawn:
Thanks for the description. Are there open issues this resolves?

I see the possibilities for new code, but is this going to required updating every existing program with custom `.usage`?

> I have not yet figured out a way to automatically detect the use of nested git-style sub-commands.

Yes, good comment! A possibility might be to stick something in process.env, but a bit special.

holmrenser:
@shadowspan I ran into this myself, but I don't think I opened an issue for this. In any case, I don't think existing programs will have to be refactored to accommodate this change.

shadowspawn:
For interest, there is a PR open for improving subcommand help for action based handlers: #740

shadowspawn:
I confirmed the change in behaviour I was concerned about for an existing custom usage:

```
program.usage("xyz");
```

and `node index.js --help` produces:

```
before: Usage: index xyz
after: Usage: xyz
```

But no matter, as I think there may be support for supplying desired command name to get consistent behaviour however called?

```
program.usage("xyz").name("abc");
```

```
Usage: abc xyz
```

(I am only experimenting in a single toy program but think this will apply.)

shadowspawn:
(And to be fair, `.name` has not been documented in README. Sigh. One of my current goals is improving coverage in the README!)

shadowspawn:
I think the original problem is hopefully covered by `.name()`. No feedback in a month. 

PR #829, opened 2018-07-23
Document default vs. coercion function order

edemaine:
Add examples and descriptive text about the effects of putting the default argument before the coercion function and vice versa.  Basically, if the default goes before the coercion function, then the function gets no second argument, and you can use coercion functions like `parseInt`.  But if the default goes after the coercion function, then the function gets a second argument, useful for accumulation.  The word "default" is essentially not mentioned in the current readme, so I didn't even realize that it was a feature for a while.

I had to discover this the hard way, and open issue #686 suggested it was a bug, possibly fixed by #757, but I believe that PR changes the behavior.  This PR just documents the existing behavior, which seems pretty functional.

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -105,6 +105,8 @@ A command's options are validated when the command is used. Any unknown options 
 
 ## Coercion
 
+Options can have a default value and/or a coercion function (usually in that order) passed in after the description string.  For example, a coercion function of `parseInt` or `parseFloat` will transform arguments into integers or floats, respectively.  If the default value is specified after the coercion function, then the function can take a second argument which is the current argument value so far (starting with the default value), which enables accumulating repeated arguments e.g. into an array.
+
 ```js
 function range(val) {
   return val.split('..').map(Number);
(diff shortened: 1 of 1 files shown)


shadowspawn:
From code inspection and experimentation, unfortunately I don't think changing the order fixes the problem. When the default value is specified before the function, the function is ignored.

https://github.com/tj/commander.js/blob/291fc04a405605e056b4e31af64927b8f811f0d7/index.js#L389

shadowspawn:
I am going to have another go at documenting the existing behaviour.

shadowspawn:
The options coverage in the README has been updated and includes new description of custom option processing, including a working parseInt example.

https://github.com/tj/commander.js#custom-option-processing

edemaine:
@shadowspawn Great!

Did you want to document the behavior that a value before a function acts as a default? (whereas only values after the function get passed to the function)  Or did this behavior change?

shadowspawn:
The value before a function still ends up as default with function ignored. I am hoping the new documentation means people less likely to try that combination with the rewritten custom option processing.

Issue #832, opened 2018-07-29
Help does not output newline after the help message


jaredpetersen:
Help should output a newline after the message.

Expected:
```
$ ./cli.js

  Usage: my-app [options] [command]

  Options:

    -v, --version    output the version number
    -h, --help       output usage information

  Commands:

    start [options]  start the app

$ 
```

Actual:
```
$ ./cli.js

  Usage: my-app [options] [command]

  Options:

    -v, --version    output the version number
    -h, --help       output usage information

  Commands:

    start [options]  start the app
$ 
```

Source:
```javascript
#!/usr/bin/env node

'use strict';

const program = require('commander');
const version = require('./package.json').version;

process.title = 'my-app';

program
  .name('my-app')
  .version(version, '-v, --version');

program
  .command('start')
  .description('start the app')
  .option('-d, --directory <directory>', 'directory for the thing')
  .option('-p, --port <port>', 'port number for the thing', parseInt)
  .action(() => console.log('started'));

program.parse(process.argv);

if (!program.args.length) program.help();
```

Somewhat related to #732 but the description newline issue has been fixed since 2.11.0 but the line after options/commands has not.

jaredpetersen:
This also is a problem for subcommands like:

```
$ ./cli.js start --help

  Usage: start [options]

  start the app

  Options:

    -d, --directory <directory>  directory for the thing
    -p, --port <port>            port number for the thing
    -h, --help                   output usage information
$

mojavelinux:
Why is this desirable? Now there are extra lines at the end of the output. No program I know of on Linux adds a blank line at the end of the help text. (e.g., `docker help version`). Now commander is producing non-standard output.

jaredpetersen:
@mojavelinux -

All of the output examples in the README show a newline at the end of the help output and were incorrect before the pull request. Additionally, the example in the Custom Help section of the README did not work without the change. The documentation could easily be modified, but I assumed it was desirable since the newline at the end of the output was documented.

I personally think the change is more visually appealing since it evens out the starting newline (before the "Usage" text).

I agree that there should not be a newline after normal program output. There does appear to be a standard there and it's shown in the README. However, I don't think there really is a standard when it comes to help output.

Here's a small survey of how a few programs handle help:
* Docker: 1 line before help, 0 lines after help
* Node.js: 0 lines before help, 0 lines after help
* FFmpeg: 0 lines before help, 2 lines after help
* Raspivid: 0 lines before help, 1 line after help

[...]

PR #833, opened 2018-07-29
Fixes newline output after help information

jaredpetersen:
Fixes issue #832.


jaredpetersen:
@vanesyan You can't just add it to the commands array, as doing so will mean that there won't be a newline after commands like `mycli somecommand --help` or `mycli --help` (without any configured commands) where options are the last outputted value.

vanesyan:
Yeah, sorry, you are correct.

jaredpetersen:
Thank you! How often do releases go out?

PR #849, opened 2018-08-17
add support for execute typescript subcommand via ts-node

KillWolfVlad:
`Command.prototype.executeSubCommand` doesn't support `.ts` extension for scripts. It is bad if you want run something via [ts-node](https://github.com/TypeStrong/ts-node).

https://github.com/tj/commander.js/blob/e5b27cc553c0c55eb2f8890dc83034d3a3eee531/index.js#L526

First, hardcoded `.js` extension is bad. I suggest use real file extension.

https://github.com/tj/commander.js/blob/e5b27cc553c0c55eb2f8890dc83034d3a3eee531/index.js#L544-L546

Second, I suggest use similar block of code for check exists file with `.ts` extension. I don't think that this duplication is serious.

### DIFF
--- a/index.js
+++ b/index.js
@@ -523,7 +523,7 @@ Command.prototype.executeSubCommand = function(argv, args, unknown) {
   // executable
   var f = argv[1];
   // name of the subcommand, link `pm-install`
-  var bin = basename(f, '.js') + '-' + args[0];
+  var bin = basename(f, path.extname(f)) + '-' + args[0];
 
   // In case of globally installed, get the base dir where executable
   //  subcommand file should be located at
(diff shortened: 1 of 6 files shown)


joeferraro:
Exactly what I was looking for, thanks @abetomo

vanesyan:
LGTM

taschmidt:
Should this be working in commander v2.20.0? Do I need to do anything to make it work? It works with .js files but I can't get it to work once I switch them to .ts. The top-level one works but my 'cmd-subcommand' shows nothing.

KillWolfVlad:
> Should this be working in commander v2.20.0?

Yes.

> Do I need to do anything to make it work?

No.

> It works with .js files but I can't get it to work once I switch them to .ts. 

Just have directure structure like this (see [test fixture](https://github.com/tj/commander.js/tree/master/test/fixtures-ts)):

```
.
└── bin
    ├── pm-install.ts
    ├── pm-remove.ts
    └── pm.ts
```

Thereafter you can execute script:

```bash
node -r ts-node/register bin/pm.ts
```

> The top-level one works but my 'cmd-subcommand' shows nothing.

Can you reproduce your issue?

taschmidt:
Thanks for the reply, @KillWolfVlad! I fixed my problem although I'm not really sure why. I'm doing a #! script and it wasn't working with `#!/bin/usr/env ts-node`. But based on your response, I tried `#!/bin/usr/env node --require ts-node/register` and it works!

davidonlaptop:
These solutions didn't work for my use case, nothing was getting executed and no errors were thrown.

But here's something that worked:

Changing [line 795](https://github.com/tj/commander.js/blob/master/index.js#L795) from
```js
proc = spawn(process.argv[0], args, { stdio: 'inherit' });
```
to
```js
proc = spawn(args.shift(), args, { stdio: 'inherit' });
```

However, I don't know if this would break something else, I just thought to post this here if others encounter the same problem.

davidonlaptop:
I've submitted a new PR: https://github.com/tj/commander.js/pull/1243

Issue #532, opened 2016-05-11
Allow user to specify which script a subcommand should call


matthewadams:
If the script a user wants to execute as a subcommand doesn't abide by the naming convention that's documented (like [`pm install` looks for script `pm-install`](https://github.com/tj/commander.js/#git-style-sub-commands)), the user should be able to specify the script that's to be executed.   For example:

```
// file: ./examples/pm
var program = require('..');

program
  .version('0.0.1')
  .command('install [name]', 'install one or more packages', 'myinstall') // (1)
  .command('search [query]', 'search with optional query', '../others/q.js') // (2)
  .command('list', 'list packages installed', {isDefault: true}, '/home/me/foo/bar/mylist.js') // (3)
  .parse(process.argv);
```

In `(1)`, commander would assume the same directory as the current file, look for `./myinstall`, then if not found,  append `.js`, resulting in `./myinstall.js`.
In `(2)`, commander would use the path relative to the current file.  No `.js` appended since there's already a `.js` extension.
in `(3)`, commander would use the absolute path.  No `.js` appended since there's already a `.js` extension.

b4dnewz:
2 years... will this issue never be addressed? :disappointed: it is an extremely useful feature that you can use an alternative naming convention or specify a file

abetomo:
Thanks for your suggestion.
I will add features.

shadowspawn:
Pull Request #532 adds an option for this.

I am interested in feedback on `executableFile` as the option name?

e.g.
```
program
.command("example", "example description", { executableFile: "myExampleCommand" });
```

shadowspawn:
Pull Request being considered for v3 is #999

shadowspawn:
This issue will be resolved when v3.0.0 is released. Available now as a prerelease. See #1001

shadowspawn:
`.command` now supports the `executableFile` option (thanks to @abetomo). 
Shipped in v3: https://github.com/tj/commander.js/releases/tag/v3.0.0

PR #854, opened 2018-08-30
Implement specification of subcommand

abetomo:
#532

### DIFF
--- a/index.js
+++ b/index.js
@@ -182,6 +182,7 @@ Command.prototype.command = function(name, desc, opts) {
     if (opts.isDefault) this.defaultExecutable = cmd._name;
   }
   cmd._noHelp = !!opts.noHelp;
+  cmd._specifySubcommand = opts.subcommand;
   this.commands.push(cmd);
   cmd.parseExpectedArgs(args);
   cmd.parent = this;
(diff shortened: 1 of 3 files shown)


shadowspawn:
To have behaviour consistent, I think we should calculate `isExplicitJS` when the option supplied name has an explicit suffix. (An easy way would be to call basename and fall through existing lookup code, but might be more appropriate to check the option supplied name directly.) 

The behaviour I see is:

```
file        option      spawn
foo-bar                 spawn foo-bar
foo-bar.js              spawn node foo-bar.js
foo-bar.js  foo-bar     spawn node foo-bar.js (ok)
foo-bar.js  foo-bar.js  spawn foo-bar.js (surprise! did not detect file suffix)
```

shadowspawn:
I don't think we need to wait any longer for feedback on `executableFile`, and I am still happy with that if ok with you.

shadowspawn:
Would you like example code for the `isExplicitJS` feedback?

shadowspawn:
This is being considered for v3, but some changes suggested. I am willing to help with code if wanted (in which case I'll get it reviewed before merging).

shadowspawn:
Started work on implementing my suggestions

shadowspawn:
Closing in favour of #999 which builds on this pull request.

shadowspawn:

Issue #855, opened 2018-09-02
[BUG] Help output broken for options with falsey default values


plroebuck:
## Description
Help output for options with falsey default values don't show up.

## Steps to Reproduce
Save the following script as "commander.spec.js".

```js
#!/usr/bin/env node
var program = require('commander');

function list(val) {
  return val.split(',');
}

program
  .option('-s, --str [value]', 'Specify string value', String, '')
  .option('-l, --list [items]', 'Specify list items', list, [])
  .parse(process.argv);
```

Make the script executable and invoke it as follows to display help.
```bash
$ chmod ugo+x commander.spec.js
$ ./commander.spec.js --help
```
### Expected Behavior
```console
  Usage: commander.spec [options]

  Options:
    -s, --str [value]               Specify string value (default: '')
    -l, --list [items]              Specify list items (default: [])
    -h, --help                      output usage information
```
### Actual Behavior
```console
  Usage: commander.spec [options]

  Options:
    -s, --str [value]               Specify string value (default: )
    -l, --list [items]              Specify list items (default: )
    -h, --help                      output usage information
```
## Version Information
* OS: macOS 10.13.6 (but verified elsewhere as well)
* Shell: bash
* Node: v10.3.0
* Transpiler: N/A

### Where
[Command.prototype.optionHelp](https://github.com/tj/commander.js/blob/713db77d5744a9274ee3812400243ffaae3b7abc/index.js#L1056)

### Affected releases
v2.12.0 through current (v2.17.1).

### Applicable Issues
#122
mochajs/mocha#3433

PR #856, opened 2018-09-03
Fix to change default value to string

abetomo:
fixes: #855


plroebuck:
@vanesyan ping again!

plroebuck:
@vanesyan when can Mocha expect a release containing this fix?

PR #862, opened 2018-09-17
Bugfix - Error output

### LINKED ISSUE #861: Standardize error output
Following up on #853, the error output should be standardized as well. Right now it does this:
```
$ ./cli.js -x

  error: unknown option `-x'

$
```

But it should be this:
```
$ ./cli.js -x
error: unknown option `-x'
$
```
Same is true for missing/wrong arguments.

jaredpetersen:
Removes indentation and newlines around error output.

Fixes #861.

### DIFF
--- a/index.js
+++ b/index.js
@@ -792,9 +792,7 @@ Command.prototype.opts = function() {
  */
 
 Command.prototype.missingArgument = function(name) {
-  console.error();
   console.error("error: missing required argument `%s'", name);
-  console.error();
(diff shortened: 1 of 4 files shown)


vanesyan:
Can you also prefix messages with `error` on  https://github.com/tj/commander.js/blob/master/index.js#L583 https://github.com/tj/commander.js/blob/master/index.js#L585 lines, please

PR #865, opened 2018-09-18
simplify bit of `parseExpectedArgs`

ZYSzys:
`argDetails.name = arg.slice(1, -1);` whether `arg[0]` equals `<` or `[`, so just extract it to reduce the codebase.

### DIFF
--- a/index.js
+++ b/index.js
@@ -1205,7 +1205,7 @@ function outputHelpIfNecessary(cmd, options) {
 }
 
 /**
- * Takes an argument an returns its human readable equivalent for help usage.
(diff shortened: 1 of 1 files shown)


shadowspawn:
Thanks for the suggestion @ZYSzys 

The current code is being paranoid and not assuming the the arguments only start with `<` or `[`. I am not sure there are any valid use cases, but I don't want to take the risk of breaking existing code just to make the code more compact without a higher level reason.

For example someone can write:

```
.arguments("<foo> hello <bar>")
```

Thanks for your contribution.

PR #869, opened 2018-09-28
follow links

### LINKED ISSUE #866: Make commander based npm packages compatible with homebrew install
I think this is an issue for this package. I mentioned some details here: https://github.com/Homebrew/homebrew-core/pull/32154#pullrequestreview-155949702

I try to create a homebrew Formula to install gitlab-time-tracker, however, after installation it doesn't work. It seems this package isn't able to determine the actual folder to use correctly. Maybe the symlinks should be followed in a loop to find the actual src folder (if that's possible)?

With the Formula install:

```
$ gtt config
baseDir:/usr/local/Cellar/gitlab-time-tracker/1.7.27/bin
gtt-config(1) does not exist, try --help
```

With npm install -g:

```
$ gtt config
baseDir:/usr/local/lib/node_modules/gitlab-time-tracker/src
```

KillWolfVlad:
yarn ~v1.10.x changed bin linking for global packages (I can't find commit with that changes).

How yarn link bin after `$ yarn global add package`?

```bash
$ file ~/.yarn/bin/package
~/.yarn/bin/package: symbolic link to ../../.config/yarn/global/node_modules/.bin/package
```

```bash
$ file ~/.config/yarn/global/node_modules/.bin/package
~/.config/yarn/global/node_modules/.bin/package: symbolic link to ../package/bin/package.js
```

But `commander.js` read link only once time. And sub-commands not work,  because in directory `~/.config/yarn/global/node_modules/.bin` other files doesn't exists.

My solution: follow links and get real file name.

In GNU/Linux shell command `readlink -f` implements it:

```bash
$ readlink -f ~/.yarn/bin/package
~/.config/yarn/global/node_modules/package/bin/package.js
```

For Node.js I found package [readlink](https://www.npmjs.com/package/readlink) with almost similar behavior (from description).

Close #866

### DIFF
--- a/index.js
+++ b/index.js
@@ -8,6 +8,7 @@ var path = require('path');
 var dirname = path.dirname;
 var basename = path.basename;
 var fs = require('fs');
+var readlink = require('readlink').sync;
 
 /**
  * Inherit `Command` from `EventEmitter.prototype`.
@@ -527,13 +528,12 @@ Command.prototype.executeSubCommand = function(argv, args, unknown) {
 
   // In case of globally installed, get the base dir where executable
   //  subcommand file should be located at
-  var baseDir,
-    link = fs.lstatSync(f).isSymbolicLink() ? fs.readlinkSync(f) : f;
+  var baseDir, link = f;
 
-  // when symbolink is relative path
-  if (link !== f && link.charAt(0) !== '/') {
-    link = path.join(dirname(f), link);
+  while (fs.lstatSync(link).isSymbolicLink()) {
+    link = readlink(link);
   }
+
   baseDir = dirname(link);
 
   // prefer local `./<bin>` to bin in the $PATH
--- a/package-lock.json
+++ b/package-lock.json
@@ -173,6 +173,14 @@
       "integrity": "sha1-iYUI2iIm84DfkEcoRWhJwVAaSw0=",
       "dev": true
     },
+    "async": {
+      "version": "2.6.1",
+      "resolved": "https://registry.npmjs.org/async/-/async-2.6.1.tgz",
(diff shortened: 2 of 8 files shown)


KillWolfVlad:
Rebased with master (previous commit [here](https://github.com/KillWolfVlad/commander.js/tree/feature/link-to-link-prev1))

KillWolfVlad:
Rebased with master (previous commit [here](https://github.com/KillWolfVlad/commander.js/tree/feature/link-to-link-prev2))

index.js
```diff
@@ -8,6 +8,7 @@ var path = require('path');
 var dirname = path.dirname;
 var basename = path.basename;
 var fs = require('fs');
+var readlink = require('readlink');
```
vanesyan: Can you please inline the synchronous part of the `readlink` module; as module itself introduces an `async` module which is unnecessary for the commander and only introduce a ton of unused kb.

KillWolfVlad: I don't think that <kbd>CTRL+C</kbd>+<kbd>CTRL+V</kbd> code from `readlink` is good idea, because we must also provide tests and support this legacy code. If don't like `async` module maybe better send PR to `readlink` repo to remove it from dependencies.

shadowspawn:
Whether to introduce dependencies is a tricky question. I'm still thinking about the tradeoffs so won't offer an opinion at this time, just add some size info.

```
$ npm init -y
$ npm install commander
$ du -d 1 -h node_modules/
 76K	node_modules//commander
 76K	node_modules/
$ npm install readlink
$ du -d 1 -h node_modules/
4.8M	node_modules//lodash
 76K	node_modules//commander
844K	node_modules//async
 24K	node_modules//readlink
5.8M	node_modules/
```

KillWolfVlad:
fixed in #935

shadowspawn:
Thanks @KillWolfVlad

Issue #47, opened 2012-01-31
A way to override the version and help message


jeromedecoster:
Hello,

It can be nice to have the possibility to modify the default version message 'output the version number'
and the default help message 'output usage information'

Thanks

fegemo:
+1 to this

digitalsadhu:
+1

zhiyelee:
Sound good. PR on this is welcome

hb20007:
It would also be nice to be able to override error messages, e.g., for custom formatting of messages like "error: missing required argument `name'".

shadowspawn:
This issue will be resolved when v3.0.0 is released. Available now as a prerelease. See #1001

shadowspawn:
Support for customising the version and help descriptions has shipped in v3 (thanks @idmontie): https://github.com/tj/commander.js/releases/tag/v3.0.0

Closing the oldest open bug. Woop!

PR #870, opened 2018-09-28
Enable custom help description messages for -V and -h

zentus:
### Enable custom help description messages for -V and -h

* Add description argument to .version()

* Add command .helpDescription()

For overriding default -h and -V flag help descriptions.

* Capitalize default help descriptions
	-V: 'Output version number'
	-h: 'Output usage information'

* Add example (help-description.js)

* Add test (test.command.helpDescription)

* Update Readme

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -254,8 +254,8 @@ Usage: pizza [options]
 An application for pizzas ordering
 
 Options:
-  -h, --help           output usage information
-  -V, --version        output the version number
+  -h, --help           Output usage information
+  -V, --version        Output the version number
   -p, --peppers        Add peppers
   -P, --pineapple      Add pineapple
   -b, --bbq            Add bbq sauce
@@ -308,8 +308,8 @@ Yields the following help output when `node script-name.js -h` or `node script-n
 Usage: custom-help [options]
 
 Options:
-  -h, --help     output usage information
-  -V, --version  output the version number
+  -h, --help     Output usage information
+  -V, --version  Output the version number
   -f, --foo      enable some foo
   -b, --bar      enable some bar
   -B, --baz      enable some baz
--- a/Readme_zh-CN.md
+++ b/Readme_zh-CN.md
@@ -192,8 +192,8 @@ Usage: pizza [options]
 An application for pizzas ordering
 
 Options:
-  -h, --help           output usage information
-  -V, --version        output the version number
+  -h, --help           Output usage information
+  -V, --version        Output the version number
   -p, --peppers        Add peppers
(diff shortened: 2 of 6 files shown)


HaoChuan9421:
Can't change `-v` and `-h` description is bad practice for localization. Waiting for this PR merged!

shadowspawn:
Just making some notes.

There are requests to be able to change the flags for version and help, conflicts with user defined flags, change the description (like this PR), or the behaviour, or disable the help, et al. I am wondering if the best approach is to allow turning off either of the default `version` and `help` and leave it up to the user to handle it entirely themselves and get whatever behaviour they need. I have not worked through what that would look like to see if it is simple enough to expect people to do it.

shadowspawn:
There is a new Pull Request covering same ground in #963. 

Would you like to continue working on this @zentus, or shall I work with the new Pull Request? I'll wait a week for a reply.

shadowspawn:
Adding reference to related issue: #47

zentus:
@shadowspawn

I am not sure I understand what changes you wish me to make. 

1. No, unfortunately I have no knowledge of TypeScript. Should I keep the `description` argument to .version() the way it is?

2. So, change default descriptions back to lowercase?

3. So I should remove the implementation of .helpDescription() all together? Could you give an example of what a better way of changing the help description for the -h flag would look like?

shadowspawn:
Thanks for reply @zentus. 

The other Pull Request has leaped ahead, so I'll switch to working with it for now. So no action necessary here unless #963 does not work out. 

FYI:
1) Description argument is fine. The _three_ forms of documentation we have for the API are the README (+ example), JSDoc, and TypeScript definition. Arguably a matching test is a fourth. Credit to you for adding README + example + test.
2) In general with Pull Requests it is helpful to do just what is needed for what is being fixed. Avoid making other indentation changes or spelling changes etc, just fix the issue.
3) In the longer term we might want something like `.help(helpFlags, helpDescription, helpOptions)` and I don't want partial solutions that will get in way of a more complete solution. 

Closing in meantime. Will reopen if needed.

shadowspawn:

crystalfp:
Almost there.
Using `.helpOption()` I can localize the help text for `"-h, --help"`. Now if only it was available a `.versionOption()` I could localize also the `-V` help.
This done, only the strings `Usage:` and `Options:` are missing for localization.
Now I'll take a look at the PR.
Thanks!
mario

shadowspawn:
See `.version()` which supports modifying flags and description, like `.helpOption()` does.

Issue #838, opened 2018-08-03
Cannot use inspect in git-style mode


tacnoman:
Hello guys.
Isn't possible to use the debugger with inspect.

Example:

```js
program
  .version(version)
  .command('command1', 'command').alias('c1')
  .parse(process.argv);
```

In file `index-command1.js`:

```js
const a = 1;
debugger
const b = 2;
```

And run: `node --inspect index.js command1`.

I know that the git style is a child_process. But is possible to pass the --inspect to another command?

Thanks for this.

jackellenberger:
I have the same issue.

ericnewton76:
That is how I debug the git-style-subcommands... just point node directly at the subcommand's .js file.

shadowspawn:
Work-around already suggested. 

Another option might be using `NODE_OPTIONS`?
https://nodejs.org/api/cli.html#cli_node_options_options

shadowspawn:
This looks like might be a solution on node side?

nodejs/node#5025

> bind to random port with --debug-port=0

zcfan:
@shadowspawn After reading the link you provide I didn't realize how it can help this situation. Is the --debug-port option assigning a random port to the new created child process?

zcfan:
Oh I didn't notice the reference before your comment. increment debug port looks good to me, finding the subcommand binary path and execute works but tedious.

shadowspawn:
Closing this one in favour of #533. Open PR with possible approach is #874.

shadowspawn:
This issue will be resolved when v3.0.0 is released. Available now as a prerelease. See #1001

shadowspawn:
If you are using the node inspector for debugging git-style executable (sub)commands using node -inspect et al, the inspector port is incremented by 1 for the spawned subcommand.

Shipped in v3: https://github.com/tj/commander.js/releases/tag/v3.0.0

Thanks to suggestion by @tcf909

Issue #533, opened 2016-05-13
error Error: listen EADDRINUSE :::5858 when using node debug


alexklibisz:
I'm using node 6.0.0 and commander ^2.9.0. I'm trying to use the built-in node debugger to inspect at some breakpoints.
## Case 1: If I run a script that doesn't use commander, it works:

src/cli/works.js:

```
var x = 5;
setTimeout(() => {
  debugger;
  console.log('world');
}, 1000);
console.log('hello');
```

```
$ node debug src/cli/works.js 
< Debugger listening on port 5858
connecting to 127.0.0.1:5858 ... ok
break in src/cli/works.js:1
> 1 var x = 5;
  2 setTimeout(() => {
  3   debugger;
debug> cont
< hello
break in src/cli/works.js:3
  1 var x = 5;
  2 setTimeout(() => {
> 3   debugger;
  4   console.log('world');
  5 }, 1000);
debug> cont
< world
debug> 
```
## Case 2: If I run my commander script without `debug`, it also works:

src/cli/cli.js

```
const
  cli = require('commander'),
  pkg = require('../../package.json');

cli
  .version(pkg.version)
  .command('grid', 'creates grid-style visualization')
  .parse(process.argv);
```

src/cli/cli-grid.js

```
debugger;
console.log('grid');
```

```
$ node src/cli/cli.js grid
grid
```
## Case 3: But if I run a script with the commander _and_ the `debug` option, it gives the following error:

```
$ node debug src/cli/cli.js grid
< Debugger listening on port 5858
connecting to 127.0.0.1:5858 ... ok
break in src/cli/cli.js:1
> 1 const
  2   cli = require('commander'),
  3   pkg = require('../../package.json');
debug> cont
< Error: listen EADDRINUSE :::5858
<     at Object.exports._errnoException (util.js:896:11)
<     at exports._exceptionWithHostPort (util.js:919:20)
<     at Agent.Server._listen2 (net.js:1246:14)
<     at listen (net.js:1282:10)
<     at Agent.Server.listen (net.js:1378:5)
<     at Object.start (_debug_agent.js:21:9)
<     at startup (node.js:85:44)
<     at node.js:444:3
debug> cont
debug> cont
debug> cont
```

So it seems that there's some sort of issue being cause by using the debugger and commander at the same time. I've searched other issues, but found nothing directly relevant. Thanks in advance.

rebeccapeltz:
I am experiencing the same problem after an upgrade today.

[...]

PR #874, opened 2018-10-02
Increment debug port for each child process.

tcf909:
When trying to debug a sub-command style program I ran into a port conflict with --inspect. This increments the debug port for spawned processes.

### DIFF
--- a/index.js
+++ b/index.js
@@ -558,7 +558,24 @@ Command.prototype.executeSubCommand = function(argv, args, unknown) {
     if (isExplicitJS) {
       args.unshift(bin);
       // add executable arguments to spawn
-      args = (process.execArgv || []).concat(args);
+      args = ((process.execArgv && process.execArgv.map(function( v ) {
+
+          var debugPort, matches;
+
+          if( (matches = v.match(/^(--debug=|--debug-brk=|--inspect=|--inspect-brk=)(?:(\b(?:(?:25[0-5]|2[0-4][0-9]|1[0-9][0-9]|[1-9]?[0-9])\.){3}(?:25[0-5]|2[0-4][0-9]|1[0-9][0-9]|[1-9]?[0-9])\b):)?(\d+)/)) && matches.length > 0 ){
+
+              debugPort = + matches[3];
+
+              if( ! Number.isFinite(debugPort) )
+                  throw new Error('Debug port is not a finite number.');
+
+              return ( matches[2] ) ? matches[1] + matches[2] + ':' + (++debugPort) : matches[1] + (++debugPort);
+
+          }
+          else
+              return v;
+
+      })) || []).concat(args);
 
       proc = spawn(process.argv[0], args, { stdio: 'inherit', customFds: [0, 1, 2] });
     } else {


abetomo:
CI has failed.
Could you check it.

shadowspawn:
The code looks a bit wild on first read and specific to node options. However, the port issue does cause confusion when it occurs so worthy of further consideration. #533 #838

shadowspawn:
This looks like might be a solution on node side? 

https://github.com/nodejs/node/pull/5025

> bind to random port with --debug-port=0

tcf909:
the debug-port=0 is a good option, but will only support node 8.x later branches. 

If this is something that people will want I'll refine it (I have done so on my own project) to also include handling environment variables.

Port conflicts are definitely an issue with the way commander handles sub commands (given the spawn nature of the process).

tcf909:
@abetomo The CI is failing on syntax not on functionality. The lint seems extremely specific.

shadowspawn:
I am leaning towards noting the port issue for debugging and including the two available debugging approaches in the README. i.e. debug sub-command directly, or use node support.

I like the idea of suggesting a standard node solution that works for other products too. 

LTS maintenance for node 6 ends this month so I don't think node 8 is a problem moving forward.

However, I only just found `--debug-port=0` and planning to wait for a while and see if there are comments on that as a solution. I have added it to each of the related issues I am aware of. I have not tried it myself yet.

shadowspawn:
And I forgot to say explicitly @tcf909 , thanks for the contributions so far, whatever we decide!

jamime:
@shadowspawn I would say that using --inspect 0 (which is already supported) is not sustainable for development. Each time you start the process you need to attach to a new port. If we use incrementing ports you'll only need to define a few connections.

shadowspawn:
Note to self: [Visual Studio Code](https://code.visualstudio.com/docs/nodejs/nodejs-debugging) detects these patterns:

>Whether or not a process is in "debug mode" is determined by analyzing the program arguments. Currently, we detect the patterns --inspect, --inspect-brk, --inspect-port, --debug, --debug-brk, --debug-port (all optionally followed by a '=' and a port number).

shadowspawn:
The nodejs `cluster` module has a `spawn` which increments debug port:
https://nodejs.org/docs/latest-v8.x/api/cluster.html

Long thread for node with people having port conflicts with spawn/fork. There is mention of debugging inspectors which can auto-detect child processes in debug mode (terminology?), initially in Chrome dev tools:
https://github.com/nodejs/node/issues/9435

shadowspawn:
After seeing that nodejs `cluster` does port incrementing, I am more open to that approach.

@tcf909
You have a complex pattern for numeric host ip (?) in this PR. I note that the Visual Studio Code flag check just looks for pure port, and the nodejs documentation says it could be host:port. Not seeing the point in looking for subclass of numeric IP when host could equally be any numeric IP or `localhost` or a dns name. Is there a reference and/or reason for that complex pattern?

tcf909:
No specific reason. We don't accept host names in our code base for listening ports and wanted to make sure valid ips are passed through or caught at the point of identification and thrown (rather than passing down the stack and wondering what is wrong). We found different versions of node handle various scenarios differently when malformed data was passed.

shadowspawn:
This is being considered for v3, but will need some work and to pass reviews. I am interesting in having a go (original author is not currently active).

shadowspawn:
Closing favour of #991

PR #891, opened 2018-11-28
resolves #877, adds ability to pass execution args on Windows platform

### LINKED ISSUE #877: should pass node options or v8 options to the  subcommend exec process
```js
Command.prototype.executeSubCommand = function(argv, args, unknown) {
```
when subcommand executing, the unknow args will passed to the sub process which spawn by the function above. but if the args is belong to node（

node options or v8 options）, it will fail。

for example, when the subcommand deal with something which needs large mem, so we need to setup the v8 options  `--max_old_space_size =2048`,  for now the `max_old_space_size ` will be treated 
as unknow options, so the process which running the subCommand can not receive these options.

erolkskn:
Hi,

This pull request resolves the issue #877 
Besides, current code is only passing execution arguments if the platform is not equal to 'win32'. This change adds the capability of passing execution arguments in Windows platform.

### DIFF
--- a/index.js
+++ b/index.js
@@ -558,7 +558,7 @@ Command.prototype.executeSubCommand = function(argv, args, unknown) {
     if (isExplicitJS) {
       args.unshift(bin);
       // add executable arguments to spawn
-      args = (process.execArgv || []).concat(args);
+      args = this.getExecutableArguments().concat(args);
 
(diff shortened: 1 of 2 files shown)


erolkskn:
hey @vanesyan , is there any update ?

PR #896, opened 2018-12-11
Error out if unknown command

freemany:
If we have commands `foo bar coo`, once you type with unknown subcommand such as:
$ foo invalid 
it will output error ''Error: unknown command invalid' with help info.

$ foo bar invalid 
it will output error ''Error: unknown command invalid' with help info.

Tests are added.

### DIFF
--- a/index.js
+++ b/index.js
@@ -498,7 +498,7 @@ Command.prototype.parse = function(argv) {
 
   // Output unknown command error
   if (args.length > 0) {
-    console.log('error: unknown command %s', args[0]);
+    console.error('error: unknown command %s', args[0]);
     this.outputHelp();
   }
 
--- a/index.js
+++ b/index.js
@@ -496,6 +496,12 @@ Command.prototype.parse = function(argv) {
     return this.executeSubCommand(argv, args, parsed.unknown);
   }
 
+  // Output unknown command error
+  if (args.length > 0) {
+    console.log('error: unknown command: ', args[0]);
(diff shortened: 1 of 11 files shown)


shadowspawn:
Lots of little comments from first read through:
- there are some requests for this in open issues, adding references to issues this resolves is helpful for reviewing
- musing: I wonder if the historical lack of warning is for simple case where user only processing options and does command detection themselves
- I was bit surprised to see explicit test for `NODE_ENV`, wondering if this is avoidable with sinon
- there are a number of additions of descriptions to tests. What is the reason for that?
- nice to see tests around the new behaviour (thanks)

shadowspawn:
See also: #941

shadowspawn:
I am leaving this one out of v3 as the error handling is quite subtle and fragile, and this Pull Request changed quite a lot of files. Keeping open, but no action likely in meantime.

shadowspawn:
I would like to offer easy handling of unknown commands, because I expect most people would like it. However, there will also be people who have written fancy suggestions for misspellings, or be relying on processing themselves, so might need a way to opt-out.

Changes to the error detection and handling are hard, with lots of code paths for .action and git-style executables, programs with and without .argument, default commands, et al. 

This PR incorrectly generates error messages when the program has `.arguments`.

Adding support for automatic detection of unknown commands is likely to be quite a big job, or require quite a bit of analysis of use cases. Or both!

PR #907, opened 2019-01-17
Add autocomplete based on omelette library

### LINKED ISSUE #385: Tab completion?
Has anyone implemented tab completion with commander? [Complete](https://github.com/hij1nx/complete) looks like a working module, but it seems like this would be perfect territory for commander.

Another module: [completion](https://github.com/twolfson/completion)

bencao:
Add built-in autocomplete for commander.js.

## Interface Design

The current proposed developer API looks like this

### single command mode

```js
program
  .arguments('<a> <b>')
  .option('--verbose', 'verbose')
  .option('-n, --name <name>', 'specify name')
  .option('--description <desc>', 'specify description')
  .complete({
    options: {
      '--name': function() { return ['kate', 'jim']; },
      '--description': ['desc1', 'desc2']
    },
    arguments: {
      a: function() { return ['a-1', 'a-2']; },
      b: ['b-1', 'b-2']
    }
  })
```

### sub command mode

```js
program
  .command('sub1 <arg1> <arg2>')
  .description('sub command 1')
  .option('--verbose', 'verbose')
  .option('-n, --name <name>', 'specify name')
  .option('--description <desc>', 'specify description')
  .complete({
    options: {
      '--name': function() { return ['kate', 'jim']; },
      '--description': ['d1', 'd2']
    },
    arguments: {
      arg1: function() { return ['a-1', 'a-2', 'a-3']; },
      arg2: ['b-1', 'b-2']
    }
  });

program
  .command('sub2 <arg1>')
  .description('sub command 2')
  .option('--verbose', 'verbose')
  .option('-n, --name <name>', 'specify name')
  .option('--description <desc>', 'specify description')
  .complete({
    options: {
      '--name': function() { return ['lucy', 'linda']; },
      '--description': ['db1', 'db2']
    },
    arguments: {
      arg1: function() { return ['a-11', 'a-12']; },
    }
  });
```

And end users need to manually enable autocompletion by running

```sh
eval "$(command-name --completion)"
```

## Implementation

Under the hood, we added a new production dependency to [omelette](https://github.com/f/omelette), which is a lean library for autocompletion that has no further dependencies.

What we basically do is by mapping user-defined autocompletion rules into a form that is supported by omelette. We utilized the "Global complete Event" from omelette and should handle all completion cases gracefully:

```js
var completion = omelette('git');

completion.on('complete', function(f, event) {
  command.autocompleteHandleEvent(event);
});

// omelette will call process.exit(0) after processing
completion.init();
```

closes #385.

Test and document also updated accordingly. 

Please let me know if any more work should be included for this MR ❤️ .


index.js
```diff
@@ -452,6 +686,9 @@ Command.prototype.allowUnknownOption = function(arg) {
  */
 
 Command.prototype.parse = function(argv) {
+  // trigger autocomplete first
+  this.autocomplete(argv);
```
bencao: parse will always call autocomplete.

For normal cases, overhead should be acceptable since we only have several string comparison operations. 

If argv shows that this request is likely for autocomplete, omelette will take over and program.exit(0) after processing.

package.json
```diff
@@ -24,7 +24,9 @@
     "index.js",
     "typings/index.d.ts"
   ],
-  "dependencies": {},
+  "dependencies": {
+    "omelette": "^0.4.12"
```
bencao: added production dependency to omelette, which target supports for node 5+, I guess it should be ok?

bencao:
@abetomo Would you mind help review this PR? Thank you so much.

abetomo (approved):

bencao:
@vanesyan Would you mind help look into this PR when you're available?

bencao:
@abetomo sorry to bother, but we haven't yet heard back from @vanesyan for a while, can you help somehow reach out to him?

index.js
```diff
@@ -443,6 +459,228 @@ Command.prototype.allowUnknownOption = function(arg) {
...
+  var RESERVED_STARTING_KEYWORDS = [
+    '--completion',
+    '--completion-fish',
+    '--compzsh',
+    '--compbash',
+    '--compfish'
+  ];
+  var firstArg = argv[2];
+
+  if (RESERVED_STARTING_KEYWORDS.includes(firstArg)) {
```
vanesyan: I worry here about breaking change if I understand correctly now reserved flags will be handled internally by commander and breaks users' code which use one/some of those options before.

index.js
```diff
@@ -1225,3 +1466,93 @@ function exists(file) {
...
+
+/**
+ * Tests whether the given parameter is an array.
+ *
+ * @param {Object} arg
+ * @return {Boolean}
+ * @api private
+ */
+
+function isArray(thing) {
```
vanesyan: Simply use `Array.isArray(...)` instead of user-defined helper

index.js
```diff
@@ -1225,3 +1466,93 @@ function exists(file) {
...
+
+/**
+ * Tests whether the given parameter is a function.
+ *
+ * @param {Object} arg
+ * @return {Boolean}
+ * @api private
+ */
+
+function isFunction(thing) {
```
vanesyan: Maybe just inline the function in place where it's necessary?

index.js
vanesyan: Maybe we should better hide auto-completion feature behind the flag and set it to `false` by default?

index.js
```diff
@@ -1225,3 +1466,93 @@ function exists(file) {
...
+  var count = 0;
+  var curr = 0;
+
+  while (curr < typedArgs.length) {
+    var currStr = typedArgs[curr];
+
+    if (optionRules[currStr]) {
+      curr = curr + 1 + optionRules[currStr].arity;
+    } else {
+      count = count + 1;
```
vanesyan: and here

index.js
```diff
@@ -1225,3 +1466,93 @@ function exists(file) {
...
+
+  // find out how many args have already been typed
+  var count = 0;
+  var curr = 0;
+
+  while (curr < typedArgs.length) {
+    var currStr = typedArgs[curr];
+
+    if (optionRules[currStr]) {
+      curr = curr + 1 + optionRules[currStr].arity;
```
vanesyan: styling suggestion replace `a = a + b` with simply `a += b`

index.js
```diff
@@ -1225,3 +1466,93 @@ function exists(file) {
...
+    } else {
+      count = count + 1;
+      curr = curr + 1;
```
vanesyan: and here

segfaultmedaddy (changes requested):

vanesyan:
Hello there, sorry for me make you waiting so long. I've partially reviewed the per, average PR looks good, I'll try to cover another part of PR in review.

Also I worry about breaking changes, as now we reserve some flags which may be used in userland code, so maybe we can hide behind a flag in commander API.

bencao:
Thanks @vanesyan for the careful review. I'll address the styling issues soon.

For the breaking changes of autocompletion flag, how about if we add a guardian condition that only if we notice **at least one autocompletion rules are registered**, we trigger autocompletion logic, otherwise we simply skip this function so even if previously user had defined options in those reserved keywords commander will still work as expected?

index.js
bencao: Added a guarding condition statement around `this.autocomplete()` call in parse function. only if at least one rule has been defined we'll run into argv checks.

In this case, we can assume that the user is aware of the completion feature and will solve the confliction of test flags if there's any.

vanesyan:
@abetomo What do you thing about this change? Should we hide it explicitly behind the flag, or enable it once user defines rules for autocomplete?

JohnRGee:
There is a lot of interesting work in this Pull Request, and credit for including documentation and tests. But I want to raise a concern that might have been missed. Commander currently has zero production dependencies. Although omelette is described as a "lean library for autocompletion that has no further dependencies", it is currently 20x larger than commander. 

- https://github.com/tj/commander.js/pull/646#issuecomment-329307273
- https://github.com/tj/commander.js/pull/486#issuecomment-313772381

```
$ npm init -y
$ npm install commander
$ npm install omelette
$ du -d 1 -h node_modules
 76K	node_modules//commander
1.5M	node_modules//omelette
1.5M	node_modules/
```

I dug a little deeper and found Omelette package includes everything including images for README, and could be much leaner. I found an open issue along these lines:

- https://github.com/f/omelette/issues/34

(I have yet to try the code in this Pull Release. This comment based on inspection of the Pull Request.)

bencao:
Thanks for pointing out the size problem. I have created an PR https://github.com/f/omelette/pull/36 to shrink the size of upstream omelette package, with the new version the size (24.2kB) issue should be less concerned.

shadowspawn:
There is a lot of good work in this pull request, but I have concerns that this adds maintenance complexity and code and a dependency for a feature that only some people will benefit from. I do appreciate that some people writing a CLI program would like completion.

I am going to take a closer look, but letting you know I may recommend no. In which case, I wonder if an alternative is an add-on package which we can reference from the README rather than built-in. (One of the first packages I came across when I first looked a couple of years ago was the defunct [commander-tabtab](https://github.com/bencevans/commander-tabtab.js) ).

shadowspawn:
I will say again, this is a very well written Pull Request! And thanks for addressing some of the concerns, including submitting a Pull Request in Omelette (not yet released).

However, this is not something we want to add into Commander itself at this time, due to the added complexity and maintenance we would be taking on, including reliance on a third party.

If you decide to publish this in some form yourself, please let us know.

bencao:
I've used commander.js for a lot of interesting projects now. It's my honor to help contribute something back to Commander. I'll look into some other ways of bringing the autocomplete capability to the community.

lkloon123:
i would sugguest commander to allow plugin extensions and this would surely be great to be a plugin 👍

jchook:
@bencao Did you have any luck extracting this PR into an extension / module?

gutenye:
For anyone seeking a solution, I’ve developed [@gutenye/commander-completion-carapace](https://github.com/gutenye/commander-completion-carapace) using Carapace. It's fast, works well, and supports numerous shells like Bash, Zsh, Fish, Nushell, and others.

shadowspawn:
I came across a new completion package with a Commander integration. Supports zsh, bash, fish, and powershell. In development at version 0.0.17, but being used in a few products and getting almost a million weekly downloads on npmjs.org.
- https://github.com/bombshell-dev/tab
- https://www.npmjs.com/package/@bomb.sh/tab

PR #915, opened 2019-02-14
fix: errors quote

zce:
```diff
- "error: missing required argument `%s'"
+ "error: missing required argument '%s'"
```

### DIFF
--- a/index.js
+++ b/index.js
@@ -792,7 +792,7 @@ Command.prototype.opts = function() {
  */
 
 Command.prototype.missingArgument = function(name) {
-  console.error("error: missing required argument `%s'", name);
+  console.error("error: missing required argument `%s`", name);
   process.exit(1);
 };
 
@@ -806,9 +806,9 @@ Command.prototype.missingArgument = function(name) {
 
 Command.prototype.optionMissingArgument = function(option, flag) {
   if (flag) {
-    console.error("error: option `%s' argument missing, got `%s'", option.flags, flag);
+    console.error("error: option `%s` argument missing, got `%s`", option.flags, flag);
   } else {
-    console.error("error: option `%s' argument missing", option.flags);
+    console.error("error: option `%s` argument missing", option.flags);
   }
   process.exit(1);
 };
@@ -822,7 +822,7 @@ Command.prototype.optionMissingArgument = function(option, flag) {
 
 Command.prototype.unknownOption = function(flag) {
   if (this._allowUnknownOption) return;
-  console.error("error: unknown option `%s'", flag);
+  console.error("error: unknown option `%s`", flag);
   process.exit(1);
 };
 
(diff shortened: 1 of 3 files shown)


kira1928:
@zce 
Your pr broke the lint test.
https://travis-ci.org/tj/commander.js/jobs/493226140#L464-L478
I think those single quotes were inserted to cheat eslint, which is a bad practice :(
Could you change the https://github.com/tj/commander.js/blob/master/.eslintrc.json file to allow only double quotes or change all those strings to be wrapped by single quotes?
(BTW, I personally prefer double quotes :p)

shadowspawn:
I have some comments too @zce. Sorry, you probably through it was an easy pull request!

1) What problem are you fixing? Personally I do not like how the left/right quotes look and it is unusual, is that what you are fixing? e.g. currently

```
error: missing required argument `text'
```

2) what cli commands have you looked at? The examples I found all used straight single quotes:

```
$ git silly 
git: 'silly' is not a git command. See 'git --help'.
$ hg silly
hg: unknown command 'silly'
$ cc silly
clang: error: no such file or directory: 'silly'
```

3) An observation for how much this affects people (i.e. semver). The change does not affect functionality, but some people will have designed their own error commands to match this style and will care if we change the style.

kira1928:
@shadowspawn
Indeed `'word'` looks better than ``wold`` in a cli program.
As for the semver, it seems that there isn't any functional commit since last release, so by now the next version can only be v2.19.1.

BTW, I noticed that there are many unresponsive pull requests and issues. Will you close them and then waiting for authors to reopen and continue, or just leave all of them opened forever?
187 opened issues and 58 opened pull requests are very scary for people who want to participate in maintaining them...

zce:
```
error: missing required argument `text'
```
__`text'__ starts with right quotes, but ends with straight single quotes

shadowspawn:
I found an OLD reference to this very issue: https://www.cl.cam.ac.uk/~mgk25/ucs/quotes.html

I would prefer 

```
'unexpected'
```

rather than

```
`unexpected`
```

(although I have to admit I am now recognise the left quote from markdown, but keep it simple!)

shadowspawn:
Although this does not change the functionality, some people will have written their own error messages to match commander style (so I suggest at least `semver:minor`), and some smaller number may be parsing error messages to do custom handling (so we could be paranoid and include in a `semver:major`). 

Consider whether you think semver patch or minor or major when you review @abetomo or @vanesyan, and comments welcome from any readers.

shadowspawn:
Merging into `release/3.0.0` branch.

shadowspawn:
Already had two approvals. Merged into v3 and will be included in that release. Closing to make it clear that should not be merged into master.

shadowspawn:

shadowspawn:

PR #916, opened 2019-02-14
Add Azure Pipelines for CI (Linux, Mac, and Windows)

### LINKED ISSUE #635: Enable tests on Windows CI
https://www.appveyor.com/docs/

kaylangan:
Azure Pipelines gives open source projects 10 build pipelines with unlimited minutes for Linux, Mac, and Windows.

This PR adds support to run CI on all three platforms and also makes changes so that the tests pass on Windows (fixes #635). I did have to special case some tests; for example, `search` works on Windows whereas the tests expect an `EACCES` error for Linux/Mac. Any pointers on preferred ways to make the tests Windows-friendly would be much appreciated!

I set up a commander.js pipeline [here](https://dev.azure.com/commanderjs/commander.js/_build/results?buildId=2). The builds seem to be slightly faster than on Travis. I'm happy to transfer ownership over if this is something you want to pursue. Here's a screenshot of the UI:
image

Azure Pipelines also has rich test reporting which works well with commonly used test runners, including mocha and ava which are being considered in #649. Here's an [example from prettier](https://dev.azure.com/prettier/prettier/_build/results?buildId=545&view=ms.vss-test-web.build-test-results-tab):
image

Let me know if there are any questions or suggestions. I'm a Program Manager for Azure Pipelines and happy to help!

### DIFF
--- a/.azure-pipelines-steps.yml
+++ b/.azure-pipelines-steps.yml
@@ -1,0 +1,31 @@
+#
+# Steps for building and testing commander.js. See jobs defined in .azure-pipelines.yml.
+#
+
+steps:
+  # Clones the repo. This is an explicit step to preserve symlinks for Windows.
+  - checkout: self
+
+  - task: NodeTool@0
+    inputs:
+      versionSpec: $(NODE_VERSION)
+      checkLatest: 'true'
+    displayName: 'Install Node.js'
+
+  # Run npm ci by default unless NPM_INSTALL is set
+  - script: npm ci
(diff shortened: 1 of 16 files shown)


shadowspawn:
Good coverage and helpful information in the description including helpful info on mocha and ava. Thanks for references to related issues, always appreciated.

I think we will come back to this after deciding where we are going with sinon, so closing this in the meantime to clear out Pull Requests we are not ready to act on.

I have added a comment to #635 back to this Pull Request.

PR #930, opened 2019-03-14
Update index.d.ts

kira1928:
Add the optional parameter `argsDescription` to description() function

### DIFF
--- a/typings/index.d.ts
+++ b/typings/index.d.ts
@@ -226,9 +226,10 @@ declare namespace local {
      * Set the description to `str`.
      *
      * @param {string} str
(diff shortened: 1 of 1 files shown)


shadowspawn:
@kira1928 
Would it be appropriate to have type matching what I think is the expected use:

`{{[argName: string]: string}}`

rather than the pure object style definition:

`{{[key: string]: any}}`

kira1928:
@shadowspawn 
Thank you for your advise.
I have fixed the problem you mentioned.
Does it look better now?

abetomo:
Thank you!

PR #935, opened 2019-03-21
fix: resolve symbolic links completely when hunting for subcommands

MarshallOfSound:
Depending on your npm / yarn set up a global install may have to resolve through several symlinks before it correctly arrives the the baseDir that actually contains the other subcommands.  The links go between directories so if we stop after the first link `baseDir` is looking in the wrong place

The changes the resolution logic to resolve symlinks until their are no more to resolve by using `realpath` which will resolve links all the way to the final destination.


index.js
```diff
@@ -527,14 +527,11 @@ Command.prototype.executeSubCommand = function(argv, args, unknown) {
...
-  var baseDir,
-    link = fs.lstatSync(f).isSymbolicLink() ? fs.readlinkSync(f) : f;
+  var baseDir;
 
-  // when symbolink is relative path
-  if (link !== f && link.charAt(0) !== '/') {
-    link = path.join(dirname(f), link);
-  }
-  baseDir = dirname(link);
+  var resolvedLink = fs.realpathSync(f)
```
abetomo: A semicolon is a required rule, can you add it?

MarshallOfSound: Yup, missed the lint config, sorry 'bout that 👍

shadowspawn:
Might take me a few days, but I am looking forward to reviewing this. Hopefully fixes #866 (and without adding a dependency as in #869).

@MarshallOfSound could you please suggest some reproduction steps that match how a user might end up with the broken symlink? It sounded like you had something in mind:

> Depending on your npm / yarn set up a global install may have to resolve through several symlinks before it correctly arrives the the baseDir that actually contains the other subcommands.

I did find mention of yarn install in https://github.com/tj/commander.js/issues/866#issuecomment-426032789 and will be trying that too.

MattCheely:
> @MarshallOfSound could you please suggest some reproduction steps that match how a user might end up with the broken symlink? It sounded like you had something in mind:

I'm not the OP, but I have a package that you can replicate the issue with via:

`yarn global add elm-app-gen`

and then run:

`elm-app-gen quickstart test-app`

in some temporary directory.

MarshallOfSound:
@shadowspawn Anyone using subcommands and the latest version of `yarn` will have this issue afaik.

Yarn installs the binary to `/usr/local/bin/my-cmd` which symlinks to `$YARN_DIR/.bin/my-cmd` which then symlinks to `$YARN_DIR/node_modules/my-package/bin/my-cmd`.

You can repro this quite easily with `yarn global add @electron-forge/cli && electron-forge init t` in some temporary directory

KillWolfVlad:
@MarshallOfSound I did not know about `fs.realpath`. Your solution looks better and more native.

shadowspawn:
(My review progress: examined code and related issues and tested on Mac, looking good, Windows still to go.)

MarshallOfSound:
@shadowspawn What needs to happen to get this merged and released?

shadowspawn:
@MarshallOfSound 
I joined as collaborator recently and not involved in the release process itself yet. Nothing more needed from you at the moment though.

abetomo:
Thank you!
We will release it soon.

PR #941, opened 2019-03-29
Fix to fail if required arguments don't exist.

yossan:
The `missingArgument` method isn't called now.

```
// <file> should be required but it works now.
program
  .argument('<file>')
  .parse(['node', 'test'])
```

These commits call the `missingArgument` method. 

```
error: missing required argument `file`
```


shadowspawn:
I have not worked through the code yet. A couple of questions @yossan (but you don't have to answer, it won't block my review, just make it easier):
- does this work correctly with multiple arguments too? `.argument('<file> <env>')`
- does this improve same situation with subcommands?

yossan:
@shadowspawn 
> does this work correctly with multiple arguments too? .argument('<file> <env>')

```
// error: missing required argument `env'
program
    .arguments('<file> <env>)
    .parse(['node', 'test', 'setup'])
```

```
program
    .arguments('<file> <env>)
    .parse(['node', 'test', 'setup', 'path'])
```

> does this improve same situation with subcommands?

This don't work. 

```
program
    .command('cmd <file>')
    .parse(['node', 'test', 'cmd'])
```
I apologize. This commit may have caused subcommands not to work well.
I will fix it as soon as possible.

shadowspawn:
See also: #896

yossan:
@shadowspawn 
I issued https://github.com/tj/commander.js/issues/995.

I think a fundamental review is necessary.

PR #951, opened 2019-04-16
WIP: Fix option namespace pollution

shadowspawn:
Work in progress. Added a feature flag to get new behaviour, where option values are not stored on the Command object but in a separate field and accessed using new accessor methods. The new behaviour is opt-in, unchanged clients get unchanged old behaviour for first release.

Builds on work in #515 and #673

Feedback welcome (from anyone). Still a work in progress, I am still coding to try ideas so anything could change. 🛠 

I will move to branch on tj/commander if other maintainers want easier access to code.

Still under consideration:
1) ~~should new accessor methods be get+has (like Map) or getOption+hasOption for clarity?~~ get/has for simplicity, like original properties
2) ~~better name for mapStyleOptions?~~ storeOptionsAsProperties
3) ~~setFeatureFlag or setFeatureFlags?~~ setFeatureFlags

Still to prototype:
- special case get/has/setFeatureFlag options so they do not affect existing clients who do not opt-in
- pass feature flag to commands so new behaviour is global
- Typescript declarations for new routines
- README
- include existing `.opts()` in README description of options too
- tests for new behaviour (and tests expecting old behaviour do not change)

Code fragment:
```
program
   .setFeatureFlags("storeOptionsAsProperties:false")
   .option("-o, once", "description");

if (program.has("once") console.log("Once upon a time");
```

Other changes:
- took `version` out of new-behaviour option values, because it is not really an option value (?)
- added warning for an old-behaviour option when can detect there is a conflict
- used null object as prototype for new-behaviour options (so close to a pure map)

### DIFF
--- a/index.js
+++ b/index.js
@@ -103,7 +103,7 @@ function Command(name) {
   this._allowUnknownOption = false;
   this._args = [];
   this._name = name || '';
-  this._optionValues = Object.create(null); // pure object as map
+  this.optMap = new Map();
   this._storeOptionsAsProperties = true; // legacy behaviour by default
 }
 
@@ -403,9 +403,9 @@ Command.prototype.option = function(flags, description, fn, defaultValue) {
     if (!option.bool) defaultValue = true;
     // preassign only if we have a default
     if (defaultValue !== undefined) {
-      self._optionValues[name] = defaultValue;
+      self.optMap.set(name, defaultValue);
       if (self._storeOptionsAsProperties) {
-        self[name] = self._optionValues[name];
+        self[name] = defaultValue;
       }
       option.defaultValue = defaultValue;
     }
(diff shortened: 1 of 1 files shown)


shadowspawn:
A side note: if we did not already have the somewhat undocumented `.opts()` routine I would have suggested having a property called `.opts`! Only two extra characters and arguable better separation to have a routine than a property, but I am used to process.argv et al.

shadowspawn:
My current thinking is:
- a new map property is the canonical source of truth
- configuration to enable/disable legacy options-as-properties
- keep existing `.opts()` to reduce breaking changes

Considering:
- configuration to control whether `.action` handler is passed Command, options object, or map

shadowspawn:
I have some new ideas and going to start again, from latest code.

Issue #943, opened 2019-03-30
Triage: Coercion vs Accumulation


shadowspawn:
#198 changed the parameters passed to the "coercion" function to allow it to also act as an "accumulator" function such as `collect` and `increaseVerbosity` in the README. However, it also means an extra value is passed to a "plain" coercion function which is unexpected and breaks the older `parseInt` example.

Issues:
- #201 2.2.0 isn't compatible any more with parseInt used with default arguments
- #400 Can't use default together with coercion (example from front-page broken)
- #523 option() default value doesn't work with parseInt
- #524 Default coercion functions not compatible with parseInt and default
- #587 Bug with Coercion?
- #686 option doesn't work correctly when given four arguments
- #834 Option Coercion - parseInt when combined with a default value will result in NaN

Pull Requests:
- #829 Document default vs. coercion function order
- #757 Fix Coercion for Repeated Options
- #784 Fix Coercion for Non Collections

shadowspawn:
Updated README to clarify option types and custom processing, including parseInt

Issue #937, opened 2019-03-26
Throw an error when the short flag for an option for longer than one character


HopefulLlama:
In the `.option(...)` method, it takes a string flag which is comma separated, and describes a short and a long flag for passing arguments, such as:

```
  .option('-v, --version', 'Version information');
```

However, when given a short flag which is longer than one character, the program will still parse and execute, but cannot be accessed by its short flag. 

For example:

```
  .option('-nv, --next-version', 'The next version');
```

It is impossible to use `-nv` on the command-line, as commander interprets this as two short flags concatenated together: `-n` and `-v`. 

It would be better to throw an error to avoid the misleading set up of commander options.

shadowspawn:
The documentation does mention the cli behaviour:

> Short flags may be passed as a single arg, for example -abc is equivalent to -a -b -c.

but I agree it would be nice if commander warned you about the setup problem.

HopefulLlama:
It does mention that, but it took me a long time to put two and two together.

I.e.: It does take some thinking to realise that a two character flag **cannot** work due to the short flag feature. It doesn't _explicitly_ say so in the documentation.

I think an addition to the documentation, and an error or warning given when the misleading setup is actioned would be suitable.

shadowspawn:
It now _explicitly_ says single character in the documentation. (But still no warning about coding error, so feel free to keep this issue open.)

shadowspawn:
Some more errors or warnings for coders (rather than users) is something we can improve in future. Closing this as have made _an_ improvement with the addition to the README.

Issue #884, opened 2018-10-28
Provide info about default values on the main github page


mprinc:

JohnRGee:
What sort of info about default values are you looking for?

There are a coupe of examples of default values for options that are covered without additional explanation in the README:

      .option('-c, --cheese [type]', 'Add the specified type of cheese [marble]', 'marble')

      .option('-s --size <size>', 'Pizza size', /^(large|medium|small)$/i, 'medium')

mprinc:
Thanks. Yes I discovered them eventually, but at the moment I was interested to provide a default value in my app, I couldn't search for any explicitly mentioned default use case, but rather hitting:

`  .command('list', 'list packages installed', {isDefault: true})`

or

`If you want to display help by default ..`

It would be nice to have more clear directions

Stogoh:
And how can i set a option on by default. For example I have an option `--databaseHost` which should have a default value. How can I do this? This is without an enum.

ozbillwang:
I am looking similar to set default value for the parameters.

I used package `argparse` before, and it is easy to set the default value, how can I do the same with `commander`?

```
bar.addArgument(
  [ '-f', '--foo' ],
  {
    action: 'store',
    defaultValue: "test.txt",
    help: 'foo3 bar3'
  }
);
```

shadowspawn:
@ozbillwang 

```
  .option('-f, --foo <value>', 'foo3 bar3', 'test.txt')
```

shadowspawn:
I have proposed a large update to options coverages in README in #953

shadowspawn:
Updated README to clarify option types, including section for default option value
https://github.com/tj/commander.js#default-option-value

Issue #790, opened 2018-04-18
Document `Command` class in readme


nicojs:
I've noticed that by using the documented `const program = require('command')` syntax you're actually polluting a global `Command` object. This can be shared by other components resulting in weird issues (like https://github.com/nicojs/node-link-parent-bin/issues/4#issuecomment-378506690)

After inspecting the source code, i see that a `Command` class is exported! Kudo's to you sir! I'll use this from now on. It also helps quite a bit with testability (injecting a mock object for `Command` at test time). Let's promote this way of working a bit by documenting it in the readme

My idea is to update the readme and use `const Command = require('commander').Command; const program = new Command();` everywhere. I would like to place a remark to let people know that they could also use `const program = require('commander');` but that it is not preferred and explain why it is not.

If you agree, I'm willing to create the PR.

I would personally like it if the `const program = require('commander');`-way-of-working would become deprecated and removed in a major release. What are your thoughts on that?

shadowspawn:
We do need to get something into the README about the "new" pattern, thanks for raising.

I don't think we will deprecate the old way as such, but it should be the my-program-is-so-simple-globals-are-ok way of working.

shadowspawn:
I have added the new pattern in a large update to README in #953

nicojs:
Awesome! LGTM

shadowspawn:
Updated README with section right at start showing use of Command class 
https://github.com/tj/commander.js#declaring-program-variable

PR #953, opened 2019-04-27
WIP Rework options description to clarify option types and custom processing

shadowspawn:
Mainly rewriting options coverage in README:
- rewrite description of function for custom processing (#943), previously called "Coercion"
- focus on common option types, boolean and option taking `<value>`
- cover option with optional `[value]` separately under "other" types
- mention short option is single character (#937)
- explicitly cover defaults (#884)
- added explanation of declaring program using `new Command` (#790)
- added some extra info about boolean options (#574)

Still to come: small changes to version option description

Comments welcome from anyone.


shadowspawn:
I have added about 100 lines to length of README, but hopefully clearer!

(I didn't refactor the regular expression support in with the rest of the options. Future work.)

(I think we should add a table of contents, but after tidying more of the README so the headings make better sense. Future work.)

ozbillwang:
Looks much better. :+1:

shadowspawn:
I had a look at "Insights > Traffic" and am surprised how much traffic `examples/` gets. In last 2 weeks the README got 2912 and `examples/` got 518 unique visitors.

Since the examples are being used, I think I'll add some more info to the new example files I added like "as shown in README", and add the expected output in a comment. However, this PR can be reviewed or merged before I do that, as examples can be added separately.

ozbillwang:
One suggestion.
```
.option('-c, --cheese <type>', 'add the specified type of cheese', 'blue');
```
Until now, I still don't fully understand how this option `<type>` to be used. 

I accidentally wrote a code as below (without <type>)

```
 .option('-c, --cheese', 'add the specified type of cheese', 'blue');
```
this option doesn't work as expected. 

But I am not sure what should be put, currently just copy it as `<type>`

shadowspawn:
Thanks @ozbillwang. I have in the early description:

```
The two most used option types are a boolean flag, and an option 
which takes a value (declared using angle brackets)
```

but do not clarify when describing default values:

```
You can specify a default value for an option which takes a value.
```

Is it mainly the default values case you tripped over? Or do you think the "takes a value" part could be expanded in both places?

shadowspawn:
(Updated examples to include README reference and example output.)

Issue #921, opened 2019-02-25
Pass unknown also as a second argument to the command:* event


negiDharmendra:
We are trying to use [this](https://github.com/tj/commander.js) awesome library to handle command line arguments and flags in [Taiko](https://github.com/getgauge/taiko/blob/master/README.md).
```
Example:

taiko [option] --> this opens an interactive repl.
taiko <js file> [option]  --> executes the js script.
```

There seems to be an issue related to unknown flags. When we provide the `js file` argument and pass an unknown flag commander does not complain about that. But when we pass only unknown flag without arguments commander prints a nice message that provides flag was invalid.

It would be great if the commander complains about the unknown flags even if they are passed after arguments.

negiDharmendra:
In order to fix the issue, we need to pass the unknown parameter to the while emitting the `'command:*'` event on the following file. 
https://github.com/tj/commander.js/blob/68545af7130967bc1a35ae30e9567d78c5e56d8a/index.js#L656

shadowspawn:
Have you got a small example program?

(I do see the warning for unknown flags in my program which has subcommands, but looks like you have just an optional argument and options and no subcommands?)

clarsen:
Attached is a PR that does the suggested behavior and adjustment of tests.  I'm a little hesitant about it though because it breaks the existing default behavior of ignoring unknown options, instead resulting in an error.

While this is the desired behavior for my programs, others might be relying on the silent default handling of unknown options.

shadowspawn:
I have reproduced this, and realised it is a duplicate of #561. Closing in favour of that open issue.

Thank you  for your contributions.

shadowspawn:
After deeper analysis I discovered that there are different behaviours depending on whether there is an action handler attached to the program. #965 is targeting the case when there is an action handler, and original report in #561 covering no action handler. Not sure which case applies to `taiko` but have an open PR and an open issue to cover the two possibilities, so will still leave this closed.

PR #965, opened 2019-05-23
complain about unknown options if there are >0 arguments on command line

clarsen:
As per https://github.com/tj/commander.js/issues/921  this passes `unknown` to `command:*` listener and as a result, the default command will also complain about unknown options.  Before, it wouldn't.

The test in test.arguments.js needed to be adjusted since it would silently succeed even though an unknown option `--setup_mode` was passed and eaten.

### DIFF
--- a/index.js
+++ b/index.js
@@ -653,7 +653,7 @@ Command.prototype.parseArgs = function(args, unknown) {
     if (this.listeners('command:' + name).length) {
       this.emit('command:' + args.shift(), args, unknown);
     } else {
-      this.emit('command:*', args);
+      this.emit('command:*', args, unknown);
     }
   } else {
     outputHelpIfNecessary(this, unknown);
--- a/test/test.arguments.js
+++ b/test/test.arguments.js
@@ -7,6 +7,7 @@ var program = require('../')
 
 var envValue = "";
 var cmdValue = "";
+var optionsValue = {};
 
 program
   .version('0.0.1')
@@ -21,10 +22,11 @@ program
 
 program
   .command("setup [env]")
-  .option("--setup_mode <mode>")
-  .action(function (env) {
+  .option("--setup_mode <mode>", "the setup mode")
+  .action(function (env, options) {
     cmdValue = "setup";
     envValue = env;
+    optionsValue = options
   })
 
 program.parse(['node', 'test', '--config', 'conf']);
@@ -36,3 +38,4 @@ program.parse(['node', 'test', '--config', 'conf1', 'setup', '--setup_mode', 'mo
 program.config.should.equal("conf1");
 cmdValue.should.equal("setup");
 envValue.should.equal("env1");
+optionsValue.setup_mode.should.equal("mode3");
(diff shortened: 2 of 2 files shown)


shadowspawn:
This looks like the same code change as #654, although for different reasons.

Note to self: see also useful comment from @clarsen in issue: https://github.com/tj/commander.js/issues/921#issuecomment-495319899

shadowspawn:
#654 has been closed, so this PR is unblocked

shadowspawn:
Open related issue is #561.

shadowspawn:
After deeper analysis I discovered that there is different unknown option detection depending on whether there is an action handler attached to the program. This PR detects the unknown option when there is an action handler, but does not change the behaviour for the original report in #561 which needs a different fix.

shadowspawn:
I think I understand! The option handling is very subtle. This fix covers the case when there is an action handler attached to the program (or default command) object and makes the behaviour consistent with subcommands with action handlers. I think this is reasonable.

It is a potentially breaking change in behaviour so needs to wait for a major release. There is an argument that ignoring options was a bug, but it has been that way for a long time and some people may have assumed it was intended and be relying on it.

I want to change the fix to the test, and I'll add a review about that.

@clarsen are you still around and interested in doing more work on this, or shall I take it over?

shadowspawn:
Reminder to myself:
- add example code and fixed behaviour while I understand
- add review about test, remove `--setup_mode`

clarsen:
> @clarsen are you still around and interested in doing more work on this, or shall I take it over?

I've been quietly watching the discussion on this but don't have a deep understanding of the legacy behavior issues.  Certainly would like to contribute back to this...

shadowspawn:
This PR makes the error handling for an action attached to the program work the same way as an action attached to a subcommand. In particular unknown options now cause an error even when an argument is supplied.

Given a program:

```js

program
  .arguments("<file>")
  .action(function (fileParam) {
    console.log(`program called with: ${fileParam}`);
  });

program
  .parse(process.argv);
```

Before:

```sh
$ node pm0.js 
program
$ node pm0.js --silly
error: unknown option '--silly'
$ node pm0.js x
program
# And the one that changes...
$ node pm0.js x --silly
program
```

After:

```sh
$ node pm0.js 
program
$ node pm0.js --silly
error: unknown option '--silly'
$ node pm0.js x
program
# Now checks the options
$ node pm0.js x --silly
error: unknown option '--silly'
```

clarsen:
Thanks... I'll try using this in the CLI app I had that originally motivated the change to check it works 'ok'.

shadowspawn:
I am going to pick this up and rework after the move to Jest. I'll add a co-author on the commit.

shadowspawn:
Closed with implementation moved to #1049

shadowspawn:
#1049 has been merged to develop and will be released in v4.0.0. 

clarsen:
> #1049 has been merged to develop and will be released in v4.0.0.

Thanks, I verified that commander@4.0.0-1 (prerelease) properly warns about unknown options with one of my scripts.  I'll wait until the final release before switching over.

shadowspawn:
v4.0.0 has been released. 

shadowspawn:
(Accidentally opened branch making comment.)

PR #970, opened 2019-05-31
docs: ✏️ update chinese readme

banli17:
✅ Closes: #967


shadowspawn:
Thanks @banli17, very quick!

Closes: #970 (I forgot that putting it in title does not make link)

PR #973, opened 2019-06-09
Add CONTRIBUTING

shadowspawn:
Added CONTRIBUTING to offer information about Issues and Pull Requests. 

(It could be added at the top level if we like it more obvious in the file structure. I put it down in the `.github` directory and it will get offered in the GitHub web interface.)


shadowspawn:
(Just got a Pull Request with no description, so will merge now instead of waiting for next release.)

PR #980, opened 2019-06-23
Add parent command as prefix of subcommand

### LINKED ISSUE #739: Usage for command should include parent commands

shadowspawn:
This is a minor rework of #740 to add parent commands into usage help for subcommands:

- updated to match current code style (lint)
- fix nested subcommand order, as noted by @simonbuchan

Resolves #739

### DIFF
--- a/index.js
+++ b/index.js
@@ -1116,11 +1116,12 @@ Command.prototype.helpInformation = function() {
   if (this._alias) {
     cmdName = cmdName + '|' + this._alias;
   }
-  var ancestor = this;
-  var prefix = [];
-  while ((ancestor = ancestor.parent)) prefix.push(ancestor.name() + ' ');
+  var parentCmdNames = '';
+  for (var parentCmd = this.parent; parentCmd; parentCmd = parentCmd.parent) {
+    parentCmdNames = parentCmd.name() + ' ' + parentCmdNames;
+  }
   var usage = [
-    'Usage: ' + prefix.join('') + cmdName + ' ' + this.usage(),
+    'Usage: ' + parentCmdNames + cmdName + ' ' + this.usage(),
     ''
   ];
 
--- a/test/test.command.usage.js
+++ b/test/test.command.usage.js
@@ -5,4 +5,4 @@ program
   .name('test')
   .command('info [options]')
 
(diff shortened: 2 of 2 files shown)


shadowspawn:
Looking for a review before merging from anyone who looked at original PR: @abetomo @simonbuchan @mojavelinux

index.js
```diff
@@ -1116,8 +1116,12 @@ Command.prototype.helpInformation = function() {
   if (this._alias) {
     cmdName = cmdName + '|' + this._alias;
   }
+  var parentCmdNames = '';
+  for (var parentCmd = this.parent; parentCmd; parentCmd = parentCmd.parent) {
+    parentCmdNames = parentCmd.name() + ' ' + parentCmdNames;
+  }
   var usage = [
-    'Usage: ' + cmdName + ' ' + this.usage(),
+    'Usage: ' + parentCmdNames + cmdName + ' ' + this.usage(),
```
simonbuchan: Tiny nit, you can shorten this by having the loop above mutate `cmdName` instead of , since it's not used below.

test/test.command.usage.js
```diff
@@ -0,0 +1,8 @@
+var program = require('../')
+  , should = require('should');
+
+program
+  .name('test')
+  .command('info [options]')
+
+program.commands[0].helpInformation().should.startWith('Usage: test info [options]')
```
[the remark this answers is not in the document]
shadowspawn: Nested commands are not currently supported! But since the code included a loop, I thought it should do it right for the possible future. :-)

index.js
shadowspawn: Thanks. I did wonder about that too, but went for long and clear.

shadowspawn:
Thanks @abetomo and @simonbuchan

shadowspawn:

shadowspawn:

PR #987, opened 2019-06-30
Allow boolean default for flag option

### LINKED ISSUE #928: Support defaults for boolean options (including --no-*), for env var set defaults
It doesn't appear we can set boolean option default values, so that we can set a default value for such options via environment variables, then override these values from the command line. I can do this for options which take a value, and would like this sort of override behavior to be consistent for both option types.

We have two use cases. Default true vs default false. For the description below, assume:

**Default true use-case**: We want to have a method to monitor progress of the program, the options.monitor boolean should default to false. If the user has a MONITOR environment variable set to true (or 1), this boolean should instead default to true. The user can then, on a case-by-case basis, override the false default with --monitor, or override the true (set via env var) default with --no-monitor. I'd like an ability to just turn on monitoring system-wide via an env var, not have to specify --monitor each time I run a CLI command. But, having changed the default system-wide, I may want to do the opposite on a single command.

**Default false use-case**: We want to have a method to perform additional validation checks, the options.check boolean should default to true. If the user has a NOCHECK (or NO_CHECK) environment variable set to true (or 1), this should instead default to false. The user can then, on a case-by-case basis, override the true default with --no-check, or override the false (set by env var) default with --check. As above, I'd like to turn off these extra checks system-wide via an env var, not have to specify --no-check each time I run a CLI command.

I was asked to open this issue as a new ticket, based on a comment to a related issue (#736). I've provided a cleaner description of the two use-cases above, but below is the original description of my use-cases.

- options.monitor should default to false or undefined, but have a way to be set to true
- options.check should default to true, but have a way to be set to false.

Current design as I understand it seems to support what I want as explicit command line options:

- Add "--monitor" to CLI to get options.monitor = true, otherwise it's undefined
- Add "--no-check" to CLI to get options.check = false, otherwise it's true

But I can't see how to make this work with default values, allowing me to set the default behavior via environment variables, but still allow for a command-line override.

- If MONITOR env var exists and is 1 or true, options.monitor should default to true, and "--monitor" should have no effect. Otherwise, options.monitor should default to undefined or false, and "--monitor" would then set it to true.
- If NOCHECK env var exists and is 1 or true, options.check should default to false, and "--check" should override this default to set it to true. Otherwise, options.check should default to true, and "--no-check" should override this and set it to false.

I thought this should work, but it doesn't. If there's not a way to do this, there should be.
```
let MONITOR = (process.env.MONITOR && (process.env.MONITOR == 'true' || process.env.MONITOR == '1')) ? true : false;
let CHECK = (process.env.NOCHECK && (process.env.NOCHECK == 'true' || process.env.NOCHECK == '1')) ? false : true;

program.command('test')
  .option('-m, --monitor', 'Monitor build process', MONITOR)
  .option('-C, --no-check', 'Do not check prerequisites', CHECK)
```

shadowspawn:
(Have not added anything to README yet, but code and tests as intended.)

This allows supplying  default value for boolean flags, in particular for a default value when both `--foo` and `--no-foo` options are defined, such as a default value from environment. Only possible after #795.

Resolves #928. 

```

var colourDefault;
if (process.env.COLOUR === "0" || process.env.COLOUR === "false")
  colourDefault = false;
if (process.env.COLOUR === "1" || process.env.COLOUR === "true")
    colourDefault = true;

program
  .option('-c,--colour', 'enable colour', colourDefault)
  .option('-C,--no-colour', 'disable colour');

program.parse(process.argv);

console.log(`colour is ${program.colour}`);
```

```
$ node index.js -h
Usage: index [options]

Options:
  -c,--colour     enable colour
  -C,--no-colour  disable colour
  -h, --help      output usage information
$ node index.js 
colour is undefined
$ node index.js -C
colour is false
$ node index.js -c
colour is true

$ COLOUR=1 node index.js -h
Usage: index [options]

Options:
  -c,--colour     enable colour (default: true)
  -C,--no-colour  disable colour
  -h, --help      output usage information
$ COLOUR=1 node index.js 
colour is true
$ COLOUR=1 node index.js -C
colour is false
$ COLOUR=1 node index.js -c
colour is true

$ COLOUR=0 node index.js -h
Usage: index [options]

Options:
  -c,--colour     enable colour (default: false)
  -C,--no-colour  disable colour
  -h, --help      output usage information
$ COLOUR=0 node index.js 
colour is false
$ COLOUR=0 node index.js -c
colour is true
$ COLOUR=0 node index.js -C
colour is false
```


shadowspawn:
To clarify one breaking aspect of this change.

There is a subtle and I think undocumented behaviour I only discovered working through this code that you can set a value for a flag to take when specified (instead of true). This may even be completely unintentional.

```
program.option('--capital', 'set capital', 'Wellington');
...
console.log(program.capital);
```
```
$ example
undefined
$ example --capital
Wellington
```

So there was an existing behaviour that these two calls were actually the same, and specifying a boolean default of true would have no affect previously, and now it will have an affect:
```
program.option('--silly', 'desc');
program.option('--silly', 'desc', true);
```

I _think_ the new behaviour is reasonable, and the old behaviour is undocumented and perhaps even accidental.

shadowspawn:
@usmonster  FYI, the future change I mentioned in #795.

usmonster:
Thanks for the ping, @shadowspawn. FYI, I'd like to review this, but I probably can't before the end of the week.

Regarding the old behavior you mention, I'm pretty sure it was intentional to allow any option (even flags) to have a default value specified, especially given that there's a test for it. I imagine some folks may rely on this behavior even if undocumented, but your change does make sense as well.

shadowspawn:
> I imagine some folks may rely on this behavior even if undocumented

Yes indeed. I actually looked up the following quote when I joined as a maintainer and have been expecting an opportunity to use it and/or suffer from it. :-)

> With a sufficient number of users of an API,
> it does not matter what you promise in the contract:
> all observable behaviors of your system
> will be depended on by somebody.
 
http://www.hyrumslaw.com

usmonster:
@shadowspawn, my only concern is that I'm not sure how many users would prefer the new feature over the existing behavior, or vice-versa. If you have more data on this beyond the initial request, feel free to share. I just don't want more issues to pop up over this breaking change.

shadowspawn:
I do not have any metrics on preferred behaviour. I do like making it convenient to have an externally supplied default that can be overridden, with`--foo/--no-foo` actually functional (thanks!).

From code inspection and experimentation I think the old behaviour was that a boolean default value (third parameter) was ignored, which I do see a use case for?

```

program
  .option("-t,--truthy", "truthy", true)
  .option("-f,--falsey", "falsey", false)
  .option("-s,--surprise", "surprise", "surprise");

program.parse(process.argv);

console.log(`truthy ${program.truthy}`);
console.log(`falsey ${program.falsey}`);
console.log(`surprise ${program.surprise}`);
```

Old behaviour:
```
$ node index.js --help
Usage: index [options]

Options:
  -t,--truthy    truthy
  -f,--falsey    falsey
  -s,--surprise  surprise
  -h, --help     output usage information
$ node index.js 
truthy undefined
falsey undefined
surprise undefined
$ node index.js -t -f -s
truthy true
falsey true
surprise surprise
```

New behaviour. (Reminder, the intended use case is for `--foo/--no-foo` but highlighting change in behaviour.)
```
$ node index.js --help
Usage: index [options]

Options:
  -t,--truthy    truthy (default: true)
  -f,--falsey    falsey (default: false)
  -s,--surprise  surprise
  -h, --help     output usage information
$ node index.js 
truthy true
falsey false
surprise undefined
$ node index.js -t -f -s
truthy true
falsey true
surprise surprise
```

shadowspawn:
Thanks for review @abetomo.

usmonster:
Thanks for the further clarification, @shadowspawn. I misunderstood the current behavior. I'm on board with this change, but will leave an inline comment or two.

test/test.options.defaults.given.js
```diff
@@ -8,15 +8,19 @@ var program = require('../')
...
+  .option('-O, --no-onions', 'No onions')
+  .option('-t, --tomatoes', 'Add tomatoes?', false)
+  .option('-T, --no-tomatoes', 'No tomatoes')
   .option('-v, --olives', 'Add olives? Sorry we only have black.', 'black')
   .option('-s, --no-sauce', 'Uh… okay')
   .option('-r, --crust <type>', 'What kind of crust would you like?', 'hand-tossed')
   .option('-c, --cheese [type]', 'optionally specify the type of cheese', 'mozzarella');
 
-program.parse(['node', 'test', '--anchovies', '--onions', '--olives', '--no-sauce', '--crust', 'thin', '--cheese', 'wensleydale']);
+program.parse(['node', 'test', '--anchovies', '--onions', '--tomatoes', '--olives', '--no-sauce', '--crust', 'thin', '--cheese', 'wensleydale']);
```
usmonster: Can you please add a few tests involving negated options?

E.g., what's the expected behavior if someone specifies a default for `--no-sauce`?

Or something arguably invalid/redundant, e.g. for `--no-onions`, but not `--onions`, or for both `--tomatoes` and `--no-tomatoes`? Thanks!

shadowspawn: Good comment, I only slightly expanded the tests, and do want to improve the coverage. I will look at that.

Honestly, I don't have an expected behaviour for a default value for a negative option! But that also reminds me I want to add to the README now other changes have landed, to cover the intended usage.

shadowspawn:
Expanded tests to have more complete coverage of intended boolean flag permutations.

shadowspawn:
Added written description of possibility for default value for --foo/--no-foo to README.

test/test.options.bool.js
```diff
@@ -1,15 +1,28 @@
-/**
- * Module dependencies.
- */
+const commander = require('../');
+// eslint-disable-next-line no-unused-vars
+const should = require('should');
```
usmonster: You might just want to do `require('should');` (unless that violates another ESLint rule...)

test/test.options.bool.js
```diff
@@ -1,15 +1,28 @@
...
+simpleFlagNoOptions.parse(['node', 'test']);
+simpleFlagNoOptions.should.not.have.property('pepper');
+simpleFlagNoOptions.cheese.should.be.true();
+
+const simpleFlagLong = simpleFlagProgram();
+simpleFlagLong.parse(['node', 'test', '--pepper', '--no-cheese']);
+simpleFlagLong.pepper.should.be.true();
+simpleFlagLong.cheese.should.be.false();
+
+const simpleFlagShort = simpleFlagProgram();
```
usmonster: This test seems a little redundant here since it's really testing flags, but fine to keep it.

Readme.md
```diff
@@ -101,8 +101,9 @@ cheese: stilton
 
 You can specify a boolean option long name with a leading `no-` to set the option value to false when used.
 Defined alone this also makes the option true by default.
```
usmonster: In rereading this and the edits, I'm not sure we're using consistent vocabulary on this specific line; i.e. it can be unclear whether "define" (or "specify", "add", "set", etc.) means "when defining the program" or "when invoking the program via the CLI."

It could just be that the sentence is slightly awkwardly worded (and sorry to comment on something that isn't a change in this commit).

test/test.options.bool.js
shadowspawn: Oh nice suggestion, matches intentions, I'll try that.

test/test.options.bool.js
shadowspawn: Actually I deleted the separate file that was testing boolean short flags, but I am regretting that. I'll put the file back and remove this explicit test!

Readme.md
shadowspawn: I agree with the awkward and potential for misinterpretation. I have struggled with the wording a couple of times, but at least we do have examples to consult. I'm not planning to have another go in this PR.

test/test.options.bool.js
shadowspawn: Done

test/test.options.bool.js
shadowspawn: Restored. Done.

shadowspawn:

shadowspawn:

PR #988, opened 2019-07-03
Remove API link, which is not being updated

### LINKED ISSUE #985: Drop API link from README?
There is a link in the README to the [API documentation](http://tj.github.com/commander.js/)

In the early days this may have been the most authoritative documentation. We now have a more comprehensive README, IDEs with built-in support for JSDoc, and TypeScript. In particular I suggest the primary use of the JSDoc is now for built-in editor support in the IDE.

I think the documentation is updated by merging to the `gh-pages` branch and running `make`, and then doing "something" to update GitHib Pages. Looks like last updated four years ago. #182 #171 #299

I suggest we delete the link from the README and focus on the other forms of documentation now.

shadowspawn:
Resolves #985

We are maintaining the README, and now have TypeScript too. We will keep maintaining the JSDoc but for built-in suggestions for IDEs that support it  rather than for generating a stand-alone reference (which was not being updated).

Removing the API link from both READMEs.

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -5,7 +5,7 @@
 [![NPM Downloads](https://img.shields.io/npm/dm/commander.svg?style=flat)](https://npmcharts.com/compare/commander?minimal=true)
 [![Install Size](https://packagephobia.now.sh/badge?p=commander)](https://packagephobia.now.sh/result?p=commander)
 
-  The complete solution for [node.js](http://nodejs.org) command-line interfaces, inspired by Ruby's [commander](https://github.com/commander-rb/commander).
+The complete solution for [node.js](http://nodejs.org) command-line interfaces, inspired by Ruby's [commander](https://github.com/commander-rb/commander).
 
 ## Installation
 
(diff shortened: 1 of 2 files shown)


Readme.md
```diff
@@ -5,8 +5,7 @@
 [![NPM Downloads](https://img.shields.io/npm/dm/commander.svg?style=flat)](https://npmcharts.com/compare/commander?minimal=true)
 [![Install Size](https://packagephobia.now.sh/badge?p=commander)](https://packagephobia.now.sh/result?p=commander)
 
-  The complete solution for [node.js](http://nodejs.org) command-line interfaces, inspired by Ruby's [commander](https://github.com/commander-rb/commander).  
-  [API documentation](http://tj.github.com/commander.js/)
+  The complete solution for [node.js](http://nodejs.org) command-line interfaces, inspired by Ruby's [commander](https://github.com/commander-rb/commander).
```
abetomo: It may be better to erase the leading space.

shadowspawn: I missed that, thanks. Done.

PR #992, opened 2019-07-07
Add helpInformation to typings

jharrilim:
Missing type for [Commander.prototype.helpInformation](https://github.com/tj/commander.js/blob/1c1ffca5d691ada5980ccfd81f42ecc006c0f62c/index.js#L1085-L1129):

https://github.com/tj/commander.js/blob/1c1ffca5d691ada5980ccfd81f42ecc006c0f62c/index.js#L1085-L1129

### DIFF
--- a/typings/index.d.ts
+++ b/typings/index.d.ts
@@ -277,6 +277,13 @@ declare namespace local {
      * @param {(str: string) => string} [cb]
      */
     help(cb?: (str: string) => string): never;
+
+    /**
+     * Returns the help information text as a string.
+     * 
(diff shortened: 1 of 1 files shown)


shadowspawn:
This is as intended since `helpInformation` is marked in the JSDoc with `@api private`. We do not document the private api calls in the TypeScript.

jharrilim:
Ok makes sense. I was thinking maybe it should be declared as public so that I may get the string and not always write to stdout (perhaps another stream), but maybe I am designing my program wrong.

PR #993, opened 2019-07-12
Upgrade packages

abetomo:
## Upgrade packages
Because there was a major upgrade.

## Drop Node.js 6 from .travis.yml
Node.js 6 is EOL.
I get the following error in a library that does not support Node.js 6.

```
node_modules/eslint/lib/cli-engine/cli-engine.js:257
        ...calculateStatsPerFile(messages)
        ^^^
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,5 @@
 language: node_js
 node_js:
-- "6"
 - "8"
 - "10"
 - "12"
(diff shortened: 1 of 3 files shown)


abetomo:
Is it all right to merge to master?

shadowspawn:
Yes. Do not need to wait for v3.
(But we should bump patch version too as changed from v2.20.0.)

abetomo:
Thank you for the review!

Issue #936, opened 2019-03-25
Run action for all commands


dland512:
Is there a way to have an action that always runs, no matter what command you are running? I tried this but it's not running the top level action attached to `program`:

```
var program = require('commander');

program
  .option('-c, --chdir <path>', 'change the working directory')
  .action(x => {
    // always run this
    console.log('chdir:', x.chdir);
  });

program.command('go')
  .option('-o --opt <whatev>', 'some option')
  .action(o => {
    console.log('opt:', o.opt);
  });

program.parse(process.argv);
```

shadowspawn:
I don't think is quite what you are asking, but these things came to mind:
- can use an option and [custom event listeners](https://github.com/tj/commander.js#custom-event-listeners)
- there can be a default action for [git-style sub-commands](https://github.com/tj/commander.js#git-style-sub-commands)

shadowspawn:
An answer was provided, and no further activity in a month. Closing this as resolved.

Feel free to open a new issue if it comes up again, with new information and renewed interest.

Issue #76, opened 2012-07-26
program, post option parse, pre action execute function


arieljake:
program
    .version('0.0.1')
    .option('-d, --debug', 'debug mode');
        .prepare(function() {});

program
       .command()
       .action()...

It would be nice to have a "prepare" or init function that runs after the options have been parsed from the command line but before the correct action is executed, for some common initialization common to all actions, like initialization of server connection given provided host and port options. Am I offbase or is this worth writing?

abetomo:
It seems to have been solved, so I will close it.

PR #996, opened 2019-07-16
Adds before:command and before:option events

holvonixAdvay:
These events are called with the name (or `*`) prepended to the
same arguments as the command:X and option:X events.

This allows you to run things after all top-level options are parsed, but before any commands are run.

### DIFF
--- a/index.js
+++ b/index.js
@@ -653,8 +653,11 @@ Command.prototype.parseArgs = function(args, unknown) {
   if (args.length) {
     name = args[0];
     if (this.listeners('command:' + name).length) {
-      this.emit('command:' + args.shift(), args, unknown);
+      args.shift();
+      this.emit('before:command', name, args, unknown);
+      this.emit('command:' + name, args, unknown);
     } else {
+      this.emit('before:command', '*', args);
       this.emit('command:*', args);
     }
   } else {
@@ -667,6 +670,7 @@ Command.prototype.parseArgs = function(args, unknown) {
     }
     if (this.commands.length === 0 &&
         this._args.filter(function(a) { return a.required; }).length === 0) {
+      this.emit('before:command', '*');
       this.emit('command:*');
     }
   }
@@ -732,6 +736,7 @@ Command.prototype.parseOptions = function(argv) {
       if (option.required) {
         arg = argv[++i];
         if (arg == null) return this.optionMissingArgument(option);
+        this.emit('before:option', option.name(), arg);
         this.emit('option:' + option.name(), arg);
       // optional arg
       } else if (option.optional) {
@@ -741,9 +746,11 @@ Command.prototype.parseOptions = function(argv) {
         } else {
           ++i;
         }
+        this.emit('before:option', option.name(), arg);
         this.emit('option:' + option.name(), arg);
       // bool
       } else {
+        this.emit('before:option', option.name());
         this.emit('option:' + option.name());
       }
       continue;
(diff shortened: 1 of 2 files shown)


shadowspawn:
Interesting, but I am not sure of the benefits, so exploring the idea.

- What is an example of what you might want to do in `before:command`?
- What is an example of what you might want to do in `before:option`?
- I am familiar with the `pre` and `post` extra scripts in npm which was the first thing I thought of. Is the pattern you have suggested used in some other context?
- Does this help with any open issues, or an independent suggestion?

holvonixAdvay:
What I wanted is code that runs after all options were processed, but before all commands were.  This is called middleware in the yargs package for example (we use it to configure logging and authentication).  For my own specific case, I had to return to yargs due to lack of support for async event handling.  So my own personal need is no longer - feel free to close it out if you aren't interested!

holvonixAdvay:
I am happy to change the PR to remove the before: emitters and keep the tests though - it doesn't seem events were being tested.

shadowspawn:
Thanks for reply. I'll take a look at the middleware idea, and tests alone too.

holvonixAdvay:
The most important thing is in the existing setup there is no way to run a single piece of code common to all commands - node event handlers have no notion of regex event matchers.

shadowspawn:
Yes, I think the common code was being asked for in #936

shadowspawn:
I am not ready for a PR to add more events at this stage. I want a clear idea of what problem we are solving and some demand for a solution. I found a 2012 issue asking for this in #76, but there was no further interest in the six years it was open.

I am interested in testing the events directly if you would like to do a new PR for that. The events are implicitly tested by running all of the other tests, but we are not checking the consistency and backwards compatibility for listeners so a useful additional test.

One change in style though. Calling parse again on the same Command object has the potential for a lot of side-affects as there is lots of state left behind. Please create a fresh Command object for each call to parse. (I know some of the existing tests do not currently do this!)

prokopsimek:
@shadowspawn I have the similar feature proposal. I would like to show an update notification in CLI while running any command. 
There could be a possibility to add custom event listener for `close` here: https://github.com/tj/commander.js/blob/master/index.js#L796-L802

I could set a function to be called at the end of all common commands then.

Something like:

```
program.on('close', () => this.notifyUpdate);
```
It does not work for now.

prokopsimek:
Ok, I have just found that it's possible to chain actions.

```typescript
program
      .action(MyCommand.run)
      .action(this.notifyUpdate);
```

shadowspawn:
A warning, action chaining like that will break in Commander v5. That is not a usage pattern I had thought of!

Is the chaining a solution for what you want to do, or are you just exploring possibilities? (I think a listener for the named command will still work in Commander v5 and can check.)

prokopsimek:
@shadowspawn I was just exploiting possibilities to implement what I wanted to do - check for updates and notify a user of our CLI after the end of each run. You can see PR  here: https://github.com/DXHeroes/dx-scanner/pull/231

prokopsimek:
I just don't want to remember that I have to add a new event listener if I add a new subcommand. Is there any listener for `--help`? Is it `options:help` listener?

shadowspawn:
Thanks for link, I had a glance through but didn't grasp the full context. 

There is an event for `--help`, and you are already using it with `.on('--help'`. 

If you have any more questions, please open up a new issue. We should get off this closed PR. 😊

PR #1010, opened 2019-08-10
Add .name and .usage, TOC, and tidy README

shadowspawn:
- add .name and .usage
- lint
- add TOC (using Markdown All in One in Visual Studio Code)
- rework the title levels so the help is a section in TOC

Main point of the PR is adding `.name` and `.usage` (#902).

Trying a table of contents. See what you think.

PR #1013, opened 2019-08-11
Add TOC and some lint, .usage and .name, to Chinese README

shadowspawn:
This includes the format changes from #1010, but initially just copies the english text for the section for .name and .usage.


shadowspawn:
Now updated with Chinese translation, thanks to @oGsLP.

PR #1016, opened 2019-08-11
change var to let and const, reguler fn exp to arrow fn

nocord:


shadowspawn:
Have you checked in what version of node those features were introduced? I expect they are pretty old, but we have not officially moved forward our minimum supported node version yet. (Intending to do that for next major version.)

There are some inappropriate substitution for `variadic` et al, but you might like to wait until I decide whether we want these changes now.

shadowspawn:
I checked my favourite support reference:
- https://node.green/#ES2015-bindings-const
- https://node.green/#ES2015-bindings-let
- https://node.green/#ES2015-functions-arrow-functions

There is only partial support for `const` and `let` in node 4 and 5. Arrow functions go back to node 4. (NB: we can not actually run our tests on versions of node that old, and do want to drop them!)

What do we currently have in index.js...
- const: just 1, introduced 3 months ago (has not caused reports yet, which is a good sign)
- let: 0
- arrow functions: just 1, introduced 1 month ago

I am leaning towards saying no for now for two reasons:
- we only just released v3 and I don't want high code churn so early in the release stream
- we do not yet have a public policy on supported node version

Is closing this ok with you @abetomo ?

shadowspawn:
A few friendly suggestions for future Pull Requests @nocord :
- look for a CONTRIBUTING document with guidelines for the project. GitHub should hopefully suggest reading one if available.
- add a description why you are opening the Pull Request
- look for "test" and "lint" scripts in the `package.json`. Lots of packages use these two script names, and you can run these tests locally before opening Pull Requests.
    - `npm run test`
    - `npm run lint`

nocord:
@shadowspawn  ok, thank you.

PR #1026, opened 2019-08-23
Unify readme code examples with 'const'

abetomo:
## Problem

'var' and 'const' are mixed in Readme code example.

## Solution
* It is better to be unified.
* Recently, 'const' is often used.


shadowspawn:
Oops, either my web page was out of date or I was not paying attention. Was intending to as myself for review (soon) but was already requested. :-)

shadowspawn:
On a slightly related note (`const`), three structure things I am thinking of for v4:
- only officially support LTS versions of node that we test with CI (but not deliberately break recent versions)
- switch testing framework
- add Windows CI testing

abetomo:
For testing on Windows, you can use GitHub Actions.

Example #1027

abetomo:
Thank you for the review!

Issue #1000, opened 2019-07-20
How can I ensure that exactly one argument has been supplied?


roxspring:
I've tried defining a command with a single required argument, and attempted to validate that exactly one source specified:

```
import * as program from 'commander';
import { version } from 'pjson';

program
  .version(version)
  .arguments('<source.md>')
  .action((source: string): void => {
    console.log('source=' + source);
    program.source = source;
  })
  .parse(process.argv);

console.log(program.args);
console.log(program.args.length);

if (program.args.length !== 1) {
  console.log('Invalid arguments');
  program.help();
}

console.log(program.source);
```

And running the program with no arguments, correctly prints 'Invalid arguments' since `program.args.length` is 0. But with 1 argument (`program a`) `program.args.length` is 2!? and outputs the following, claiming that `program.args.length` is 2!?

```
source=a
[
  'a',
  Command {
    ...
  }
]
2
Invalid arguments
```

Weirdly running with four arguments `program a b c d` also appears to drop the `b` argument somehow:

```
source=a
[
  'a',
  Command {
    ...
  },
  'c',
  'd'
]
4
Invalid arguments
```

So I guess I have two questions:

1. Where's the second argument gone?
2. (more importantly) how can I ensure that exactly one argument has been supplied?

shadowspawn:
There is not currently error detection for excess arguments, and it is not currently easy to handle it yourself in a general way. In simple cases you can check the argument count or the original arguments (`process.argv` or `program.rawArgs`).

Open issue is #749 and has some other ideas.

shadowspawn:
There are improvements on the develop branch for the next release (#1030 and #1048).

shadowspawn:
v4.0.0-0 prerelease published: #1067

shadowspawn:
v4.0.0 has been released with a fix for `program.args` which was getting the command appended (matching the arguments passed to the action handler).

Hopefully more useful now!

Feel free to open a new issue if still problems, with new information and renewed interest.

Issue #386, opened 2015-03-31
Problem with nested commands


foxx:
This seems to also be affecting another user on [SO](http://stackoverflow.com/questions/27066200/commander-cant-handle-multiple-command-arguments).

The output of `options.parent.args` seems to be malformed, the second element in the list appears to be a circular reference to itself, where as `rawArgs` seems to contain the correct items. Had a look at `parseOptions` in the source but the code isn't very clear in terms of documented logic.

At the very least this would be worthy of a docs update, but as I'm not too sure why the problem is happening, I'm reluctant to sent a PR fix.

Thoughts?
## Sample code

```
  var program = require('commander');

  program
    .version(version)
    .option('-v, --verbose', 'Enable verbose output')
    .usage('[options] [command]')

  program
    .command('runserver <args>')
    .description('run websockets server')
    .action(function(env, options){
      console.log(options.parent.rawArgs);
      console.log(options.parent.args);
    });

  program.parse(process.argv);
```
## Output

```
$ node bin/boards-server.js -v runserver 1 2 3 4 5 6 7 8 9
[ 'node',
  '/vagrant/server/bin/boards-server.js',
  '-v',
  'runserver',
  '1',
  '2',
  '3',
  '4',
  '5',
  '6',
  '7',
  '8',
  '9' ]
[ '1',
  { commands: [],
    options: [],
    _execs: [],
    _allowUnknownOption: false,
    _args: [ [Object] ],
    _name: 'runserver',
    parent:
     { commands: [Object],
       options: [Object],
       _execs: [],
       _allowUnknownOption: false,
       _args: [],
       _name: 'boards-server',
       Command: [Function: Command],
       Option: [Function: Option],
       _version: <Buffer 30 2e 30 2e 31>,
       _events: [Object],
       _usage: '[options] [command]',
       rawArgs: [Object],
       verbose: true,
       args: [Circular] },
    _description: 'run websockets server' },
  '3',
  '4',
  '5',
  '6',
  '7',
  '8',
  '9' ]
```

arboleya:
I'm not using nested commands but still I'm facing this `[Circular]` thing on my remaining args.

# Program

````javascript
var program = require('commander');

program
  .version(require('../package.json').version)
  .usage('<input> <output> [vars...]')
  .arguments('[input] [output]')
  .action(function(input, output, vars){
    console.log('input', input);
    console.log('output', output);
    console.log('vars', vars);
  })
  .parse(process.argv);
````
# Execution

````bash
[...]

PR #1030, opened 2019-08-24
WIP: Keep command out of program.args and add unknown to .action params

shadowspawn:
# Pull Request

I have not written any new tests or updated docs yet. This is a draft PR to show current code. 

## Problem

There are problems with the processing of arguments to call `.action`:
- extra arguments are ignored and the first one is overwritten
- the passed array of arguments is mutated, but that ultimately affects program.args, including especially inappropriately adding the command object into that array
- the combination of issues makes it hard for client code to detect extra arguments

See: #386 #749 #1000 

```

  program
    .option('-v, --verbose', 'Enable verbose output')
    .usage('[options] [command]')

  program
    .command('sub <something>')
    .action(function(oneArgument, cmd, extraArguments) {
      console.log(`Expected argument is: ${oneArgument}`);
      console.log('Extra arguments: ', extraArguments);
      console.log('program.args: ', program.args);
    });

program.parse(["node", "some-file.js", "-v", "sub", 1, 2, 3, 4, 5, 6, 7, 8, 9]);
```

```
$ node index.js 
Expected argument is: 1
Extra arguments:  3
program.args:  [ 1,
  Command {
    ... dump of command object!
  3,
  4,
  5,
  6,
  7,
  8,
  9 ]
```

## Solution

Changed code to prepare parameters for `.action` in a separate array without changing passed array. Pass unexpected parameters to `.action` as an array rather than just tagged on end.

```
$ node index.js 
Expected argument is: 1
Extra arguments:  [ 2, 3, 4, 5, 6, 7, 8, 9 ]
program.args:  [ 1, 2, 3, 4, 5, 6, 7, 8, 9 ]
```

### DIFF
--- a/index.js
+++ b/index.js
@@ -279,16 +279,13 @@ Command.prototype.action = function(fn) {
       }
     });
 
-    // Always append ourselves to the end of the arguments,
-    // to make sure we match the number of arguments the user
-    // expects
-    if (self._args.length) {
-      args[self._args.length] = self;
-    } else {
-      args.push(self);
-    }
+    // The .action callback adds the command itself after the expected arguments.
+    var expectedArgsCount = self._args.length;
+    var actionArgs = args.slice(0, expectedArgsCount);
+    actionArgs[expectedArgsCount] = self;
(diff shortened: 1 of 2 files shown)


shadowspawn:
Closing in favour of #1048 where reworked changes after move to Jest.

Issue #927, opened 2019-03-08
Jasmine Mocha Support


vlikin:
Convert tests to a more standard way. It is very hard to add new features to for external developers. Your library provides enough functionality. But it is not so well maintainable.

shadowspawn:
Thanks for the feedback. I was actually doing some research on this just yesterday, and can refer you to #649

shadowspawn:
@mojavelinux converted a couple of the test files to ava in #755 to show what it looks like, which I thought was a useful way of seeing real code for a modest time investment. If someone is interested in showing what it would look like using mocha, you could do a draft Pull Request with the same two files and a mocha setup.

shadowspawn:
Closing this in favour of #649

Issue #945, opened 2019-04-01
Triage: Exit


shadowspawn:
Commander calls `exit` under various circumstances, such as after displaying an error or the help. This is pretty convenient for the caller of a one-shot parse in a CLI, but is a problem when Commander is being used as part of a long running program or if caller wants more control over the failure handling.

Issues:

- #188 optional exit value for Help. (feature request)
- #435 process.exit(64)
- #443 Explicitly return when an error occurs
- #575 Make calls to process.exit() overridable
- #934 Parse multiple commands without exiting the process

Pull Requests:

- #436 Add support for exit codes.
- #444 Explicitly return from error methods
- #576 Make process.exit() injectable.

ngocdaothanh:
Yes, please simply throw errors. The outer program can catch the errors and handle them properly.

shadowspawn:
I think being able to opt-in to using a throw makes good sense. But I think it is a bit untidy for default handling of expected errors with the source line number and code? Compared with just a message intended for humans.

ngocdaothanh:
> a bit untidy for default handling

To be backward compatible, we can add an option to:
```Command.prototype.parse = function(argv)```

like this:
```Command.prototype.parse = function(argv, exitOnError = true)```

**Comment by shadowspawn** [comment]
I am experimenting with a routine to optionally replace calls to `process.exit`. Since the most likely implementation is to throw, it is passed an Error ready to throw. e.g. 

```
program.exitOverride((err) => { throw err; });
```

The custom error has properties for `exitCode`, `code`, and `message`. For example, the `--version` handling creates an error like this:

```
CommanderError(0, 'commander.version', this._version)
```

I'm currently just replacing the call to `process.exit` and not changing where or how the preceding messages are displayed. One thing at a time.

shadowspawn:
There are improvements on the develop branch for the next release (#1040).

shadowspawn:
v4.0.0-0 prerelease published: #1067

shadowspawn:
v4.0.0 has been released. It does not cover all the exit related issues found in triage, but addresses the most common issue of the unconditional call to exit.

Feel free to open a new issue if still issues, with new information and renewed interest.

PR #1035, opened 2019-08-31
WIP: Prototype Jest as testing framework

shadowspawn:
Following lead of #755 which converts two of the tests from should.js to AVA, this initally converts the same two tests to Jest

Related issues: #661, #649, #927

(Previous PR got closed when I deleted release/3.0.0 branch.)

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -1,5 +1,6 @@
 {
-  "extends": "standard",
+  "extends": ["standard", "plugin:jest/recommended"],
+  "plugins": ["jest"],
   "rules": {
     "one-var": "off",
     "semi": ["error", "always"],
--- a/package-lock.json
+++ b/package-lock.json
@@ -13,6 +13,112 @@
         "@babel/highlight": "^7.0.0"
       }
     },
+    "@babel/core": {
+      "version": "7.5.5",
+      "resolved": "https://registry.npmjs.org/@babel/core/-/core-7.5.5.tgz",
+      "integrity": "sha512-i4qoSr2KTtce0DmkuuQBV4AuQgGPUcPXMr9L5MyYAtk06z068lQ10a4O009fe5OB/DfNV+h+qqT7ddNV8UnRjg==",
+      "dev": true,
+      "requires": {
+        "@babel/code-frame": "^7.5.5",
+        "@babel/generator": "^7.5.5",
+        "@babel/helpers": "^7.5.5",
+        "@babel/parser": "^7.5.5",
+        "@babel/template": "^7.4.4",
+        "@babel/traverse": "^7.5.5",
+        "@babel/types": "^7.5.5",
+        "convert-source-map": "^1.1.0",
+        "debug": "^4.1.0",
+        "json5": "^2.1.0",
+        "lodash": "^4.17.13",
+        "resolve": "^1.3.2",
+        "semver": "^5.4.1",
+        "source-map": "^0.5.0"
+      },
+      "dependencies": {
+        "@babel/code-frame": {
+          "version": "7.5.5",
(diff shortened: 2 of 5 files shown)


shadowspawn:
I am happy with how this is going, and I like Jest. It is popular, comes with everything you need without extra installs or decisions,  and I can find what I want in its documentation.

Pushing on with converting tests to Jest.

shadowspawn:
Some notes on approach:
- making new Command per test and not using global
- naming tests "when <do this> then <this happened>"
- structuring test code keeping in mind Arrange Act Assert
- (usually) one assert per test
- doing some refactoring (intend similar coverage, but not preserving original tests or files)

shadowspawn:
Currently working through all the `test.options` files.

shadowspawn:
I have added `_exitOverride(callback)` as an experiment to allow suppressing calls to `process.exit` and instead throwing an error. I'll separate it out into a separate Pull Request, but wanted to try it after needing to mock `process.exit` when testing `.version`. I'll use it as doing the remaining tests and see how it goes for practical use. Relates to #945.

shadowspawn:
Finished converting existing tests with some refactoring. Tests run about 4 times faster on my computer, due to multithreading and rewrite of signal tests.

To finish up, I plan to replace remaining calls to process.exit and write tests for coverage of the new exitOverride.

shadowspawn:
Suggested CHANGELOG entry:

### Changed

- switched tests from Sinon and Should to Jest

shadowspawn:
Failing on node 8. I was not expecting node version problems. Hurrah for tests!

shadowspawn:
This is a big change, although does not affect a production install, so I was planning to land this in  a major version. 

We could make a `release/4.0.0` branch like we did for 3.0.0 or just work on `develop`. Do you have a preference @abetomo ?

I am happy either way. I am leaning slightly towards just working on develop because I don't think there are small fixes worth shipping before 4.0.0 and this PR is the biggest piece of work I plan for v4.

shadowspawn:
Giant PR! But not expecting you will look at every file. :-)
Questions welcome.

tests/options.flags.test.js
```diff
@@ -0,0 +1,76 @@
+const commander = require('../');
+
+// Test the various ways flags can be specified in the first parameter to `.option`
+
+test('when only short flag defined and specified then value is true', () => {
+  const program = new commander.Command();
+  program
+    .option('-p', 'add pepper');
+  program.parse(['node', 'test', '-p']);
+  expect(program.P).toBe(true);
```
abetomo: I'm sorry if I sound picky.
It seems good to use `toBeTruthy`.
https://jestjs.io/docs/ja/expect#tobetruthy

shadowspawn: No problem. I did consider `toBeTruthy`, but I decided important to test for `true` in at least some places. Say if we changed the implementation to return `1` or `default` or some other truthy value, some of the tests should fail because it would break some clients.

```

shadowspawn:
Are you still reviewing @abetomo? (Just checking in case comment was only question. No rush, please take your time to check what you want.)

abetomo:
What do you think about checking the test code with eslint?

shadowspawn:
In `npm run lint`? Good idea, yes.

abetomo:
LGTM!
Thank you for a wonderful job.

When eslint is over, I will review that part again.

shadowspawn:
Pushed, with eslint errors fixed (and the deliberately skipped tests renamed to avoid a persistent warning).

shadowspawn:
(GitHub Action tests failed on macOS 8, but I can't reproduce locally in first try.)

Issue #934, opened 2019-03-19
Parse multiple commands without exiting the process


Bernd-L:
# Goals

- Parse a final (immutable) string and take action accordingly
- Keep the process open (including on `-h` and `-v`, and even on incorrect input)
- Use commander.js as a library (will get called sometimes later, triggered by events)

## Edit:
Goals continuation:
- Get the output from `-v` and `-h` into a string instead of into the terminal

# Use case

A bot which should listen for new messages in a live-chat channel.
Upon the arrival of a new message, it will look for a specific prefix, then send the command to jommander.js to be parsed, and will take action accordingly.

Notice: The bot itself runs on a server in an automated manner and doesn't need nor expect any command-line arguments. The goal here is to parse several commands without the node process exiting because commander.js feels like it.

When a `prefix -v` comes along, it should respond. And it does - great!
But now my bot has made a `clean exit`.
Turns out commander.js decides when my app is done executing.

This behaviour is, of course, unacceptable, as the next user typing `prefix -v` into the chat will wait forever, for the bot is now shut down.

The fact that the bot needs to be reconnected to the chat-API and a database every time it starts doesn't make this situation easier.

# TL;DR

How does one parse multiple commands without having their app shut down in between?
How to get the output from `-v` and `-h` into a string instead of into the terminal?

---
PS: #443 Talks about something somewhat-related

shadowspawn:
Thanks for the use case, and for the succinct `TL;DR` too. 

Related open issues about not calling exit: #443 #575 

And issues about allowing custom exit code, so similar but different use case: #188 #435

shadowspawn:
I have started work on an exit override: https://github.com/tj/commander.js/issues/945#issuecomment-527412375

shadowspawn:
There are improvements on the develop branch for the next release (#1040).

shadowspawn:
v4.0.0-0 prerelease published: #1067

shadowspawn:
v4.0.0 has been released.

PR #1040, opened 2019-09-09
Add override to allow throw instead of process.exit

shadowspawn:
# Pull Request

## Problem

Commander has multiple calls to `process.exit` which is hard to override to use as a library and parse multiple strings, as parsing errors or --help terminate the process. Doing a throw instead would allow appropriate handling and carry on.

Issues: #945 #575 #934 

## Solution

Add `.exitOverride(callback)` routine to do custom handling of code that would otherwise call `process.exit()`. The callback is optional and the default behaviour is to throw an instance of CommanderError which includes the exitCode, a string code, and a message.

This allows writing code like:

```
program
  .exitOverride()
  .option(-p, --pepper <type>, 'add pepper');

try {
  program.parse(['node', 'test', '--pepper']);
} catch (err) {
  // custom processing
}
```

(The routine name `exitOverride` is a bit implementation specific rather than friendly, but this is an optional feature and not sure how much or how it will be used.)

### Limitations

- code flow is not changed, so override still expected to throw or exit and not fall through
- executable subcommand calls spawn as async, not currently overriding those `process.exit` as different pattern required
- writing to stdout and stderr as normal before override

## ToDo

- TypeScript
- README
- not doing tests here as have extensive tests in `feature/jest` branch

Making the code visible for discussion and review.


shadowspawn:
One of the prompts for adding this feature was to simplify writing unit tests. The test file for exitOverride itself (on a different branch) is: 

https://github.com/shadowspawn/commander.js/blob/feature/jest/tests-jest/command.exitOverride.test.js

shadowspawn:
Added support for async calls in executableSubcommand.

shadowspawn:
Allows custom exit codes: #188 #436

shadowspawn:
The implementation already landed in the Jest PR. This PR now effectively promotes it from private `_exitOverride` to public `exitOverride`.

Added description to README.

Suggested line for CHANGELOG:

* `.exitOverride()` allows override of calls to `process.exit` for additional error handling and to keep program running ([#1040])

LXSMNSYC:
Hello, I was looking for a solution to prevent "--help" from exiting the CLI.

I tried using `exitOverride`, but it doesn't really prevent it from exiting. Is there any way for this?

shadowspawn:
@LXSMNSYC 
`exitOverride` would be my suggestion. Open a new issue with some more details. (I am not sure what you mean by "doesn't really prevent it from exiting".)

LXSMNSYC:
> @LXSMNSYC
> `exitOverride` would be my suggestion. Open a new issue with some more details. (I am not sure what you mean by "doesn't really prevent it from exiting".)

I have found a work around to my solution.

I have to take a look at commander's index js to know how it executes commands, and ti understand the behavior in general :)

PR #1047, opened 2019-09-21
Call default subcommand with unknown options

shadowspawn:
Finished off #614, see that PR for description and discussion


shadowspawn:
Open issue is #461

Issue #7, opened 2011-08-24
Show help if no command was executed


ghost:
Hi

I wanted my utility to show help if no command was executed.

This was how I monkey patched my code to solve it.

```
parentAction = commander.Command::action
commander.Command::action = (fn) ->
    newFn = ->
        fn.apply this, arguments
        commander.commandExecuted = this
    parentAction.apply this, [newFn]

# later after parse statement
if !commander.commandExecuted?
    process.stdout.write commander.helpInformation()
```

Is this an acceptable solution, or is there something I'm missing in the API?

Chris

foxx:
PR sent, sorry for the delay

mgenereu:
What if I want the argument missing messages and not the help?

mgenereu:
I should leave an example.  Given this code:
```js
program
  .version('1.0.0')
  .arguments('<abc> <def> <ghi>')
  .parse(process.argv)
```
my experience is:
```
$ program
$ program first
  error: missing required argument `def'
$ program first second
  error: missing required argument `ghi'
```
I was expecting:
```
$ program
  error: missing required argument `abc'
  error: missing required argument `def'
  error: missing required argument `ghi'
$ program
  error: missing required argument `def'
  error: missing required argument `ghi'
$ program first second
  error: missing required argument `ghi'
```

pedro93:
+1 for this

lancedolan:
For posterity wondering specifically about the case where a _wrong_ command is given, such as a case where a user gives a bunch of arguments but forgets the command itself, this still isn't handled in the latest commander and the user will get blank unhelpful output. 

However, the command docs provide a very easy to handle the case explicitly yourself:

```
// error on unknown commands
program.on('command:*', function () {
  console.error('Invalid command: %s\nSee --help for a list of available commands.', program.args.join(' '));
  process.exit(1);
});
```

anantanandgupta:
here is what i have written

```
program.allowUnknownOption(false);

const parsed = program.parse(process.argv);
if (!(parsed.args && parsed.args.length > 0 && (typeof (parsed.args[0] === 'object')))) {
  program.outputHelp();
}
```

antonkulaga:
It is really annoying that there is not flag to automatically display help after parse errors when it crashes. If it will not be added I will just switch to another CLI lib in my next project

shadowspawn:
Open a new issue with more details if you want visibility on this @antonkulaga

[...]

PR #1048, opened 2019-09-21
Keep command out of program.args and add unknown to .action params

shadowspawn:
This reimplements #1030 after the move to Jest. See #1030 for description and discussion.

### DIFF
--- a/index.js
+++ b/index.js
@@ -346,16 +346,16 @@ Command.prototype.action = function(fn) {
       }
     });
 
-    // Always append ourselves to the end of the arguments,
-    // to make sure we match the number of arguments the user
-    // expects
-    if (self._args.length) {
-      args[self._args.length] = self;
-    } else {
-      args.push(self);
+    // The .action callback takes an extra parameter which is the command itself.
+    var expectedArgsCount = self._args.length;
+    var actionArgs = args.slice(0, expectedArgsCount);
+    actionArgs[expectedArgsCount] = self;
+    // Add the extra arguments so available too.
+    if (args.length > expectedArgsCount) {
+      actionArgs.push(args.slice(expectedArgsCount));
     }
 
-    fn.apply(self, args);
+    fn.apply(self, actionArgs);
   };
   var parent = this.parent || this;
   var name = parent === this ? '*' : this._name;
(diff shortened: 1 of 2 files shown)


xPaw:
This PR now makes it impossible to detect whether any command has run, like the example in: https://github.com/tj/commander.js/issues/7#issuecomment-475485954

We want to print help when no command has run, and it doesn't really seem to be possible (process just exists without printing anything). Checking `process.argv.length` is not really a good option either.

And on top of that `command:*` is not triggered either if you pass no command.

shadowspawn:
@xPaw 
Sorry you hit issues.

This PR does change the behaviour in an area that people have written code based on undocumented details of the previous behaviour, so will cause some breakages. Hopefully we can work out a more robust way to achieve the same behaviours, and where appropriate add migration steps to the Release Notes.

I test `process.argv.length` in my own CLI program for the specific case that user ran my command with no arguments:
```
// Show help if no command specified.
if (process.argv.length === 2) {
  program.help();
}
```

Please open a new issue with an example of what you are trying to do and some example code.

xPaw:
`process.argv.length` is not really an elegant solution.

Issue was already made in 2011 about this: #7, you could just re-open that.

shadowspawn:
That one was closed, with some up-voted example code supplied which tests `program.args` rather than `process.argv`: https://github.com/tj/commander.js/issues/7#issuecomment-32448653

Does that work for your case?

xPaw:
This PR specifically "broke" testing `program.args` as it not longer pushes the found command to it. That's what I did before.

My current "fix" was to use `program.rawArgs`

https://github.com/thelounge/thelounge/pull/3488/commits/e09599aeae474c854a15b3b66b665ad01c610399#diff-25f700bb76bbf957569bd288fbf7d328L63-R63

shadowspawn:
Ok, I see the problem. Not sure about best solution yet, but that is reasonable.

shadowspawn:
I have added a migration tip to the Release Notes, but it might be an issue that needs a better solution. The detection of unrecognised commands and missing commands when using action handlers is a bit fragile.

Issue #1045, opened 2019-09-14
Only officially support node 8 and higher from Commander v4


shadowspawn:
We have been planning for a long time to use more modern javascript features, but changes will break older versions of node and we do not have a documented policy.

Node 8 is the oldest version in LTS. I suggest we only officially support node 8 and higher from Commander v4. 

However, I suggest we actually target node 6 when considering language features and not deliberately break node 6 until there is good reason. That should cover most people using a current version of Commander while node 6 usage drops off. We no longer run tests on node 6 though so it won't be officially supported.

shadowspawn:
I think the README addition is useful to be clear about what we are testing, but based on #1073 I think in practice we will still  need to bump major version if we deliberately make code changes to (say) require node 6 or 8. Without being noisy about the required version using `engines` we will still break people who might reasonably expect compatibility when updating within a major version.

 - https://docs.npmjs.com/files/package.json#engines

shadowspawn:
I noticed today that that I used `class` for defining new `CommanderError` so we have indeed started requiring newer version of node, and not just a future consideration! Class was fully supported from node 6.

- https://node.green/#ES2015-functions-class
- https://docs.npmjs.com/files/package.json#engines

I'll do a PR adding engines field set to require node 6 to be explicit that Commander v4 onwards requires node 6 or higher.

shadowspawn:
v4.0.0 has been released.

PR #1053, opened 2019-09-22
Add note that only supported on node 8 and above

shadowspawn:
# Pull Request

## Problem

We do not have a stated policy for node support, and have long wanted to be able to use some new javascript language features. I would like to be clear for myself what we can and can't use without breaking clients.

Related issue is #1045

## Solution

Say that we only support node 8 and higher in the README.

As per #1045, I suggest we actually stick to features supported by node 6 for now.

(I deliberately have not put node 8 as an actual requirement in the `engines` field in the `package.json` since we are still work on older versions at least currently.)

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -42,8 +42,6 @@ The complete solution for [node.js](http://nodejs.org) command-line interfaces, 
 npm install commander
 ```
 
-Commander is supported on Node 8 and above (and may still work on older versions but is not tested below 8).
-
 ## Declaring _program_ variable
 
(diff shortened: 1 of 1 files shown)


shadowspawn:
I had a second go, and moved this under the Support section instead of Installation. I also added a comment about the Issues page, partly to balance the TideLift info.

(I don't think Support would be the right place if we enforced the allowed Node version, but feels a reasonable fit since we are saying what we support rather than what is compatible.)

Issue #843, opened 2018-08-07
Regression between v1.16 and v1.17 when no command are passed and '*' is registered


Skywalker13:
There is a regression in the version v1.17

See this sample:
```js
var program = require("commander");

program
  .version("0.0.1")
  .command("*")
  .description("test")
  .action(function(cmd) {
    cmdValue = cmd;
  });

program.parse(process.argv);

if (typeof cmdValue === "undefined") {
  console.error("no command given!");
} else {
  console.log("command:", cmdValue);
}
```

## v1.16
```bash
$ node test.js
no command given!
```

## v1.17
```bash
$ node test.js
command: Command {
  commands: [],
  options: [],
  _execs: {},
  _allowUnknownOption: false,
  _args: [],
  _name: '*',
  _noHelp: false,
  parent:
   Command {
     commands: [ [Circular] ],
     options: [ [Object] ],
     _execs: {},
     _allowUnknownOption: false,
     _args: [],
     _name: 'test',
     Command: { [Function: Command] super_: [Object] },
     Option: [Function: Option],
     _version: '0.0.1',
     _versionOptionName: 'version',
     _events:
      { 'option:version': [Function],
        'command:*': [Function: listener] },
     _eventsCount: 2,
     rawArgs:
      [ 'C:\\Program Files\\nodejs\\node.exe',
        'C:\\Users\\schroeterm\\Desktop\\test\\test.js' ],
     args: [] },
  _description: 'test',
  _argsDescription: undefined }
```

abetomo:
Thanks for the report!
It is a bug when I fixed this issue (#778).
I will fix it.

Issue #742, opened 2018-01-06
Allow a non-executable to be set as the default command


mojavelinux:
When designing a CLI that uses command functions, it should still be possible to set a command as the default command if no command is specified. This is pretty much the same functionality as the existing default command, except it will be used for non-executable commands.

```js
cli
  .command('build', { isDefault: true })
  .description('Run the build')
  .action(() => {
    console.log('build!')
  })
```

mojavelinux:
I'd be fine with saying that the two types of defaults are mutually exclusive if that turns out to be a challenge. Chances are, you are either designing one type of CLI or another.

shadowspawn:
Support for setting a non-executable as the default command is included in the published pre-release of Commander v5.0.0 (#1163)

shadowspawn:
Commander v5.0.0 has been released.

https://github.com/tj/commander.js/releases/tag/v5.0.0

PR #258, opened 2014-09-09
Add ability to specify arguments syntax for top-level command

rrthomas:
* index.js: Add arguments method, support top-level command in action.

This would fix issue #257.

PR #1062, opened 2019-09-29
Run default command if it's not just *

zephraph:
# Pull Request

## Problem

I'm trying to build a command like the one below

```
cli [optional]
cli subcommand [optional]
```

The implementation loosely looks something like this...

```
program
	.argument('[optional]')
	.option('-d', 'do a thing')
	.action(() => { console.log('whoop') })

program
	.command('subcommand [optional]')
	.option('-b', 'blah')
	.action(() => { console.log('beh') }

program.parse(process.argv)
```

When trying to run `cli -d` the action for the main command won't run. Using `program subcommand -b` runs fine.

## Solution

I traced it down to https://github.com/tj/commander.js/pull/844 which contains a fix for this issue: https://github.com/tj/commander.js/issues/843. In said issue, @Skywalker13 reported that it was expected that a program that only had one command registered that was the `*` command shouldn't trigger an action. The fix in #844 just changed it so that if there were _any_ commands registered then the default command won't run. 

I preserved the behavior from #843 while also ensuring that valid usecases (i.e. when commands are present besides `*`) will still cause the base action to trigger. 

## ChangeLog

Run actions when no optional arguments are provided and real commands are registered.


shadowspawn:
This might already be fixed by #1047 and #614

zephraph:
I don't run them as separate executables. The commands are all configured in the same file/process.

shadowspawn:
Oh right, sorry. I was just working on v4 so fired off a quick reply!

zephraph:
I do appreciate the quick reply! 😄

shadowspawn:
I did a little reading and experimentation. Crikey, we have a lot of little quirks and special cases around whether there are subcommands, whether there are git-style executables, whether there are parameters...

At the moment I am not sure if this is a change making things more consistent, although I agree it does not break the existing tests. This change feels a bit too special case to me on first impression.

I'm not going to accept this PR yet as I do not have a clear idea (in my own mind) of sensible high-level story for what we intend for non-subcommands calls with:
- program with and without subcommands using action handlers
- program with and without subcommands using git-style executables
- program with and without action handler on program
- program with and without `command( '*')`
- call with and without arguments
- call with and without options

I'm not planning to fix everything at once, but want to be more confident that heading in right direction.

zephraph:
There's certainly a lot going on, so I understand that sentiment. I have a patch (using [patch-package](https://github.com/ds300/patch-package#readme)) to unblock me for now. 

> This change feels a bit too special case to me on first impression

Again, I completely understand this. This is a unique issue that occurs in a very specific circumstance. Here's a minimal reproduction of what I ran into:

```
// command.js
const commander = require("commander")

// This should be called, but won't be because there's another command registered
program.arguments('[option]').action(() => {
  console.log('called')
})

program.command('something').action(() => {
  console.log('does\'t matter')
})

program.parse(process.argv)
```

```
node command.js
```

If the program takes _only_ optional arguments _and_ it has an action, that action will not be called if _any_ commands are registered. The program's action here is treated basically the same as `command(*)`. If there's a required argument or there are no commands registered, it's totally fine.

shadowspawn:
I have been doing some research today into general command handling. 

The original behaviour was that you could define `.command('*')` to get called as a default handler for the runtime command/argument. The ability to add action handler and pass arguments to program using `.arguments()` got added in #258. The new support *also* uses the event `command:*` as it is is somewhat a replacement or alternative.

I think when program has an action handler with defined arguments, it is reasonable and expected it get called when the calling signature matches, which would mean it should be called in @zephraph example program in opening comment. (I don't think using `.arguments()` is required, but it does make it more obvious that the user knows what they want!)

I am not sure whether we need to treat `.command('*')` as different, but I think can look to the default command handling for executable (sub)commands for inspiration. For example, that has been extended to include calling when just options and no other arguments. (See also #843 for using a defined command as the default handler.) 

So I think there is a "bug" here, but I don't this is the right fix.

@zephraph would you like to explore a better fix here in this PR, or open a new issue for future consideration with your problem description?

zephraph:
I’m open to explore other options here. 

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

shadowspawn:
Listening for `command:*` is still a documented use case for detecting unknown commands. I do not think the event is expected to be emitted when there are no arguments. (So effectively the `*` matches any string but not a missing string.)

I think to some extent `.command('*')` is legacy now, but we could treat it as a fancy way of declaring a `command:*` listener and keep it backwards compatible, even if we remove it from the documentation.

With an action handler on the main program, I think there is a reasonable expectation that it is called if the signature matches the declared `.arguments()`, including with no parameters.

To allow separate behaviours for `command:*` and program action handler, we would either use a new event for the program action handler, or just make it possible to tell it was defined if they are not usefully both defined.

I haven't though of a use case where I would want both an action handler on main, and a separate listener on `.command:*`?

zephraph:
> To allow separate behaviours for command:* and program action handler, we would either use a new event for the program action handler, or just make it possible to tell it was defined if they are not usefully both defined.

Yeah, I think this is the right direction. If they're both defined that's indeterministic functionality and should be a runtime error. I don't think having them both would ever make sense, because you can't  differentiate between the arguments on main and an invalid command.

shadowspawn:
Renamed PR while Work In Progress. 🚧 🛠 
😄

zephraph:
Okay @shadowspawn I think I got something decent. I'm sure there are details we can work through. 

- There's a `program-action` <sub>(name?)</sub> now that's for the main program action handler. That's what gets registered if there's no parent. 
- Having both `program-action` and `command:*` handlers registered at the same time is considered an error. 
- If a command doesn't match, it'll emit `command:*" _if there's a registered handler_. Otherwise it'll emit `program-action`. 

The only thing I'm not really confident in is what happens if `args.length` is `0`. 
Specifically this line: https://github.com/tj/commander.js/blob/6a87525ef966b5a2259fddc88df5c09dbf30b5e9/index.js#L821

I removed that funky `command:*` call logic on the `this._arg` check, but I might need to add it back? Open to feedback.

zephraph:
I went back through the old issue that added some of the previous logic and made sure we had a test that matched that logic exactly. It adds a _little_ more confidence at least.

index.js
```diff
@@ -793,25 +797,28 @@ Command.prototype.normalize = function(args) {
 Command.prototype.parseArgs = function(args, unknown) {
   var name;
 
+  if (this.listeners('program-action').length && this.listeners('command:*').length) {
```
shadowspawn: I do not think need to error? My earlier comment about wondering what use case there was for both was about understanding existing patterns rather than about preventing both.

index.js
```diff
@@ -793,25 +797,28 @@ Command.prototype.normalize = function(args) {
...
+    console.error("Can't have listeners for both command:* and for the main program");
+    this._exit(1);
+  }
+
   if (args.length) {
     name = args[0];
     if (this.listeners('command:' + name).length) {
       this.emit('command:' + args.shift(), args, unknown);
-    } else {
```
shadowspawn: The old behaviour was emit `command:foo` if defined, else emit `command:*`. There could be multiple listeners on `command:*`.

I think the backwards compatible change is: emit `command:foo` if defined, else emit `command:*` and also emit `program-action` if defined. (Don't actually need to test for `program-action` being defined but it feels wrong to dispatch that event in subcommands.)

So I suggest:

```
} else {
  if (this.listeners('program-action').length) {
    this.emit('program-action', args, unknown);
  }
  this.emit('command:*', args, unknown);
}
```

index.js
```diff
@@ -793,25 +797,28 @@ Command.prototype.normalize = function(args) {
...
     // If there were no args and we have unknown options,
     // then they are extraneous and we need to error.
     if (unknown.length > 0 && !this.defaultExecutable) {
       this.unknownOption(unknown[0]);
     }
-    if (this.commands.length === 0 &&
-        this._args.filter(function(a) { return a.required; }).length === 0) {
-      this.emit('command:*');
-    }
+    this.emit('program-action');
```
shadowspawn: The filter was to avoid calling program action handler when it has a required option which is missing, like:
`.arguments(<file>)`. So I think that is still needed.

I am not sure about the `this.commands.length` check, still working out what that was covering and why.

shadowspawn: Oh, we have an explicit test that program action handler not called if there is a subcommand and no arguments are specified. The test name is missing part of the scenario: "when no arguments then asterisk action not called'.

I want to read back through the issues to see if that was a key behaviour, or just the safest way of adding the call to program handler in special circumstances.

shadowspawn: Ah ha! The `this.commands.length` was to avoid calling `.command('*')` handler when no parameter from https://github.com/tj/commander.js/pull/844/files

I do not think we need that check now, I do not see a reason change emit of `program-action` bases on whether there are commands.

tests/command.asterisk.test.js
```diff
@@ -10,6 +10,16 @@ test('when no arguments then asterisk action not called', () => {
   expect(mockAction).not.toHaveBeenCalled();
 });
 
+test('when no arguments with asterisk handler then asterisk action not called', () => {
```
shadowspawn: Good. 

This test from #844 got broken/lost when I moved the tests to Jest. I didn't understand quite what it was checking. Now after all my research in this issue, I *think* I understand the scenarios, and should add some more tests while it is making some sense!

shadowspawn:
> program-action (name?) event

Hmm, I was wondering what to call this too. I would prefer something closer to style of `command:foo`.
Perhaps `command-program` or `program-command`?

shadowspawn:
(The code changes are surprisingly small, which makes me hopeful that this is a good direction.)

tests/command.asterisk.test.js
zephraph: Feel free to add any other tests you can think of.

index.js
zephraph: Preventing both does make it easier though. We can’t really differentiate between the two. If someone has both and it’s not an error either they both get called which seems unexpected, or only one gets called and it won’t be immediately obvious which. 

I really do think it’d be good to error for this case.

index.js
zephraph: Hmmm.. the current logic would call `command:*` if there are no matches for the provided command name and there’s at least one * listener. 

That does seem to back up the current behavior, right?

index.js
shadowspawn: If someone has both then calling them both is backwards compatible, so that is my preferred behaviour.

I don't see why people would want to define both, but I don't see a significant benefit in blocking it. I don't recall seeing reported issues about people wanting different behaviours when defining both — either wanting just one called,  or being confused by calling both.

I do empathise with wanting to reduce the number of possible scenarios since we had trouble reasoning about it, but not at the cost of breaking code patterns we didn't think of.

index.js
shadowspawn: I didn't quite understand your comment. Does my suggestion make more sense in the light of the separate comments about calling both being backwards compatible? I suspect we had different expectations about calling one/both.

Otherwise, comment again as I missed the point! (Thanks.)

shadowspawn: This discussion is helping my understanding. I am updating my suggestion to be more self-consistent. The two cases are symmetrical so should be treated the same. Simply:

```
} else {
  this.emit('program-action', args, unknown);
  this.emit('command:*', args, unknown);
}
```

shadowspawn:
Added lots of tests in two Pull Requests onto your branch @zephraph, but I can merge them separately and no problem if you don't want them in your branch.

index.js
shadowspawn: Note: I added the filter back in one of my Pull Requests.

tests/command.asterisk.test.js
shadowspawn: Added lots in Pull Request!

zephraph:
More than happy to add them. Thanks!

index.js
zephraph: 👍 I updated this as suggested

index.js
zephraph: I updated this as suggested

zephraph:
I updated the PR based on our convos and renamed `program-action` to `program-command`.

zephraph:
I just noticed the semver major label. If we consider this a breaking change, I'd rather go back a step and simplify the two states. If we _don't_ consider it breaking, then I'm good with it as it is in its current state.

shadowspawn:
> I just noticed the semver major label.

This changes emit calling behaviour for `command:*` and calling behaviour for program action handler under some circumstances. The changes are not simply bug fixes as such. So I think should go in a major version.

> If we consider this a breaking change, I'd rather go back a step and simplify the two states.

In what sort of way?

(I am a little delighted that after me worrying about how to approach this properly and how complicated it might be, you implemented the separation in a few lines of code! 🎉 )

index.js
```diff
@@ -360,8 +360,13 @@ Command.prototype.action = function(fn) {
     fn.apply(self, actionArgs);
   };
   var parent = this.parent || this;
-  var name = parent === this ? '*' : this._name;
-  parent.on('command:' + name, listener);
+  if (parent === this) {
+    parent.on('program-command', listener);
+    parent.on('command:*', listener);
```
shadowspawn: Ah no, I think should just be adding listener for `'program-command'` and not `'command:*'` or the listener will get run twice.

The listeners for `command:*` get created when caller explicitly adds one (e.g. for detecting unknown commands) or caller adds a command called `*` (i.e. `.command('*')`

zephraph:
```suggestion
```

zephraph:
Having both `program-command` and `command:*` listeners be valid at the same time feels like a indeterministic state that would be hard to reason about. I just believe if we're making it a breaking change it might be good to consider if there are simplification opportunities here that make sense. 

Of course, if you'd rather just push forward with this solution I'm 👍.

shadowspawn:
Thanks. I am happy with current solution, pushing forward. 😄 

A reason for not doing more with the two states is my tentative long term goal is to replace `command:*` with clearer patterns. I opened #1090 when I was stuck thinking about how to move forward on this PR.

- this PR makes the program action handler stand alone and no longer use `.command:*`
- #1088 might replace use of `.on('command:*')` for detecting unknown commands
- #742 might replace use of `.command('*')` as a default command

As for when to release this... Maybe February! The last major release was 1 November and I think we should have at least 3 months between major releases, even when they are small ones.

I'll move out of WIP and finish a review.

Thanks for your patience and contributions. 👍

shadowspawn:
We will merge this onto `release/5.x` branch and not `develop`.

shadowspawn:
FYI: I am completely rewriting the command handling in #1149 informed in part by the hard work we did here to understand the interactions of the "normal" command handling and the `*` handling and how they might be used.

shadowspawn:
This is hopefully behaving as desired in published pre-release of Commander v5.0.0 (#1163)

shadowspawn:

Issue #997, opened 2019-07-18
Could have error() output to stderr?


crystalfp:
I have a command line application with the output redirected to a file.
```
node app.js mandatory-argument > output.dat
```
If `mandatory-argument` is not present, commander `help()` function is called. Unfortunately the help text goes to the output file due to the redirection. In the commander source file, the help output is hardcoded to go to stdout (line 1147).

A workaround is to call `help()` this way:
```
	const cb = (hlpText: string): string => {
		process.stderr.write(hlpText);
		return "";
	};
	program.help(cb);
```
But should be simpler if all commander messages go to stderr and not to stdout.
Thanks for looking!
mario

shadowspawn:
I saw a related article tonight: https://www.jstorimer.com/blogs/workingwithcode/7766119-when-to-use-stderr-instead-of-stdout

cengizhanbasak:
If you like me to do, I'll change the stdout to stderr.

crystalfp:
Yes, please.

shadowspawn:
@crystalfp Are you displaying the help yourself? Or is Commander displaying the help when you call `.parse`?

crystalfp:
It is commander that displays the help.
I had to add the following code to display help when the program is called without arguments:
```

	// To send the help text to stderr
	const cb = (hlpText: string): string => {
		process.stderr.write(hlpText);
		return "";
	};
	program.help(cb);
```

shadowspawn:
That isn't a bad work-around. 

When using git-style executable subcommands, the help gets displayed automatically when there are no arguments (unless there is a default command). I was wondering if this was the behaviour you had encountered.

I don't think it is appropriate for all Commander messages to go to `stderr` though. When the help is explicitly requested, the caller is likely to expect the output to be on `stdout`, e.g.:

```
my-command --help | grep port
```

crystalfp:
This was the only way to have the help text visible when stdout is routed to a file. Anyway, for my specific program, this workaround solved the problem. Ah, and I don't use subcommands. 

I understand your points. I will continue with my workaround any time I have to redirect stdout.
Thanks!
mario

shadowspawn:
An idea for future is is to have the optional parameter to the help calls be the `logger`, and this could be passed through to the help event listeners too. But listener might want to detect it it is a tty to turn on/off colour, so tradeoffs with function vs stream...

e.g. `program.help(console.error);` or `program.help(process.stdout)`

shadowspawn:
[...]

PR #1069, opened 2019-10-03
Write help text to stderr instead of stdout

cengizhanbasak:
# Pull Request

https://github.com/tj/commander.js/issues/997

## Problem

Informative functions were writing to stdout, causing problems in piping the output of the script.

## Solution

Changed writing stdout to stderr.

## ChangeLog

- Changed stdout to stderr for informative outputs.

### DIFF
--- a/index.js
+++ b/index.js
@@ -880,7 +880,7 @@ Command.prototype.version = function(str, flags, description) {
   this._versionOptionName = versionOption.long.substr(2) || 'version';
   this.options.push(versionOption);
   this.on('option:' + this._versionOptionName, function() {
-    process.stdout.write(str + '\n');
+    process.stderr.write(str + '\n');
     process.exit(0);
   });
   return this;
@@ -1172,7 +1172,7 @@ Command.prototype.outputHelp = function(cb) {
   if (typeof cbOutput !== 'string' && !Buffer.isBuffer(cbOutput)) {
     throw new Error('outputHelp callback must return a string or a Buffer');
   }
-  process.stdout.write(cbOutput);
+  process.stderr.write(cbOutput);
   this.emit(this._helpLongFlag);
 };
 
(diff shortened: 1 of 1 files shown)


shadowspawn:
There is a typo!

But changing over to `stderr` properly will require more work than this. When the help is explicitly asked for, the display should continue to go to stdout as now. It is only when the help is displayed due to an input error that it should go to stderr.

cengizhanbasak:
Added a parameter for outputHelp, allowing users to specify if they want to output to stderr or not. Output is to stderr as default, "--help" outputs to stdout though.

shadowspawn:
The failing test is due to one change you made which I think is not correct. I'll let you look into it. 

Note 1: you can run the tests locally before you push.
Note 2: there have been *big* changes to the tests on the develop branch, which has the code for the next version of Commander.

I do not want to change the public interface for `.outputHelp`, but that is a reasonable place to experiment as it is a bottleneck for the current calls. You have added a parameter to `.outputHelp`, but everywhere it is called you are passing `false` so I don't think you have changed the error case.

I think getting the help-as-error to stderr in a tidy way is a tricky problem and harder than it looks. If you find the associated `emit` call, you will see a further complication...

shadowspawn:
I added a question for original poster. I was thinking about the help-as-error case, but that may not have been what the original poster was commenting on.

shadowspawn:
Thanks @cengizhanbasak , I see both your original simple change and second try were consistent with what original poster was asking for. (I had a different case in mind.)

However, there is a work-around, and I don't think this change offers enough to justify the extra parameter to `.outputHelp` for now. 

#997 is one person's issue and has not attracted any likes in a couple of months of being open, and this change would not resolve the couple of previous issues that have mentioned `stderr`. The help related functions are currently a somewhat odd collection of routines that have limited use. I would like a clearer idea of what changes will make them more widely useful and not make one-off additions.

Issue #1038, opened 2019-09-08
POLL: API for mandatory options


shadowspawn:
The most upvoted and second oldest open issue is #230 to add support for mandatory options which must be specified on the command line. 

(This is in addition to the existing support for options with a required value when the option is specified.)

There have been a number of possible forms for the API proposed. Which one would you like to see implemented? You can vote 👍  for more than one, and can vote 👎 against any you dislike.

shadowspawn:
`program.requiredOption('-f,--foo <value>', 'foo must be supplied');`

shadowspawn:
`program.optionEx('-f,--foo <value>', 'foo must be supplied', { required: true });`

shadowspawn:
program.requiredOption('-f,--foo* <value>', 'foo must be supplied');

shadowspawn:
Thanks for the feedback. Implementing `.requiredOption()` in #1071.

shadowspawn:
Added `.requiredOption()` to v4.0.0 prerelease: #1067

PR #1071, opened 2019-10-05
Add .requiredOption() for mandatory options

shadowspawn:
# Pull Request

## Problem

Some programs have options which are mandatory. This usually means they must be specified on the command line. This is not directly supported by Commander, and is a popular request. (It is often confused with options which have a required value when the option is specified.)

Issues: #230 #1038

## Solution

The most popular form for the API (#1038) is to add a method called `.requiredOption()`. For simplicity it is has the same signature as `.option()` and supports all the same parameters.

(This new functionality is in addition to the existing support for options with a required value when the option is specified, like `'-f,--foo <requiredValue>'`.)

Commander will check that each mandatory option has a value after parsing, and display an error if the option value has not been supplied. Requiring a value rather than requiring the option on the command line allows use of environment variables to optionally supply default values.

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -245,7 +245,7 @@ $ custom --list x,y,z
 
 ### Required option
 
-You may specify a required (mandatory) option using `.requiredOption`, which must have a value after parsing. This is otherwise the same as `.option` in format, taking flags and description, and optional default value or custom processing.
+You may specify a required (mandatory) option using `.requiredOption`. The option must be specified on the command line, or by having a default value. The method is otherwise the same as `.option` in format, taking flags and description, and optional default value or custom processing.
 
 ```js
 const program = require('commander');
--- a/Readme.md
+++ b/Readme.md
@@ -15,6 +15,7 @@ The complete solution for [node.js](http://nodejs.org) command-line interfaces, 
     - [Default option value](#default-option-value)
     - [Other option types, negatable boolean and flag|value](#other-option-types-negatable-boolean-and-flagvalue)
     - [Custom option processing](#custom-option-processing)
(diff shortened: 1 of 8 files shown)


shadowspawn:
Pretty tidy now, but I'll leave as a draft for another day in case I think of anything more. I made some improvements today after thinking on it overnight. :-)

I think this will be safe to add to 4.0.0 when (if) approved since it is a separate new feature that should not affect any existing parsing.

shadowspawn:
ToDo: add call to `.requiredOption` to TypeScript test program.

shadowspawn:
I did a little research into the help conventions for required options. There are common conventions for the short description in the usage with optional options in square brackets, but I didn't see a common convention for the longer list of options. 

Plan to leave it up to the user, for now. Obviously the user can change the option description itself. We also support the user customising the "usage" description, which is a good place to show required options as part of the calling signature. e.g.

    Usage: foo --host HOST [--debug] [--port PORT]

An interesting lightweight idea I saw in `picocli` was optional support for prefixing lines with required options with a character (like '*'). This would be very easy to add later. e.g.

```
Usage: foo --host HOST [--debug] [--port PORT]

  --debug            include debugging information
* --host <name>      mandatory host name
  --port <number     port number
```

shadowspawn:
Ready for review. No rush. Feedback welcome.

shadowspawn:
Thanks @abetomo 

I am thinking of doing another prerelease version of 4.0.0 with this change included?

abetomo:
I think it can be included in 4.0.0.
If you want to emphasize the addition of features, you can release it in 4.1.0.

Issue #1073, opened 2019-10-10
2.20.1 Not Backward Compatible


terrancesnyder:
https://github.com/tj/commander.js/blob/master/index.js#L100

// new Set();

Is not compatible. This hit us with pm2 where pm2 includes commander, and pm2 runs on node 0.10.x however it includes updated commander which is now using new Set() which is not supported under 0.10.x. We can not upgrade. Please ensure semantic versioning and move Set() to a different version.

shadowspawn:
Thanks for report. I'll do some investigation.

shadowspawn:
`Set` goes back to Node 0.12.18 but not to 0.10.48 according to:
- <https://node.green>

Node v0.12.0 (Stable) was released 2015-02-06
- <https://nodejs.org/en/blog/release/v0.12.0/>

Commander v2.0.0 was released on Jul 19, 2013, so Commander 2.x started before node 0.12 and so reasonable that 2.x.y should not break in this way despite the age.

The fix I made was not essential. I suggest we reverse the change I made in 2.x, if you agree @abetomo ? (I'm happy to do it, but likely to be a couple of days before I have time.)

abetomo:
Fixed.
Could you try using v2.20.3?

shadowspawn:
Release a new version, and no further activity in a month. Closing this as resolved.

Thanks @abetomo for quick action on this one.

People will probably upgrade from Command 2.x to 4.x now, but if anyone hits same issue with 3 we could make the same change there. 

Feel free to open a new issue if it comes up again, with new information and renewed interest.

PR #1059, opened 2019-09-28
Improve 2.x executable subcommand tracking and update dependencies

shadowspawn:
This one is optional. I put together a last update for 2.x before we move onto 4.x.

If you are ok with this PR and a 2.x release @abetomo , I could publish this one.

PR #1056, opened 2019-09-26
Improve executable subcommand tracking

shadowspawn:
Improve executable subcommand tracking, and update dependencies. A last release for 3.x series?!

This can be released if you are happy with it @abetomo 

I'll update PR and release notes with credit details after release.

PR #1074, opened 2019-10-11
Support Node.js 0.10 (Revert #1059)

abetomo:
# Pull Request

## Problem

#1073

## Solution

Revert #1059


index.js
```diff
@@ -484,7 +484,7 @@ Command.prototype.parse = function(argv) {
     })[0];
   }
 
-  if (this._execs.has(name)) {
+  if (this._execs[name]) {
```
abetomo: Previously it was the following code.

```js
if (this._execs[name] && typeof this._execs[name] !== 'function') {
```

The function check was deleted because it was a code when it was an Array.
reference: #249

```js
> a = []; a['filter']
[Function: filter]
> o = {}; o['filter']
undefined
```

abetomo:
test failed: https://github.com/tj/commander.js/pull/1059#issuecomment-536227242

index.js
shadowspawn: The function check is to avoid treating built-in functions like `toString` as though they are subcommands, so I recommend a revert to the full previous code.

(The problem the Set was fixing was correctly handling a built-in property that is _not_ a function, which the old code gets wrong.)

shadowspawn: I was just thinking something similar! Yes. Simple and elegant.

abetomo:
I'll do `npm publish`.

shadowspawn:
Thanks Abetomo. (I think I used publish tag 2_x with the previous 2.x publish, but then deleted it after publish. No need to delete it though. The publish tag needs to look different than a version number because they share same namespace.)

abetomo:
I'm sorry!
I was impatient and unpublished 2.20.2.

If you unpublish, you cannot publish again with that version.
Will you release in 2.20.3?

shadowspawn:
Bumping it to 2.20.3 is fine.

shadowspawn:
This issue was responsibly reported by the Checkmarx Application Security Research Team. It was fixed in  v2.20.3 (#1074) and v3.0.2(#1056).

The Checkmarx vulnerability library lists this as: https://devhub.checkmarx.com/cve-details/Cx435a6fda-ca38/

Issue #1037, opened 2019-09-02
TS error when using imported Command type


marcopetreri:
As described in the example below I cannot use **Command** type in types definitions because of this error:
```
Exported variable 'getBuildCommand' has or is using name 'local.Command' from external module "***/node_modules/commander/typings/index" but cannot be named
```
Example
```typescript
import { Command } from 'commander';

export const getBuildCommand = (
   ...
) => (project: string, cmd: Command) => { ... };
```

shadowspawn:
I was not able to reproduce this with typescript 3.6.2 and this program fragment. Someone else might recognise the error, or you could add some more context.

fson:
I have made a minimal repoducible example for this issue: https://github.com/fson/commander-ts-repro.

When you compile the code using `npm run build`, it results in the following errror: 
> index.ts:3:1 - error TS4082: **Default export of the module has or is using private name 'local.Command'.**

fson:
I also found a workaround: in some cases, the error went away after changing the exported function from an arrow function like this:
```ts
export default (program: Command) => { 
```
to a function expression:
```ts
export default function(program: Command) {
```

shadowspawn:
Reproduced issue, and workaround, from @fson. (Thanks!)

Still trying to make sense of the local namespace pattern being used in the Commander definitions file.

shadowspawn:
I've had a go at reworking the TypeScript declaration file and fixed this error, but the changes are tricky. If any TypeScript experts see this then a review would be very welcome.

shadowspawn:
Fixed in Commander v4, which has been released.

shadowspawn:
I was reproducing this again and a historical note. This issue can be reproduced using TypeScript 3.8.3, but is fixed from 3.9.2.

PR #1081, opened 2019-10-27
Switch from class to class-like to include all types in commander namespace

shadowspawn:
# Pull Request

## Problem

TypeScript errors in some client use cases because not all types are exported. See #1037

## Solution

Rework to use class-like interfaces and constructors rather than explicitly using `class` so all the types can be included in the commander namespace. (This is made hard by the existing default export of the global Command object so can not use multiple exports or export explicit classes.)

This is beyond my TypeScript comfort zone! We might need a TypeScript expert to confirm this is a reasonable approach.

### DIFF
--- a/typings/index.d.ts
+++ b/typings/index.d.ts
@@ -5,18 +5,17 @@
 
 ///<reference types="node" />
 
-declare namespace local {
+declare namespace commander {
 
-  class CommanderError extends Error {
+  interface CommanderError extends Error {
     code: string;
     exitCode: number;
     message: string;
     nestedError?: string;
-
-    constructor(exitCode: number, code: string, message: string);
   }
+  type CommanderErrorConstructor = { new (exitCode: number, code: string, message: string): CommanderError };
 
-  class Option {
+  interface Option {
     flags: string;
     required: boolean; // A value must be supplied when the option is specified.
     optional: boolean; // A value is optional when the option is specified.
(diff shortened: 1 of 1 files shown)


shadowspawn:
@alan-agius4 (#646)
@jsamr (#713) 
Thanks for your past contributions to the Commander TypeScript declaration file.

In case you are interested, I have a PR in flight and would welcome any feedback from TypeScript contributors.

I have done a refactor to get the Command and Option types fully into the `commander` namespace to fix a reported bug. I did a lot of reading and  trial and error, but I seem to have it working nicely now with a manual class-like declaration (interface + constructor).

All my attempts to use a true class failed as declaring a class adds a value (the constructor) to the otherwise type-only namespace, which then conflicts with the export of the global command object.

shadowspawn:
Thanks @alan-agius4 !

jsamr:
@shadowspawn I think grouping all the declarations in the same namespace is a good idea, and I was a bit skeptical to the former local namespace, and I would guess it would fix #1037 . I just wonder however why you had to get rid of class declarations in favor of instance interface + constructor interface + constant declaration? It's OK, but a bit over-engineered.

shadowspawn:
With the single `export =`, adding a class to the namespace causes an error. Stripped down example:


```
declare namespace commander {
  class Foo {
  }
}
declare const commander: object
export = commander;
```

> error TS2300: Duplicate identifier 'commander'.

jsamr:
@shadowspawn This is expected :-) The namespace is **a value**! A POJO which fields are the names of declared values.

So you should get rid of `declare const commander: object` and you'll be all good.

jsamr:
@shadowspawn If you give me access to your fork, I can offer a PR within your fork to show you what I mean.

jsamr:
@shadowspawn see my edited answer above

shadowspawn:
Thanks @jsamr. I did consult the TypeScript documentation, especially the Declaration Files pages, and the Deep Dive was indeed one of the most relevant pages.

My understanding is the previous and PR solution allow a single export to include both meanings of `commander`, the "value" which is the global program object, and the (non-value) "namespace" which includes the supporting "types". The shape of the global program object is the Command class/interface with a few additions.

I am not sure how to declare the global/default program object without exporting a "value" for it. Did you have a pattern in mind?

For example, client code could look like:
```
import * as commander from 'commander';
commander
   .name('global-program');
const myProgram = new commander.Command();
myProgram
   .name('local-program');
```

jsamr:
@shadowspawn Now I get it. Typescript cannot merge a constant value with a namespace which has values within, such as classes. So your trick is perfectly legitimate :slightly_smiling_face: 

LGTM! :+1:

shadowspawn:
Thanks @alan-agius4 and @jsamr

PR #1091, opened 2019-11-09
Display help despite missing required option

### LINKED ISSUE #1089: Help option does not display help when using requiredOption
Hi,

I wanted to have one option required. Once I added one, the help message will no longer show unless you include the required option. I would have expected it to show the help.

I created some sample code, from one of your samples.

Test Code:

```
const commander = require("commander");

program
  .option("-d, --debug", "output extra debugging")
  .requiredOption("-s, --small", "small pizza size")
  .option("-p, --pizza-type <type>", "flavour of pizza");

program.parse(process.argv);

if (program.debug) console.log(program.opts());
console.log("pizza details:");
if (program.small) console.log("- small pizza size");
if (program.pizzaType) console.log(`- ${program.pizzaType}`);
```

Output:
```
$ node test.js -h
error: required option '-s, --small' not specified
```
If I remove the requiredOption I get the expected output:
```
$ node test.js -h
Usage: test [options]

Options:
  -d, --debug              output extra debugging
  -s, --small              small pizza size
  -p, --pizza-type <type>  flavour of pizza
  -h, --help               output usage information
```

shadowspawn:
# Pull Request

Fixes #1089 #1092

## Problem

When ask for help but there is a missing required option, getting error instead of help, so can't find out more.

e.g.
```
$ node test.js -h
error: required option '-s, --small' not specified
```

## Solution

Check for missing required options after checking for displaying help. This fix covers help for commands with action handlers, and help on the main program.

(It does not cover program with a required option and calling help on an executable subcommand, which seems an unlikely case. Deal with it if and when it comes up.)

### DIFF
--- a/index.js
+++ b/index.js
@@ -324,6 +324,7 @@ Command.prototype.action = function(fn) {
 
     // Output help if necessary
     outputHelpIfNecessary(self, parsed.unknown);
+    self._checkForMissingMandatoryOptions();
 
     // If there are still any unknown options, then we simply
     // die, unless someone asked for help, in which case we give it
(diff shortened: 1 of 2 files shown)


shadowspawn:
Thanks @abetomo. I'll prepare a PR for a patch release.

PR #1095, opened 2019-11-11
Fix brackets in CHANGELOG entry

shadowspawn:
Fix minor typo for PR link.


shadowspawn:
Drat, but minor typo! 🙂 

I am not sure why that test is failing, I don't think it is real. If keeps causing problems I could try switching to a synchronous call (or even remove as quite heavyweight to install ts-node for that one somewhat special check.)

Thanks for release.

PR #1096, opened 2019-11-14
Remove trailing blanks from wrapped help text

chdh:
# Pull Request

## Problem

Auto-wrapped lines in the generated help text have an extra space character at the end. This leads to unwanted empty lines in the help text display at the console, when a resulting line is exactly 80 characters long.

The trailing space characters can be seen in the changes of this PR for tests/helpwrap.test.js.

## Solution

A trimRight() is added in wrap().

### DIFF
--- a/index.js
+++ b/index.js
@@ -1429,7 +1429,7 @@ function wrap(str, width, indent) {
     if (line.slice(-1) === '\n') {
       line = line.slice(0, line.length - 1);
     }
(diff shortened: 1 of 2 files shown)


shadowspawn:
I'll do some testing, but looks sensible, and I like that it removes (invisible) trailing spaces from the unit tests.

shadowspawn:
4.1 has shipped, thanks @chdh

Issue #933, opened 2019-03-18
Triage: Namespace pollution


shadowspawn:
The current code uses the same namespace for user specified options, user specified arguments, and command.Command internals. Opened this issue to collect the related issues found during triage.

Issues:
- #19 Cannot name options/subcommands after commander public methods or JavaScript built-ins
- #105 '--domain' option may cause problem
- #183 Option values should not be added to the command instance.
- #284 "name" option is broken as of 2.4.0
- #370 Breaks when using "domain" as an option
- #377 Command name "inspect" causes weird behaviour
- #404 argument 'once' (and probably others) cause naming collision
- #483 It causes unexpected result if arguments are the same as option name
- #514 Move parsed options into a data object
- #516 Using .name and .description has unexpected results.
- #584 program.opts() can be overridden by an option
- #592 Error when option --alias is defined
- #629 Error when argument has same name as a flag
- #644 Options object is not strictly the user input
- #648 Init value of option '--name' is a [Function]
- #1020 Can not use action as option name

Pull Requests:
- #515 Data overload
- #673 Fix options key name conflict with commander object's' key

Fixing this will be a major breaking change. To manage the disruption perhaps we can make a new approach opt-in (non-breaking), then make the new behaviour default (breaking), before finally deprecating the old behaviour.

vanesyan:
Should we open an issue to collect all wanted breaking changes and finally decide on v3?

shadowspawn:
(Started v3 conversations in #666)

shadowspawn:
Draft code for options refactor in #951

shadowspawn:
My biggest stumbling block at the moment is what to pass into the action handler. Currently the command is passed in, but most people probably assume it is just the options. I did at first. 

Should it be the command, with properties for access to clean options? Or an option object. Or an option map. Or configurable... Still thinking...

One a somewhat related note, there is an existing `.opts()` method that grabbed the nicest short name! I have been wondering about optsM and opsO for map and object, somewhat in style of argV and argC, but probably better to be more readable. We should not be afraid of a few extra characters...

shadowspawn:
Current plan:
- legacy default behaviour is options stored as properties on program/command, action handler gets passed program/command.
- modern default behaviour is options separate from program/command, action handler gets passed options
- allow override to get command passed in modern for advanced uses
- offer options as object (not map) as familiar and fairly backwards compatible code for action handler

shadowspawn:
[...]

PR #1102, opened 2019-11-22
Opt-in behaviour to avoid name pollution

shadowspawn:
# Pull Request

## Problem

Commander has traditionally stored the option values on the command object. This gives a compact syntax especially when working with the program object, but means there are potential naming conflicts between user defined options and properties on Command. This catches people out and there are not good work-arounds.

Collected issues: #933

For example, code with a conflict on both program and a command::

```

program
  .option('--name <value>', 'specify name');
program
  .command('sub')
  .option('--action <value>', 'specify action')
  .action((options) => {
    console.log(options.action);
  });

program.parse(process.argv);

options = program.opts();
console.log(options.name);
```

```
% node pollution.js sub
[Function]
[Function]
% node pollution.js --name NAME sub --action ACTION
ACTION
NAME
```

## Solution

Add a way of configuring Command to behave differently. Implement "modern" configuration which stores the option values separately, and passes just the options to action handlers. The main code change is accessing program options through (existing) `.opts()` member.

- backwards compatible. Existing behaviour unchanged.
- often no changes required in action handlers. Most people probably think of the parameter as for just the options anyway!
- configurable for special cases or existing advanced usage requiring command object passed to action handler

With the addition of couple of lines to the example program above, no other code changes:

```
...
program
  .storeOptionsAsProperties(false)
  .passCommandToAction(false);
...
```

Now get:

```
% node pollution.js sub                            
undefined
undefined
% node pollution.js --name NAME sub --action ACTION
ACTION
NAME
```

## ToDo

Add info to README.

## Notes

I considered using a Map for the option values rather than a plain Object, but it would mean changing all existing client code accessing option values. Continuing to use an object means action handling code is largely unaffected.

I considered creating option values object with `Object.create(null)` to further reduce name collisions, but felt it was too unusual for the small benefits. 

I would have exposed a data property called `opts` if there was not already a function with that name! However, using the existing function does allow code that works with both the old way and the new way, which is nice.

We could change all the code examples to use `program.opts()` so they work with both old and modern behaviour.

In a future major version we could change the default behaviour. The routines are named for the future, where you need to opt back into the old behaviour.


shadowspawn:
ts-node test still unreliable.

Previously noted: https://github.com/tj/commander.js/pull/1095#issuecomment-552376208

shadowspawn:
I am rethinking "modern" name as not suitable for the longterm, and considering other names for this combo configuration.

shadowspawn:
My current idea is a named combo like `{ combo: 'safeOptions' }` instead of `{ modern: true }`. This is clearer that it is just a combination of  other settings and not something new, and allows other combinations in the future especially for backwards compatibility when we change default behaviour, like say `{ combo: 'v2Compatible' }`.

shadowspawn:
Still not happy with the interface using `configureCommand`. Will experiment with routines in more typical Commander style, like `. storeOptionsAsProperties(false)`.

shadowspawn:
Old behaviour:
```
```

New behaviour, safe options, which are passed to action handler:
```
program
   .storeOptionsAsProperties(false)
   .passCommandToAction(false);
```

(The routines are named for the future when the default behaviour has changed, and would pass true to get the legacy behaviour.)

shadowspawn:
Added section in README covering new routines and rationale, and example files for the three expected code patterns.

Ready for feedback/review.

shadowspawn:
Woo hoo! 🎉

index.js
```diff
@@ -531,6 +540,70 @@ Command.prototype.allowUnknownOption = function(arg) {
...
+ *
+ * @param {String} key
+ * @return {Object} value
+ * @api private
+ */
+
+Command.prototype._getOptionValue = function(key) {
+  if (this._storeOptionsAsProperties) {
+    return this[key];
+  } else {
```
abetomo: This is a personal preference, but this 'else' can be deleted.

shadowspawn:
I am thinking perhaps do next release in early January?

(I am starting to get busy leading into Xmas so my time will be limited.)

index.js
shadowspawn: Short version: I'll look into this a little, and probably delete the else.

Long version: I do not use an else when it is an early return like for an error, but I sometimes keep the if/else when they are both normal situations. However, I know some people prefer not to have the redundant else and am ok with that. I think I saw a lint option for this recently. I'll have a look and see if I am in the minority in this code, and we could enforce it. :-)

shadowspawn: There were only two cases, and they were both introduced by me in this pull request! I have changed them both and added a lint rule.

abetomo: Thank you!

shadowspawn: Thanks-you for the suggestion. 🙂

shadowspawn:
Shipped in Commander v4.1

Issue #806, opened 2018-05-21
Async typescript functions as actions


sr258:
When I pass an async typescript function into a command's action() method, the async function is called correctly, but is not run to the end, as the program does not wait when awaiting another function. It simply ends execution. I've seen a similar question [here](https://github.com/tj/commander.js/issues/505), but it was not answered, as far as I understand.

Consider this example:
```
#!usr/bin/env node

import * as commander from 'commander';
import axios from 'axios';

async function run() : Promise<void>  {  
    var response = await axios.get("https://dummyimage.com/600x400/000/fff", { "responseType": "arraybuffer" });
    if (response.status === 200) {
        console.log(`Downloaded file: ${response.data.byteLength} bytes.`)
    }
    else {
        console.error('Error downloading');        
    } 
}

async function main() {
  commander
  .command("do", "Test")
  .action(run)
  .parse(process.argv);
//  await run();
  console.log('finished');
}

try {
  main();
}
catch (error) {
  console.error(error);
}
```

If you run the example above, you can see that finished is called but the download is never finished. Calling await run() without the help of commander works, though.

So, is it possible to use commander with typescript's async functions?

sr258:
For future readers: I've moved to yargs, as it supports async actions.

bigtfromaz:
I investigated yargs as an option but frankly I think commander.js offer the best design.  However that doesn't mean that commander.js does not have an issue in this regard.

I am very experienced but quite new to JavaScript.  It leaves a lot to be desired when compared to most other languages.

With that said, I find myself in a position where I have to build an enterprise-quality command-line app that must run under node.js.  The app will also offer a REST API interface to some functions. Obviously neither it, nor the CLI can be allowed to block the message/UI thread.

commander.js seems to be a little schizophrenic in this regard.  Is it a parser or a command dispatcher? It's a really nice parser.  Dispatcher? No so much.  Command actions definitely do not block the thread however there is no reasonable way to return anything from the handling function, including a Promise<myType>.

I have evaluated a few options but none of them will allow me to sleep well at night after this thing goes live.  So I have a suggested change:

1. Offer a version of command.parse that doesn't actually dispatch an action but rather, returns the objects that would have been supplied when calling the action.  This would allow the consuming app to dispatch it's own method using either Promise or TypeScript async/await.

2. Offer an awaitable and sync versions of command.parse that returns whatever value is returned by the action function.

I would suggest option 1 as it provides a clean separation between parsing and functional execution.

[...]

PR #1118, opened 2019-12-11
Add parseAsync

shadowspawn:
# Pull Request

## Problem

People are using async routines from TypeScript and Javascript, and want to declare async action handlers. For the caller to properly wind up the async calls, Commander needs to expose something to allow handling the asynchronous calls.

See #806

## Solution

Add a `parseAsync()` call. It calls the normal parse, but then returns a Promise which includes the possibly async action handler. The client can then deal with the Promise in the normal way. 

It is ok to have a mixture of async and non-async action handlers. Calling parseAsync works with either. Basically, if you have any async action handlers then you should call parseAsync instead of parse.

With this TypeScript code:
```ts
import * as commander from 'commander';

function sleep(ms: number) {
  return new Promise(resolve => setTimeout(resolve, ms));
}

async function run() {
  console.log('run+');
  await sleep(2000);
  console.log('run-');
}

program
  .command("wait")
  .action(run);


program.parseAsync(['node', 'foo', 'wait']).then(() => {
  console.log('parseAsync success');
}).catch(err => {
  console.log('parseAsync failed');
});

console.log('Reached end of program');
```

See this:
```bash
$ npx tsc async.ts && node async.js
run+
Reached end of program
run-
parseAsync success
```

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -11,7 +11,7 @@ Read this in other languages: English | [简体中文](./Readme_zh-CN.md)
 
 - [Commander.js](#commanderjs)
   - [Installation](#installation)
-  - [Declaring _program_ variable](#declaring-program-variable)
+  - [Declaring program variable](#declaring-program-variable)
   - [Options](#options)
     - [Common option types, boolean and value](#common-option-types-boolean-and-value)
     - [Default option value](#default-option-value)
@@ -33,7 +33,7 @@ Read this in other languages: English | [简体中文](./Readme_zh-CN.md)
   - [Bits and pieces](#bits-and-pieces)
     - [Avoiding option name clashes](#avoiding-option-name-clashes)
     - [TypeScript](#typescript)
-    - [Node options such as `--harmony`](#node-options-such-as---harmony)
+    - [Node options such as --harmony](#node-options-such-as---harmony)
     - [Node debugging](#node-debugging)
     - [Override exit handling](#override-exit-handling)
   - [Examples](#examples)
@@ -375,6 +375,19 @@ program
   })
 
 program.parse(process.argv)
+```
+
+You may supply an `async` action handler, in which case you call `.parseAsync` rather than `.parse`.
+
+```js
+async function run() { /* code goes here */ }
+
+async function main() {
+  program
+    .command('run')
+    .action(run);
+  await program.parseAsync(process.argv);
+}
 ```
 
 A command's options on the command line are validated when the command is used. Any unknown options will be reported as an error. However, if an action-based command does not define an action, then the options are not validated.
--- a/index.js
+++ b/index.js
@@ -129,6 +129,7 @@ function Command(name) {
   this._optionValues = {};
   this._storeOptionsAsProperties = true; // backwards compatible by default
   this._passCommandToAction = true; // backwards compatible by default
+  this._actionResults = [];
 
   this._helpFlags = '-h, --help';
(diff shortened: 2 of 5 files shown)


shadowspawn:
To do:
- add tests
- investigate compatibility with node 6
- README

jdmarshall:
I think I agree that the problem traces to the end of the event listener in action(), but if I back up for a second I have two other concerns.  

The simplest one is, wouldn't it be simpler to register a new event that is emitted after fn.apply() resolves/returns.
The harder question is why action.apply() in an event handler in the first place? That makes propagating errors (and **_actions will have errors_**, especially when you are writing them!) much more problematic.

shadowspawn:
I think we have two different problems @jdmarshall. 

I was focused on caller being able to manage async action handlers, and (say) `await` actions being completing before proceeding with other code. My intention was to make it possible and the rest was up to the caller. 

Your problem (with setInterval) is presumably that a pending promise alone does not keep node running, which is a bit nasty:
- https://github.com/nodejs/node/issues/22088

I didn't encounter the early-exit problem because I was using `setTimeout` as my async example. I am not sure if it is up to Commander to keep node running forever, but it might be reasonable to keep it running for some default period (like say 10s, which is the default timeout on docker stop).

shadowspawn:
@jdmarshall Do you have a small example of a problem async action that would not run correctly with the addition of an await? (i.e. what is it doing asynchronously that node does not recognise will resolve later?)

jdmarshall:
Not at present.  What drew me here was that I had an async action that seemed to be working fine, performing a lot of async calls (throttled uploads), and then I added a second async task to the end of that action and it never ran.  

I haven't been back to it yet to come up with a repro case.

shadowspawn:
Here is a javascript example for node 8 or up with `await`. Lots of logging!

```

function sleep(ms) {
  return new Promise(resolve => setTimeout(resolve, ms));
}

async function doSomethingAsync() {
  console.log('doSomethingAsync+');
  await sleep(1000);
  console.log('doSomethingAsync-');
}

async function main() {
  console.log('main+');
  program
    .command('example')
    .action(doSomethingAsync);
  console.log('Calling parseAsync');
  const promise = program.parseAsync(['node', 'script', 'example']);
  console.log('Working before waiting');
  await promise;
  console.log('main-');
}

console.log('start program');
main().then(() => {
  console.log('main succeeded');
}).catch(() => {
  console.log('main failed');
})
console.log('finish program');
```


```
$ node example.js
start program
main+
Calling parseAsync
doSomethingAsync+
Working before waiting
finish program
doSomethingAsync-
main-
main succeeded
```

shadowspawn:
Thanks @blendsdk

shadowspawn:
Added README. Have a test and typings. 
Moving out of draft.

This is a non-breaking addition so can go in v4, but I am happy to leave it for v5 if preferred.

shadowspawn:
This is the only PR I am interested in including before v4.1, but happy to leave it for v5 whether for safety or to allow more time to review @abetomo . No worries.

shadowspawn:
Thanks @abetomo. I'll start preparing 4.1 soon.

shadowspawn:
CHANGELOG updated for 4.1 on develop branch. Ready to release to master, version number and tag still to be updated. 

```
npm version minor
```

I'll add an issue for last change to Chinese README but we can add that after release.

PR #1119, opened 2019-12-12
fix option type definition to match usage.

UrielCh:
# Pull Request

```bash
  npm run test
Test Suites: 32 passed, 32 total
Tests:       223 passed, 223 total
Snapshots:   0 total
Time:        3.109s
Ran all test suites.

npm run lint
no changes.
```
## Problem

```
// I want to use this kind of option
    .option('--hostname <hostname>', 'hostname used for binding http server', (value: string, previous: string[]) => { previous.push(value); }, [] as string[])
// currently the typing using any, any, any
// argument are named arg1, arg2
// now it's value, previous
```

## Solution

- enforce typings of option.
- add some extra typing test.

## ChangeLog

- improve option() typing enforcing function usage.

### DIFF
--- a/tests/options.custom-processing.test.js
+++ b/tests/options.custom-processing.test.js
@@ -13,6 +13,10 @@ function collect(value, previous) {
   return previous.concat([value]);
 }
 
+function collectVoid(value, previous) {
+  previous.concat([value]);
+}
+
 function commaSeparatedList(value, dummyPrevious) {
   return value.split(',');
 }
@@ -130,6 +134,22 @@ test('when collect -c a -c b -c c then value is [a, b, c]', () => {
     .option('-c, --collect <value>', 'repeatable value', collect, []);
   program.parse(['node', 'test', '-c', 'a', '-c', 'b', '-c', 'c']);
   expect(program.collect).toEqual(['a', 'b', 'c']);
+});
+// freeze void returning collector behaviour
+test('when collect returning Void -c a -c b -c c then value is void', () => {
+  const program = new commander.Command();
+  program
+    .option('-c, --collect <value>', 'repeatable value', collectVoid, []);
+  program.parse(['node', 'test', '-c', 'a', '-c', 'b', '-c', 'c']);
+  expect(program.collect).toBeUndefined();
+});
+// freeze void returning collector behaviour
+test('when collect returning Void -c a -c b -c c -c d then value is []', () => {
+  const program = new commander.Command();
+  program
+    .option('-c, --collect <value>', 'repeatable value', collectVoid, []);
+  program.parse(['node', 'test', '-c', 'a', '-c', 'b', '-c', 'c', '-c', 'd']);
+  expect(program.collect).toEqual([]);
 });
 
 test('when commaSeparatedList x,y,z then value is [x, y, z]', () => {
--- a/tests/options.custom-processing.test.js
+++ b/tests/options.custom-processing.test.js
@@ -13,10 +13,6 @@ function collect(value, previous) {
   return previous.concat([value]);
 }
 
-function collectVoid(value, previous) {
-  previous.concat([value]);
-}
-
 function commaSeparatedList(value, dummyPrevious) {
   return value.split(',');
 }
@@ -134,22 +130,6 @@ test('when collect -c a -c b -c c then value is [a, b, c]', () => {
     .option('-c, --collect <value>', 'repeatable value', collect, []);
   program.parse(['node', 'test', '-c', 'a', '-c', 'b', '-c', 'c']);
   expect(program.collect).toEqual(['a', 'b', 'c']);
-});
-// freeze void returning collector behaviour
-test('when collect returning Void -c a -c b -c c then value is void', () => {
-  const program = new commander.Command();
-  program
-    .option('-c, --collect <value>', 'repeatable value', collectVoid, []);
-  program.parse(['node', 'test', '-c', 'a', '-c', 'b', '-c', 'c']);
-  expect(program.collect).toBeUndefined();
-});
-// freeze void returning collector behaviour
-test('when collect returning Void -c a -c b -c c -c d then value is []', () => {
-  const program = new commander.Command();
-  program
-    .option('-c, --collect <value>', 'repeatable value', collectVoid, []);
-  program.parse(['node', 'test', '-c', 'a', '-c', 'b', '-c', 'c', '-c', 'd']);
-  expect(program.collect).toEqual([]);
 });
 
 test('when commaSeparatedList x,y,z then value is [x, y, z]', () => {
--- a/typings/commander-tests.ts
+++ b/typings/commander-tests.ts
@@ -10,6 +10,9 @@ const errorInstance = new program.CommanderError(1, 'code', 'message');
 
 const name = program.name();
 
+program.storeOptionsAsProperties(true);
+program.passCommandToAction(true);
+
 program
     .name('set name')
     .version('0.0.1')
--- a/typings/commander-tests.ts
+++ b/typings/commander-tests.ts
@@ -37,6 +37,12 @@ function collect(val: string, memo: string[]) {
     memo.push(val);
     return memo;
 }
+/*
+ * Collector without return all so work
+ */
+function collect2(val: string, memo: string[]) {
+    memo.push(val);
+}
 
 function increaseVerbosity(v: any, total: number) {
     return total + 1;
@@ -51,6 +57,7 @@ program
     .option('-l, --list <items>', 'A list', list)
     .option('-o, --optional [value]', 'An optional value')
     .option('-c, --collect [value]', 'A repeatable value', collect, [])
+    .option('--collect2 [value]', 'A repeatable value void return', collect2, [])
     .option('-v, --verbose', 'A value that can be increased', increaseVerbosity, 0)
     .parse(process.argv);
 
--- a/typings/commander-tests.ts
+++ b/typings/commander-tests.ts
@@ -40,9 +40,9 @@ function collect(val: string, memo: string[]) {
 /*
  * Collector without return all so work
  */
-function collect2(val: string, memo: string[]) {
-    memo.push(val);
-}
+//function collect2(val: string, memo: string[]) {
+//    memo.push(val);
+//}
 
 function increaseVerbosity(v: any, total: number) {
     return total + 1;
@@ -57,7 +57,7 @@ program
     .option('-l, --list <items>', 'A list', list)
     .option('-o, --optional [value]', 'An optional value')
     .option('-c, --collect [value]', 'A repeatable value', collect, [])
-    .option('--collect2 [value]', 'A repeatable value void return', collect2, [])
+    // .option('--collect2 [value]', 'A repeatable value void return', collect2, [])
     .option('-v, --verbose', 'A value that can be increased', increaseVerbosity, 0)
     .parse(process.argv);
 
--- a/typings/commander-tests.ts
+++ b/typings/commander-tests.ts
@@ -51,7 +51,6 @@ program
     .option('-l, --list <items>', 'A list', list)
     .option('-o, --optional [value]', 'An optional value')
     .option('-c, --collect [value]', 'A repeatable value', collect, [])
-    // .option('--collect2 [value]', 'A repeatable value void return', collect2, [])
     .option('-v, --verbose', 'A value that can be increased', increaseVerbosity, 0)
     .parse(process.argv);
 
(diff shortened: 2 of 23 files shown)


shadowspawn:
I agree, the current typing does not offer much help for intended usage!

typings/index.d.ts
```diff
@@ -165,12 +165,14 @@ declare namespace commander {
...
-     * @param {((arg1: any, arg2: any) => void) | RegExp} [fn] function or default
+     * @param {((value: string, previous: any) => void) | RegExp} [fn] function or default
      * @param {*} [defaultValue]
      * @returns {Command} for chaining
      */
-    option(flags: string, description?: string, fn?: ((arg1: any, arg2: any) => void) | RegExp, defaultValue?: any): Command;
     option(flags: string, description?: string, defaultValue?: any): Command;
+    option(flags: string, description: string, regexp: RegExp, defaultValue?: any): Command;
+    option<T>(flags: string, description: string, fn: (value: string, previous: T) => (T | void), defaultValue?: T): Command;
+    option<T>(flags: string, description: string, fn: (value: string, previous: T) => void, defaultValue: T): Command;
```
shadowspawn: Does the last overload ever get considered? Won't any overload that could match
`option<T>(flags: string, description: string, fn: (value: string, previous: T) => void, defaultValue: T): Command;`

already be matched by:
`option<T>(flags: string, description: string, fn: (value: string, previous: T) => (T | void), defaultValue?: T): Command;`

typings/index.d.ts
```diff
@@ -165,12 +165,14 @@ declare namespace commander {
      *
      * @param {string} flags
      * @param {string} [description]
-     * @param {((arg1: any, arg2: any) => void) | RegExp} [fn] function or default
+     * @param {((value: string, previous: any) => void) | RegExp} [fn] function or default
      * @param {*} [defaultValue]
      * @returns {Command} for chaining
      */
-    option(flags: string, description?: string, fn?: ((arg1: any, arg2: any) => void) | RegExp, defaultValue?: any): Command;
     option(flags: string, description?: string, defaultValue?: any): Command;
```
shadowspawn: I think this overload should come last, otherwise it matches any call that does not have a default value because of the any on the third parameter. Reference: https://www.typescriptlang.org/docs/handbook/declaration-files/do-s-and-don-ts.html#function-overloads

`option(flags: string, description?: string, defaultValue?: any): Command;`

typings/commander-tests.ts
```diff
@@ -37,6 +37,12 @@ function collect(val: string, memo: string[]) {
     memo.push(val);
     return memo;
 }
+/*
```
shadowspawn: I was wondering about the void and had to look at the code to realise why it is there. The coercion routine can return void to indicate the coercion failed and the default value should be used. e.g.

```
function shortString(value: string) {
   if (value.length < 4)
      return value;
}

program
   .option('-s <str>', 'string less than four characters long', shortString, '...');
```

shadowspawn:
I am still thinking about whether it is ok to try and enforce that the previous, the return type, and the default value are all the same type. I think that will often be the case.

typings/index.d.ts
```diff
@@ -165,12 +165,14 @@ declare namespace commander {
      *
      * @param {string} flags
      * @param {string} [description]
-     * @param {((arg1: any, arg2: any) => void) | RegExp} [fn] function or default
+     * @param {((value: string, previous: any) => void) | RegExp} [fn] function or default
```
shadowspawn: I have actually deleted the JSDoc param lines on the `develop` branch.

typings/commander-tests.ts
UrielCh: allowing void for me allow using shorter code:
```javascript
function collect2(val: string, memo: string[]) { memo.push(val);}
// instead of
function collect2(val: string, memo: string[]) { memo.push(val); return memo;}
// of cours this last version should be sorted, but not working, push returning a length
const collect2 = (val: string, memo: string[]) => memo.push(val);
```
so prefer keep it as it, since, no alternative have real advantage.
the current version can store collection in readonly collection.
or also make the following collector fonctionnal:
```javascript
const collect2 = (val: string, memo: string[]) => [...memo, val ];
```

typings/index.d.ts
UrielCh: If I keep the current way your collector working:
- If the collector function return void, the defaultValue MUST be provide to store parameters, because, if not this option is useless an will be keep as undefined/void all the time.
- if default value is provided, so no problem the default value can be use as a storage, and the collector function can return void.

UrielCh:
`requiredOption` is now also updated.
Function Overloads order updated.

typings/index.d.ts
UrielCh: final version:
```javascript
   option(flags: string, description: string, regexp: RegExp, defaultValue?: any): Command;
    option<T>(flags: string, description: string, fn: (value: string, previous: T) => void, defaultValue: T): Command;
    option<T>(flags: string, description: string, fn: (value: string, previous: T) => T, defaultValue?: T): Command;
    option(flags: string, description?: string, defaultValue?: any): Command;
```

typings/commander-tests.ts
shadowspawn: I read the code and not sure of existing behaviour, so I tried it.

I do not think this example is working as you might expect. And to be fair, my own example did not work as I assumed either! The Commander code tests the function return value against null (for some reason) but a void function returns undefined. 

After that, I am not sure under what circumstances a void function makes sense for the coercion routine, if any.

```

function collect2(val, memo) {
  memo.push(val);
}

program
  .option('-c,collect <value>', 'A repeatable value void return', collect2, []);

program.parse(process.argv);

console.log(program.collect);
```

```
$ node void.js                    
[]
$ node void.js -c a          
undefined
$ node void.js -c a -c b
[ 'a', 'b' ]
$ node void.js -c a -c b -c c
undefined
$ node void.js -c a -c b -c c -c d
[ 'a', 'b', 'c', 'd' ]
```

typings/index.d.ts
shadowspawn: A couple of thoughts:
1) The regular expression support is no longer documented in README and soft deprecated. 
2) I am not sure if void makes sense, after experimenting tonight.

Would just two overloads work for all expected cases? (NB: this suggestion is untested as yet.)

```
option<T>(flags: string, description: string, fn: (value: string, previous: T) => T, defaultValue?: T): Command;
option(flags: string, description?: string, defaultValue?: any): Command;
```

(I think legacy code using regular expression will not cause a TypeScript error since will match signature of second overload. Untested,)

shadowspawn: By experimenting I confirmed that the `defaultValue?: any` allows malformed functions if there is no default value to force the match with the first overload. Trying a tighter overload after some experimentation with TypeScript, still experimenting...

```
option<T>(flags: string, description: string, fn: (value: string, previous: T) => T, defaultValue?: T): Command;
option(flags: string, description?: string, defaultValue?: string | boolean | RegExp): Command;

```

typings/commander-tests.ts
UrielCh: So, I should had I read the implementation.

Is the current implementation is logic ?
I can change It, to make it match my last typing.

I'm not sure I understood the way you expect it to works.

shadowspawn: I am interested in improving the TypeScript to better match the implementation and the examples in the README.

I do not want to change the implementation. In particular, I don't want to make the example from your original post work (sorry):
```
    .option('--hostname <hostname>', 'hostname used for binding http server', (value: string, previous: string[]) => { previous.push(value); }, [] as string[])
```

Are you still interested in improving the types without changing the code? No problem if are not interested in just the types, I will open an issue or PR with some of what we learnt and your suggestions.

I have learnt a lot looking into this PR!

UrielCh:
so If wee Keep the actual implementation, I can add this to the test :
```javascript
function collectVoid(value, previous) {
  previous.concat([value]);
}

...

test('when collect returning Void -c a -c b -c c then value is [a, b, c]', () => {
  program
    .option('-c, --collect <value>', 'repeatable value', collectVoid, []);
  program.parse(['node', 'test', '-c', 'a', '-c', 'b', '-c', 'c']);
  expect(program.collect).toBeUndefined();
});

test('when collect returning Void -c a -c b -c c -c d then value is [a, b, c, d]', () => {
  program
    .option('-c, --collect <value>', 'repeatable value', collectVoid, []);
  program.parse(['node', 'test', '-c', 'a', '-c', 'b', '-c', 'c', '-c', 'd']);
  expect(program.collect).toEqual([]);
});

```

UrielCh:
new version enforcing current implementation.
remove void returning collector from typing.
force string type as default value is no conversion function is given

tests/options.custom-processing.test.js
```diff
@@ -13,6 +13,10 @@ function collect(value, previous) {
   return previous.concat([value]);
 }
 
+function collectVoid(value, previous) {
```
shadowspawn: I think this is not a valid custom processing function since it does not return a value, so I don't want to check that it fails in the "expected" way. Im articular, if we change the implementation in the future in a way that changes this behaviour, I won't consider it a breaking  change.

typings/index.d.ts
```diff
@@ -162,24 +162,22 @@ declare namespace commander {
...
-     *
-     * @param {string} flags
-     * @param {string} [description]
-     * @param {((arg1: any, arg2: any) => void) | RegExp} [fn] function or default
-     * @param {*} [defaultValue]
-     * @returns {Command} for chaining
      */
-    option(flags: string, description?: string, fn?: ((arg1: any, arg2: any) => void) | RegExp, defaultValue?: any): Command;
-    option(flags: string, description?: string, defaultValue?: any): Command;
+    option(flags: string, description: string, regexp: RegExp, defaultValue?: string): Command;
```
shadowspawn: 1) A boolean flag can take a default value, which can be a string _or a boolean_.
2) Where the overload order does not matter, I prefer the more common case first.

So I suggest: (add boolean and reorder):
```js
     option(flags: string, description?: string, defaultValue?: string | boolean): Command;
     option<T>(flags: string, description: string, fn: (value: string, previous: T) => T, defaultValue?: T): Command;
     option(flags: string, description: string, regexp: RegExp, defaultValue?: string): Command;
```

UrielCh: Okey, nothing to add ?

ready to merge ?

tests/options.custom-processing.test.js
UrielCh: so I remove this test ?

UrielCh: Dropted.

typings/commander-tests.ts
```diff
@@ -37,6 +37,12 @@ function collect(val: string, memo: string[]) {
...
+ * Collector without return all so work
```
shadowspawn: This can be deleted now

typings/index.d.ts
```diff
@@ -162,24 +162,20 @@ declare namespace commander {
...
-     *
-     * @param {string} flags
-     * @param {string} [description]
-     * @param {((arg1: any, arg2: any) => void) | RegExp} [fn] function or default
-     * @param {*} [defaultValue]
-     * @returns {Command} for chaining
      */
-    option(flags: string, description?: string, fn?: ((arg1: any, arg2: any) => void) | RegExp, defaultValue?: any): Command;
-    option(flags: string, description?: string, defaultValue?: any): Command;
+    option(flags: string, description?: string, defaultValue?: string): Command;
```
shadowspawn: The boolean needs to be added to `option(flags: string, description?: string, defaultValue?: string)` not to the regexp overload.

See calling example in tests: https://github.com/tj/commander.js/blob/b25cc3abf209a285576b4d117d6dbff54a2a6100/tests/options.bool.combo.test.js#L49

shadowspawn: As in: https://github.com/tj/commander.js/pull/1119#discussion_r361771928

shadowspawn:
Thanks for your enthusiasm in improving the types! It has taken some digging to cleanly describe the intended usage patterns.

typings/index.d.ts
UrielCh: oups, sorry.

shadowspawn:
This does now describe the intended uses much better and in a more helpful way. It is more opinionated than typical for Commander, and people may report reasonable uses we have not thought of! In that case, we can change or weaken the typing as we learn more.

shadowspawn:
(Don't worry about the merge conflict, I will resolve that by hand when has passed second review. In comments rather than in code.)

shadowspawn:
Would you like to review this one too @abetomo , or shall I go ahead and merge?

shadowspawn:
This PR is included in the published pre-release of Commander v5.0.0 (#1163)

shadowspawn:

PR #1120, opened 2019-12-15
Improvements to JSDoc and constructors

shadowspawn:
# Pull Request

## Problem

We do not have any lint checks on the JSDoc, which is a bit arcane. We do not generate JSDoc based documentation any more, but it may be used by smart editors for autocomplete et al.

## Solution

TypeScript supports processing javascript files. I ran the following command and fixed a number of issues it picked up. The two most common were properties missing from constructor (and so unrecognised as valid), and issues with JSDoc declarations being inaccurate compared with the code.

`npx tsc --allowJS --checkJS index.js --noEmit`

### DIFF
--- a/index.js
+++ b/index.js
@@ -47,10 +47,11 @@ function Option(flags, description) {
   this.optional = flags.indexOf('[') >= 0; // A value is optional when the option is specified.
   this.mandatory = false; // The option must have a value after parsing, which usually means it must be specified on command line.
   this.negate = flags.indexOf('-no-') !== -1;
-  flags = flags.split(/[ ,|]+/);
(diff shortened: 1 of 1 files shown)


shadowspawn:
Can also turn on TypeScript checking in Visual Studio Code by adding line:
```js
// @ts-check
```

but VSC does not currently support Command constructor function extending EventEmitter so multiple warnings about `.on()` and `.listener(), and a couple of other problems. Down from 53 issues to 16 with this PR, but not down low enough to add the line permanently.

shadowspawn:
I'll redo this without the code changes. Safer to leave the code alone until more compelling reason to change.

Issue #1123, opened 2019-12-17
tests/command.executableSubcommand.lookup.test.js consistently timing out


BridgeAR:
We check `commander` for failures on [citgm](https://github.com/nodejs/citgm) to verify that new Node.js releases do not break anything. Recent changes to commander or to the test suite made one test consistently fail on our infrastructure with a timeout.

It would be great to either increase the maximum time allowed for that test to at least 10 seconds or (better) to split the test into multiple ones to reduce the individual run time, if possible.

Refs: https://ci.nodejs.org/view/Node.js-citgm/job/citgm-smoker/2134/nodes=ubuntu1604-64/testReport/junit/(root)/citgm/commander_v4_0_1/
(Available for ~7 days)

shadowspawn:
Thanks for including Commander! We saw a timeout issue and increased the timeout on the develop branch, but has not been merged onto master yet: #1108

I'll merge it across soon instead of waiting for the next release.

Not sure why the test is taking so long and a fallback is just removing it!
- https://github.com/tj/commander.js/pull/1095#issuecomment-552376208

shadowspawn:
Timeout increased. Let us know if still causing )false positive) problems.

BridgeAR:
Thank you very much for the fast response!

BridgeAR:
I am going to close this as resolved.

shadowspawn:
Oh, testing against latest release rather than master branch? Foiled.

FYI: next release is likely to be early January: https://github.com/tj/commander.js/pull/1102#issuecomment-563882950

shadowspawn:
Reworked test in #1125, not released yet.

shadowspawn:
@BridgeAR 
Removed direct use of `ts-node` in the problem tests and have not seen any timeouts since the change. 

Commander v4.1 released.

BridgeAR:
Awesome, thank you very much!

PR #1124, opened 2019-12-17
Increase timeout for test using ts-node

shadowspawn:
# Pull Request

## Problem

Test timing out.

Affecting node citgm, see #1123

## Solution

As quick fix have increased timeout, as per change on develop branch: #1108


shadowspawn:
Thanks @abetomo

PR #1132, opened 2019-12-30
Improve JSDoc to match code

shadowspawn:
# Pull Request

Using Typescript to check the Javascript identifies a lot of minor issues. This PR addresses most errors in the JSDoc.

(Previous abandoned #1120 also included code changes.)

## Problem

JSDoc issues are displayed when check `index.js` with TypeScript:

```
npx tsc --allowJS --checkJS index.js --noEmit
...
Found 53 errors.
```

or by adding comment to `index.js` in Visual Studio Code:

```
// @ts-check
```

## Solution

Make minor changes to the JSDoc to fix parameter names and types to match code.

```
npx tsc --allowJS --checkJS index.js --noEmit
...
Found 35 errors.
```

### DIFF
--- a/index.js
+++ b/index.js
@@ -115,7 +115,7 @@ exports.CommanderError = CommanderError;
 /**
  * Initialize a new `Command`.
  *
- * @param {String} name
+ * @param {String} [name]
  * @api public
  */
 
@@ -694,7 +694,7 @@ Command.prototype.parse = function(argv) {
  * @param {Array} argv
  * @param {Array} args
  * @param {Array} unknown
- * @param {String} specifySubcommand
+ * @param {String} executableFile
  * @api private
  */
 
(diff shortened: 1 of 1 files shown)


index.js
```diff
@@ -927,7 +927,7 @@ Command.prototype._checkForMissingMandatoryOptions = function() {
  * void of these options.
  *
  * @param {Array} argv
- * @return {Array}
+ * @return {{args: Array, unknown: Array}}
```
abetomo: https://github.com/tj/commander.js/pull/1132/files#diff-168726dbe96b3ce427e7fedce31bb0bcR966

`return this.optionMissingArgument(option);`

Need to mention this 'return'?

shadowspawn: `optionMissingArgument` calls `process.exit` or throws so the `return` never actually returns anything (the remaining warning is about returning void).

 I think the correct fix is to delete the `return` in this line. I didn't do that this time because I was only changing comments and not code, but I could do that again?

shadowspawn: FYI: this is fixed in #1138

PR #1133, opened 2019-12-30
Refactor Option from prototype to class

shadowspawn:
# Pull Request

We have long been wanting to adopt more modern conventions including ES2015 additions, and can start doing so as we have dropped support for older versions of node.

I am starting small with Option to discuss any questions this raises.

## Problem

Using prototype for implementing Option class is unfamiliar to new programmers, and noisy.

## Solution

Refactor code using `class`.

### DIFF
--- a/index.js
+++ b/index.js
@@ -27,66 +27,68 @@ exports = module.exports = new Command();
 
 exports.Command = Command;
 
+class Option {
+  /**
+   * Initialize a new `Option` with the given `flags` and `description`.
+   *
+   * @param {String} flags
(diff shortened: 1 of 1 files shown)


shadowspawn:
The code compare is very noisy with whitespace included, but looks tidy with whitespace ignored.

I had to search to find the whitespace controls: you can ignore whitespace from the cog in the "Files changed" view, or add `?w=1` to the URL.

https://github.com/tj/commander.js/pull/1133/files?w=1

PR #1134, opened 2019-12-31
add failing test for argument labels with spaces

vith:
# Pull Request

## Problem

The argument label is cut off in the help information.

```
 FAIL  tests/command.help.test.js
  ● when argument name includes space it should be printed correctly in helpInformation

    expect(received).toMatch(expected)

    Expected pattern: /^Usage: +\[options\] <some argument name>$/m
    Received string:  "Usage:  [options] <som>·
    Options:
      -h, --help  output usage information
    "

      101 |     .arguments('<some argument name>');
      102 |   const helpInformation = program.helpInformation();
    > 103 |   expect(helpInformation).toMatch(/^Usage: +\[options\] <some argument name>$/m);
          |                           ^
      104 | });
      105 | 

      at Object.toMatch (tests/command.help.test.js:103:27)
```

### DIFF
--- a/tests/command.help.test.js
+++ b/tests/command.help.test.js
@@ -94,3 +94,11 @@ test('when command sets noHelp then not displayed in helpInformation', () => {
   const helpInformation = program.helpInformation();
   expect(helpInformation).not.toMatch('secret');
 });
+
+test('when argument name includes space it should be printed correctly in helpInformation', () => {
+  const program = new commander.Command();
+  program
+    .arguments('<some argument name>');
+  const helpInformation = program.helpInformation();
+  expect(helpInformation).toMatch(/^Usage: +\[options\] <some argument name>$/m);
+});


shadowspawn:
A failing test. 🤔  
An interesting way of showing the problem, and can be kept for regression testing. Thanks. 👍 

I'll have a look into it, but if it isn't a quick fix then will likely ask you to open an issue. I have been adding reproduce tests for open issues to `openIssues.test.skip` so we have somewhere to add the test if the issue stays open.

shadowspawn:
There is an assumption in the code that the arguments can be separated using space. Both `.arguments` and `.command` spit up the parameter on spaces. This dates back to the original commit.

 e.g. https://github.com/tj/commander.js/blob/b25cc3abf209a285576b4d117d6dbff54a2a6100/index.js#L171

```
Command.prototype.arguments = function(desc) {
  return this.parseExpectedArgs(desc.split(/ +/));
};
```

shadowspawn:
It isn't a quick fix to change the way `.arguments` (and `.command`) parse the string. After looking, I think this is an enhancement request rather than a bug.

Please open an issue if you would like to keep this issue visible for consideration. Thanks.

Issue #1039, opened 2019-09-08
*Bug* When specify none-defined options followed by values don't exit program with error


snowman:
```js
// File: index.js
const program = require('commander');

program
  .version('0.0.1')
  .parse(process.argv);

const args = program.args;

console.log(args);
```

```bash
$ node index.js --bug
error: unknown option '--bug'

$ node index.js --bug 0
error: unknown option '--bug'

$ node index.js --bug 0 1 2 3
[ '1', '2', '3' ]
```

dependency:
```json
$ cat package.json
{
  "dependencies": {
    "commander": "^3.0.1"
  }
}
```

Expecting exit code with an error instead of continue of it.

shadowspawn:
Reproduced.

shadowspawn:
This old PR looks like same issue: #121

shadowspawn:
Likely to be same issue as #561

shadowspawn:
Closing this as duplicate of #561

shadowspawn:
Opened a PR to rework option parsing to avoid dropping options and arguments: #1138

Issue #1032, opened 2019-08-28
Passing command option leaves content in args


oliversturm:
As [discussed here](https://github.com/tj/commander.js/issues/978#issuecomment-525692913), opening as a new issue.

I'm trying to find out whether a command has been executed. The simple idea is that if a command is given, it should be executed (and nothing else). If no command is given, the default logic at the bottom should execute (but not otherwise). In all cases, the global option --load should be supported. Here's the test code:

```js
var program = require('commander');

program.version('0.0.1').option('-l, --load <name>', 'load something');

program
  .command('doit [id]')
  .description('do something')
  .option('--better', 'do it better')
  .action((id, cmd) => {
    console.log('executing doit action');
    console.log('id is', id);
    console.log('global load option is ', cmd.parent.load);
    console.log('cmd.better is', cmd.better);
  });

program.parse(process.argv);

if (program.args.length === 0)
  console.log('started without command, executing main logic');
```

This works correctly when I execute like this:

```shell
➜ node test.js -l loadname doit myid        
executing doit action
id is myid
global load option is  loadname
cmd.better is undefined
```

However, when I pass the --better option to the command, the whole thing breaks and the "main logic" is executed in addition to the command:

```shell
➜ node test.js -l loadname doit --better myid
executing doit action
id is myid
global load option is  loadname
cmd.better is true
started without command, executing main logic
```

Unexpectedly, `args` contains information after the action has been executed, *if* the command option `--better` is passed.

shadowspawn:
(Thanks @oliversturm )

shadowspawn:
This will be fixed by #1138

shadowspawn:
The contents of `program.args` is now the arguments with known options removed in the published pre-release of Commander v5.0.0 (#1163)

There isn't an explicit way to directly tell if a command has been executed, but more edge cases are now handled by default (i.e. missing command, unknown command).

shadowspawn:
Commander v5.0.0 has been released with significant changes the argument parsing.

https://github.com/tj/commander.js/releases/tag/v5.0.0

Issue #962, opened 2019-05-15
Subcommand destroys ordering of options


laurentseal:
Commander.js version: 2.13.0

The feature "subcommand in dedicated binary" destroys option ordering for boolean longopts. The bug is that, for unknown options, commander always assumes they have an argument. Boolean longopts don't have an argument, so commander assumes the next option to be an argument to the boolean option. Unknown options get sorted to the end of the argument list, so option ordering is nonsensical after the aforementioned wrong assumption has taken place.

Example:

```
$ ls
somecommand              somecommand-subcommand1
```

Contents of somecommand:
```
program.command('subcommand1', 'execute subcommand1').
```

Triggering the bug:
`$ ./somecommand subcommand1 --bool-option subcommand2 --subcommand2-option subcommand2-argument`

should result in 

`./somecommand-subcommand1 --bool-option subcommand2 --subcommand2-option subcommand2-argument`

being called but commander executes

`./somecommand-subcommand1 subcommand2-argument --bool-option subcommand2`

This is a serious bug because the usage output cannot be made correct for the user, it is confusing the user and it results in the program not working without user feedback in combination with bug #961 .
It also could lead to random logic being executed if the options are implemented in a specific manner in the subcommand.

MatthewRalston:
Seconded. I'm having a similar issue with 'git-style subcommands'. I have to admit it's only happening for one of my subcommands. For example, I'm using a `--verbose` flag to enable debug level logging in a subcommand. Node v12.2.0, commander 2.20.0


## Expectation

I expect that the process.argv from the 'git-style subcommand' is identical in order to process.argv from the main command

## Observation

I observe a reordering of the process.argv array that interferes with positional argument parsing *AND* doesn't throw an error about non-flag arguments being passed/parsed. Perhaps this is not important, but the order of all flags/positional arguments is rearranged. Also, this is not being handled by 

## Steps taken to debug

Only affects one of my subcommands, 'probability' (sub2) which is correctly spelled in both, and no similar command or alias exists. Also, rearrangement of the command declarations in `mycommand` had no effect.

```bash
bin/mycommand sub1 --verbose positionalArg1 # works fine
bin/mycommand sub2 --verbose positionalArg1 positionalArg2 #Causes the process.argv order to be rearranged 
```

To find *where* the rearrangement is being done, I did the following


mycommand
```node
program.description(...)
   .version(...)
   .command('sub1 <positionalArg1> <positionalArg2>', 'sub1 does distance')
[...]

Issue #903, opened 2019-01-05
Provide access to the allowUnknownOption complaints


JoshuaKGoldberg:
There's no way I can find to find _which_ args are unsupported. In this 

```typescript
program..allowUnknownOption();

// error on unknown commands
program.on('command:*', function () {
  console.error('Invalid command: %s\nSee --help for a list of available commands.', program.args.join(' '));
  process.exit(1);
});
```

The function receives no arguments and `command.args` is empty in my usage. Can there be a `command.unknownOptions` added to store the ones added?

shadowspawn:
Closing in favour of #802

Thank you  for your contributions.

PR #1138, opened 2020-01-05
parseOptions rework phase 1

shadowspawn:
# Pull Request

Rework option handling to preserve argument order and make more consistent in face of unknown options, including during Commander's own processing before dispatching to subcommands.

## Problem

The previous handling of unknown options assumed that unknown options can be removed and tracked separately by assuming that they all take a value. This leads to problems particularly since the unknown option may turn out to be a plain boolean flag. This causes bugs with reordered or dropped arguments after unknown options are encountered.

Issues: #508 #962 #1032 #1039

Historical issues: #56 #75 

When the undocumented `.allowUnknownOption` was used, the unknown options were not available from an action handler or after `.parse` called.

Issue: #802 #903

## Solution

`.parseOptions()` removes the known options and splits the remaining arguments into _operands_ (non-options) and _unknown_ without changing the order of the remaining arguments, and without making assumptions about whether unknown options take values or are flags. 

The unknown options are included along with other arguments to action handlers et al for processing by later commands.

NB: this does significantly change the value of `.args` after `.parse()` and the behaviour when the undocumented `.allowUnknownOption()` is used, and affects client usage of those features.

### Fixed

- preserve argument order in subcommands (#508 #962 #1138)
- program.args has arguments with just top level options removed (#1032 #1138)
- unknown options included in `.args` and action handler arguments (#802 #1138)

### Changed

- *Breaking* `.args` contains command arguments with just recognised options removed (#1138)
- *Breaking* `.parseOptions()` (#1138)
    - `args` in returned result renamed `operands` and does not include anything after first unknown option
    - `unknown` in returned result has arguments after first unknown option including operands, not just options and values
- *Breaking* `.on('command:*', callback)` and other command events passed (changed) results from `.parseOptions`, i.e. operands and unknown  (#1138)
- *Breaking* when `.allowUnknownOption()` used, any unknown options are included in arguments passed to command action handler (#1138)


shadowspawn:
Thanks @abetomo ! 👍 

I'll update branches and merge with some care...  🛠

PR #1145, opened 2020-01-09
parseOptions rework phase 2

shadowspawn:
# Pull Request

## Problem

`.parse()` called a private routine `.normalize()` to split short options, turn `--long=value` into `--long value`, et al. This caused a few subtle problems:

- the extra functionality was not available when called public routine `.parseOptions()`
- the expansion was done before the types of the flags for subcommands were known, so support for including value in argument like `-fValue` (added in #599) only worked for top level command 
- an argument into a subcommand option could get incorrectly processed as a global flag due to early expansion, like `do sub --flag=--version`

## Solution

1. roll the support into `.parseOptions` and delete `.normalise`
2. expand arguments only for known options
3. as a bonus, add support for multiple boolean short flags followed by a flag expecting an argument (in subcommands as well as top command)

So for example:
```
program
  .command('sub')
  .option('-w,--www')
  .option('-s,--secure')
  .option('-p,--port <number>')
  .action((cmd) => {
    console.log(cmd.opts());
  });
program.parse(process.argv);
```

Can now call like:
```
# my-program sub -wsp443
{ www: true, secure: true, port: '443' }
```

### Added

- combined short options as single argument may include boolean flags and value flag and value (e.g. `-a -b -p 80` can be written as `-abp80`) (#1145)
- `.parseOption()` includes short flag and long flag expansions (#1145)

### Changed

- only recognised option short flags and long flags are expanded (e.g. `-ab` or `--foo=bar`) (#1145)

### Fixed

- combining option short flag and value in single argument now work for subcommands (e.g. `-p80`) (#1145)

### Removed

- private function `normalize` (the functionality has been integrated into `parseOptions`)


shadowspawn:
(I ended up doing a merge and not a rebase. It has resolved the branch conflicts, but still showing extra commits that have already been squashed and merged so apologies for the noise.)

"Files changed" is looking sensible.

shadowspawn:
Had another go at the rebase, and succeeded. See #1152

Issue #1105, opened 2019-11-23
Change Command to not inherit from EventEmitter?


shadowspawn:
While working on the name pollution issue, I started thinking it could be much tidier if Command "has-a" EventEmitter rather than "is-a" EventEmitter.

I expect the primary use of EventEmitter by clients is for adding `.on()` callbacks. This could be offered with a delegate routine for just that, and the other EventEmitter functionality done internally using a private property.

Advantages of changing the EventEmitter usage are:

- reduce noise in code completion suggestions in editors
- reduce name collisions when storing option values as properties on Command
- if clients do not need access to EventEmitter, can remove implicit dependency on TypeScript definitions for node

Do any gentle readers use more than `.on()` from EventEmitter? 

Do you rely on `this` being the command object in your listener?

shadowspawn:
Better to stay with conventional pattern without strong case for changing, or good examples of other patterns

PR #974, opened 2019-06-12
fix: Add node types to typings

ffflorian:

PR #1146, opened 2020-01-10
Remove EventEmitter from TypeScript definition file

shadowspawn:
# Pull Request

## Problem

Need to also install `@types/node` or get an error due to explicit inheritance of Command from EventEmitter.

Related: #646 #974 #1105

## Solution

Don't say Command extends `EventEmitter`! Instead list the `.on()` function manually.

A reason for not doing this would be if there are likely use cases where people are using more of the methods from EventEmitter on Command, or passing around a Command as an EventEmitter. I have not encountered this in any programs or examples. For rare cases it is easy enough to override the type so this does not prevent use.

Ideally I would like to actually change Command so it _has-a_ EventEmitter and not _is-a_ EventEmitter (#1105 ), but that is a much bigger change. This PR is a simple typings change to achieve a similar outcome for TypeScript now without changing the implementation at all.

### Changed

- removed EventEmitter from TypeScript definition for Command

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -605,12 +605,7 @@ console.log(programOptions.name);
 
 ### TypeScript
 
-The Commander package includes its TypeScript Definition file, but also requires the node types which you need to install yourself. e.g.
-
-```bash
-npm install commander
-npm install --save-dev @types/node
-```
+The Commander package includes its TypeScript Definition file.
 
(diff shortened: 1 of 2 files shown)


shadowspawn:
This is more opinionated than Commander usually is about how Commander might be used, but is consistent with common and expected usage and simplifies installation.

Issue #1088, opened 2019-11-02
Triage: detecting missing or unknown commands


shadowspawn:
What is the best way to test for missing or unknown commands? Is the existing support good enough?

Research prompted by Commander v4 which changes `program.args` and will break some client code which was dependent on the previous behaviour: #1048

Example of affected code: https://github.com/tj/commander.js/issues/7#issuecomment-32448653

Related issues:
- #7 Show help if no command was executed
- #240 Support to know if any known command was processed, or any output written
- #338 How to detect an unsupported command? [git-style]
- #432 running an unknown subcommand does not fail
- #603 Show help when command is unknown
- #809 Using sub-command always invokes *
- #940 argments are empty
- #942 Throw error when no sub-command specified
- #977 It looks not way to know has action is run?
- #978 Top level command action does not work
- #1032 Passing command option leaves content in args

shadowspawn:
A direct test for no arguments at all (including options) is used in the README. This works the same in Commander v4.
```
if (!process.argv.slice(2).length) {
  program.outputHelp(make_red);
}
```

or without the slice:
```
if (process.argv.length < 3) {
  program.help();
}
```

or equivalently using `rawArgs:
```
if (program.rawArgs.length < 3) {
  program.help();
}
```

shadowspawn:
The example code from the README for detected unknown commands (in a program using action handlers) works the same in Commander v4:

```
program.on('command:*', function () {
  console.error('Invalid command: %s\nSee --help for a list of available commands.', program.args.join(' '));
  process.exit(1);
});
```

shadowspawn:
I am somewhat interested in adding support for testing for an unhandled command, which has not triggered any action handlers or external commands. This would implicitly cover both missing and unknown commands with a single test.

shadowspawn:
People are also interested in having Commander do the messaging in missing and unknown cases, but probably treat that as a separate enhancement.

shadowspawn:
By default, an unknown subcommand shows an error and a missing subcommand shows the help in the published pre-release of Commander v5.0.0 (#1163)

Issue #995, opened 2019-07-15
The Command.prototype.parse() method has issues.


yossan:
This method has the following issues:

* when registered `subcommand`, thier arguments and options isn't parsed to check valid.
* when registered `action`, `this.parseArgs()` called subcommand without parsing their arguments and options.

As a result, this method's arguments are not evaluted correctly in subcommand.
In the following code expect to occur `error: missing required argument "file"` but not now.

```
parse
  .command('install <file>')
  .parse('node', 'main.js', 'install') // Expect error but not.
```

Also now `the required argument` will not correctly work. 
[Fix to fail if required arguments don't exist.](https://github.com/tj/commander.js/pull/941)

So it seems that this method should be reconsidered fundamentally.

Thank you for the wonderful library.

# Related 

https://github.com/tj/commander.js/pull/941

shadowspawn:
Thanks @yossan. There are some deliberate differences in behaviour between `.action` and separate executable sub-commands, but I think there are also a number of shortcomings and accidental differences. Bring them closer together and detecting more errors is on my wish list.

yossan:
@shadowspawn 

Thank you for the review.
I want "the required argument" to work correctly.
It's a really great library.

shadowspawn:
A missing required argument for a command now shows an error in the published pre-release of Commander v5.0.0 (#1163)

shadowspawn:
Commander v5.0.0 has been released with significant changes to the parsing, and adds default detection of a missing required argument.

https://github.com/tj/commander.js/releases/tag/v5.0.0

Issue #948, opened 2019-04-04
Built-in help subcommand does not work with aliased git-style commands


MattTheNub:
# Example

```js
#!/usr/bin/env node

const program = require('commander');

program
  .version('0.0.1')
  .description('Fake package manager')
  .command('install [name]', 'install one or more packages').alias('i');

program.parse(process.argv);
```
When running the built-in help command on the alias (`pm help i`), this happens:
## Expected Behavior
The alias gets resolved to `install`, and the information is read from `pm-install`
## Actual Behavior
The command attempts to read `pm-i` and fails, as it does not exist.

shadowspawn:
The `npm help foo` pattern has never been documented in the README (#892) and would appear to be buggy making me wonder how implemented it was.

With your sample program it appears to be working as expected for `--help` though, given I only created index.js:

```
$ node index.js install --help
error: index-install(1) does not exist, try --help
$ node index.js i --help
error: index-install(1) does not exist, try --help

```

shadowspawn:
I came across the (simple) implementation code today and had another look at behaviour. Commander is blindly turning "help foo" into "foo --help", but reasonable it should work with alias. 

Categorising this as a bug.

shadowspawn:
This is fixed in the published pre-release of Commander v5.0.0 (#1163)

shadowspawn:
Commander v5.0.0 has been released.

https://github.com/tj/commander.js/releases/tag/v5.0.0

PR #1149, opened 2020-01-12
Parse rework for nested commands

shadowspawn:
# Pull Request

Going wide and trying to unify and improve behaviours at same time as adding support for nested commands.

## Problem

The parsing is a collection of accumulated behaviours, making it difficult to add new features, and in particular nested commands. For example:

- action handlers and executable subcommands have individual historical behaviours
- validation of options done in action handler support rather than independent
- the parsing is scattered between multiple routines

Using the Event Emitter for internal processing is a complication and introduces possible interaction with user defined listeners in middle of internal processing.

Commander started out being flexible and leaving things to caller, but as Commander has become more structured and full featured there is an expectation that some behaviours should be included (e.g. error when subcommand not found, due to typo).

### References

Commander does not support nested commands (without using git-style executables):

- Issue: #1 #63 #655 #764
- Related PR: #245 #1024

Allow a non-executable to be set as the default command: #742

Implicit help command: #205 #256 #350 #892
but not when an argument: #1153

Implicit help not working for alias: #948 

Built-in error when unrecognised command: #432 #1088

Help displayed for missing command: #1088 

Built-in error for missing arguments: #995 

Asterisk handler getting called for executable subcommands: #809

## Solution

Rework parsing, big time:

- do not use events for internal command processing (still used for option processing)
- explicitly handle just one of action vs executable
- parse command independent of whether action handler attached
- recursive parsing, for nested command support
- recursive parsing, with intention top level command is less special
- treat action/executable subcommands more symmetrically
- add error and/or help for missing and unknown commands (for all subcommands, not just executable)

### Added

- allow a non-executable to be set as the default command (#742 #1149)
- support nested sub-commands with action-handlers (#1 #764 #1149)
- automatic implicit help command for all subcommands (not just executables) (#1149)
- customise implicit help command with `.addHelpCommand()` (#1149)
- _Breaking_ display error message for unknown subcommand (#432 #1088 #1149)
- _Breaking_ display help for missing subcommand (#1088 #1149)

### Changed

- _Breaking_  display error if required argument for command is missing (#995 #1149)

### Fixed

- only add implicit help command when it will not conflict with other arguments (#1153 #1149)
- do not emit `command:*` for executable subcommands (#809 #1149)
- implicit help command works with aliases (#948 #1149)
- options are validated whether or not there is an action handler (#1149)

### Removed

- `parseExpectedArgs` is now private (#1149)

### Migration tips

If you use `.on('command:*')` or more complicated tests to detect an unrecognised subcommand, you may be able to delete the code and rely on the default behaviour.

If you use `program.args` or more complicated tests to detect a missing subcommand, you may be able to delete the code and rely on the default behaviour.

If you use `.command('*')` to add a default command, you may be be able to use `isDefault:true` with a normal command instead.


byCedric:
Awesome work! 🚀

shadowspawn:
I am thinking of moving onto `develop` after this PR?

The other things I am considering for 5.0 are largely non-breaking:
- change all setup errors by author to `throw` rather than error+exit (about half already do), and add tips to the infamous accidental executable situation
- move `Command` from prototype to class (technically non-breaking, but core change)
- switch from `var` to `const` and `let` (non-breaking)
- make `.helpInformation()` public and rework README section on help (non-breaking)
- consider more options to .parse for electron et al (technically non-breaking)

index.js
```diff
@@ -124,19 +122,28 @@ exports.CommanderError = CommanderError;
...
   this.options = [];
-  this._execs = new Set();
   this._allowUnknownOption = false;
   this._args = [];
   this._name = name || '';
   this._optionValues = {};
   this._storeOptionsAsProperties = true; // backwards compatible by default
   this._passCommandToAction = true; // backwards compatible by default
   this._actionResults = [];
+  this._actionHandler = undefined;
```
abetomo: I'm sorry if I sound picky.
Here we 'define' it.So I think it's not 'undefined'.
I think 'null' is better, but how about?

shadowspawn: I did feel "defining" it to be "undefined" was a bit odd. I was really just adding it to be clear it was a property. I'll do a little reading on null vs undefined conventions. Good question.

shadowspawn: Changed all to null except the boolean that is lazy initialised where undefined felt better (and added a comment for that one).

shadowspawn:
Woo hoo! 🎉 
Thanks @abetomo , it was a big and complex Pull Request to review.

I'll merge it in soon and start work on preparing CHANGELOG / Release Notes for a prerelease.

shadowspawn:
I gave the nested command support a try-out emulating the Cordova CLI. I implemented one nested subcommand using inline code, and one using `.addCommand` and a module:
https://github.com/shadowspawn/commander-does-cordova.git

PR #1156, opened 2020-01-27
Minor Readme Update for Debugging Information

jamesrenaud:
# Pull Request

## Problem

Debugging within the git style sub(commands) using VSCode is a little tricky, and took me a while to figure out how to handle that case.

## Solution

Just a quick miscellaneous addition to the `bits and pieces` section of the documentation to show the correct launch flag.


shadowspawn:
Thanks @jamesrenaud . I think I'll rework the debugging section to include this new information rather than add a separate section, but no action required from you.

shadowspawn:
I have included this information in #1166

PR #1157, opened 2020-01-29
fix(ts): action can be an async function

prokopsimek:
Since we can use `parseAsync` for async functions so the action can be asynchronous.

related to https://github.com/tj/commander.js/issues/806

# Pull Request

## Problem
```
Promise returned in function argument where a void return was expected.eslint(@typescript-eslint/no-misused-promises)
```

## Solution
add `Promise<void>` to the action argument

- TypeScript typings

### DIFF
--- a/typings/commander-tests.ts
+++ b/typings/commander-tests.ts
@@ -45,6 +45,14 @@ function increaseVerbosity(v: any, total: number) {
     return total + 1;
 }
 
+function syncCall() {
+    console.log("Sync success!");
+}
+
+async function asyncCall() {
+    return;
+}
+
 program
     .version('0.0.1')
     .usage('[options] <file ...>')
@@ -111,7 +119,11 @@ program
 
 program
     .command("name1", "description")
-    .command("name2", "description", { isDefault:true })
+    .command("name2", "description", { isDefault:true });
+
+program
+    .command("name3").action(syncCall)
+    .command("name4").action(asyncCall);
 
 program
     .exitOverride();
--- a/typings/index.d.ts
+++ b/typings/index.d.ts
@@ -115,7 +115,7 @@ declare namespace commander {
      *
      * @returns Command for chaining
      */
-    action(fn: (...args: any[]) => void): Command;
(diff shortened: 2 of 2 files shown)


shadowspawn:
Thanks.

Would you like to add an example usage to `typings/commander-tests.ts`? (This file is compiled to check for errors, but not run.)

prokopsimek:
@shadowspawn Hi, I'll add the test tomorrow :)

typings/commander-tests.ts
```diff
@@ -45,6 +45,16 @@ function increaseVerbosity(v: any, total: number) {
     return total + 1;
 }
 
+function syncCall() {
```
shadowspawn: Sync has void return? 😄

prokopsimek: Exactly .)

typings/commander-tests.ts
```diff
@@ -112,6 +122,8 @@ program
 program
     .command("name1", "description")
     .command("name2", "description", { isDefault:true })
+    .command("name3", "description").action(syncCall)
```
shadowspawn: Commands with action handlers should not supply description. Which we can not enforce with typing, sadly! I like adding it here as name3, name4 though. But name 3 and name 4 need to be added to program alone. e.g.
```
program
  .command("name3)
  .action(syncCall);
```

prokopsimek: thanks! :) Fixed

shadowspawn:
Thanks @prokopsimek 
We have changes for a v5 underway, but I am thinking we will release this in a v4 patch too as was not as intended ( I missed the TypeScript changes to the action parameter when adding .parseAsync()).

shadowspawn:
To confirm the nature of the fix, this PR resolves an eslint error, rather than a TypeScript error?

(I am having difficulty reproducing the error condition myself.)

shadowspawn:
This has been released:
https://github.com/tj/commander.js/releases/tag/v4.1.1

PR #1167, opened 2020-02-04
Update dependencies and rewrite tests

shadowspawn:
# Pull Request

## Problem

Update dependencies, but lots of lint errors introduced by default tests in updated eslint test.

## Solution

Rework tests and much reduce use of callback to tests. Actually tighter nicer code on the whole!

Disabled lint in a couple of places where not sure how to rewrite Jest test.


shadowspawn:
I deliberately left @types/node on 12 as the LTS version, but would be ok to use current too.

PR #1168, opened 2020-02-05
Replace anonymous functions with arrow functions

abetomo:
# Pull Request

## Solution

You can use the arrow function to delete `self`.
I like this style, but close it if you don't need it.

### DIFF
--- a/index.js
+++ b/index.js
@@ -274,8 +274,7 @@ class Command extends EventEmitter {
 
   _parseExpectedArgs(args) {
     if (!args.length) return;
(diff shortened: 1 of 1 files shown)


shadowspawn:
Oops, bad merge in the web user interface by me. I'll try fixing from command-line...

shadowspawn:
Manually merged. (I was using self in a recently added `function` that I have switched to arrow function too.)

abetomo:
Thank you!

Issue #512, opened 2016-03-23
Commander incorrectly assumes that process.argv[0] is 'nodejs' and argv[1] is script name


sedwards2009:
Command assumes that the first value in `process.argv` is the name of the nodejs executable being used, and the the second value is the name of the script which was passed to nodejs.

This is mostly fine except in the situation where a packaged electron app is used. For a packaged app a custom binary is used (typically with the name of the app) and it automatically calls the `main.js` script at start up. It's `process.argv` starts with the name of the exe and immediately the first of the user supplied arguments.

A possible solution may be to add an extra method which only takes the list of arguments. The calling application would then be responsible for extracting the real arguments from `process.argv`. (The method could also just return the parsed results and not modify the `commander` object. See #183 )

cheton:
I fixed my issue with Electron app with this workaround:
```js
program
  .version('0.1.0')
  .option('-c, --config <filename>', 'set config file');

if (process.argv.length > 1) {
  program.parse(process.argv);
}
```

bolinfest:
There should be a way to tell `parse()` to treat the arguments as-is. This is the offending line:

https://github.com/tj/commander.js/blob/v2.9.0/index.js#L455

If you have the args as a value, `args`, then it seems like you have to do:

    program.parse([null, null].concat(args))

to get the expected behavior. If you still want to be clever without changing the API, then instead of:

```
var parsed = this.parseOptions(this.normalize(argv.slice(2)));
```

you could do:

```
var effectiveArgs = argv === process.argv ? argv.slice(2) : argv;
var parsed = this.parseOptions(this.normalize(effectiveArgs));
```

That way, if `argv` is a user-defined array, it does not need the prefix, but if it is the canonical `process.argv` argument, then `slice(2)` is applied as normal.

Though I guess this would break anyone who has already hacked around this behavior?

hvmonteiro:
I've catched this problem today, the solution below only works if you are executing the application with 'node start', not if you are directly executing an electron-builder executable (number of parameters change).

```
if (process.argv.length > 1) {
  program.parse(process.argv);
}

```

cheton:
This works for both scenario. pkg.name is the application name specified in your package.json.
```js
const normalizedArgv = ('' + process.argv[0]).indexOf(pkg.name) >= 0
    ? ['node', pkg.name, ...process.argv.slice(1)]
    : process.argv;
if (normalizedArgv.length > 1) {
    program.parse(normalizedArgv);
}
```

[...]

PR #1172, opened 2020-02-06
Parse argument conventions

shadowspawn:
# Pull Request

## Problem

When the parse arguments are not from node, it is somewhat clumsy massaging the arguments to match the expected conventions. The case I encounter myself is when writing tests. A case that has been reported is when hosting in electron.

Separately, in most casual cases the argument to parse is `process.argv`.

Related: #512

## Solution

- default to `process.argv` when arguments array not supplied
- add `from` option to make it easy to  handle other conventions

---

More code and longer README than I hoped, but useful and straight forward, see whether you think it it is useful enough. Could shift content from README to an example  to make it shorter.


shadowspawn:
Thanks. I'll do another prerelease this weekend.

Issue #1015, opened 2019-08-11
Have in-house support in-order to recommend matching commands


jamesgeorge007:
[yargs](https://github.com/yargs/yargs) comes shipped with a `recommendCommands()` method as part of the public API. It would be great if commander.js would have it as well :clap: 

```ts
program
  .name('sample')
  .usage('<command>)
  .version(version)
  .recommendCommands();

program
  .command('init')
  .action(() => {});

program
  .command('create');
  .action(() => {});
```

`sample creat`
```bash
Unknown command creat.
Did you mean create?
```

jamesgeorge007:
@shadowspawn I would love to submit a PR if you could throw light on the implementation strategy :+1:

jamesgeorge007:
@shadowspawn @oGsLP thoughts :thinking:

oGsLP:
Only for me, I think it's not necessary. People can realize it by 
```js
program.command("*").action(cmd => {
  suggestCommand(cmd);
});
```
And we can find a npm package named `didyoumean` to customzie `suggestCommand(cmd)` easily.

I'm not familiar with yargs, it seems that it doesn't have an api to match regex cmd like '*' so it has `recommendCommands()` to handle this. But commander do.

shadowspawn:
I do not think it likely I would accept a PR for this currently, as we still have some improvements to make in whether Commander warns about unrecognised commands in the first place. (So not "in-house support".)

I am interested in how to add possible matches, and currently that would be in own code or in an add-on package. Also interested in how popular an idea this is, so feel free to leave this issue open even if you decide not to investigate yourself.

My limited knowledge (not researched) is most computer programs use the Levenshtein distance algorithm. I was casually looking at didyoumean and didyoumean2 in weekend. (Using these as examples rather than recommendations.)

(Both roll-your-own solutions and didyoumean also suggested by @oGsLP, thanks for comment )

shadowspawn:
I came across an implementation in comments:  https://github.com/tj/commander.js/issues/432#issuecomment-175785162

jamesgeorge007:
@shadowspawn what about documenting how to achieve it with [didyoumean](https://www.npmjs.com/package/didyoumean).
I would happily submit a PR :+1:

shadowspawn:
Same feedback as building it in, I do not think it likely I would accept a PR for this currently. If the code is not huge, you could simply add it as a comment here though.

jamesgeorge007:
A typical implementaion would be:-
```js
[...]

PR #1176, opened 2020-02-07
Update README example for .on('command:*')

shadowspawn:
# Pull Request

## Problem

A handler for `.on('command:*')` is no longer needed just to show an error message, as we do that by default, so would like a new example in the README.

It was suggested we add built-in support to recommend matching command, but do not want to that for now since it would add a dependency: #1015 

The example code directly calls `process.exit()` which is not recommended.

Was:
```
// error on unknown commands
program.on('command:*', function () {
  console.error('Invalid command: %s\nSee --help for a list of available commands.', program.args.join(' '));
  process.exit(1);
});
```

## Solution

Use actual code for `didYouMean` as the example.

Follow best practice and do not directly call `.exit()`

```
program.on('command:*', function (operands) {
  console.error(`error: unknown command '${operands[0]}'`);
  const availableCommands = program.commands.map(cmd => cmd.name());
  const suggestion = didYouMean(operands[0], availableCommands);
  if (suggestion)
    console.error(`Did you mean '${suggestion}'?`);
  process.exitCode = 1;
});
```

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -554,11 +554,13 @@ program.on('option:verbose', function () {
   process.env.VERBOSE = this.verbose;
 });
 
-// custom error on unknown command
 program.on('command:*', function (operands) {
-  console.error(`Invalid command '${operands[0]}'. Did you mean:`);
-  console.error(mySuggestions(operands[0]));
-  process.exit(1);
+  console.error(`error: unknown command '${operands[0]}'`);
+  const availableCommands = program.commands.map(cmd => cmd.name());
+  const suggestion = didYouMean(operands[0], availableCommands);
+  if (suggestion)
+    console.error(`Did you mean '${suggestion}'?`);
+  process.exitCode = 1;
 });
 ```
 


Readme.md
```diff
@@ -554,11 +554,13 @@ program.on('option:verbose', function () {
...
-// custom error on unknown command
 program.on('command:*', function (operands) {
-  console.error(`Invalid command '${operands[0]}'. Did you mean:`);
-  console.error(mySuggestions(operands[0]));
-  process.exit(1);
+  console.error(`error: unknown command '${operands[0]}'`);
+  const availableCommands = program.commands.map(cmd => cmd.name());
+  const suggestion = didYouMean(operands[0], availableCommands);
+  if (suggestion)
+    console.error(`Did you mean '${suggestion}'?`);
```
jamesgeorge007: This would be misleading. Say for instance, a first time user makes use of this code snippet to handle unknown commands ending up with an error saying that `didYouMean` is not defined.

shadowspawn: An alternative I considered was to use a clearly write-your-own routine like:

```
const availableCommands = program.commands.map(cmd => cmd.name());
const suggestion = myFindBestMatch(operands[0], availableCommands);
console.error(`Did you mean '${suggestion}'?`);
process.exitCode = 1;
```

Do you think this would be preferable?

jamesgeorge007: https://github.com/tj/commander.js/issues/1015#issuecomment-583810519

PR #1179, opened 2020-02-08
Suggest help when unknown command

shadowspawn:
# Pull Request

## Problem

New users may not know how to get help. We have a new error in Command v5 for unknown commands so a good place to try an improvement.

We have:
```
$ node nested.js silly     
error: unknown command 'silly'
```

Other programs:

```
$ git unrecognised
git: 'unrecognised' is not a git command. See 'git --help'.
$ hg unrecognised
hg: unknown command 'unrecognised'
(use 'hg help' for a list of commands)
$ docker unrecognised
docker: 'unrecognised' is not a docker command.
See 'docker --help'
```

## Solution

Suggest `--help`. 

Notes:
- suggesting help option rather than implicit help command, since always available

```
$ node nested.js silly
error: unknown command 'silly'. See 'nested --help'.
```

### DIFF
--- a/index.js
+++ b/index.js
@@ -1124,7 +1124,12 @@ class Command extends EventEmitter {
    */
 
   unknownCommand() {
-    const message = `error: unknown command '${this.args[0]}'`;
+    const partCommands = [this.name()];
+    for (let parentCmd = this.parent; parentCmd; parentCmd = parentCmd.parent) {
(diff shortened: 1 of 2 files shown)


shadowspawn:
Thanks @abetomo and @jamesgeorge007

Issue #1002, opened 2019-07-25
Support subclassing Command


ngocdaothanh:
Currently, the return type of the methods in class `Command` is `Command`:

```js
public command(name: string, desc?: string, opts?: commander.CommandOptions): Command;
```

I cannot extend `Command` to add more methods, like this:

```js
class MyCommandWithUtilities extends Command {
  public util1(): this {
     // ...
     return this;
  }
}

const cmd = new MyCommandWithUtilities();
cmd
  .command('foo')
  .util1()  // Syntax error, because `command` returns `Command`, not `MyCommandWithUtilities`
  .option(...);
```

So, please change the return type of the methods in `Command` to `this`.

```js
public command(name: string, desc?: string, opts?: commander.CommandOptions): this;
```

It seems that simply changing the definitions in index.d.ts is not enough. The main source code must also be changed.

See:
https://github.com/microsoft/TypeScript/pull/4910

ngocdaothanh:
```
public command(name: string, desc?: string, opts?: commander.CommandOptions): this;
```

It's not that simple, because the current design is to return a new `Command` instance, unless `desc` is specified; the return type is `Command | this`.

So to support subclassing, we need to change the design of `command`.

shadowspawn:
Yes, for action handler based subcommands (no description), `.command` returns a new object representing the new command. (That might be doable using a factory method to create the new object, but definitely more complicated than just changing the TypeScript definition for `.command`.)

ngocdaothanh:
Workaround: use a wrapper.

Example (TypeScript):

```
export class MyCommand {
  private readonly cmd: commander.Command;

  constructor(cmd?: commander.Command) {
    this.cmd = cmd ? cmd : new commander.Command();
  }

  public command(name: string): MyCommand {
    const cmd = this.cmd.command(name);
    return new MyCommand(cmd);
  }

  public description(desc: string): this {
    this.cmd.description(desc);
    return this;
  }

  public util1(): this {
     // ...
     return this;
  }

  ...
}
```

shadowspawn:
See #1191 for work in progress which resolves this issue.

PR #1180, opened 2020-02-08
Use this instead of Command as a return type in TypeScript

vonagam:
# Pull Request

## Problem

```ts

  customMethod() {
    return this;
  }
}

const custom = new CustomCommand();

custom
  .description('a custom method will be treated as `any` after that')
  .customMethod(); // no type checks here or after that
```

Because right now `description` and other chainable methods return `Command` and not `this`, type information of subclasses gets inaccessible after a call to an inherited method (and would have otherwise caused an error if not for `[key: string]: any` in `Command` for storing parsed options on a instance). 

## Solution

Changed return type to `this` in chainable methods.

Also made `parseAsync` return `Promise<this>` instead of returning `Promise<_actionResults>` to be consistent with `parse`. (This  change is little off topic, can remove if you want to expose `_actionResults`)

(This PR to `develop` branch, not `master`)

### DIFF
--- a/index.js
+++ b/index.js
@@ -712,7 +712,7 @@ class Command extends EventEmitter {
 
   parseAsync(argv, parseOptions) {
     this.parse(argv, parseOptions);
-    return Promise.all(this._actionResults);
+    return Promise.all(this._actionResults).then(() => this);
   };
 
   /**
--- a/typings/index.d.ts
+++ b/typings/index.d.ts
@@ -40,7 +40,7 @@ declare namespace commander {
      * 
      * You can optionally supply the  flags and description to override the defaults.
      */
-    version(str: string, flags?: string, description?: string): Command;
+    version(str: string, flags?: string, description?: string): this;
 
     /**
      * Define a command, implemented using an action handler.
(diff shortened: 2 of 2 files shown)


shadowspawn:
My first impressions are favourable. 😃 

1) Related #1002

2) Could also change return type of command() signature for adding stand-alone executable?

```
 command(nameAndArgs: string, description: string, opts?: commander.CommandOptions): Command;
```

3) parseAsync: when I added `.parseAsync()`, `.parse()` did not have a well behaved return type and I deliberately did not match it. However, now on the develop branch `.parse()` is well behaved and I am interested in your change.

vonagam:
About second point. As you said in the mentioned issue:

> That might be doable using a factory method to create the new object, but definitely more complicated than just changing the TypeScript definition for .command

Because if you do not redefine `command` in subclass, then you should expect `Command` class and not your subclass, since [that](https://github.com/tj/commander.js/blob/30ae4ac91bc7806497021d5531e2476406c5f3ad/index.js#L163) what you would get.

```ts

  customMethod() {
    return this;
  }
}

const custom = new CustomCommand();

custom
  .command('subcommand') // this returns Command, not CustomCommand
  .customMethod(); // this is untyped and causes a runtime error
```

But if you redefine `command` in subclass to return that subclass, then it will work and typescript allows it.

```ts

  command( ...args: any[] ) {
    const hack = super.command( ...args );
    const command = new CustomCommand();
    Object.assign(command, hack);
    return command; 
  }

  customMethod() {
    return this;
  }
}

const custom = new CustomCommand();

custom
  .command('subcommand') // this is CustomCommand type
  .customMethod(); // this is properly typed and works
```

shadowspawn:
I was referring to the "other" signature for `.command()` when it is passed a description, and when it does just return `this`.

```
const custom = new CustomCommand();

custom
  .command('sub1', 'implemented in stand-alone executable')
  .command('sub2', 'implemented in stand-alone executable')
  .customMethod();
```

vonagam:
Ah, sorry, misunderstood.

shadowspawn:
The two different behaviours of `.command()` are a great source of confusion, but this change will in a small way make it clearer.

shadowspawn:

PR #1181, opened 2020-02-08
[Docs]: require didyoumean.js lib before use

jamesgeorge007:
# Pull Request

## Problem

`didyoumean.js` was used directly in the example without importing the respective package.

Follow up of #1176 
Addresses #1015 

## Solution

Added the necessary `require` statement.


shadowspawn:
Thanks. I am happy without the require for now. It is just a fragment, and more an example of what is possible than runnable code. The focus of the example is on the listeners.

Issue #1155, opened 2020-01-23
Add way to collect inherited/global options from parents


shadowspawn:
## Problem

When there are global common options on the program, and subcommands using action handlers, there are options in two places. In a simple program you may have easy access to both the "program" and the "command", but that pattern is not always convenient, and does not scale very well to nested subcommands or to the action handler being passed just options.

Current:
```
program
  .option('-g,--global')
  .command('sub')
  .option('-l,--local')
  .action((cmd) => {
    console.log(program.global); // using program
    console.log(cmd.parent.global)); // using parent
    console.log(cmd.local);
  });
```

## Possible Solution

@a-fas offered a routine or an option to `.opt()` to collect all the options:
https://github.com/tj/commander.js/pull/1024#issuecomment-568403615

Enhancing `.opts()` is a nice idea. Adding this issue to make it visible as a possible enhancement.

e.g. trying some wording:
```
const opts1 = cmd.opts({ deep: true });
// or
const opts2 = cmd.opts({ collectAllOptions: true });
```

shadowspawn:
Related historical: #243

shadowspawn:
No activity in a month. I did some research and rethinking the higher level goals. Closing in favour of #1229.

PR #1184, opened 2020-02-10
Type commands property

vonagam:
# Pull Request

## Problem

Since `parent` and `commands` properties do not start with underscore i assume that they are not private. And i see no potential issue with adding types for them, unless i am mistaken in my initial assumption.

## Solution

Added types for them, marking them as readonly to show that a user needs to use `addCommand` if he wants to modify them.

### DIFF
--- a/typings/index.d.ts
+++ b/typings/index.d.ts
@@ -32,6 +32,8 @@ declare namespace commander {
 
     args: string[];
 
+    commands: Command[];
+
     /**
      * Set the program version to `str`. 
      *


shadowspawn:
Are you adding these for completeness, or because you want to use them yourself?

vonagam:
I want to use parent access to customise options collection like in #1155 (in example you use `cmd.parent` inside an action) and other stuff.

I want to use commands access for lookup by name or path like in `_findCommand` method (but it is a private method).

shadowspawn:
Long answer...

There isn't a strong sense of private and public data properties from the original code, other than some use of underscore. Explicit routines were added for accessing some properties like `.name()`.  Very few got added to the TypeScript and I am using my judgement as to what makes sense for maintenance reasons when cases come up.

I am happy adding `commands`. In fact I added code to the README recently which uses that property! Also, without the readonly(s). The readonly only goes so deep, and no protection in javascript anyway, so I am comfortable with user beware. We have `args there without qualification and I would rather follow the same pattern.

I would rather not add `parent`. 
1) It is interesting you used  #1155 as an example as part of the point of that enhancement is to remove a possible case where parent might be used. 
2) `parent` is an implementation detail and I think for internal use rather than public consumption.
3) In particular, I missed `parent` when I was moving property initialisation into the constructor and would likely make it `null` rather than `undefined` if/when I do. We did that for most of the properties which are objects.

vonagam:
Interesting, i though that you would be ok with `parent` since it is a simple property and against `commands` since in the future it might change from an array to an object for example.

`readonly commands: ReadonlyArray<Command>;` - just says that a) you can't replace `commands` with your own array, b) that you can't add or remove commands from array. (You still can modify commands themselves.) Because if you do allow it then there will be a problem with missing or wrong `parent` value.

I am ok with removing readonly restrictions, i just opted for more strict version.

Well with public `parent` people will be able to customise merging of options from above however they like, not just being limited to presets. Also adds ability to get a root from deep nested commands.

So, should i remove `parent` then?

shadowspawn:
> Interesting, i though that you would be ok with parent since it is a simple property and against commands since in the future it might change from an array to an object for example.

Well, you could talk me out of that too! 😃 

> I am ok with removing readonly restrictions, i just opted for more strict version.

I think that was a good thing to suggest and appreciate that you did, but would prefer without for now.

> So, should i remove parent and readonly restrictions then?

Yes thanks, I'm happy to add plain commands.

@abetomo will be reviewing too before it is accepted.

abetomo:
@vonagam Thank you for your contribution!

@shadowspawn Thanks for the review!

shadowspawn:

PR #1187, opened 2020-02-12
Add `fail-fast: false` to GitHub actions settings

abetomo:
# Pull Request

## Problem

If a test fails in one environment, it is all canceled.
You may not know if it will fail in other environments.

## Solution

All tests are run, so it is easier to see which environment failed.


abetomo:
Certainly, this project is unlikely to fail in a specific environment, so this setting may not be necessary.

PR #1195, opened 2020-02-15
Add program to exports

shadowspawn:
# Pull Request

Trying out an idea. See what you think...

## Problem

commander started out with just a default export of a root command. The ability to create a new Command was added later in a backwards compatible way, and is now the recommended way to use commander. 

The combination of a global default export of the root command and named export of Command is a little clumsy.

```

vs

```

## Solution

Add a named export for `program` so a more explicit way of using the global root command and more symmetry with named exports.

```
const program = require('commander').program;
```

(Might eventually make this the only way of accessing the global, one day...)

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -56,7 +56,7 @@ Commander exports a global object which is convenient for quick programs.
 This is used in the examples in this README for brevity.
 
 ```js
-const program = require('commander');
+const program = require('commander').program;
 program.version('0.0.1');
 ```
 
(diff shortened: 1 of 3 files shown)


shadowspawn:
Thanks for review. I'll update the other examples and the Chinese README before I merge.

shadowspawn:
On second thoughts I'll merge and open another PR.

Personally I usually include a whole package, but I think a more common pattern for using a small number of parts of a package is just the parts that are wanted.

So in the README we could use:
```
const { program } = require('commander);
```

instead of:
```
const program = require('commander').program;
```

PR #1199, opened 2020-02-19
Define option multiple times on command line

jonasholtkamp:
# Pull Request

## Problem

This is not related to any issue, since I'd rather file a PR with a (proposed) solution than creating an issue.

I looked for the possibility to define multiple options of the same type/name, like this:

```bash
$ pizza-options --cheese stilton --cheese cheddar --cheese gouda
```

Currently the only way to define multiple values for an options is this way:

```bash
$ pizza-options --cheese stilton, cheddar, gouda
```

The rationale behind my idea is that it way cleaner and easier to use on the command line if you can define an option/parameter multiple times (with different values) rather than define an array of values for one option. A real-world example would be the `-e` option of `docker run`.

## Solution

My (proposed) solution is to create a new `this` attribute, called `multiple`. It is switched to true with this syntax:

```js
program
  .option('--cheese <flavour>+', 'cheese flavour')
  .option('--cheese[flavour]+', 'cheese flavour')
```

I'm not sure if this syntax (and especially the plus sign) is the best way to express this behaviour; I followed regular expression syntax, where a plus sign means basically _one or more occurrences of a token_.

This PR is to be considered as "focus on concept"-first.

## ChangeLog

That's hard. My first try:

```
### Added

* plus sign (`+`) after an option argument indicates that this option can be used multiple times
```

### DIFF
--- a/index.js
+++ b/index.js
@@ -609,7 +609,7 @@ class Command extends EventEmitter {
   /**
    * Store option value
    *
-   * @param {String} key
+   * @param {string} key
    * @param {Object} value
    * @param {boolean} multiple
    * @api private
@@ -629,7 +629,7 @@ class Command extends EventEmitter {
   /**
    * Remove option value from storage
    *
-   * @param {String} key
+   * @param {string} key
    * @param {Object} value
    * @api private
    */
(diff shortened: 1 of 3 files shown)


shadowspawn:
Thanks for the description and context. 

There is a relevant example in the README in the section on https://github.com/tj/commander.js#custom-option-processing

See the `collect` example, does what you are looking for?

jonasholtkamp:
Damn, that's exactly what I built. Didn't find it by searching.

shadowspawn:
There is a lot of information in the README! Easy to miss a small section or example, especially if you are calling it something different than how it is described.

jonasholtkamp:
Yes, that README is so awesome. But, you know, forest and trees. Also, the `collect` is just an example how to use the custom option processing. Maybe the README would benefit from a section that lists best-practice/helpful functions for custom option processing.

shadowspawn:
Closing as collecting multiple option values is already supported by using a custom option processing function.

(Nicely written Pull Request description.)

PR #1200, opened 2020-02-19
Tidier use of global program with require

shadowspawn:
# Pull Request

Changes to README to make use of exported `program` added in #1195

## Problem

The export from commander is a combination of a global Command object, and classes, especially Command. Global object exports are somewhat out of fashion these days.

The require examples in the README were all using the global object.

## Solution

Use the new `program` export and the destructure syntax to more explicitly import a global Command object,


shadowspawn:
I'll do a prerelease soon to get `program` available so matches README on develop branch.

PR #1205, opened 2020-03-02
Add eslint settings for TypeScript

abetomo:
# Pull Request

## Problem

https://github.com/tj/commander.js/pull/1191#issuecomment-593185320
> I suspect we aren't running lint on the TypeScript. 

## Solution

I tried setting TypeScript as a trial.

(@shadowspawn Close it if you are already working on it.)


shadowspawn:
Nice. Good start being able to lint the TypeScript separately. 👍 

I tried turning it on for editor too by combining in `eslintrc` but ran into some issues, not sure whether it will be sensible to do combined settings. Running it separately like this is an ok fallback, we only have two TypeScript files.

I see there is <https://github.com/standard/eslint-config-standard-with-typescript> which could be appropriate  since we are using `standard` for JavaScript.

I suggest if we keep two config files then use same filetype for `.eslintrc.json` and `.typescript-eslint.js`. I don't mind which — maybe `.js` because it is the first file that eslint looks for, and more flexible, and more familiar?

abetomo:
Modified to use 'standard-with-typescript'.

I prefer setting with `js` because it is more flexible than json.

As you mentioned, the file type is different from `.eslintrc.json`.
I tried to `js` the file type as a suggestion.
Which one do you prefer?

shadowspawn:
Let's go with `.js` format.

shadowspawn:
`standard-with-typescript` produces more warnings, but I still like using standard for a start to be consistent with our eslint JavaScript setup.

I suggest add one rule to restore this to default and avoid a lot of errors. We have our own "semi" rule for javascript, so reasonable we need one for TypeScript too. Feel free to make layout a bit more compact, I used fully expanded layout here.

```
    // Set member-delimiter-style back to default, we use semicolons
    // https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/member-delimiter-style.md
    '@typescript-eslint/member-delimiter-style': ['error', {
      "multiline": {
          "delimiter": "semi",
          "requireLast": true
      },
      "singleline": {
          "delimiter": "semi",
          "requireLast": false
      }
    }]
```

As otherwise get errors for the semicolons in this:

```
  interface CommanderError extends Error {
    code: string;
    exitCode: number;
    message: string;
    nestedError?: string;
  }
```

shadowspawn:
Might be able to get the linting working with single configuration file using `overrides`. I have not tried this yet. Main reason I am interested is to see if works in editor too. 🙂 

- https://eslint.org/docs/6.0.0/user-guide/configuring#specifying-processor
- https://stackoverflow.com/questions/57597142/how-to-run-typescript-eslint-on-ts-files-and-eslint-on-js-files-in-the-same-pr

abetomo:
Thank you for the review!

I have modified it based on your feedback.

shadowspawn:
I can now see the errors in VisualStudio Code too when I open the TS files, nice! Quite a few things to fix after this lands, but mostly easy. 😅

Could lint both the typescript files:, script:

```
    "typescript-lint": "eslint 'typings/*.ts'",
```

shadowspawn:
👍 
I'm happy for this to be merged into `develop` now (before 5.0) if you like.

Nice changes to bring our TypeScript style support up to level of our JavaScript for future work with a lint script and editor support.

abetomo:
Thanks for the review!

Taylerpaul (commented):
Stealth Roasting

PR #1206, opened 2020-03-02
Refactor codebase to ES6 and update dependencies

khaledosman:
# Pull Request
- Updated dependencies
- Fixed lint issues with tests using `done` callback
- Split Command, CommanderError, Option into their own files and moved helper functions to utils.js
- Refactored codebase to ES6, classes, let/const, arrow syntax, etc..


khaledosman:
looks like this is already in the develop branch

Issue #531, opened 2016-05-10
Allow for multiple aliases


cbetta:
I'd love for commander allow to allow for multiple aliases. I've found them to be useful for many reasons varying from default subcommands (`records:list` and `records`, `record:show` and `record`), preventing typos (`record:list` vs `records:list`) and just shortcuts (`records:list` and `rl`).

Currently I'm resorting to duplicating my `command` statements which feels far from effective.

cbetta:
I'd be happy to add this but would love feedback on the syntax. 

For one I don't want to show all aliases.

```js
import commander from 'commander';

commander
  .command('link:create <a> <b>')
  .alias('l', 'lc', 'link')
  .action(...);
```

Then `--help` should look something like:

```
> example --help
   Usage: example [options] [command]

   Commands:

      link:create|l|lc|link <a> <b>                         
```

Which might get a bit excessive. Instead some aliases might be hidden instead:

```js
import commander from 'commander';

commander
  .command('link:create <a> <b>')
  .alias('link')
  .hidden_alias('l', 'lc')
  .action(...);
```

Which would then result in:

```
> example --help
   Usage: example [options] [command]

   Commands:
      link:create|link <a> <b>                         
```

The alternative would be to pass an object with the keys of the aliases and the values of the visibility:

```js
import commander from 'commander';

commander
  .command('link:create <a> <b>')
  .alias({ link: true, lc: false, l: false})
  .action(...);
```

kodie:
Would love to see support for multiple aliases.

FireCubez:
Agreed with @tdreid and @kodie , it's the most natural in my opinion

ganobrega:
This is already being supported?

shadowspawn:
@ganobrega Current status is:
- commands support a single alias
- options do not support an alias as such

DuckyCrayfish:
Was a consensus ever reached on the option to support? If not I have yet another suggestion that is a bit of  a middle-ground.

shadowspawn:
Not consensus, throw in another idea @DuckyCrayfish

I personally lean towards just repeating `.alias` or allowing array, and not offering ability to hide. Keep it simple.

For the help, an approach I have copied is to list the aliases in a separate help section. This avoids cluttering up the main command help with the variations. e.g.

```
Aliases:
  i: install
  list: ls
  lsr, list-remote: ls-remote
  remove: rm
```

(The "real" commands are listed on the right.)

nrgapple:
[...]

PR #1236, opened 2020-04-02
Add support for multiple aliases

shadowspawn:
# Pull Request

See #531

## Problem

Some people would like to add more than one alias for a command, but not have them all visible in the help.

## Solution

Allow adding multiple aliases. Only display the first one in the built-in help, and leave any mention of the rest up to the caller.

This is intended to be backwards compatible. There is a small change from v5.0 behaviour, calling `.alias()` when there is no alias will now return `undefined` instead of `null`, but seems unlikely to affect much code and can be made fully backwards compatible if necessary.

To provide public access to the possibly multiple aliases, added `.aliases()` which works with array rather than string.

## ChangeLog

- add support for multiple aliases

### DIFF
--- a/index.js
+++ b/index.js
@@ -117,7 +117,7 @@ class Command extends EventEmitter {
     this._executableFile = null; // custom name for executable
     this._defaultCommandName = null;
     this._exitCallback = null;
-    this._alias = null;
+    this._alias = [];
 
     this._hidden = false;
     this._helpFlags = '-h, --help';
@@ -937,7 +937,7 @@ class Command extends EventEmitter {
    */
   _findCommand(name) {
     if (!name) return undefined;
-    return this.commands.find(cmd => cmd._name === name || cmd._alias === name);
+    return this.commands.find(cmd => cmd._name === name || cmd._alias.includes(name));
   };
 
   /**
@@ -1201,7 +1201,7 @@ class Command extends EventEmitter {
    *
    * @param {string} str
    * @param {Object} [argsDescription]
-   * @return {String|Command}
+   * @return {string|Command}
    * @api public
    */
 
@@ -1213,10 +1213,12 @@ class Command extends EventEmitter {
   };
 
   /**
-   * Set an alias for the command
+   * Set an alias for the command.
    *
-   * @param {string} alias
-   * @return {String|Command}
+   * You may call more than once to add multiple aliases, Only the first alias is shown in the auto-generated help.
+   *
+   * @param {string} [alias]
+   * @return {string|Command}
    * @api public
    */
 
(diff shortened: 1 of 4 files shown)


kodie:
This is awesome!

typings/index.d.ts
```diff
@@ -275,6 +275,8 @@ declare namespace commander {
     /**
      * Set an alias for the command.
      *
+     * You may call more than once to add multiple aliases, Only the first alias is shown in the auto-generated help.
```
victornpb, line 278:
```suggestion
     * You may call more than once to add multiple aliases, only the first alias is shown in the auto-generated help.
```
just catched a small typo, a capitalized word after a comma.

index.js
```diff
@@ -1231,7 +1233,25 @@ class Command extends EventEmitter {
...
+   *
+   * @param {string[]} [aliases]
+   * @return {string[]|Command}
+   * @api public
+   */
+
+  aliases(aliases) {
+    // Getter for the array of aliases is the main reason for having aliases() in addition to alias().
+    if (aliases === undefined) return this._aliases;
+
```
victornpb, line 1253: What do you think about this?
```suggestion
    if (arguments.length > 1) aliases = Array.from(arguments);
```

shadowspawn:
Changes: added `.aliases()` mainly to give public access to full list of aliases (e.g. for code adding command-line completion).

index.js
shadowspawn: Fair question. Using `arguments.length` means some smart editors list multiple arguments in the auto-detected signature. I actually removed all uses just recently! #1235

An alternative would to use separate arguments _instead_ of array. I prefer array in this case for the symmetry with the getter/setter signatures.

victornpb: Oh ok, I don't even know what would be the ts definition for that.
Array makes more sense from a typing pov, and varargs is just a preferred lazy typing person.

shadowspawn:
I'll leave PR in draft for a few more days for any other feedback. 

Thanks @kodie and @victornpb and @abetomo

shadowspawn:
No further feedback. Moving out of draft. 

I have deliberately not added to README for now, hopefully anyone who wants the feature will be able to find it easily enough, and the JSDoc and TSDoc cover the support for multiple aliases.

shadowspawn:
Included in Commander v5.1 which has been released.

https://github.com/tj/commander.js/releases/tag/v5.1.0

viceice:
This is blocking us from upgrading, since we use `aliasses` as an options.

Wha'ts the best compatible way to solve this.

shadowspawn:
Sorry for the clash @viceice 

See the section in the README about avoiding name clashes, and have a look at `storeOptionsAsProperties` and `passCommandToAction`:
- https://github.com/tj/commander.js#avoiding-option-name-clashes

viceice:
@shadowspawn Ok, thanks for the info. Works with the suggested changes.

Issue #571, opened 2016-09-18
Add support for variadic args on options


develmts:
Seems pretty obvious to me that is perfectly possible ( and even reasonable) to be able to send a unkown length list of values for an option, and agree that being that the scenario, it could only be applied to last option ( or, optionally, until we find some kind of "terminate" char combo ( "--" would be great, but probably mess with some command shells)

it's doable?

amonks:
https://github.com/tj/commander.js?#variadic-arguments

:)

develmts:
It specifies exactly "The last argument of a command can be variadic"
And If there isn't a command ?
I was asking for something like

`myapp --list-of-files-to-use file1 file2 file3`

amonks:
Could you use the `args` value?

```bash
myapp --option value file1 file2 file3 --optionb value file4

# //=> command.parse().args === ['file1', 'file2', 'file3', 'file4']
```

If you needed to, you could add a --use-a-list-of-files toggle, if it's set, you treat `args` as a list of files.

develmts:
That's what I'm using right now. but it would be nice  to do 

myapp --list-of-something item1 item 2 item 3

and commander returns something like  obj['list-of-something"] = ['item1','item2', 'item3']

lastmjs:
I'm seeing the need for this as well.

sntran:
> I was asking for something like
> 
> `myapp --list-of-files-to-use file1 file2 file3`

Another option is to use Coercion (as noted in examples):

```js
program.
  .option('-f, --files <file>', 'A path for file.', function(file, files) {
    files.push(file);
    return files;
  }, [])
```

When called with:

```shell
myapp --files file1 -f file2 --files file3`
```

Will give you an array of files:

```js
console.log(program.files);
// ['file1', 'file2', 'file3']
```

rousan:
I think support for variadic args on options should be added.

lastmjs:
I've gotten around this by taking in a comma separated list for an option. I take the option value and split the list to get my variadic arguments. I kind of disagree that it's not normal to want variadic values. I have a very legitimate need for them and I'm sure there are many more use cases.

ericnewton76:
A command separated list is a fine solution for your problem.

The real issue is when you have a variadic option, and no options between that variadic option and other arguments, they bleed into the variadic.  Using a comma separated list helps define for the parser that these arbitiary characters showing up are still part of my --mods option, and not input file arguments (for example)

shadowspawn:
[...]

PR #1250, opened 2020-04-19
Add support for variadic options

shadowspawn:
# Pull Request

## Problem

People would like to be able to use variadic options, an option which can take more than one value from the following arguments.

See #571

## Solution

1) Syntax:

```js
program
   .option('-m, --mods <mod...>'),
   .option('-o, --optional [opts...]')
```

Being cautious, this makes the change a breaking change in case people have put dots after a comma separated list for example.

2) If a value or values are specified then the option value is an array.

3) The first option argument is parsed in the same way as if the `...` was not specified.

4) Subsequent option arguments are processed as for optional options (`[value]`), so a leading `-` or `--` indicates an option and not a further argument. A lone `--` stops the option processing as normal.

5) The variadic behaviour triggers for stand-alone options with the first value in the next argument, and not for options where the option value is included in the argument. 

These do not trigger variadic processing:

```js
$ program -mvalue operand
$ program --optional=value operand
```

6) Repeated use of the option will accumulate the values.

These are equivalent:

```js
$ program --mods first second
$ program --mods first --mods second
```

Example:

```
$ transpile --include *.js --exclude *.test.js -- operand
```

## ChangeLog

- add support for variadic options (#1250)

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -44,6 +44,9 @@ const typescriptSettings = {
 
 module.exports = {
   plugins: ['jest'],
+  parserOptions: {
+    ecmaVersion: 8
+  },
   overrides: [
     javascriptSettings,
     typescriptSettings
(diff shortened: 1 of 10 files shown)


shadowspawn:
Using variadic options will cause some complications for caller to clarify their intent under some circumstances, in similar ways to options with optional values. So I don't want to promote it too much, it should only be used when compelling benefits.

I am thinking it will be mentioned briefly in the README with a longer explanation in a separate document.

shadowspawn:
Added README and ready for review.

Fixed a few spelling mistakes as well, so some extra files touched.

Issue #1249, opened 2020-04-18
Support options with only a short flag


shadowspawn:
Commander originally expected there would always be both short and long flags for each option. Support got added for long-only (#18).

Things almost work if you do try defining only a short flag, but by accident, and the value for `-a` ends up getting stored for `A`.

```
.option(`-d`, `debug`)
```

(The simple work-around is to add a long flag, so this may not be much of an issue in practice as has not been raised in issues here. It did get mentioned  as a minor annoyance in a good comparison article in 2018: https://pantas.net/node_argument_parsers/ )

shadowspawn:
Released with Commander v6.0.0 : https://github.com/tj/commander.js/releases/tag/v6.0.0

PR #1256, opened 2020-04-25
Support options with only a short flag

shadowspawn:
# Pull Request

## Problem

Users may wish to declare an option which has only a short flag and no long flag.

Defining just a short flag for an option appeared to almost work, but was not actually supported internally.

See: #1249

## Solution

Extend option flag parsing to allow short, or long, or both. Extend the support to the custom option flags for `.helpOption()` and `.version()` too.

## ChangeLog

- allow options to be added with just a short flag (#1256)

### DIFF
--- a/index.js
+++ b/index.js
@@ -1709,19 +1709,24 @@ function humanReadableArgName(arg) {
 }
 
 /**
- * Pulling the flags out of something like '-m,--mixed <value>'
+ * Parse the short and long flag out of something like '-m,--mixed <value>'
+ *
  * @api private
  */
 
 function _parseOptionFlags(flags) {
   let shortFlag;
   let longFlag;
+  // Use original very loose parsing to maintain backwards compatibility for now,
+  // which allowed for example unintended `-sw, --short-word` [sic].
   const flagParts = flags.split(/[ |,]+/);
-  // single character after '-'
-  if (flagParts.length > 0 && /^-[^-]$/.test(flagParts[0])) shortFlag = flagParts.shift();
-  // long flag after '--'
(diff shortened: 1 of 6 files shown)


shadowspawn:
Side note: I hit a merge conflict when updating the branch which prompted me to look at how `negate` was set in the option constructor. I tightened the test to match the README and expectation in other code that it is for a long option that `startsWith` `--no-`.

shadowspawn:
I created a `release/6.x` branch since this is a semver major change, although hopefully only breaking for people who were using lone short flags and working around the bugs.

I am thinking we might do a pre-release in June for v6, with short flags and variadic options.

index.js
```diff
@@ -1708,6 +1711,28 @@ function humanReadableArgName(arg) {
...
+ *
+ * @api private
+ */
+
+function _parseOptionFlags(flags) {
+  let shortFlag;
+  let longFlag;
+  // Use original very loose parsing to maintain backwards compatibility for now,
+  // which allowed for example unintended `-sw, --short-word` [sic].
+  const flagParts = flags.split(/[ |,]+/);
```
glensc, line 1725: this pattern seems buggy if the comment is still accurate, did you intend to say `/( |,)+/` there? or just `/[ ,]+/` to match space or comma, or really wanted to add a pipe `|` to the mix?

shadowspawn: The pipe is really in the mix. From the README:

> Each option can have a short flag (single character) and a long name, separated by a comma or space or vertical bar ('|').

The comment is about `-sw` which is neither a short flag nor a long name.

PR #1257, opened 2020-04-27
Added A Minified index.js For Current Release

SmintGaming:
# Pull Request


shadowspawn:
Why? What is the problem this is solving?

Note: if we did add a minified version, we would do it as part of our release process and not from a Pull Request.

PR #1264, opened 2020-05-13
Replace indexOf with includes

abetomo:
# Pull Request

## Problem

There was some code that was easier to understand using 'includes' than 'indexOf'.

## Solution

* Replace 'indexOf' with 'includes'
    * Changed the relevant if statement to an early return.

### DIFF
--- a/index.js
+++ b/index.js
@@ -20,10 +20,10 @@ class Option {
 
   constructor(flags, description) {
     this.flags = flags;
-    this.required = flags.indexOf('<') >= 0; // A value must be supplied when the option is specified.
-    this.optional = flags.indexOf('[') >= 0; // A value is optional when the option is specified.
+    this.required = flags.includes('<'); // A value must be supplied when the option is specified.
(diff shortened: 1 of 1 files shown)


index.js
```diff
@@ -1722,35 +1722,35 @@ function incrementNodeInspectorPort(args) {
...
-          // e.g. --inspect=1234
-          debugPort = match[3];
-        } else {
-          // e.g. --inspect=localhost
-          debugHost = match[3];
-        }
-      } else if ((match = arg.match(/^(--inspect(-brk|-port)?)=([^:]+):(\d+)$/)) !== null) {
-        // e.g. --inspect=localhost:1234
-        debugOption = match[1];
+    if (!arg.includes('--inspect')) {
```
shadowspawn, line 1725: This would be better as `startsWith` rather than `includes` to match the intent of `indexOf === 0`

abetomo: Thank you.
I'm sure you're right

Issue #1225, opened 2020-03-21
Collect: help


shadowspawn:
Collecting help issues that have come up multiple times.

Override or disable help option:

- <strike>#242 Make auto help optional</strike>
- <strike>#542 Hiding auto generated help options?</strike>
- #689 Don't Show options when there are no options
- <strike>#893 Override --help?</strike>
- <strike>#1215 Prevent/override automatic help generation</strike>
- #1370 Add method to get help option description

Struck out items included in Command v6.1.0

shadowspawn:
Customisation, especially prefix/banner:

- #46 Add Support For Title?
- #578 Option to prefix help with text
- #1023 Customize help message for a command

shadowspawn:
For customisation (e.g. prefix/banner), an implementation idea is:

Currently there is a single event which is emitted after the help is displayed, `--help`.

A possible approach for allowing more control might be to emit three events:

- `help:header`
- `help:body`
- `help:footer`

If there is a listener for `help:body` then it is called _instead_ of the normal help.

DevNico:
Would be great to have the ability to hide options from the help command. eg. I have three commands:
- server:setup
- server:start
- server:stop

Now those commands all have the option -v to enable verbose output. (I know that I could just make this a global option but in some cases I don't want to do that) In this case in the general help command the -v option is shown three times. It would be great if I could just add the option in a way so it only shows up in the help for that specific command. Is this already somehow possible?  If not I would really like this as a feature

shadowspawn:
Expanding the events to include the "group" of a command and its subcommands, which is effectively global if listened to on the root command:

1. `preGroupHelp`
2. `preHelp`
3. `help` | `groupHelp`
4. `postHelp`
5. `postGroupHelp`

shadowspawn:
A possible syntax for disabling the built-in help option could be:

```js
program.helpOption(false);
```

with some symmetry with:

```js
program.addHelpCommand(false);
```

shadowspawn:
I didn't like how the event listeners worked out as felt too complicated, and currently trying more direct api:

```js
program.addHelpText('after', 'See web site for further help');
```
```

shadowspawn:
Trying `.addHelpText()` which takes a position as first parameter, and a string to add or a function returning a string. Position is 'before' or 'after' to affect just this command, and 'beforeAll' or 'afterAll' to affect this command and all its subcommands. 

[...]

PR #1267, opened 2020-05-27
Add support for "local" and global help:header, help:body and help:footer

JCQuintas:
# Pull Request

## Problem

This adds the possibility to customize the **header, body and footer** of the help command. In theory this will also allow for the "suppression" of the help command if you supply an empty function to the `"help:body"` listener.

It solves part of the problems listed in https://github.com/tj/commander.js/issues/1225

## Solution

There are two solutions on this PR:

### Local or Scoped Help Listeners

For "local" or "scoped" help, we now have three listenable events. Which are called in the following order:

1. `help:header`
2. `help:body`
3. `help:footer`

When you add a listener to `help:body`, the default help text **won't** be written to the `stdout`.

All emitters send their scoped `this` to the listeners as a parameter.

```js
program
  .on('help:header', () => {
    console.log('Custom Header\n');
  })
  .on('help:body', (instace) => {
    console.log('------------');
    console.log(instace.helpInformation().replace(/\n$/, '\n------------\n'));
  })
  .on('help:footer', () => {
    console.log(`Commander - ${new Date().getFullYear()}`);
  })

// Custom Header
//
// ------------
// Usage: help-listeners [options]
//
// Options:
//   -V, --version  output the version number
//   -b, --bar      enable some bar
//   -h, --help     display help for command
// ------------
//
// Commander - 2020
```

### Global Help Listeners

> I would like more input on this, as I believe right now it might have side effects when using multiple commands? But I'm not sure and this would require testing.

The global listeners behave similarly to local listeners and use the same keys, but affect **all** the help calls. Root and children commands. Local commands will still be called and the order is:

1. **global** `help:header`
2. local `help:header`
3. **global** `help:body`
4. local `help:body`
5. local `help:footer`
6. **global** `help:footer`

### Typings, tests and documentation

- [x] TypeScript typings
- [ ] JSDoc documentation in code
- [x] tests
- [ ] README
- [x] examples/


shadowspawn:
1) Is this something you want to use yourself, or just coding it?

2) The global support is a challenge. It is harder than doing the local ones, but in practice is something people are quite likely to want to "brand" their whole program, so I think a good idea you have attempted it. Some initial thoughts.

The local listeners could be added manually everywhere with customisation via `.createCommand`, but that would be a barrier to usage, so more a work-around than a solution.

I don't think using a true global is the way to go for holding the listeners. The problem is if different packages in a program use Commander internally, they interact via any globals. Commander did start out as global only and have this problem, but we are now encouraging the use of `new Command()` to make things independent.

A possible approach would be differently name events which are emitted by walking the command tree at runtime, following the `parent` links. Say `help:global-header` or `global-help:header`. Still thinking about the terminology and the approach, so this is a suggestion and not a recommendation. The extra events would actually be in the tree hierarchy so not truely global, but would affect all subcommands.

shadowspawn:
(And thanks for the contribution!)

shadowspawn:
> With that said, I think your comments on the "local" listeners wasn't clear on your comment, do you think it can work the way I implemented?

Yes in general, that was what I had in mind, but l have a lot of suggestions on the details.

typings/index.d.ts
```diff
@@ -27,6 +27,15 @@ declare namespace commander {
     from: 'node' | 'electron' | 'user';
   }
 
+  // (string & {}) is sort of a hack so we able to show the string types autocompletion
+  // as well as allow any string to be used
+  export type Event =
+  | 'help:header'
+  | 'help:body'
+  | 'help:footer'
+  | (string & {})
```
shadowspawn, line 36: This does indeed seem to work in Visual Studio Code. Do you have any documentation or a source for what `(string & {})` is doing to the type? An intersection of a string and an object with no properties?

examples/help-listeners
```diff
@@ -0,0 +1,21 @@
...
+  .on('help:header', () => {
+    console.log('Custom Header\n');
+  })
+  .on('help:body', (instance) => {
+    console.log('------------');
+    console.log(instance.helpInformation().replace(/\n$/, '\n------------\n'));
+  })
+  .on('help:footer', () => {
+    console.log(`Commander - ${new Date().getFullYear()}`);
+  })
```
shadowspawn, line 18: The replace and `\n` are a bit confusing here. I eventually recalled that `helpInformation()` has a trailing return for legacy reasons because it is used with `process.stdout.write`. I suggest a simpler more symmetrical pattern like:

```
    .on('help:header', () => {
      console.log('Custom Header\n');
    })
    .on('help:body', (instance) => {
      console.log('------------');
      process.stdout.write(instance.helpInformation()); // has trailing eol
      console.log('------------');
    })
    .on('help:footer', () => {
      console.log(`\nCommander - ${new Date().getFullYear()}`);
   })
```

shadowspawn:
Event parameter. I had been thinking we would pass the `helpInformation` as a parameter to the `help:body` event, but I do like the symmetry with passing the command object to all the help listeners taking the same parameters. So preferring what you did at the moment.

index.js
```diff
@@ -128,6 +129,9 @@ class Command extends EventEmitter {
     this._helpCommandName = 'help';
     this._helpCommandnameAndArgs = 'help [command]';
     this._helpCommandDescription = 'display help for command';
+    this._helpHeader = 'help:header';
+    this._helpBody = 'help:body';
+    this._helpFooter = 'help:footer';
```
shadowspawn, line 134: These can be used as literals rather than having a property for them. You were probably copying the pattern for `_helpLongFlag` but it is a property because it can be changed by the user.

shadowspawn:
Possible suggestion for the "global" flavour of the events:
- `help:headerAll`
- `help:bodyAll`
- `help:footerAll`

index.js
```diff
@@ -1530,12 +1534,23 @@ class Command extends EventEmitter {
...
     }
-    process.stdout.write(cbOutput);
-    this.emit(this._helpLongFlag);
+    const globalBodyListeners = globalEventEmitter.listeners(this._helpBody);
+    const bodyListeners = this.listeners(this._helpBody);
+    if (bodyListeners.length === 0 && globalBodyListeners.length === 0) {
+      process.stdout.write(cbOutput);
+    } else {
+      globalEventEmitter.emit(this._helpBody, this);
+      this.emit(this._helpBody, this);
```
shadowspawn, line 1549: I think only one of the body helper should get called, the local one by preference, falling back to the global one.

index.js
```diff
@@ -1530,12 +1534,23 @@ class Command extends EventEmitter {
...
+    globalEventEmitter.emit(this._helpHeader, this);
+    this.emit(this._helpHeader, this);
     const cbOutput = cb(this.helpInformation());
     if (typeof cbOutput !== 'string' && !Buffer.isBuffer(cbOutput)) {
       throw new Error('outputHelp callback must return a string or a Buffer');
     }
-    process.stdout.write(cbOutput);
-    this.emit(this._helpLongFlag);
+    const globalBodyListeners = globalEventEmitter.listeners(this._helpBody);
+    const bodyListeners = this.listeners(this._helpBody);
```
shadowspawn, line 1544: I have been moving to a different pattern for testing listeners, since we don't call the listener directly we do not need the list. Just check the count.

```js
if (this.listenerCount('help:body')) {
   this.emit('help:body', this);
} else {
   process.stdout.write(cbOutput);
}
```

index.js
```diff
@@ -1530,12 +1534,23 @@ class Command extends EventEmitter {
         return passthru;
       };
     }
+    globalEventEmitter.emit(this._helpHeader, this);
+    this.emit(this._helpHeader, this);
     const cbOutput = cb(this.helpInformation());
```
shadowspawn, line 1539: In theory we should try and include or make available cb in the processing of the listener too, but that callback is a pattern I would like to deprecate. Instead, the processing of the callback can shift into the non-override body block since no need to call if not being displayed.

shadowspawn:
That is probably enough comments from me for today!

shadowspawn:
Prompted by this PR, I had a rethink of the event names to include the extra level, as I don't think the page layout terms work as well for that (i.e. header, body, footer). This is my favourite so far:

1.` preHelpGlobal`
2. `preHelp`
3. `help` | `helpGlobal`
4. `postHelp`
5. `postHelpGlobal`

typings/index.d.ts
JCQuintas: I went to hunt for some sort of explanation on this but couldn't find any. So I'll try to explain what I believe happens.

An empty interface `{}` is almost like an "any", except it doesn't allow the `undefined | null`, we can see it if you parse the code below on an editor:

```typescript
interface Empty {}

const n:Empty = 0
const s:Empty = 'a'
const a:Empty = []
const o:Empty = {}
const t:Empty = true
const f:Empty = false
const sy:Empty = Symbol()
const u:Empty = undefined //error
const nu:Empty = null //error
```

So the typescript intersection `(string & {})` would mean it has to be a string but can't be null or undefined.

When you declare `const foo: 'a' | 'b' | string = ''` typescript seems to coerce all the values into the `string` type. But when using `(string & {})`, it doesn't, because it is not expecting a "true" string.

The two drawbacks I can find or thing of are:

In some typeguard functions like the one below, it will throw an error, though it is easily bypass-able. 
```typescript
function foo(arg:  'a' | 'b' | (string & {})): 'a' | undefined {
  if (arg === 'a') {
    return arg; // Type '(string & {}) | "a"' is not assignable to type '"a"'
    //return arg as 'a'; // this works
  }
  return
}
```

The other drawback might be that this could be patched in the future, since I couldn't find any documentation or explanation for it. Though I suspect if patched, it would behave just like a `string` type instead. But who knows...

examples/help-listeners
JCQuintas: yeah, what you suggested was close to my first approach I believe, but I wanted the output to be neat. But indeed it is much better to read.

index.js
JCQuintas: Yeah I think it might be better too, I was in doubt of what to do there, thanks :)

index.js
JCQuintas: How is `this.litenerCount` supposed to know how many listeners there are? Or is this just an indirect call to `this.listeners('whatever').length`?

JCQuintas:
> Event parameter. I had been thinking we would pass the `helpInformation` (or even `cbOutput`) as a parameter to the `help:body` event, but I do like the symmetry of passing the command object to all the help listeners and using the same parameters. So preferring what you did at the moment.

I decided to pass "this" to the listener so you would be able to check all the commands/parsed arguments or anything really if you wanted to customize the help while using an arrow function. In theory the "local" listeners already pass "this" to `function () {}`, but it doesn't work with `() => {}`

JCQuintas:
> Prompted by this PR, I had a rethink of the event names to include the extra level, as I don't think the page layout terms work as well for that (i.e. header, body, footer). This is my favourite so far:
> 
> 1.` globalPreHelp`
> 2. `preHelp`
> 3. `help` | `globalHelp`
> 4. `postHelp`
> 5. `globalPostHelp`

I think it is better than header/footer, but I prefer the `:` as divider, though this is mostly a cosmetic change, so it is ultimately your choice. :P

1.` pre:help:global`
2. `pre:help`
3. `help` | `help:global`
4. `post:help`
5. `post:help:global`

typings/index.d.ts
JCQuintas: An alternative to this would be to overload the "on" function with

```typescript
    on(event: 'help:header' | 'help:body' | 'help:footer' | symbol, listener: (...args: any[]) => void): this;
    on(event: string, listener: (...args: any[]) => void): this;
```

index.js
shadowspawn: `listenerCount` is an EventEmitter method: https://nodejs.org/api/events.html#events_emitter_listenercount_eventname

JCQuintas: Oops. Yeah is missed that. 😅

JCQuintas:
I've been looking more through the code to see how to improve this PR. And I had not notice that we always have access to `this.parent`, so traversing the tree is easy. Now your suggestion of a listener tree is clearer to me. :)

So the idea would be to search for every `globalPreHelp` in the parent tree, but trigger only the one that is closest to root? This might be a bit misleading as they would not be trully global.

```js
- command('foo')
  - command('bar').addListener('globalHelp')
    - command('bas')
  - command('xoo')
    - command('yoo')
```

```bash
foo            # not
foo bar        # trigger
foo bar bas    # trigger
foo xoo        # not
foo xoo yoo    # not
```

What are the advantages of this vs always triggering the global listener only the "root" command?

shadowspawn:
> So the idea would be to search for every globalPreHelp in the parent tree, but trigger only the one that is closest to root? This might be a bit misleading as they would not be trully global.

All of the globalPreHelp would be called. Only the closest globalHelp would be called.

Yes, the listener is only "global" from that point in the tree, and not truly global. I am not entirely happy with it myself! I think the most command usage will be adding it to the root command, but I would prefer more accurate wording which is still easy to understand.

The qualifying word could go on either the "global" flavour or the "local" flavour.

Alternative suggestions or examples welcome!

> What are the advantages of this vs always triggering the global listener only the "root" command?

This does allow a custom behaviour for a subcommand and all its subcommands (say to add a description of shared options)

shadowspawn:
I spent some time wondering about other words and looking for other uses. I came across "group" to describe a command and its subcommands. So:

1. preGroupHelp
2. preHelp
3. help | groupHelp
4. postHelp
5. postGroupHelp

shadowspawn:
I'll mention the *other* thing I want to consider at the same time as part of adding new events. When the help is being displayed due to a usage error, it should be displayed on stderr rather than stdout. This has only come up once I am aware of in #997, so not much demand.

To make it easy for implementers I am thinking probably want easy access to both a `console.log()` equivalent and a `process.stdout.write()` equivalent. Interested if `write` is actually important to implementers, but as we know from this PR the way `helpInformation` is currently constructed with a trailing `\n` makes only offering a `log` a (small) complication.

Since most people will only want one or the other, I was thinking they could go in a "context" object. I have looked for the stdout/stderr problem being solved elsewhere and not found examples. I might be missing a simpler solution!

shadowspawn:
Trying out the context pattern idea, parameter could look like:

```
const helpContext = {
  'command': program,
  'write': (...args) => process.stdout.write(...args),
  'log': (...args) => console.log(...args),
}
```

And be used like: 

```
  .on('help', (context) => {
    context.write(context.command.helpInformation());
  })
```

JCQuintas:
Why would someone use `context.write` if they have access to their own loggers? Or is this the internal implementation?

shadowspawn:
`context.write` might call `process.stdout.write` or `process.stderr.write` depending on why the help is being displayed. This is especially relevant when someone is letting Commander display the normal help and just adding a banner or footer, so that everything goes to the same stream.

shadowspawn:
Ok if I work on this too @JCQuintas , or would you prefer I leave it to you for now?

JCQuintas:
Sure, go ahead. :)

shadowspawn:
Experimenting on branch: https://github.com/shadowspawn/commander.js/tree/feature/on-help

shadowspawn:
I'll close this in favour of #1296, with especially the global listener idea here inspiring the current form of that PR.

I am intrigued by the TypeScript trick to get Intellisense suggesting event strings in Visual Studio Code, but without some backing references it feels a bit too hacky.

Issue #1237, opened 2020-04-02
Adding `--no-emit` option results in error


jonhue:
Adding the option `.option('--no-emit', 'description')` to a command results in the following error when running `bin cmd -h`:

```
/mnt/c/code/tony-lang/cli/node_modules/commander/index.js:1495
    this.emit(this._helpLongFlag);
         ^

TypeError: this.emit is not a function
    at Command.outputHelp (/mnt/c/code/tony-lang/cli/node_modules/commander/index.js:1495:10)
    at outputHelpIfRequested (/mnt/c/code/tony-lang/cli/node_modules/commander/index.js:1645:9)
    at Command._parseCommand (/mnt/c/code/tony-lang/cli/node_modules/commander/index.js:893:7)
    at Command._dispatchSubcommand (/mnt/c/code/tony-lang/cli/node_modules/commander/index.js:860:18)
    at Command._parseCommand (/mnt/c/code/tony-lang/cli/node_modules/commander/index.js:877:12)
    at Command.parse (/mnt/c/code/tony-lang/cli/node_modules/commander/index.js:712:10)
    at Object.<anonymous> (/mnt/c/code/tony-lang/cli/dist/index.js:26:20)
    at Module._compile (internal/modules/cjs/loader.js:1151:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:1171:10)
    at Module.load (internal/modules/cjs/loader.js:1000:32)
```

Is this expected?

shadowspawn:
The problem is due to a clash of your option name with the properties of the Command object, in this case  the `emit` method from `EventListener`. The value for the `--no-emit` flag is stored as the `emit` property and overwrites the expected method. 

To resolve the problem, you may opt-in to a new way of storing the option values using `.storeOptionsAsProperties(false)` which was added specifically to avoid this sort of problem.

See https://github.com/tj/commander.js#avoiding-option-name-clashes

jonhue:
Thanks, that makes sense! 🙂

shadowspawn:
Opened a PR to add a warning for option name clashes: #1275

Issue #1226, opened 2020-03-22
TypeError: this.name is not a function


RoXioTD:
Hey everyone,

i currently updated to commanderJs 5.0.0 and now some of my command are not working fine when providing the option `name`.

Given is the following command with the option name:
`program.command('test').description('Logs something').option('-n, --name <value>').action(args => console.log(args.name))`

If i run `test` the command works as expected.
But if i run `test --name Foo` in my console Commander throws the following error:
`TypeError: this.name is not a function`
    `at Command._parseCommand (....\node_modules\commander\index.js:910:37)`
    `at Command._dispatchSubcommand (....\node_modules\commander\index.js:860:18)`
    `at Command._parseCommand (....\node_modules\commander\index.js:877:12)`
    `at Command.parse (....\node_modules\commander\index.js:712:10)`

I already tried this [Avoiding option name clashes](https://github.com/tj/commander.js#avoiding-option-name-clashes)

Anyone has any suggestions

shadowspawn:
`name` is a method on Command, so you will hit issues with an option `--name` when using default behaviour of storing the value as a property. It seems to work ok by adopting the new style processing for me?

```
const { Command } = require('commander');
const program = new Command();

program
  .storeOptionsAsProperties(false)
  .passCommandToAction(false)
  .name('my-name')
  .option('--name <value>', 'Enter name')
  .action((options) => {
    console.log(`Name option is ${options.name}`);
  });

program.parse();
```

```
$ node index.js           
Name option is undefined
$ node index.js --name foo
Name option is foo
$ node . --help
Usage: my-name [options]

Options:
  --name <value>  Enter name
  -h, --help      display help for command
```

RoXioTD:
Your solution fixed the issue. Thanks for your answer.

counterbeing:
Just ran into this one as well. Seems like it would be nice to throw an error if someone specifies a `--name` option. I can imagine this one being run into a fair bit, name is probably a pretty popular choice.

shadowspawn:
Opened a PR to add a warning for option name clashes: #1275

PR #1275, opened 2020-06-07
Throw for likely option name problems

shadowspawn:
# Pull Request

One day we might change the default behaviour to make options safer! In the meantime, we can try harder to warn about possible problems.

## Problem

While the default behaviour is to store options values as properties, people may hit issues with clashes with existing properties:

Collected issues: #933
Recent `--name`: #1226
Recent `--no-emit`: #1237

## Solution

Throw an error if the user has not called `storeOptionsAsProperties` and the property matching the option is already defined. 

Takes a reasonable amount of code to avoid incorrect warnings, but hopefully helps programmer resolve the problem.

## ChangeLog

- throw an error with advice on how to resolve if there might be a clash between option name and a Command property

## Example

```js
program
  .option('-E,--no-emit')
```

```shell
...
      throw new Error(`option '${option.flags}' clashes with existing property '${option.attributeName()}' on Command
      ^

Error: option '--no-emit' clashes with existing property 'emit' on Command
- call storeOptionsAsProperties(false) to store option values safely,
- or call storeOptionsAsProperties(true) to suppress this check,
- or change option name
...
```

### DIFF
--- a/index.js
+++ b/index.js
@@ -441,8 +441,8 @@ class Command extends EventEmitter {
     const name = option.attributeName();
     option.mandatory = !!config.mandatory;
 
-    if (!this._storeOptionsAsPropertiesCalled && this[oname] !== undefined && !option.negate) {
-      throw new Error(`option name '${oname}' clashes with existing property on Command
+    if (!this._storeOptionsAsPropertiesCalled && this[name] !== undefined && !option.negate) {
+      throw new Error(`option name '${name}' clashes with existing property on Command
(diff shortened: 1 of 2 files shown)


shadowspawn:
Adding as draft for a few days in case I think of more cases to cover.

shadowspawn:
Moved out of draft, ready for review.

shadowspawn:
Thanks @abetomo

Issue #1033, opened 2019-08-29
Conflict with global options and command specific options


dennisdupont:
When .command() is called with options that are the same as the global options (aka top level options), then command specific option is parsed as if it were the top level option instead and the command specific option value is never defined.
```
commander
    .version(version)
    .description('tool')
    .option('-f, --format <format>', 'output format')

commander
    .command('fetch [version]')
    .alias('f')
    .description('fetch things')
    .option("-f, --file <destFile>", "file name to save output")
    .option("-d, --dir <destDir>", "destination directory for output")
    .action(function(version, options) {
    	// looking for something in destFile fails *always*
    	if (options.destFile) {
    		...
    	}
    }
```

If the above is invoked (assume linked or globally installed) as 
```> app f -f someFile```
then I would expect options.destFile to have the value "someFile", but it is always null and instead format has the value "someFile".
If invoked as 
```> app -f someFormat f -f someFile```
then I would expect options.destFile to have the value "someFile" and format to have the value "someFormat", but the behavior is the same again.

_NOTE: does not matter whether the alias or the full command is used_

This is due to code in parseOptions that tries to match all options instead of stopping at a command and allowing the child object to parse its own options.  

There is an easy fix, although it may cause side effect due to my ignorance of the design and other features. That is in parseOptions, once a arg is not matched to an option *and* does not look like an option, simply stop processing and add the remainder to the unknown list:
```


    // option is defined
    if (option) {
    	...
      continue;
    }

    // looks like an option
    if (arg.length > 1 && arg[0] === '-') {
    	...
      continue;
    }

    // arg
    args.push(arg);

    // ============= Uninformed neophyte change starts here  ===================
    // not an option and doesn't look like one, must be a subcommand 
    // 		- stop collecting options at this level
    for (var u = i + 1; u < len; ++u) {
      unknownOptions.push(argv[u]);
    }
    break;
    // ============= Uninformed neophyte change ends here  ===================

  }
```

**Comment by shadowspawn** [comment]
Global options are not positional, and are processed before checking the command options. This could be thought of as similar to options coming before or after arguments.

So these are effectively the same:
```
app --format=short fetch
app fetch --format=short
```
and
```
app fetch --dir=build 1.2.3
app fetch build --dir=1.2.3
```

There is not currently support for positional global options.

shadowspawn:
An answer was provided, and no further activity in a month. Closing this as resolved.

Feel free to open a new issue if it comes up again, with new information and renewed interest.

[...]

Issue #886, opened 2018-11-07
How to run multiple commands together


pldg:
I'm trying to build a CLI that need to run multiple commands together, some commands even have specific options. I also need to store the value passed on each commands in a variable for later use.

**cli/pm.js**

```js
#!/usr/bin/env node

const program = require('commander')
const pkg = require('../package.json')
const numbers= require('./pm-numbers.js')
const say = require('./pm-say.js')

program
  .version(pkg.version)
  .option('--flag')

program.parse(process.argv)

console.log(program.flag, numbers, say)
```

**cli/pm-numbers.js**

```js
const program = require('commander')

let numbers;

program
  .command('numbers <items...>')
  .action(items => numbers = items.map(Number))

program.parse(process.argv)

module.exports = numbers
```

**cli/pm-say.js**

```js
const program = require('commander')

let say;

program
  .command('say <word>')
  .option('--font <type>')
  .action((word, options) => {
    return say = {
      say: word,
      font: options.font
    }
  })

program.parse(process.argv)

module.exports = say
```

Test:

- If I run a single command like `node cli/pm.js say hello` or `node cli/pm.js numbers 1 2 3` everything works fine
- `node cli/pm.js --flag` print out `error: unknown option '--flag'`
- If I run multiple commands like `node cli/pm.js numbers 1 2 3 say hello` log out `numbers === [ 1, 2, 3, NaN, NaN ]` and `say === undefined`
- `node cli-test/pm.js numbers 1 2 3 say hello --font italic` print out `error: unknown option '--font'`

What am I doing wrong? Anyone can help?

JohnRGee:
`commander` does not expect or support specifying multiple commands on the same command line, and what you are doing here is processing the whole of `process.argv` three times. None of the three programs know about each other, and have no way of parsing out the _other_ commands and options. 

So when you run 

``` bash
node cli/pm.js --flag
```
 
both `numbers.js` and `say.js` would complain about an unknown option `--flag`.

When you run 

``` bash
node cli/pm.js numbers 1 2 3 say hello
```

`numbers.js` treats all the arguments as numbers (including `say` and `hello`), and `say.js` does nothing because the (first) command is not `say`.

I am not sure if other parsing packages support multiple commands on single line, but it is unusual for a command line tool to do this. You might be able to still use `commander` for some of the work by preprocessing `process.argv` yourself and splitting the args up into multiple calls to your combined parser. I am imagining a command line with an extra argument to split the commands like:

``` bash
[...]

PR #1276, opened 2020-06-07
Implement lazily created sub commands.

mshima:
# Pull Request

## Problem

Some programs implements plugin like sub commands.
The plugin options isn't known at the time the root commander is created.

## Solution

- Implement lazily created sub commands.
- Promisifying some parts of current private api is required.

### DIFF
--- a/index.js
+++ b/index.js
@@ -187,6 +187,7 @@ class Command extends EventEmitter {
     cmd._passCommandToAction = this._passCommandToAction;
 
     cmd._executableFile = opts.executableFile || null; // Custom name for executable file, set missing to null to match constructor
+    cmd._actionHandler = opts.action || null; // A handler for sub commands.
     this.commands.push(cmd);
     cmd._parseExpectedArgs(args);
     cmd.parent = this;
@@ -228,8 +229,8 @@ class Command extends EventEmitter {
     // Fail fast and detect when adding rather than later when parsing.
     function checkExplicitNames(commandArray) {
       commandArray.forEach((cmd) => {
-        if (cmd._executableHandler && !cmd._executableFile) {
-          throw new Error(`Must specify executableFile for deeply nested executable: ${cmd.name()}`);
+        if (cmd._executableHandler && !cmd._executableFile && !cmd._actionHandler) {
+          throw new Error(`Must specify executableFile or action for deeply nested executable: ${cmd.name()}`);
         }
         checkExplicitNames(cmd.commands);
       });
@@ -668,6 +669,11 @@ class Command extends EventEmitter {
    */
 
   parse(argv, parseOptions) {
+    this._parseProgram(argv, parseOptions);
+    return this;
+  }
+
+  _parseProgram(argv, parseOptions) {
     if (argv !== undefined && !Array.isArray(argv)) {
       throw new Error('first parameter to parse must be array or undefined');
     }
@@ -714,9 +720,7 @@ class Command extends EventEmitter {
     this._name = this._name || (this._scriptPath && path.basename(this._scriptPath, path.extname(this._scriptPath)));
 
     // Let's go!
-    this._parseCommand([], userArgs);
-
-    return this;
+    return this.parseCommand([], userArgs);
   };
 
   /**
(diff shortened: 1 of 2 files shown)


shadowspawn:
Interesting, thanks. I have quite a few details I want to work through, but think I get the idea.

> Promisifying some parts of current private api is required.

In case I am missing something, I think lazy-loading and promise support are two different things? (i.e. might be desirable but not required?)

mshima:
It’s required for my use case now.
Since I am using network and it doesn’t provides a sync api alternative.

mshima:
@shadowspawn let me know if this feature is desired.

See the implementation:
https://github.com/mshima/environment/blob/cli/cli/index.js

It currently uses @npmcli/arborist (npm 7 beta) for testing purpose.

shadowspawn:
Thanks for the example.

shadowspawn:
> @shadowspawn let me know if this feature is desired.

Yes and no. I am interested in exploring the details further, but won't be accepting the PR in its current form if that is what you are asking.

Due to this PR I have been doing some research and thinking on lazy loading, and promisifying, and how best to add new functionality to command creation. I will post more.

For my own reference on the command creation, took me a while to find a previous PR where a similar addition was being made to the command "options" like `executableFile`. #1185 suggested adding `construct`, but it shipped as a new factory method instead in #1191.

mshima:
Probably its possible to implement this feature with multiple independent commands, like you recommended at https://github.com/tj/commander.js/issues/886, but I always like to try the upstream approach first.

At https://github.com/mshima/commander.js/tree/multiple I have a implementation (poor) that allows to run multiple commands and once using `--` (can be make configurable).
One side effect is that https://github.com/tj/commander.js/issues/1033 can have a workaround:
```
myCli --globalAndLocalOption -- command --globalAndLocalOption
```

mshima:
The end result is:
```
yo --local-only -- generator1 --generator1-config -- generator2 --generator-2-config
```

- Create a environment with localOnly option
- Install and register generator1 to run with generator1Config option
- Install and register generator2 to run with generator2Config option
- When program.parseAsync promise resolves, run the environment.

shadowspawn:
Your example is a bit different than I expected. It isn't just lazy-loading, but is loading dynamic commands based on the arguments to the declared command (i.e. declared `run` which loads a command based on argument `namespace`):

```
  .command('run <namespace>', 'run the generator', {isDefault: true, subCommandBuilder: subCommandHandler})
  .command('help <namespace>', 'shows the generator help', {subCommandBuilder: subCommandHandler})
```

And you mentioned running multiple commands too in the following comments. 

These are just words so might not be clear, but are you interested in "dynamic" loading of commands, or "lazy" loading of commands? (I think your proposed implementation supports both, but some of the changes I was going to explore do not!)

mshima:
IMO the created command isn't dynamic, once it's created the definitions are all there.
The `run` command is.
Probably the feature request definition is `Add support to dynamic commands` which allows to create lazily created sub commands.

index.js
```diff
@@ -850,45 +855,58 @@ class Command extends EventEmitter {
...
     if (subCommand._executableHandler) {
-      this._executeSubCommand(subCommand, operands.concat(unknown));
-    } else {
-      subCommand._parseCommand(operands, unknown);
+      if (subCommand._actionHandler) {
+        const actionPromise = subCommand._actionHandler({ command: this, operands, unknown });
+        if (actionPromise instanceof Promise) {
+          return actionPromise;
+        }
+        return Promise.resolve(actionPromise);
```
shadowspawn, line 875: It would probably be more consistent with JavaScript conventions to check for "thenable" than for a Promise as such, but in fact can just drop the test entirely as `Promise.resolve` handles Promises too.

index.js
```diff
@@ -736,13 +740,14 @@ class Command extends EventEmitter {
...
    * @param {string} parseOptions.from - where the args are from: 'node', 'user', 'electron'
-   * @return {Promise}
+   * @return {Promise} promise `this`
    * @api public
    */
 
   parseAsync(argv, parseOptions) {
-    this.parse(argv, parseOptions);
-    return Promise.all(this._actionResults).then(() => this);
+    return this._parseProgram(argv, parseOptions).then(() => {
```
shadowspawn, line 748: Instead of scattering `Promise.resolve` through the code in places that do not use Promises, could call `Promise.resolve` here instead to minimise the code churn.

shadowspawn:
It is encouraging that lazy loading and dynamic loading using Promises did not require too much change. However, this PR is doing too many things that each need to be tackled more carefully for easy adoption and the longterm maintenance of the program.

The recursive `run` test is pretty cool as an example of this implementation.

A possible insight for future evolution of Commander API is that a key difference between an ordinary command and one implemented using a stand-alone executable is that the latter is essentially lazy loaded. The executable command name and parameters and description are all available so that the (parent) help can be generated without loading the command itself. This is also relevant for dynamic and lazy loading.

I wonder whether the dynamic loading could be implemented with current code with an async action handler.

PR #1289, opened 2020-06-24
Optimize Chinese readme typesetting

mastermay:
add whitespace between Chinese and half-width characters (alphabetical letters, numerical digits and symbols) for good readability.


shadowspawn:
I will merge this in so I can open an issue about adding the v6 changes.

Issue #1272, opened 2020-06-01
Question: Is it possible to have a splash screen?


Thanzex:
Hi there! I've been trying to make a nice ASCII art splash screen for my CLI. I'd like it to be displayed when the program is called without arguments or commands and to be displayed before the help information.

I've tried with a hidden default command, but then it's name shows up in the usage and the help is relative to the command.
Is there any clean way to do it?

shadowspawn:
It is easy to display a banner for no arguments by checking arguments before calling Commander, but not currently clean support for including the banner in normal help. Currently being discussed.

Related: 
- https://github.com/tj/commander.js/issues/1225#issuecomment-602005009
- https://github.com/tj/commander.js/pull/1267

shadowspawn:
To give an example for my comment:

> It is easy to display a banner for no arguments by checking arguments before calling Commander

```
const { program } = require('commander');

// Configure program
program
   .option('-d,--debug');

// Test for no arguments before calling parse
if (process.argv.length <= 2) {
    console.log('ASCII ART');
    program.help(); // exits
}

program.parse(process.argv);
```

shadowspawn:
Closing this as answered and/or covered by #1225

shadowspawn:
See `.addHelpText()` in Commander 7.0.0.

Issue #1234, opened 2020-03-30
usage and description display order


tdjastrzebski:
Although it makes perfect sense to display **usage** and then **description** on command help, it may make more sense to display **description** first and then **usage** on program help. Otherwise default help output looks like:
```
Usage: MyProgram MyCommand

MyProgram is a super demo program (description)

Options:
--version
--myoption, -m   demo option
```
Desired order:
```
MyProgram is a super demo program (description)

Usage: MyProgram MyCommand

Options:
--version
--myoption, -m   demo option
```

shadowspawn:
I do not think we are likely to make this change on its own. I have collected help related issues into #1225, and added this. One way of achieving what you suggest might be with a "prefix/banner" which has come up a few times.

shadowspawn:
No further activity in a month. Added to #1225. Closing as (partially) covered by that collection.

Thank you  for your contributions.

Issue #1023, opened 2019-08-18
Customize help message for a command


RonWang:
Our goal: customize the help message for the root command. 

Existing solution won't help, because we need to append the message on top of auto-gen help, and also modify the auto-gen help a little bit.

Context for our cmd app: below is the root cmd, and we have a `sub-cmd` under the root as well as a `git-style-sub-command`. 

```
commander
    .description('')
    .command('sub-cmd')
    .command('git-style-sub-cmd')
    .version(require('../package.json').version)
    .option('-h, --help', 'output help message', () => {
        console.log('custom help msg');
    })
    .parse(process.argv);
```

In the example I tried using option('-h, --help'), the console.log works for me with my demanded help output. But when I try sub-cmd and git-style-sub-cmd, the help message for the root command is always coming along when I type "cmd sub-cmd --help" and "cmd git-style-sub-cmd --help". 

I noticed #242 , where @shadowspawn  mentioned it won't be supported. Reopen it as it'd be an important feature for us.

RonWang:
lol great post! #1111111111(2)

shadowspawn:
I'll try and explain what is happening with what you tried.

```
    .option('-h, --help', 'output help message', () => {
        console.log('custom help msg');
    })
```

This adds an option which effectively overrides the built-in help since it uses the same flags. The callback is used for  "Custom option processing" and covered in the README.

Flags are treated as "global" and these three calls get processed much the same way and call your action handler:

```
my-command --help
my-command --help sub-cmd
my-command sub-cmd --help
```

This approach would let you customise the help for your main command, but you do miss out on the built-in support for help on the subcommands.

shadowspawn:
Related issue, prefix help with some text: #578

madanalogy:
Hello, just sharing a workaround I found. You can make use of ```.helpOption(flags, description)``` available in v3.0 of commander. Instead of using a string for description, you can call a function that returns a string and have your banner printed in that function (along with any information you want to prefix to the help message).

My implementation:
```
function handleHelp(txt) {
  printHeader();
  return txt;
}
program
  .helpOption('-h, --help', handleHelp('output usage information'));
```

For full implementation including context, see https://github.com/MarkBind/markbind/pull/946

shadowspawn:
@madanalogy 
Won't the header get printed every time you run the command, whether or not you show the help?

madanalogy:
> @madanalogy
> Won't the header get printed every time you run the command, whether or not you show the help?

[...]

Issue #578, opened 2016-09-24
Option to prefix help with text


jez9999:
Hi,

I would like an option to prefix the --help output with my program's banner along the top, eg.

```
=====
MyProgram version 1.2.3
=====

  Usage: myprogram [options]
  ... etc...
```

There doesn't seem to be a way to do this right now (at least not just for when the help text gets displayed) - could it be added?

b4dnewz:
this will be a super feature, so you can prefix your command line tools with (for example) ASCII art like the most of cli tools out there.

kodie:
Crazy that this hasn't been implemented yet

shadowspawn:
This issue has not had any activity in over six months. It isn't likely to get acted on due to this report.

Feel free to open a new issue if it comes up again, with new information and renewed interest. 

This request did get some 👍 so I appreciate there may be some interest. We already have a few help customisation options, but not something that does a "header". 
 
Thank you  for your contributions.

shadowspawn:
This came up again, and I have an idea for a possible implementation:
https://github.com/tj/commander.js/issues/1023#issuecomment-574451728

PR #1296, opened 2020-07-04
Expand help customisation with .addHelpText

shadowspawn:
# Pull Request

## Problem

1) can not add a splash screen/title/banner to the help
2) can not add a global epilog
3) event listener is a little clunky and unfamiliar to use
4) help is displayed on `stdout` even when being displayed as an "error"

See #1225 for collected custom help issues, and #997 for error output.

- beforeAll: #46 #78 #578 #1272
- before: #1023 #1234

## Solution

Add `.addHelpText()` which takes a position as first parameter, and a string to add or a function returning a string. Position is 'before' or 'after' to affect just this command, and 'beforeAll' or 'afterAll' to affect this command and all its subcommands.

For example:

```js
program.addHelpText('afterAll', '/nSee the web site for more information');
```

Also:
- `.help()` and `.outputHelp()` can be passed `{ error: true }` to write to stderr.
- Deprecate passing callback to `.help()` and `.outputHelp()`
- Deprecate `.on('--help')`

## ChangeLog

- use `.addHelpText()` to add text before or after the built-in help, for just current command or also for all subcommands (#1296)
- deprecate callback parameter to `.help()` and `.outputHelp()` (removed from README)
- deprecate `.on('--help')` (removed from README)


westhom:
Hi there, one thing I badly wanted from commander was the ability to group commands and options in help output, for readability. I implemented it as a quick hack here: [https://github.com/westhom/commander.js/compare/c5a5e7b7..86a1980b](https://github.com/westhom/commander.js/compare/c5a5e7b7..86a1980b).

For commands, you can call `program.command(...).helpGroup(N)`, where N is a group index the command belongs to. For options it wasn't as clean, set through a new param: `program.option('-k', 'set K', undefined, undefined, N)`. The help generator methods sort commands/options by their group before formatting the line. By default everything belongs to group 0 and if no other groups exist the help output is normal. Example:

```
Usage: help-groups [options] [command]

Options:
  -a                 set A
  -b                 set B
  -c                 set C
  
  -q                 set Q
  -r                 set R
  -s                 set S
  
  -h, --help         display help for command

Commands:
  login              log in
  logout             log out
  
  read <process>     read it
  update <process>   update it
  delete <process>   delete it
  refresh <process>  refresh it
  
  publish            publisher
  unpublish          unpublisher
  
  help [command]     display help for command
```

I was going to open a PR for this but it looks like help is being revamped. It could be nice to have a grouping option like that.

shadowspawn:
Hi @westhom , that is not covered by this PR and not the sort of issue I had collected in #1225 either. I recommend you open a new issue for discussion (with the info from your comment).

westhom:
> I recommend you open a new issue for discussion (with the info from your comment).

OK, will do!

shadowspawn:
To be safe, the change to use `stderr` for displaying the help in response to an error makes it a `semver` major change.

The other changes are backwards compatible, as the `--help` event and the `.help(cb)` callback are deprecated but not removed.

shadowspawn:
This is code complete. Leaving as Draft for now as planning to land this in v7, so leave PR open for visibility in the meantime.

shadowspawn:
Ready for review. 

My main concern is I think the event listeners are a little clumsy to use, and not familiar to people. I wonder whether adding a convenience routine which takes a string would make it easier to use for many simple cases. e.g.

```
myCommand.onHelp('postHelp', '\nSee the web site for more information');
```

shadowspawn:
Putting back into "Work in Progress" while I explore adding a more convenient routine to use. Maybe `addHelp`:

```js
myCommand.addHelp('postHelp', '\nSee the web site for more information');
```

shadowspawn:
Working on `addHelpText` and pleased with how it is working out this time, getting away from the raw event listener. Now as simple or simpler to use than original `.on(--help)` but does more. 

I'll rework the Pull Request description when I get further and delete some of these comments.

shadowspawn:
I am happier with how this is looking now with an explicit routine which is more readable than an event listener, and leaving the ability to override the built-in help out of this routine.

shadowspawn:
I deprecated the callbacks to `.help()` and `.outputHelp()` as this PR originally included being able to override the built-in help too. That is no longer included, but I would still like to call the callbacks deprecated so can remove them in future!

abetomo:
Could you merge `develop` into `release/7.x`?
This includes the 6.1.0 differences.

shadowspawn:
Oops, will do. I was merging 6.1 into the PR so the diff against develop looked better, missed doing 7.x too.

shadowspawn:
I have merged `develop` into `release/7.x`.

abetomo (approved):
LGTM :+1:

crystalfp:
Only a (minor) issue in the new version:
1. `.addHelpText("after", "\n")` adds **two** blank lines after help.
2. `.addHelpText("after", "")    ` adds **nothing** after help.
3. `.addHelpText("after", " ")  ` adds **one** blank line after help.

A little confusing. If case 1 should behave as `console.log()` then case 2 should add one blank line after help. Instead, if the function behaves like `write()` then case 1 should add one blank line, in my opinion.

Besides this minor issue, the new version is excellent. Thanks!
mario

shadowspawn:
Case 2 is a special case. If there is an empty string then nothing is output. This is support a function generating the string deciding that there is nothing to be displayed.

crystalfp:
OK, understand. Maybe adding a line to documentation stating this could be useful. Something like this at the end of the *Custom help* paragraph: “If the string or the output of the function is an empty string, nothing is output, otherwise the string is output followed by a newline”.

PR #1302, opened 2020-07-17
fix: assignment of negate property

### LINKED ISSUE #1301: BUG: defining an option which includes `no-` makes the option true by default
**Current Behavior -**
If we define an option whose name contains `no-` for e.g. `module-no-parse`, it makes the option `true` by default.

**Expected Behavior -**
It is supposed to do this only for flags with leading `no-` as mentioned in docs.  https://github.com/tj/commander.js/#other-option-types-negatable-boolean-and-flagvalue
> You can specify a boolean option long name with a leading no- to set the option value to false when used. Defined alone this also makes the option true by default.

**Steps to reproduce -**
Define an option whose name includes `no-` e.g. `module-no-parse`. 
Check its default value in the parser.
Screenshot at 2020-07-16 20-28-38

**Addition Info**
Bug found via https://github.com/webpack/webpack-cli/issues/1679

snitin315:
# Pull Request
Resolve #1301 

## Problem
We were checking for indexOf `-no-` for `negate: true` instead of `--no-`

## Solution
check index of `--no-`

### DIFF
--- a/index.js
+++ b/index.js
@@ -23,7 +23,7 @@ class Option {
     this.required = flags.indexOf('<') >= 0; // A value must be supplied when the option is specified.
     this.optional = flags.indexOf('[') >= 0; // A value is optional when the option is specified.
     this.mandatory = false; // The option must have a value after parsing, which usually means it must be specified on command line.
-    this.negate = flags.indexOf('-no-') !== -1;
+    this.negate = flags.indexOf('--no-') !== -1;
     const flagParts = flags.split(/[ ,|]+/);
     if (flagParts.length > 1 && !/^[[<]/.test(flagParts[1])) this.short = flagParts.shift();
     this.long = flagParts.shift();
--- a/tests/options.opts.test.js
+++ b/tests/options.opts.test.js
@@ -70,4 +70,13 @@ describe.each([true, false])('storeOptionsAsProperties is %s', (storeOptionsAsPr
     program.parse(['node', 'test']);
     expect(program.opts()).toEqual({ pepper: pepperDefault });
   });
+
+  test('when option with name including no- is specified then it should not have default value as true', () => {
+    const program = new commander.Command();
+    program.storeOptionsAsProperties(storeOptionsAsProperties);
+    program
+      .option('--module-no-parse <value>', 'test description');
+    program.parse(['node', 'test']);
+    expect(program.opts()).toEqual({});
+  });
 });


index.js
```diff
@@ -23,7 +23,7 @@ class Option {
     this.required = flags.indexOf('<') >= 0; // A value must be supplied when the option is specified.
     this.optional = flags.indexOf('[') >= 0; // A value is optional when the option is specified.
     this.mandatory = false; // The option must have a value after parsing, which usually means it must be specified on command line.
-    this.negate = flags.indexOf('-no-') !== -1;
+    this.negate = flags.indexOf('--no-') !== -1;
```
shadowspawn, line 26: This should test for it being at the start of the string. The other places in the code use a regular expression, but checking the index would be fine here I think.

snitin315: How about something like this?

```suggestion
    this.negate = flags.startsWith('--no-');
```

shadowspawn: Even better! (We moved away from using index to using routines in #1264 which is on the develop branch.)

tests/options.opts.test.js
```diff
@@ -70,4 +70,13 @@ describe.each([true, false])('storeOptionsAsProperties is %s', (storeOptionsAsPr
...
     expect(program.opts()).toEqual({ pepper: pepperDefault });
   });
+
+  test('when option with name including no- is specified then it should not have default value as true', () => {
+    const program = new commander.Command();
+    program.storeOptionsAsProperties(storeOptionsAsProperties);
+    program
+      .option('--module-no-parse <value>', 'test description');
+    program.parse(['node', 'test']);
+    expect(program.opts()).toEqual({});
```
shadowspawn, line 80: Thanks for adding a test.

Testing `.opts()` being the empty object is a little fragile as `.opts()` has some subtle differences between `storeOptionsAsProperties` true/false, but this approach does avoid needing to know what property name the flag turned into. Good enough for now!

shadowspawn:
I was just looking at the merge conflicts, and discovered I already fixed `--no-` on the `release/6.x` branch while reworking the option flag handling for #1256

I'll do some extra checking, but think this is already done.

shadowspawn:
Adding a regression test with @snitin315 as co-author

snitin315:
Thanks @shadowspawn, how soon can we expect the release?

shadowspawn:
We are getting ready to release v6 now, anytime from tonight to end of week.

shadowspawn:
Fixed in Commander v6.0.0 which has been released : https://github.com/tj/commander.js/releases/tag/v6.0.0

PR #1306, opened 2020-07-19
Add document links to errors

abetomo:
# Pull Request

## Problem

Not that there was a problem.

## Solution

I saw another tool that included the URL of the document in the error message.
That helped to solve the problem.

It's just one idea, so close it if you don't need it.

### DIFF
--- a/index.js
+++ b/index.js
@@ -468,7 +468,9 @@ class Command extends EventEmitter {
       throw new Error(`option '${option.name()}' clashes with existing property '${option.attributeName()}' on Command
 - call storeOptionsAsProperties(false) to store option values safely,
 - or call storeOptionsAsProperties(true) to suppress this check,
-- or change option name`);
+- or change option name
+
+Read more on https://git.io/JJc0W`);
     }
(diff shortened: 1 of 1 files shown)


shadowspawn:
Short version: 👍 

-----

Long version

The small concern I have is building a link to a particular section in the README is fragile in the longterm because the README may change. It doesn't completely break though, just loads at top of page.

We are trying hard in the throw to help the person solve the problem, and adding an URL may help them.

(If we need to offer more specific help for a throw case we could add a separate page to link to, but hopefully we are offering enough!)

shadowspawn:
I'll try it out before I finish reviewing it. Because the message appears in the middle of a throw with stack etc, I had tried to pull together the additional lines so they didn't get lost in the noise starting the lines with dash. Not sure how effective it is, so I'll try generating a throw with new link to see how it looks!

abetomo:
Thank you for the review!

shadowspawn:
Released in Commander v6.1

PR #1312, opened 2020-07-25
tests: remove eslint-disable jest/no-test-callback

michga:
# Pull Request

## Problem
I'm proposing a way to remove the use of `// eslint-disable-next-line jest/no-test-callback` in one of the tests.

## Solution

We can use a Promise to remove the use of the callback.
If `resolve` isn't called for some reason, the test will fail after 5 seconds (timeout).

### DIFF
--- a/tests/command.exitOverride.test.js
+++ b/tests/command.exitOverride.test.js
@@ -178,7 +178,7 @@ describe('.exitOverride and error details', () => {
   test('when executableSubcommand succeeds then call exitOverride', async() => {
     const pm = path.join(__dirname, 'fixtures/pm');
     const program = new commander.Command();
-    await new Promise(resolve => {
+    await new Promise((resolve) => {
       program
         .exitOverride((err) => {
           expectCommanderError(err, 0, 'commander.executeSubCommandAsync', '(close)');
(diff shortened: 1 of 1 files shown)


tests/command.exitOverride.test.js
```diff
@@ -175,19 +175,18 @@ describe('.exitOverride and error details', () => {
     expectCommanderError(caughtErr, 0, 'commander.version', myVersion);
   });
 
-  // Have not worked out a cleaner way to do this, so suppress warning.
-  // eslint-disable-next-line jest/no-test-callback
-  test('when executableSubcommand succeeds then call exitOverride', (done) => {
+  test('when executableSubcommand succeeds then call exitOverride', async() => {
     const pm = path.join(__dirname, 'fixtures/pm');
```
shadowspawn, line 179: To be explicit about having assertions (as well as calling resolve eventually), I think add a first line:

```
expect.hasAssertions();
```

(This could have been done in the previous version of the test too! 😄  )

michga: Good idea, i've added it.

shadowspawn:
Released in Commander v6.1

Issue #1311, opened 2020-07-24
README out of date for missing argument


shadowspawn:
See: https://github.com/tj/commander.js/issues/546#issuecomment-663630799

```
if (typeof cmdValue === 'undefined') {
  console.error('no command given!');
  process.exit(1);
}
```

shadowspawn:
Released in Commander v6.1
https://github.com/tj/commander.js/releases/tag/v6.1.0

PR #1320, opened 2020-07-28
Tidy up example now check for required command arguments

shadowspawn:
# Pull Request

## Problem

Example  in README and example file `env` includes unreachable code with current Commander.
See #1311

## Solution

Remove unreachable code, and simplify example.


shadowspawn:
Released in Commander v6.1

Issue #1321, opened 2020-07-31
Conveniency about parameter "argsDescription" in program.description()


ltccss:
In my case, some arguments of a command are not necessary to describe,  but currently I have to write all descriptions of them to avoid the null reference error.

I think it may be more pleasant to make a null check for "argsDescription[arg.name]" here?
https://github.com/tj/commander.js/blob/1b5ede83cee2b5bc7edbcdebde81f273095ebd8c/index.js#L1551

shadowspawn:
Reasonable suggestion.

(This is the second ever issue opened for the undocumented `argsDescription` parameter to `.description()`!)

shadowspawn:
Released in Commander v6.1
https://github.com/tj/commander.js/releases/tag/v6.1.0

PR #1323, opened 2020-08-01
Allow partial argument descriptions

shadowspawn:
# Pull Request

## Problem

If any argument descriptions are provided to `.arguments()`, then all the arguments must be described or there is a runtime error.

## Solution

Fallback to empty argument description if missing.

See #1321

## ChangeLog

- allow just some arguments in argumentDescription to `.description()`


shadowspawn:
Released in Commander v6.1

PR #1325, opened 2020-08-08
Allow disabling the built-in help option

shadowspawn:
# Pull Request

## Problem

People have asked about disabling the built-in help menu, usually to add their own help option and help handling. (This may be used less if it is possible to customise the help directly: #1296)

See:  #1225 for collected issues

## Solution

Add support for `.helpOption(false)` which disables the built-in help option. This is symmetrical with `.addHelpCommand(false)`.

Nice simple change in API and README, but lots of small code changes.

## ChangeLog

- allow disabling the built-in help option using `.helpOption(false)` (#1325)

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -559,7 +559,7 @@ from `--help` listeners.)
 
 ### .helpOption(flags, description)
 
-Override the default help flags and description.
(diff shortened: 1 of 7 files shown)


shadowspawn:
Released in Commander v6.1

Issue #1315, opened 2020-07-27
Extra documentation: tricks and traps with options with optional values


shadowspawn:
We could add a separate page, outside the README, going into details about potential challenges using options with optional values. They seem quite attractive and the README used to use them more than options with require values! In practice, they are a bit tricky and aren't a free choice.

- parsing ambiguity when use option as boolean flag but also operands and subcommands
- possible work-arounds
   - dash-dash (for expert users)
   - tell users always dash-dash before operands
   - tell users options come after operands
   - convert arguments into options! Options work pretty nicely together.
- conflict between short flag with value, and combined short flags with optional values. Work-around code to restore pre-Commander v5.0 parsing for people who prefer short-flag combining to short-flag+value.

heyjiawei:
Hi @shadowspawn I would like to help out with this, is it alright?

shadowspawn:
@heyjiawei happy to have someone else involved. Would you like to make a start on writing the page, or are you wanting to provide feedback/help with something I write?

heyjiawei:
If you have started on this than I can help provide feedback. Otherwise, I can start writing a section on this

shadowspawn:
> parsing ambiguity when also option as boolean flag but also operands and subcommands

I had a typo there, "when also" should be "when use" (fixed in original post now). 

If there is an option with an optional value `-o`

```
example -o
example -o optionalValue
example index.js
example -o whatIsThis
```

In the last example, the user might be intending `whatIsThis` as an argument or a subcommand but it will be read as the value for `-o`.

> conflict between short flag with value, and combined short flags with optional values.

Suppose there is a plain flag `-b` and an option with an optional value `-o`.

```
example -b -o
example -oValue   => same as -o Value
example -bo   => same as -b -o
example -ob   ???
```

Before Commander v5, `-ob` expanded to `-o -b`. After Commander v5 the `b` is read as the value for `-o`. I have a PR coming to allow changing this to help people coming from older versions of Commander who are used to the old behaviour: #1326

shadowspawn:
(I have been thinking about documenting complications with optional values for a long time, but have not started writing it down yet.)

heyjiawei:
Ah I see. Here is a rough draft (sorry its not complete but do comment!):

## Tricks and traps with options with optional values

There potential challenges using options with optional values. They seem quite attractive and the README used to use them more than options with require values! In practice, they are a bit tricky and aren't a free choice.

[...]

Issue #1203, opened 2020-02-26
Undocumented behavior change since 2.x


guimard:
Hi,

this simple program has 2 different behavior depending on commander version:
```javascript
var p = require('commander');
p.option('-d, --debug [level]', 'debug level')
 .option('-m, --mangle [test]', 'mangle')
 .option('--out', 'output')
 
p.parse(process.argv);
console.error('Debug args:', p.opts());
```

Resuilt of `node test.js -dm --out`:
 * 2.x : `Debug args: { debug: true, mangle: true, out: true }`
 * 4.x : `Debug args: { debug: 'm', mangle: undefined, out: true }`

This has strange effect when using uglify-js.

Cheers,
Xavier

shadowspawn:
Quick response: I think this is due to changes from #395, released in Commander 3.0.0

https://github.com/tj/commander.js/releases/tag/v3.0.0

shadowspawn:
The implementation of option parsing is refined further in Commander v5, and the updated README description is:

> Multiple short flags may optionally be combined in a single argument following the dash: boolean flags, the last flag may take a value, and the value. For example -a -b -p 80 may be written as -ab -p80 or even -abp80.

guimard:
Hi,

commander 5.0.0-4 fixes the problem.

Cheers,
Xavier

guimard:
Following the doc:
> Multiple short flags may optionally be combined in a single argument following the dash: boolean flags, the last flag may take a value, and the value. For example -a -b -p 80 may be written as -ab -p80 or even -abp80.

It becomes a bug

shadowspawn:
Drat, the new wording is not as clear as I hoped! The behaviour is as intended, but I will recheck the source documentation for behaviour and wording of description.

shadowspawn:
The standard I consult is: https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap12.html

Commander is more flexible with optional values than described there, but Commander is deliberately supporting the pattern that a value can directly follow the short flag, from item 2:

> If the SYNOPSIS shows an optional option-argument (as with [ -f[ option_argument]] in the example), a conforming application shall place any option-argument for that option directly adjacent to the option in the same argument string, without intervening <blank> characters.

And Guideline 5: 

> One or more options without option-arguments, followed by at most one option that takes an option-argument, should be accepted when grouped behind one '-' delimiter.

shadowspawn:
[...]

PR #1326, opened 2020-08-08
Add combineFlagAndOptionalValue

shadowspawn:
# Pull Request

## Problem

Previous versions of Commander always expanded combined short flags, so `-fb` always expanded to `-f -b`.  This allowed combining short flags for options with optional values as though they were boolean flags. 

The new behaviour is that an option with an optional value and the value can be combined, so `-fb` effectively expands to `-f b`.

This is a breaking change for old programs which heavily use optional values without an easy fix.

See: #1203

## Solution

Allow choosing between the two behaviours. Not included in the README. (Thinking might instead include in an future extended discussion of issues with optional values, see #1315.)

## ChangeLog

- add `.combineFlagAndOptionalValue(false)` to ease upgrade path from older versions of Commander (#1326)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -114,6 +114,9 @@ If you use `program.args` or more complicated tests to detect a missing subcomma
 
 If you use `.command('*')` to add a default command, you may be be able to switch to `isDefault:true` with a named command.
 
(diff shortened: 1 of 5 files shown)


shadowspawn:
Released in Commander v6.1

Issue #1197, opened 2020-02-15
feature request: pre run and post run hooks


majidsajadi:
How about option to add a preRun and a postRun function to each command to run before or after the main function?

shadowspawn:
This idea does come up occasionally. We need a good idea of what problem this is solving.

What problem would this help with? What is an example of what you would do in the preRun or postRun, and how would this be better than just calling it from the main function?

Related issues: #76 #936 #996

majidsajadi:
Opening database connection, creating directories or log file, etc.
We can call it from the main function but i guess it is better if we let the main function be about the command it self and not some repetitive functions.

Here is an example of a commander with support of prerun and postrun hooks

https://github.com/spf13/cobra#prerun-and-postrun-hooks

shadowspawn:
I looked at Cobra recently, has a lot of features. Good example thanks.

I see for pre/post Cobra has not two but four extra functions, with two of them specific to the command, and two of them "persistent" (I would have said inherited) and passed on to subcommands.

My first impression is pre/post hooks are going to be somewhat complex to use and of limited use.

I'm happy to leave this issue open though to see whether it generates further interest and insights.

shadowspawn:
One implementation complication with pre/post hooks would be if they should support async.

ad-m:
Hi,

We develop relatively extensive CLI (190 categories + 670 operations = 860 commands). Our CLI is contains multiple nested groups of commands. We are currently using structured-cli, which is abandoned and causes problems that require solutions in the engine. We tried to use commander.js. We quickly came across needs of plugins.

Our CLI is designed to manage cloud computing services. In practice, it is calling the platform API.

I developed my own mechanism around commander.js while trying to make a new CLI with it. So far I prepared plugins such as:
- api - provides API client initialization, including ensuring that the user has previously executed login commands, and if not - rejects the command until logging in, also supports the injection of additional login-related ones,
- formatOutput - adds parameters responsible for output formatting (similar to the "--output" and "--query" parameters available in CLI gcloud, aws and az), and also format the output accordingly,
- noWait - modifies the client-API so that it does not wait for the operation to complete after it is accepted by the API,
- validatorChoices - introduces parameter verifications in the range of acceptable values, e.g. acceptable output formats.
- formatError - format errors consuming standard error API format

The developed plugin mechanism turns out to be quite flexible in complementing the restrictions of commander.js in parameter parsing. Our plugins implementation wrap `action` in wrapper and promisify everyting:

[...]

PR #1330, opened 2020-08-13
Emit new events beforeAction and beforeExec

b-jsshapiro:
This is the promised pull request on issue #1197. Detailed explanation in the (only) commit message, but here's the summary:

1. Added `Command._emitToTree(eventName, subEvent, args)`. `Command._emitToTree()` emits events on a command and on all of its parents, progressively expanding the set of emitted events so that sub-tree filtering can be performed even though the standard EventEmitter is being used. Would be nice if the core EventEmitter supported some filtering, but it doesn't.

2. Added invocations of `Command._eventEmitter` before actionHandler and execHandler calls, with the result that a callback occurs prior to execution. The eventNames are `beforeAction` and `beforeEvent` respectively.

There may be a better or cleaner approach, but this one at least has the merit of being clear, and didn't require broadly invasive changes.

### DIFF
--- a/index.js
+++ b/index.js
@@ -296,21 +296,26 @@ class Command extends EventEmitter {
    * can select for events on any of its sub-command trees. Probably overkill,
    * but given that the depth is usually one it doesn't seem harmful.
    */
-  _emitToTree(eventName, subEvent, args) {
+  _emitToTree(eventName, args) {
+    let val = {
+      command: this,
+      name: this.name(),
+      cmdPath: this.name(), // initially
+      execArg: args.execArg,
+      args: args.args
+    };
+
     let cmd;
-    let subEvents = [subEvent, '*'];
+
     for (cmd = this; cmd; cmd = cmd.parent) {
-      subEvents.map((e) => {
-        cmd.emit(eventName + ':' + e, args);
-      });
-      // cmd.emit(eventName + ':*', args);
-      // console.log('Emitting for ' + subEvent);
-      // cmd.emit(eventName + ':' + subEvent, args);
+      cmd.emit(eventName, val);
+
       if (cmd.parent) {
-        subEvents = subEvents.map((e) => {
-          return cmd.parent._name + '.' + e;
-        });
-        subEvents.push('*');
+        const newCmdPath = cmd.parent.name() + '.' + val.cmdPath;
+
+        // Use spread to make a new emitted value, because the event structure
+        // is supposed to be read-only.
+        val = { ...val, cmdPath: newCmdPath };
       }
     }
   }
@@ -884,17 +889,17 @@ class Command extends EventEmitter {
         // add executable arguments to spawn
         args = incrementNodeInspectorPort(process.execArgv).concat(args);
 
-        this._emitToTree('beforeSpawn', this.name(), { command: this, bin: process.argv[0], args: args });
+        this._emitToTree('preSpawn', { execArg: process.argv[0], args: args });
         proc = spawn(process.argv[0], args, { stdio: 'inherit' });
       } else {
-        this._emitToTree('beforeSpawn', this.name(), { command: this, cmd: bin, args: args });
+        this._emitToTree('preSpawn', { execArg: bin, args: args });
         proc = spawn(bin, args, { stdio: 'inherit' });
       }
     } else {
       args.unshift(bin);
       // add executable arguments to spawn
       args = incrementNodeInspectorPort(process.execArgv).concat(args);
-      this._emitToTree('beforeSpawn', this.name(), { command: this, cmd: process.execPath, args: args });
+      this._emitToTree('preSpawn', { execArg: process.execPath, args: args });
       proc = spawn(process.execPath, args, { stdio: 'inherit' });
     }
 
@@ -1001,8 +1006,9 @@ class Command extends EventEmitter {
           }
         });
 
-        this._emitToTree('beforeAction', this.name(), { command: this, cmd: this.name(), args: args });
+        this._emitToTree('preAction', { execArg: this.name(), args: args });
         this._actionHandler(args);
+        this._emitToTree('postAction', { execArg: this.name(), args: args });
         this.emit('command:' + this.name(), operands, unknown);
       } else if (operands.length) {
         if (this._findCommand('*')) {
--- a/index.js
+++ b/index.js
@@ -290,6 +290,32 @@ class Command extends EventEmitter {
   };
 
   /**
+   * Emit a general event and a sub-event to this command and all of its
+   * parents in the command hierarchy, proceeding bottom-up. The number of
+   * events emitted at each level increases by one so that a given command
+   * can select for events on any of its sub-command trees. Probably overkill,
+   * but given that the depth is usually one it doesn't seem harmful.
+   */
+  _emitToTree(eventName, subEvent, args) {
+    let cmd;
+    let subEvents = [subEvent, '*'];
+    for (cmd = this; cmd; cmd = cmd.parent) {
+      subEvents.map((e) => {
+        cmd.emit(eventName + ':' + e, args);
+      });
+      // cmd.emit(eventName + ':*', args);
+      // console.log('Emitting for ' + subEvent);
+      // cmd.emit(eventName + ':' + subEvent, args);
+      if (cmd.parent) {
+        subEvents = subEvents.map((e) => {
+          return cmd.parent._name + '.' + e;
+        });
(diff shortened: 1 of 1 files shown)


b-jsshapiro:
Oh yes. lint and test pass, per instructions.

b-jsshapiro:
Closed the pull request by mistake while adding a comment. Sorry.

shadowspawn:
I have worked on a similar problem in #1296 with allowing more control over the help. The pattern being used there is a pair of events, one of which is only sent to the command, and one which is sent to the tree (group) to allow more "global" behaviours.

Your example in your commit comment is:

```
At the typescript-types node, the emitted events are
   beforeAction:*
   beforeAction:typescript-types

At the schematool node, the emitted events are
   beforeAction:*
   beforeAction:schematool.*
   beforeAction:schematool.typescript-types
```

using a pair of events like in #1296 this simplifies to:

```
At the typescript-types node, the emitted events are
   preAction
   preGroupAction

At the schematool node, the emitted events are
   preGroupAction
```

Are there use cases you have in mind that need more events? 

(On the naming, I narrowly went for "pre" over "before" for the help case, and using it here again to see how it feels in this scenario.)

b-jsshapiro:
Hey, John, thanks for looking.

I like pre/post better than before/after. I used _before_ in an attempt to follow one of the existing notification patterns.

FYI, my first proposal is already being used in near-production code. All it does is read a _yaml_ file after options (which can override config values) are parsed, and I don't think I'd have been able to use commander without _something_ to let me get in between the phases. Still easy enough to update here, but it provides at least _some_ confirmation of practical value.

Regarding the hierarchy, yes there's a need, but on re-thinking there isn't a need for this much. Let me start with the CLI code to identify the problem:

```
program
  .name(path.basename(PkgName))
  .version(PkgVersion)
  .description(PkgDescription)
  .option('--no-doc-comments', 'exclude doc comments from output')
  .option('-c, --configfile <string>', 'configuration file path', 'schematool.yml')
  .option('--odir <string>', 'output directory name')
  .option('-o, --output <string>', 'output file name')
  .on('beforeAction:*', LoadConfig)
  ;

// Elsewhere:
program
  .command('typescript-types [schema...]')  // <- CHILD Command Node
  .description('Generate typescript typings for this schema')
  .action((schema) => {
    TypescriptTypes();
  })
  .on('beforeAction:*', (stuf) => { console.log('tst before* ts-types'); }) // this was testing
  .on('beforeAction:typescript-types', (stuf) => { console.log('tst before ts-types'); })
  ;
```
For my immediate application, I need that LoadConfig handler on the root to fire no matter what command is run. The problem is, the on() attaches the callback on the root-level `Command` object in the hierarchy, while the emit() call gets made on a child `Command` object. That's what led me to `_emitToTree()`.

Here's how the rest evolved:

I realized why I wasn't seeing the event on the root-level Command object, kicked myself, and added `_emitToTree` so that it could be caught at the root. The ability to catch it at each level came for free and didn't seem like a bad thing, because tree-based filtering will be useful later for various things I had in mind. The embarrassing part of this is that I've written extensible command parsers of my own; the hierarchy of Command objects shouldn't have been a surprise.

Initially I just emitted ('beforeAction:commandName', { Command, cmdString, args }), but that hit a wall because EventEmitter's dispatcher doesn't provide any pattern matching. Sometimes I'll want to filter on the command name, so I decided (in haste and stupidly) to emit _both_ 'beforeAction:commandName' and 'beforeAction:*', simulating the _appearance_ of pattern matching. Then the hierarchy thing got my attention and I did the growing event list. Immediately saw the growing number of emits as a red flag, but it wasn't a bright enough red to get my attention in the wee hours.

Thinking about it after looking at your response, I think I would revise it to emit a single 'preAction' event at each Command node and change the payload value to be:

```
{ Command, 'leafCommandName', 'relative.path.to.leafCommandName', args }
```

Which can still be filtered in the handler if desired, still permits a subtree-based filter, doesn't suffer from a progressively increasing event count, and (considering your comment about #1269 ) subsumes the need for preGroupAction. Since the handlers might be async, and the node-relative path is changing as we go up the tree, I'd also update `_emitToTree()` to create a new payload object for each emit so that the emitted payload object doesn't get modified out from under a potentially deferred handler. Object spread is your friend.

Yeah, postAction and postExec are probably called for as well, though in some of those places there's an existing 'command' event being emitted already. Struck me as a little odd that those were post-events when I saw them, but I'm sure there was a reason.


Would you like a replacement pull request that puts all of this together? I'll add the post-events, but aside from that it's a smaller and simpler, which is usually a good sign. :-)

shadowspawn:
Thanks for info, and a positive sign you are actually using it. I wasn't sure whether this was all in theory or actually solving a problem. 🙂 

> Since the handlers might be async

Async event handlers? My current thinking is fully supporting async event functions and pausing at the right points before proceeding requires much more machinery, and eventEmitter does nothing to help.

> so that the emitted payload object doesn't get modified out from under a potentially deferred handler.

The opposite problem is that some use cases are going to want to modify the parse result and I am not sure the best way to achieve that. The most direct way is modifying the passed Command (which of course has problems with async again!) but it is not really set up for that.

> though in some of those places there's an existing 'command' event being emitted already.

That was originally the way the commands were implemented by Commander, adding its own listeners. The implementation has changed to direct dispatching but the events were retained to avoid breaking existing client code which added additional listeners.

> Would you like a replacement pull request that puts all of this together?

Thanks, but only if you are updating your own implementation anyway. I have plenty to think about.

b-jsshapiro:
@shadowspan: updated per last exchange. Definitely simpler. Not entirely sure where the right place is for the postExec event, so I haven't dropped that in yet. My opinion would be to emit that after process close/error, but thought it best to confirm with you?

Also, I wonder if the process.error case shouldn't emit a distinct event with some error information. Any thoughts on what to call that event?

shadowspawn:
@b-jsshapiro To be clear, I am finding this and discussion pretty interesting but may not merge this request. I don't want you to put work in assuming this is heading for a merge and be disappointed.

shadowspawn:
> Not entirely sure where the right place is for the postExec event, so I haven't dropped that in yet. My opinion would be to emit that after process close/error, but thought it best to confirm with you?

Short version: yes, that seems the most likely place and purpose.

Long version 

(For postExec, I assume you are referring to running stand-alone executable subcommands using `spawn`. And "process close/error" are the events being listed for on the spawned process eventEmitter.)

The high level question is what problems does this help with, what is the user trying to achieve?

For telling that the subcommand has completed or terminated, could use a `postExecutable` emit when the sub-command process emits the close/error.

shadowspawn:
> Also, I wonder if the process.error case shouldn't emit a distinct event with some error information. Any thoughts on what to call that event?

Hmm, that triggers all sorts of thoughts.

I think the errors that come back from `spawn` probably don't have very rich error information in general. Going wider though...

The support that is available for modifying error handling and process termination is to use `.exitOverride()` which is passed a `CommanderError` including the error message displayed to the user (when relevant). I did actually wonder about emit-ing the errors instead, but went with throwing an error for the much more familiar try/catch pattern for users. Commander is currently using events in a fairly light way as a convenience for some functionality rather than the scaffolding for the runtime.

EventEmitter does have an `error` event which is special. Listening for the event changes the runtime behaviour from triggering a throw to assuming the program has handled the error. We don't want the default behaviour for all errors to be a stack dump so not the right event, but interesting parallels.

b-jsshapiro:
I don't currently have a use case for exec events. I put them in from a sense of symmetry. I don't have a use case for post events, and especially not given that you already have `.exitOverride()`. I wouldn't have added the post events, except that you mentioned them. Actually, I really don't _like_ events. Given a free hand I would have added a `.beforeCommand()` callback.

As our exchange has proceeded, it turned out on my end that there was a reason to make a `GetModuleConfig()` call from the top of each command, so I can handle the config issue from there. Given which, I'm going to close out the pull request for lack of a use case.

From a design perspective, I'd encourage some form of `.beforeCommand()`-like thing. Even a cursory look at some basic UNIX or Windows commands will confirm that there are lots of cases of mutually exclusive options. In my opinion, it's cleaner to check those in one place than to replicate them in all of the relevant `.option()` callbacks. Distributing the logic rarely turns out to be maintainable.

Oh. While I have you, maybe consider moving the `argv[0]` pre-prep to `parseOptions()`? It's sitting in a kind of strange place at the moment, with the effect that `parse()` and `parseOptions()` can have different behavior, and it looks pretty trivial to relocate it. Let me know if a pull request on that would save you some work.

Issue #1283, opened 2020-06-19
Object default values display JSON under help menu


carsonreinke:
Using an any object as the default value displays a JSON version of that object.  This might not be very user friendly.

Here is an example using `IPv4` from `ip-num` package:
```
(default: {"bitSize":32,"maximumBitSize":"4294967295","type":1,"octets":[{"value":1},{"value":1},{"value":1},{"value":1}],"separator":".","value":"16843009"})
```

Consider changing `JSON.stringify` at https://github.com/tj/commander.js/blob/master/index.js#L1413 to a scheme utilizing `.toString` instead.  The impact could be pretty severe though.

shadowspawn:
Interesting idea.

Would work nicely when there is a custom `.toString`, like with `ip-num`. The fallback mode for `.toString` on uncustomised objects is not very user friendly either though. i.e. `[object Object]`

On a somewhat related note, there was an issue opened about whether a custom coercion function should be called on the default value rather than the default value being in final form.

(A work-around is to handle the default in your own code when the option is undefined.)

carsonreinke:
@shadowspawn I'm assuming you are referring to #1036.

I feel the supplied default should be previously coerced, since it may hard to represent all desired default values directly from the command line.

I'm actually not sure `toString` is even the right method for this, since the value is actually a command line friendly representation of the data.  Whatever that would be called.

Either way, I don't see a real simple solution.  My work around was this:

```js
function parseIP(value: string): IPv4 | string {
    return IPv4.fromDecimalDottedString(value);
}

program.option('--dns <ip>', 'Additional DNS server to check', parseIP, '1.1.1.1');

...

if(typeof(program.dns) === 'string') {
    program.dns = parseIP(program.dns);
}
```

shadowspawn:
I finally found the comment where someone expected the default value to be processed by the coercion function : https://github.com/tj/commander.js/issues/400#issuecomment-408164531

carsonreinke:
@shadowspawn Thank you.  Looking into this more, I thought that coercing the default value would be the best, but it appears it is more complicated.  The coercion function also takes a `previous` argument, coercing the default value, there is no previous value.

Ultimately from the test cases, the default value should be previously coerced.  The easiest would be to just add some more complex "stringify" functionality.

Maybe something like this:

```js
switch(typeof(option.defaultValue)) {
  case 'undefined': return '';
  case 'string':
  case 'boolean':
  case 'number':
    return JSON.stringify(option.defaultValue);
[...]

Issue #1106, opened 2019-11-23
Feature Request: ability to hide options


Tbhesswebber:
### Ask:

Provide a `.noHelp()` command for an option to opt-out of the default help generation _or_ update the `.allowUnknownOptions` command to take a list of accepted "unknown" options and store any of those which are passed somewhere within the program for access.

```javascript
const commander = require('../');
const program = new commander.Command();

// unknown options
program.allowUnknownOptions(['--super-special-pizza']);

program
  .option('-h, --hawaiian', 'Create a sad pizza')
	// noHelp
  .noHelp();
```

### Context: 

Currently, there is no ability to hide options other than to manually parse `process.argv` in search of the specific options that you want and then strip those out of what you pass to `program.parse`.

I am working on an internally open-sourced CLI that takes advantage of feature flags to determine whether a command is in development, beta, production, or deprecated.  The CLI is dynamically populated with the available commands depending on the user's self-declared "level" in a configuration file.

#### Use 1
In order to make our testing more robust, we'd like to be able to pass `--[dev|beta|prod]` in on the command line that will override that configuration file, making it so that we need fewer configuration files in order to run integration and e2e tests on the full suite.

#### Use 2:
Once we integrate with the API that is in development, we will want to pass a `--test [endpoint]` flag to the code during e2e tests to inform it that it needs to use that specific endpoint for the test-session and the test-specific headers that are required for a test session.

shadowspawn:
A challenge with additions to `.option()` is that we don't have an options parameter to easily specify custom behaviour. I suggested `.optionEx` in #1038 but it was not popular, at least for that case. 

The issue I have with the `.noHelp()` chain call is that it is actually being called on a command object rather than on an option object, and feels a bit fragile. (There was a similar suggestion for a chain call for mandatory options too.)

A different idea, but a bit clumsy, is a separate call to supply the extra settings: 
```
.markOption('--hawaiian', { noHelp: true }):
```

Similar past issue: #811

shadowspawn:
Exploring a couple of builder patterns for extending functionality, which might fit in with Commander chaining style.

Inline builder:

```
program
  .option('--one', 'description 1')
  .buildOption('--test <endpoint>')
    .noHelp()
    .build()
  .option('--two', 'description 2');
```

And a more conventional builder which returns an Option:

```
program
  .option('--one', 'description 1')
  .addOption(OptionBuilder('--test <endpoint>')
    .noHelp()
    .build()))
[...]

Issue #518, opened 2016-03-26
Provide Array of permitted values for an option.


joshuapinter:
I have an `.option()` that is only allowed to be one of "all", "debug" or "none".

I'd like an easy way to pass an `Array` to an `option()` and have Commander automatically check that the value passed by the user is one of the permitted values. 

Something like this: 

``` js
.option('-t, --target <target>', 'Specifies the target.', ["all", "debug", "none"])
```

Also including the permitted values in the help output could be useful, provided it's not a very large `Array`.

I'd gladly do the PR if I can get a response from @tj or a collaborator that this is useful.

Thanks!!

tmiddlet2666:
I second this, Would be great to be able to do this and validate options.

shadowspawn:
The `.option` call is quite flexible already and it is not easy to add more functionality to the current signature. I am wondering about allowing building options to make it possible to add less common functionality without complicating the existing usage. See https://github.com/tj/commander.js/issues/1106#issuecomment-558379166

I have been wondering if there could be a collection of validators available from Commander to use as custom option processing functions, but my first attempts were a bit clumsy. For a crude example of bring-your-own validator: https://github.com/tj/commander.js/issues/1130#issuecomment-569565495

shadowspawn:
I have been wondering about the syntax and here is one idea, fluent (chaining) calls on Option object like with Command object:

```
program
   .addOption(new Option('-t, --target <target>', 'Specifies the target (all | debug | none).')
      .valueIncludedIn(["all", "debug", "none"])
   )
);
```

shadowspawn:
And another idea, convenience routines for custom option processing with existing api:

```js
program
   .option( '-t, --target <target>', 
      'Specifies the target (all | debug | none).', 
      OptionType.includedIn(["all", "debug", "none"])()
   )
...

(I used `OptionType` here but the other name I considered was `OptionValue`.)

**Comment by shadowspawn** [comment]
Related link from https://github.com/tj/commander.js/issues/215#issuecomment-660456539

Python argparser has `choices` for restricting allowed values: https://docs.python.org/3/library/argparse.html#choices

**Comment by shadowspawn** [comment]
Experimenting with Option approach in #1331:

```js
program
   .addOption(new Option('-s, --size <value>').choices(['big', 'little']));
```

shadowspawn:
Included in v7.0.0 pre-release #1386

lydell:
[...]

PR #1331, opened 2020-08-15
Enhance Option class to allow hiding help, specifying choices, and change how default value displayed in help

shadowspawn:
# Pull Request

There are various requests for additional behaviours for options that are hard to add to the existing Command interface. This is an experiment to see whether turning Option into a richer class will allow adding new features in a tidy way by working directly with the option, rather than indirectly from the command.

## Problem

- Feature Request: ability to hide options #1106
- Object default values display JSON under help menu #1283
- Provide Array of permitted values for an option. #518

## Solution

Enhance Option with support for configuring custom behaviours. Add a created Option to command with `.addOption()`.

Use like:

```js
program
   .addOption(new Option('-s, --secret').hideHelp())
   .addOption(new Option('-t, --timeout <delay>', 'timeout in seconds').default(60, 'one minute'))
   .addOption(new Option('-s, --size <value>', 'drink size').choices(['big', 'little']));
```

```bash
$ node help.js -h
Usage: help [options]

Options:
  -t, --timeout <delay>  timeout in seconds (default: one minute)
  -s, --size <value>     drink size (choices: "big", "little")
  -h, --help             display help for command
```

## ChangeLog

- allow hiding options from help
- allow restricting option arguments to a list of choices
- allow setting how default value is shown in help


shadowspawn:
Tossing up between getter/setter methods which return existing value when no value specified (like Command:name) or methods which describe what they do and always act (like allowUnknownOption).

Tbhesswebber:
This looks awesome!  I'm not seeing any tests for things being chained together - what is the expected behavior if we provide defaults and hide the help like in the following?

```js
program
   .addOption(new Option('-t, --timeout').hideHelp().default(60, 'one minute'));
```

It seems to me like anything not shown in help should require a default value - should that require `.default` to be called?

Also, I realize that I was part of the impetus for this PR and I explicitly asked for a way to hide options, but should there be some way to toggle exposure of the hidden options.  Something like giving every program a `--advancedHelp` flag that shows hidden flags?  Maybe `hideHelp` should take an options object that allows it to accept default values and a boolean representing if it should show in advanced usage?

Thanks so much for all of your work on this and I'm sorry that I'm constantly poking holes in things to see if they pop!

shadowspawn:
Thanks @Tbhesswebber . No popping noises yet. 🎈 

> I'm not seeing any tests for things being chained together - what is the expected behavior if we provide defaults and hide the help like in the following?

These calls chain (like when configuring Command), so in your example the option will have a default value and be hidden.

> It seems to me like anything not shown in help should require a default value

Why is it different to a non-hidden option? If I don't look in the help, then I may not know what other options exists or whether they have a default value.

> but should there be some way to toggle exposure of the hidden options

You can arrange for this yourself by passing false into hideHelp. Say by testing an environment variable to see whether to show everything.

Tbhesswebber:
Fantastic!  I think I just naturally think of all things having defaults within a CLI in order to avoid having to handle multiple ways of tracking values (with the exception of anything that is provided by a config file), so you can ignore that point of mine!

Again, I really appreciate your efforts here!

shadowspawn:
We could also add a createOption on Command to go with the createCommand, to make it easier to subclass Command and Option.

index.js
```diff
@@ -619,7 +719,22 @@ Read more on https://git.io/JJc0W`);
    */
 
   option(flags, description, fn, defaultValue) {
```
abetomo, line 721: It's pretty much the same as `requiredOption`.
I thought it would be better to make it a function like the original code, but what do you think?

shadowspawn: It was much smaller when I first did it so seemed reasonable not sharing code, but I had left out the old regexp code! I'll take another look.

shadowspawn:
Making a series of changes, I'll again comment when done.

index.js
shadowspawn: I changed `.requiredOption()` to call `.option()` to do most of the work. Means using knowledge that added to end of `options` array, but nicely consistent with `requiredOption` just being `option` plus mandatory.

shadowspawn:
I made a bunch of tidies since review comment. (A couple related to previous PR! 😊 )

- remove duplicated code by having `.requiredOption()` implemented using `.option()`
- rename `parseArgWith` to `argParser` as more conventional name
- improve JSDoc
- describing deprecated features as legacy in the comments, since we won't delete them until there is a good reason
- added @deprecated to TSDoc so get editor feedback (e.g. latest VSCode this week!)

shadowspawn:
` npm run typescript-checkJS ` is currently showing some errors, including from my latest changes. I will investigate...

shadowspawn:
Fixed.

shadowspawn:
> describing deprecated features as legacy in the comments, since we won't delete them until there is a good reason

I revisited this today. I was worried "deprecated" meant "we will delete this" and was too strong. But deprecated also can mean "this was a bad idea" or "there is a better way now". I'll go back to calling things deprecated instead of legacy, it is the right word for the functionality we would like people to stop using and is supported with the `@deprecated` tag.

So I'll make some wording changes but not logic changes, again! Sorry for the churn, but I am pleased to get this sorted out in my mind. I have been wondering how best to talk about old code for a while. 🤔 💡

shadowspawn:
Done. Back to "deprecated"!

abetomo (approved):
awesome!

shadowspawn:
Thanks for review @abetomo , it was a big one! 🎉

PR #1339, opened 2020-08-29
Create new command with class constructor

cravler:


shadowspawn:
After some more thinking, perhaps not.

This works nicely if the subclass keeps the same signature for the constructor, which will usually be the case. However, the subclass may ignore the name or require additional parameters in its constructor.

Expecting the subclass to define createCommand is simplistic, but means the calling convention for creating subcommand is explicit. A single parameter which is the command name. The subclass can do what wants for its constructor.

shadowspawn:
The return type of `createCommand` also gets used as the return type of `.command()` for Typescript, a suggestion from:
https://github.com/tj/commander.js/pull/1191

cravler:
agree, subclass can overide this method, if needed

Issue #1335, opened 2020-08-24
Is there a way to run command as detach mode?


fdorantesm:
I'm looking how to create a http server as detached process with command but i didn't find if it's possible or not.

Regards.

shadowspawn:
The simple answer is do it yourself with calls to `child_process.spawn` et al.

(Commander subcommands implemented as external executables are run as a separate process, but it is an attached process, and does not sound like a match for what you have in mind.)

aemonge:
Hey @fdorantesm , I had the same issue/request you needed, so I went ahead and implemented it in a forked repository.

@shadowspawn I did found the subcommand issue a bit tough, but I think I've been able to include it and was able to implement the feature. Please go ahead a start a discussion with me to query about my implementation, i really hope it can be merged and it provides value to other users.

Thanks both in advance, you can see my pull request > https://github.com/tj/commander.js/pull/1340

fdorantesm:
> Hey @fdorantesm , I had the same issue/request you needed, so I went ahead and implemented it in a forked repository.
> 
> @shadowspawn I did found the subcommand issue a bit tough, but I think I've been able to include it and was able to implement the feature. Please go ahead a start a discussion with me to query about my implementation, i really hope it can be merged and it provides value to other users.
> 
> Thanks both in advance, you can see my pull request > #1340

Nice. I solved using `spawn`method of `process-child. I'm starting an express app.

fdorantesm:
> The simple answer is to do it yourself with calls to `child_process.spawn` et al.
> 
> (Commander subcommands are implemented as external executables are run as a separate process, but it is an attached process, and does not sound like a match for what you have in mind.)

Indeed, thanks.

Issue #351, opened 2015-02-18
Using commander I describe a command line based API. How about an HTTP API?


rjsteinert:
Using commander I describe a command line based API. It seems to me the same description could translate to an HTTP API. Now I know it's probably not ideal... But let's just say I'm lazy and interested in writing things that have to be a command line interface and an HTTP API that comes for free would be nice to have. Does anyone know of any work/thoughts on this idea?

SomeKittens:
Writing something that works via command-line *and* HTTP is a neat idea!  However, that's not the goal of this project.

aemonge:
I agree, but If you found any resources can you post them here ?

PR #1340, opened 2020-08-31
Feature/http client

aemonge:
# Pull Request

## Problem
Enhancement, allow the commander utility to serve as an http server version of the CLI. 

## Solution
By passing to the program main class an Express server variable, it's been able to be used as a web-client software. Allowing a easy deployement of a CLI or WebInterface tooling.

```
const express = require("express");
const app = express();
// ....................
app.use(express.json());
program.parse(app, {from: "express", port: 8080});
```
## Discussion

Documentation should be change to include the new feature, and also to recommend the usage of the `program.write` function to centralize the communication that can be either a `express.send` or a `console.log`.

Moreover this pull-request needs to be updates to include the unit testing of my new functions, to guarantee they wont break in the future

### Added
 - Added http client support, to use the create tool not only as CLI but also as a web client.

### DIFF
--- a/index.js
+++ b/index.js
@@ -128,6 +128,7 @@ class Command extends EventEmitter {
     this._exitCallback = null;
     this._aliases = [];
     this._combineFlagAndOptionalValue = true;
+    this._statuses = { error: 500, warn: 219, info: 201 };
 
     this._hidden = false;
     this._hasHelpOption = true;
@@ -1299,7 +1300,7 @@ Read more on https://git.io/JJc0W`);
     this._versionOptionName = versionOption.attributeName();
     this.options.push(versionOption);
     this.on('option:' + versionOption.name(), () => {
-      this._write(str + '\n');
+      this.write(str + '\n');
       this._exit(0, 'commander.version', str);
     });
     return this;
@@ -1648,7 +1649,7 @@ Read more on https://git.io/JJc0W`);
     if (typeof cbOutput !== 'string' && !Buffer.isBuffer(cbOutput)) {
       throw new Error('outputHelp callback must return a string or a Buffer');
     }
-    this._write(cbOutput);
+    this.write(cbOutput);
     this.emit(this._helpLongFlag);
   };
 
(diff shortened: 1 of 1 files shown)


shadowspawn:
This is probably too special-case for Commander core.

I don't really understand the use case. Why parse command-line style commands through a web interface?

I suspect there are some runtime issues:
1) By default, Commander will exit the process if the parsing fails (taking down your server)
2) Commander does not clear out state when doing a parse. You need to create a new Command to do another parse.

shadowspawn:
Related: #351 #1335

aemonge:
Hi, 

Indeed, I've found that both things are needed to be considered. Since the state wont be cleared, I thought of closing the connection after processing successfully or not any command. This way it will behave more as the CLI it is, and only delegating the input/output to a HTTP channel. And to use a `while` bash loop, or a `forever` script to keep it alive; sidestepping away from the issues we're both aware.

Also, I do agree with you in the sense that this is out of scope of the main Commander core; maybe we can work together to enable Commander to support a plug-in system in which this feature can be the first plug-in. 

To be honest, I think the only purpose to enable a webHttp I/O for the scripts is to share to none-cli programmers. In my strict case, my colleagues are afraid of opening a terminal (such a shame, i know....) but they aren't afraid of using an API.  So that's the main goal of it, to allow an easy portability of a CLI script.

shadowspawn:
Thanks for the usage comments.

I think it would be easier to have a web server call Commander, than Commander "be" a web server. `.exitOverride()` both modifies the default exit-on-error behaviour, and includes the error string in the exception, which might allow convenient passing back over the http channel.

No plugin plans currently.

Closing this as out of scope for Commander.

aemonge:
Ok..... just let me know if I can help in the future if you ever decide to implement plugins.

aemonge:
By the way, I'm using [pm2](https://pm2.keymetrics.io/) to restart the script after any process, so now it can be keep alive for every for DeamonHttp purposes

PR #1350, opened 2020-09-08
Add note to Readme related to name clashes

baparham:
Options like storeOptionsAsProperties must be specified before some other types of options. This updates the Readme.md to remind readers to keep this in mind.

# Pull Request

## Problem
Hit undocumented bug when specifying `storeOptionsAsProperties(false)` after `version()`. The program would silently crash without telling me that I couldn't call `storeOptionsAsProperties(false)` since I've already called `version()`

## Solution
Update readme to reflect requirement to be careful with the order of functions called

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -627,6 +627,8 @@ There are two new routines to change the behaviour, and the default behaviour ma
 - `passCommandToAction`: whether to pass command to action handler,
 or just the options (specify false)
 
+> It's important to note that those options must be specified before other options that would be affected by this change in behavior, such as `.version(), .option(), etc`
+
 Example file: [storeOptionsAsProperties-action.js](./examples/storeOptionsAsProperties-action.js)
 
 ```js


shadowspawn:
Under what circumstances do you see a silent crash? I see a noisy throw with an explanation, although mentions options in general and not `.version`.

```js

program
  .version('1.2.3')
  .storeOptionsAsProperties(false);

program.parse();
```

```
$ node index.js
/Users/john/Documents/Sandpits/commander/my-fork/index.js:685
      throw new Error('call .storeOptionsAsProperties() before adding options');
      ^

Error: call .storeOptionsAsProperties() before adding options
    at Command.storeOptionsAsProperties (/Users/john/Documents/Sandpits/commander/my-fork/index.js:685:13)
...
```

baparham:
Thank you for the reality check. I had been hiding my errors inadvertently with a .catch higher in my stack. I removed it and see the error you are referring to:
`(node:22692) UnhandledPromiseRejectionWarning: Error: call .storeOptionsAsProperties() before adding options`

So this can be closed since the information is there, I was just hiding it!

shadowspawn:
Whew! We could extend the README tips if people stumble over this, but it did sound like you had something extra happening. Thanks for update. 👍

PR #1351, opened 2020-09-09
Add test for addTextHelp returning void

shadowspawn:
# Pull Request

## Problem

Function passed to `.addHelpText` may return void, whether because at runtime there was nothing to write, or because handled writing itself (like just converting straight from `.on('--help'`). This was not being tested.

## Solution

Add a test.


abetomo (approved):
The addition of the test is great.

Issue #1343, opened 2020-09-04
Flag option with implicit negate alternative.


mshima:
We currently have a mix of commander.js with meow for unknown options.
I am trying to disable allowUnknownOption.
One challenge is that commander does not allow --no to flags without adding the negate flag.

Why do we need need negate?
Our documentation and usability requires it.
When the application is executed for the first time (ex `--foo`) we save the option, and then when it is executed for the second time (ex `--no-foo`), we allow that option to be changed.

Adding negate option to every flag is a workaround, but that brings others problem:
- Our help is already too big without the negate value.
- We don't provide descriptions for the negate option and we don't have plans to provide.

It would be great if commander added 1 of the following:
- Something like program.alwaysAllowNegate(true)
This would always allow a flag to have negate value.
- ~Add an option to hide the option help, eg hide when staring with '!':~
```
this.option('cheese', 'with cheese').option('no-cheese', '!with cheese')
```
- ~Hide empty negate flag help if the flag is added.~
```
this.option('cheese', 'with cheese').option('no-cheese')
```
- ~Add implicit negate when the option starts or ends with !.~
```
this.option('cheese!', 'with cheese')
```

shadowspawn:
Nice description of your problem.

An open issue to allow hiding options is #1106 and a PR with a possible approach is #1331.

mshima:
@shadowspawn after a failed attempt, we actually need negate option by default.
Reason:
- Execute `foo` command with `bar` option not settled.
- At some point it can consolidate to `bar: false` value.
- We regenerate command args becoming [`--no-bar`].
- Then we fork with child_process.fork('foo', ['--no-bar'].
- Fails with unknown `--no-bar` option.

shadowspawn:
I do not understand your failed attempt. Were you only adding either `--bar` or `--no-bar`, depending on the saved state? (I would expect you need to add both since either could be specified on the command line independent of the saved state.)

shadowspawn:
I do not know if this covers your use cases, but as an example this custom class automatically adds a hidden negative option when boolean options are added. It uses new features coming in Commander v7 (code available now on the `release/7.x` branch).

```js
const { Command, Option } = require('commander');

class MyCommand extends Command {
  createCommand(name) {
    return new MyCommand(name);
  }

  addOption(option) {
    const result = super.addOption(option);
    // Add a hidden `--no` option for boolean options
    if (option.long && !option.required && !option.negate && !option.optional) {
[...]

PR #1354, opened 2020-09-12
Implement implicitNegateOption.

mshima:
# Pull Request

Fixes https://github.com/tj/commander.js/issues/1343

## Problem

- We are currently using commander with unknown option parsing using [meow](https://github.com/sindresorhus/meow) that allows `--no-` automatically.
- We implement ~25 sub commands each providing their options.
- Almost every option contains 3 states: false, true and default (saved value).
- Adding a negate option to every option is not an option:
  - help will be too big
  - not every option is added the same way, some are generated, some has to be coded.

## Solution

- Add allowNegateOption() to Options.
- Add implicitNegateOptions() to Command that will set allowNegateOption() at Options.
- Fallback parsing `--no-foo` to `--foo` with false value if no `--no-foo` option was found.

### DIFF
--- a/index.js
+++ b/index.js
@@ -34,6 +34,13 @@ class Option {
     }
     this.description = description || '';
     this.defaultValue = undefined;
+    this.negateOption = false;
+  }
+
+  allowNegateOption(allowNegateOption) {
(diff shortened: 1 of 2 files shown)


mshima:
Reopened against develop branch.

Issue #939, opened 2019-03-27
Triage: Problems with negating boolean options


shadowspawn:
People have a variety of problems with boolean options and negating.  Opened this issue to collect the related issues found during triage.

NB: there is an open pull request and I have not yet looked at how well it addresses the issues, but likely to be some documentation work even if functionally solid.

Issues:
- #108 On and off switches are not working
- #238 when specifying options, the "--no" version of a flag must appear after the flag
- #574 Defining boolean options via Command#option() defaults to undefined, not false.
- #626 Command.opts() values are always undefined
- #630 No way to know if parameter is set with --no- options
- #928 Support defaults for boolean options (including --no-*), for env var set defaults
- #979 Adding negative option makes Commander.js to return "true" when option is not specified

Pull Requests:
- #795 fixes behavior of --no-* options

usmonster:
Thanks for the triage work here, especially concerning the readme & examples updates in #953! FYI, I've updated my open PR (#795) to include your changes.

Regarding this:
> NB: there is an open pull request and I have not yet looked at how well it addresses the issues, but likely to be some documentation work even if functionally solid.

Once you get around to finishing your review, please let me know if there's anything related to [my comment on Custom Events](https://github.com/tj/commander.js/pull/795/files#r185717938) that you think might require further functional or documentation updates. Thanks again!

shadowspawn:
This issue will be resolved when v3.0.0 is released. Available now as a prerelease. See #1001

shadowspawn:
Support for combined `--foo` and `--no-foo` has shipped in v3 (thanks @usmonster): https://github.com/tj/commander.js/releases/tag/v3.0.0

PR #1355, opened 2020-09-12
Implement allowNegateOption.

mshima:
# Pull Request

Fixes https://github.com/tj/commander.js/issues/1343

## Problem

- We are currently using commander with unknown option parsing using [meow](https://github.com/sindresorhus/meow) that allows `--no-` automatically.
- We implement ~25 sub commands each providing their options.
- Almost every option contains 3 states: false, true and default (saved value).
- Adding a negate option to every option is not an option:
  - help will be too big
  - not every option is added the same way, some are generated, some has to be coded.

## Solution

- Add allowNegateOption() to Options.
- Add implicitNegateOptions() to Command that will set allowNegateOption() at Options.
- Fallback parsing `--no-foo` to `--foo` with false value if no `--no-foo` option was found.

### DIFF
--- a/index.js
+++ b/index.js
@@ -119,7 +119,6 @@ class Command extends EventEmitter {
     this.options = [];
     this.parent = null;
     this._allowUnknownOption = false;
-    this._implicitNegateOptions = false;
     this._args = [];
     this.rawArgs = null;
     this._scriptPath = null;
@@ -499,7 +498,7 @@ Read more on https://git.io/JJc0W`);
    */
 
   _optionEx(config, flags, description, fn, defaultValue) {
-    const option = new Option(flags, description).allowNegateOption(this._implicitNegateOptions);
+    const option = new Option(flags, description);
     const oname = option.name();
     const name = option.attributeName();
     option.mandatory = !!config.mandatory;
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -34,6 +34,13 @@ class Option {
     }
     this.description = description || '';
     this.defaultValue = undefined;
+    this.negateOption = false;
+  }
+
+  allowNegateOption(allowNegateOption) {
+    allowNegateOption = (allowNegateOption === undefined) || allowNegateOption;
+    this.negateOption = !this.negate && this.long && !this.required && allowNegateOption;
```
shadowspawn, line 42: I don't think the `required` test is appropriate. Commander allows `--no-bar` to be added when there is a `--bar <value>`.

shadowspawn:
Some initial comments. I am not sure about this.

An implicit negate for a single option is interesting, but overlaps with the current implementation. The original implementation assumed there would be just one of positive or negative option, but #795 added the ability to have both work together.

Turning on implicit negate options for a whole command is quite intrusive. Every option will have to cope with the option value potentially being `false`.

The implicit negate should probably appear in the help description.

index.js
mshima: Seems you are right.

mshima:
Took your addOption from https://github.com/tj/commander.js/pull/1331.

shadowspawn:
Somewhat related issues:

- #939 Triage: Problems with negating boolean options
- #795 fixes behavior of --no-* options

shadowspawn:
There are two separate ideas in the description and original code for this PR:
- adding automatic support for `--no`
- having single option support for `--no` (rather than adding a separate `--no` option)

Both are interesting, but I would like more interest and use cases from the community before adding either. At the moment the use case is to match a specific existing implementation.

The single option support  for `--no` has a disadvantage that is overlaps with the existing support, so needs to offer some compelling value to justify supporting multiple approaches.

The code changes in `parseOptions` in this PR are reasonably tidy. To match historical usage, the custom option processing function is only called on option-arguments and not when the option is used in a boolean way. The existing handling passes in null as the value, and tests for that in the listener. To get similar behaviour with the added `--no` code the listener could perhaps check for a boolean value in addition to null, or instead of null.

There are major changes to the options coming inCommander v7 and the changes have now landed on the `release/7.x` branch. Have a look at what is possible and what is missing there.

PR #1361, opened 2020-09-25
Terminology and links

shadowspawn:
# Pull Request

Followup work from https://github.com/tj/commander.js/pull/1332#issuecomment-691615173

Moved Terminology to own document, and added links to the new documents from the README.


abetomo (approved):
Thank you for a good job.

PR #1363, opened 2020-09-26
Back to shared implementation method for .option and .requiredOption

shadowspawn:
# Pull Request

## Problem

Implementing `.requiredOption() ` using `.option()` was very compact, but having second thoughts. Some drawbacks mainly for subclassing:

- tricky to modify option creation with chain of methods being called 
- fragile reliance on new option being last in options array
- calling `.addOption` without final configuration in the passed option

(Prompted by thinking about how to add an override to add a matching `--no` option for every boolean option.)

## Solution

Just shift the shared code back into `._optionEx` like previous pattern, but with the new code.

Much as @abetomo suggested in https://github.com/tj/commander.js/pull/1331#discussion_r486710924 !

### DIFF
--- a/index.js
+++ b/index.js
@@ -667,6 +667,31 @@ Read more on https://git.io/JJc0W`);
   }
 
   /**
+   * Internal implementation shared by .option() and .requiredOption()
+   *
+   * @api private
+   */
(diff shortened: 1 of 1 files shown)


abetomo (approved):
LGTM.
Thanks!

PR #1364, opened 2020-09-26
Drop Ruby reference

shadowspawn:
# Pull Request

`commander.js` has long surpassed Ruby commander in popularity, and the reference does not help most new readers.


abetomo (approved):
🆗

Issue #1367, opened 2020-10-01
Allow tsx file extension for subcommands


brunomacf:
I'm using [ink](https://github.com/vadimdemedes/ink) to implement my cli app and right now commander does not look for sub-command files with tsx extension. I Going to open a PR to fix that.

shadowspawn:
Released in Commander v6.2.0

PR #1368, opened 2020-10-01
Allow tsx file extension for sub-commands (#1367)

brunomacf:
# Pull Request

## Problem

I'm using [ink](https://github.com/vadimdemedes/ink) to implement my cli-app which uses React and right now commander does not look for sub-command files in tsx extension. This PR add tsx as a valid file extension when looking for sub-command files.

Related Issue: #1367

## Solution

It's a very simple PR where i just added tsx as valid file extension in `_executeSubCommand` function.

### DIFF
--- a/index.js
+++ b/index.js
@@ -838,7 +838,7 @@ Read more on https://git.io/JJc0W`);
   _executeSubCommand(subcommand, args) {
     args = args.slice();
     let launchWithNode = false; // Use node for source targets so do not need to get permissions correct, and on Windows.
-    const sourceExt = ['.js', '.ts', '.mjs'];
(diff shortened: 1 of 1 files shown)


shadowspawn:
How are you running your program so that `.tsx` files can be opened by node? 

I know how this works with `ts-node` and `.ts`, not sure if the same for `.tsx`:

    node -r ts-node/register pm.ts

brunomacf:
@shadowspawn yep is essentially the same thing... and i'm using exactly that command to run it here with tsx files :)

shadowspawn (approved):
LGTM

abetomo (approved):
👍

shadowspawn:
Released in Commander v6.2.0

PR #1369, opened 2020-10-04
Add tests for some getters

shadowspawn:
# Pull Request

## Problem

We have a setter `foo(value)` getter `foo()` pattern for some properties, but not appearing in tests.

## Solution

Extend tests.


abetomo (approved):
+1

Issue #1370, opened 2020-10-11
Add method to get help option description


Melissa0x1f992:
Something like `helpDescription()`?  And maybe when someone uses `helpOption(..., 'custom description')`, subsequent calls will return the new description?

Context: I'm using `helpOption(false)` to disable the implicit help because I have a non-console help output, and I'd rather not display help redundantly on the console.  I use `helpInformation()` to get the help string in the action of my own `'-h, --help'` option.  I'd like my help description to match commander's default help option description.  My current options are just to hardcode a copy of your string, or access the `_helpDescription` field.  I've opted for the former for now, but I'd rather not have a random copy of data when the intent was to refer to the already-existing data.

shadowspawn:
Best approach currently is a copy of the string, as you have done. 

There is a big refactor of the help in progress in #1365, but it does not provide access to the default descriptions.

Will be interesting to see if this gets upvotes, now `helpOption(false)` is possible. I don't think I would add access to the default help option description alone as this string is one of many: could also want help option flags, help command description, version option flags, version option description...

shadowspawn:
Improving support for non-console help output in #1387

shadowspawn:
Improved support for redirected output in Commander 7.0.0.

PR #1365, opened 2020-09-28
Refactor help internals into separate interface/class

shadowspawn:
# Pull Request

Major help refactor (see also earlier experimentation in #1338, #1346). 

## Problem

The help generation is an organic collection of methods with custom logic embedded at various levels, and inconsistent handling of arguments, options, and commands. 

- the code is messy to maintain and extend
- it isn't easy for programs to modify or build on the built-in processing
    - nice example of complex help: https://github.com/tj/commander.js/pull/1338#issuecomment-683239306
- there isn't a natural place in API to add functionality, like sorting the help

Related issues:
- #1338 custom help
- #625 #910 sorting
- #301 #621 color

## Solution

Move the help related routines into a separate class. Inspired by discussion and examples from @cravler: https://github.com/tj/commander.js/pull/1346#issuecomment-689476637

Implement the Help methods as if they are static to allow override not only with subclass (see `.createHelp()`), but also with plain  functions (see `.configureHelp()`).

The impact on existing programs only using public methods should be small. All of the methods which are moved or deleted in the refactor were private, except for `Option.fullDescription` which is new in v7 so not released yet.

Callers can:
- subclass Help for complex cases, such as styling output with colour
- configure Help for simpler cases

For example:

```js
program.configureHelp({
   sortSubcommands: true,
   subcommandTerm: (cmd) => cmd.name()
});
```

## ChangeLog

- (added) refactor the code generating the help into a separate public Help class
    - support sorting subcommands and options in help
    - support specifying wrap width (columns)
    - allow subclassing Help class
    - allow configuring Help class without subclassing
- (changed) initialise the command description to empty string (previously undefined)
- (fixed) in the help the first line of command description was wrapping two characters early
- (fixed) in the help the pad width calculation was not including help option and help command
- (fixed) in the help the pad width calculation was including hidden options and commands

PR #1371, opened 2020-10-14
Output stream

Melissa0x1f992:
# Pull Request
Add custom output streams to commands

## Problem
stdout isn't available to my user.  I have a custom method of displaying text output to the user.  I still want to make use of the full Commander help suite.  Disabling the built-in help commands and writing wrappers becomes an increasingly complicated task when you want to make use of the entire help suite.  

## Solution
This PR will solve that problem by removing the assumption that all output is to `process.stdout`.  A user can instead pass a writeable stream that will replace the default stdout stream.  In this way, anyone can output command results in any way that works for their use case by defining their own writeable stream.

This is a more elegant solution to the problem in #1370.  Instead of using `helpOption(false)` and writing my own help option, I can use Commander's help option, and handle the output with my own writeable stream.  This is a less invasive solution than #779, as it leaves the implementation of the writeable stream to the user.

**Things to note:**
* I've intentionally not updated the README or the examples/. This is a WIP PR, and I'd like to know if this idea will be accepted before documenting it further.
* This adds a dependency on `stream.Writeable`, but that's built into Node, so I think that's fine
* `process.stdout` is a `stream.Writeable`, but it's also a `tty.WriteStream`.  Only the latter has the `.columns` field in its API.  Nonetheless, I've chosen to work with `stream.Writeable` for a few reasons.
First, `tty.WriteStream` requires a file descriptor passed to its constructor, which is too literal/constraining of an implementation for this idea; not every Commander output stream should need a file descriptor.  
Second, all of Commander's uses of `.columns` are already protected by a default if the value is Falsey. 
Finally, if someone wants a non-default column width in their custom stream, they can specify a `.columns` field themself.  
* An alternative way to address the above point would be to create a type that's just a `stream.Writeable` + `.columns`.  I think this alternative adds unneeded complexity.  Also, under this alternative, I don't know how the default `process.stdout` would fit in, as it wouldn't be of the custom type, even if it was compatible.  However, this alternative would have the pro of not relying on the existing `.columns || 80` in the code to prevent undefined behavior.
* Error output still prints to `process.stderr` by way of `console.error()`.  For my use case, I have no problem with this, but it would be straightforward to expand with an `errorStream()` method and field if we wanted.  
* Similarly, currently Commander doesn't use `console.log()` for standard output.  From this point forward, that would change from an implementation detail to a requirement.

## ChangeLog
* allow overriding the output from default (`process.stdout`) to any `stream.Writeable`

### DIFF
--- a/index.js
+++ b/index.js
@@ -128,6 +128,7 @@ class Command extends EventEmitter {
     this._exitCallback = null;
     this._aliases = [];
     this._combineFlagAndOptionalValue = true;
+    this._outputStream = process.stdout;
 
     this._hidden = false;
     this._hasHelpOption = true;
@@ -198,6 +199,7 @@ class Command extends EventEmitter {
     cmd._storeOptionsAsProperties = this._storeOptionsAsProperties;
     cmd._passCommandToAction = this._passCommandToAction;
     cmd._combineFlagAndOptionalValue = this._combineFlagAndOptionalValue;
+    cmd._outputStream = this._outputStream;
 
     cmd._executableFile = opts.executableFile || null; // Custom name for executable file, set missing to null to match constructor
     this.commands.push(cmd);
@@ -684,6 +686,20 @@ Read more on https://git.io/JJc0W`);
     if (this.options.length) {
       throw new Error('call .storeOptionsAsProperties() before adding options');
     }
+    return this;
+  };
+
+  /**
+   * The Stream to which this command writes
+   * normal output (help, usage, etc.).
+   *
+   * @param {tty.WriteStream} writeStream
+   * @returns {Command} `this` command for chaining
+   * @api public
+   */
+  outputStream(writeStream) {
+    if (writeStream === undefined) return this._outputStream;
+    this._outputStream = writeStream;
     return this;
   };
 
@@ -1287,7 +1303,7 @@ Read more on https://git.io/JJc0W`);
     this._versionOptionName = versionOption.attributeName();
     this.options.push(versionOption);
     this.on('option:' + versionOption.name(), () => {
-      process.stdout.write(str + '\n');
+      this._outputStream.write(str + '\n');
       this._exit(0, 'commander.version', str);
     });
     return this;
@@ -1500,7 +1516,7 @@ Read more on https://git.io/JJc0W`);
 
   optionHelp() {
     const width = this.padWidth();
-    const columns = process.stdout.columns || 80;
+    const columns = this._outputStream.columns || 80;
     const descriptionWidth = columns - width - 4;
     function padOptionDetails(flags, description) {
       return pad(flags, width) + '  ' + optionalWrap(description, descriptionWidth, width + 2);
@@ -1542,7 +1558,7 @@ Read more on https://git.io/JJc0W`);
     const commands = this.prepareCommands();
     const width = this.padWidth();
 
-    const columns = process.stdout.columns || 80;
+    const columns = this._outputStream.columns || 80;
     const descriptionWidth = columns - width - 4;
 
     return [
@@ -1573,7 +1589,7 @@ Read more on https://git.io/JJc0W`);
       const argsDescription = this._argsDescription;
       if (argsDescription && this._args.length) {
         const width = this.padWidth();
-        const columns = process.stdout.columns || 80;
+        const columns = this._outputStream.columns || 80;
         const descriptionWidth = columns - width - 5;
         desc.push('Arguments:');
         desc.push('');
@@ -1636,7 +1652,7 @@ Read more on https://git.io/JJc0W`);
     if (typeof cbOutput !== 'string' && !Buffer.isBuffer(cbOutput)) {
       throw new Error('outputHelp callback must return a string or a Buffer');
     }
-    process.stdout.write(cbOutput);
+    this._outputStream.write(cbOutput);
     this.emit(this._helpLongFlag);
   };
 
--- a/index.js
+++ b/index.js
@@ -693,7 +693,7 @@ Read more on https://git.io/JJc0W`);
    * The Stream to which this command writes
    * normal output (help, usage, etc.).
    *
-   * @param {tty.WriteStream} writeStream
+   * @param {stream.Writeable} writeStream
    * @returns {Command} `this` command for chaining
    * @api public
    */
--- a/tests/command.outputStream.test.js
+++ b/tests/command.outputStream.test.js
@@ -1,0 +1,14 @@
+const commander = require('../');
+const stream = require('stream');
+
+test('when command output stream is not set then default is process.stdout', () => {
+  const program = new commander.Command();
+  expect(program.outputStream()).toBe(process.stdout);
+});
+
+test('when set command output stream then command output stream is set', () => {
+  const program = new commander.Command();
+  const customStream = new stream.Writable();
+  program.outputStream(customStream);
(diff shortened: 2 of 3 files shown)


shadowspawn:
A couple of things are in flight that are relevant. I'll be interested to dig through the things you considered.

I played around with custom stream idea when working on #1296 and settled for a light weight option that just have a simple "log" routine. I left supplying a custom logger undocumented as it wasn't the focus of the work, and I didn't have use cases to check it against.

There is a big refactor fo the help in #1365. This will make it more practical to override help to change behaviour

cravler:
For me it will be cool if we have something like this:

``` js
class Command {

  /**
   * Return output stream
   * https://nodejs.org/api/stream.html#stream_class_stream_writable
   *
   * @returns {stream.Writable}
   */

  outputStream() {
    return process.stdout;
  };

  /**
   * Return error stream
   * https://nodejs.org/api/stream.html#stream_class_stream_writable
   *
   * @returns {stream.Writable}
   */

  errorStream() {
    return process.stderr;
  };

  /**
   * @param {string} message human-readable description of the error
   * @return never
   * @api private
   */

  _error(message) {
    if (!this._exitCallback) {
      this.errorStream().write('error: ' + message + '\n');
    }
  };
}
```

shadowspawn:
Some current code patterns that complicate the output customisation are:
- help may be displayed on stdout or stderr (new in 7)
- help is displayed with `process.stdout.write` (and stderr), while errors are displaying using `console.error`
- `.helpInformation()` has a trailing linefeed (and displayed using `write`)

I was tempted to remove the trailing linefeed from `.helpInformation()` when it was made public so it could more conveniently be displayed with either log or write, but kept it the same at the time for backwards compatibility.

cravler:
for example, if I want show all errors with date prefix and in red color:

``` js
const { Writable } = require('stream');

const errStream = new Writable({
  write(chunk, encoding, callback) {
    const str = chunk.toString();
    const prefix = '[' + (new Date()).toISOString() + '] ';
    const applyStyle = str => '\x1b[31m' + str + '\x1b[0m'; // red
    process.stderr.write(applyStyle(prefix + str));
    callback();
  }
});
```

Melissa0x1f992:
> There is a big refactor for the help in #1365. This will make it more practical to override help to change behaviour

Looking at this refactor, I think the `out` and `err` stream customization would get used in the `_getHelpContext()` method.  Would it be easier if this commit had branched from something closer to that code?  Lesson learned that I should have developed from the `develop` branch 😅

I think @cravler's example shows how letting the user provide their own stream opens up a lot of customization without having to bloat Help with features.  My use case involves displaying text to users completely away from a terminal environment.  It'd be out of scope to have Commander directly support such diverse use cases, but just by accepting `stream`s from the user, we get the ability to integrate Commander to a broad set of use cases.

I agree that it's better to replace `console.error()` with `.write()`ing to an `err` stream.  That would keep the console output story consistent.

shadowspawn:
I prefer duck typing to an explicit class like `stream.Writeable`. Bit torn whether to allow both `console.log()` and `process.stdout.write()` styles, implicit line-feed and explicit line-feed. Over-the-top support would be for plain function or object-with-function, with either style! For example:

```js
program.setOutput({
  log: (str) => console.log(str),
  error: (str) => console.error(str)
});

program.setOutput({
  log: console, // uses console.log
  error: console // uses console.error
});

program.setOutput({
  writeOut: (str) => process.stdout.write(str),
  writeErr: (str) => process.stderr.write(str)
});

program.setOutput({
  writeOut: process.stdout, // uses process.stdout.write
  writeErr: process.stderr // uses process.stderr.write
});
```

shadowspawn:
Hmm. I wanted to try out the object-with-function approach to mirror passing a class/object, but at least in my attempt, I think it is less clear than a simple function and didn't add much value!

cravler:
I vote for this:

``` js
const util = require('util');

program.setOutput({
  log: (...args) => process.stdout.write(util.format(...args, '\n')), //same as console.log
  error: (...args) => process.stderr.write(util.format(...args, '\n')) //same as console.error
});
```

shadowspawn:
`columns` and `tty` got mentioned in first comment, and I realise there are a few issues in `release/7.x`. Old code is using `stdout.columns` even when displaying on `stderr` now, and should probably test `isTTY` too. I haven't worked out how to improve the default behaviour yet. Hopefully solution will mean this PR does not need to directly deal with it.

shadowspawn:
I'm going to have play around with this signature. Because of the changes to help, I think people doing custom solutions affecting the help will need to supply both `stdout` replacement and `stderr` replacement. The columns property will allow number or function, so can plug in runtime calculation if needed. 

```js
outputOverride({
  log: (str) => console.log(str),
  error: (str) => console.error(str),
  columns: 120
});
```

(This could be used to add timestamp to error calls as in the example above, but that would affect help being displayed on stderr too. A bottleneck routine for non-help errors is probably easier for that.)

cravler:
> I'm going to have play around with this signature. Because of the changes to help, I think people doing custom solutions affecting the help will need to supply both `stdout` replacement and `stderr` replacement. The columns property will allow number or function, so can plug in runtime calculation if needed.
> 
> ```js
> outputOverride({
>   log: (str) => console.log(str),
>   error: (str) => console.error(str),
>   columns: 120
> });
> ```
> 
> (This could be used to add timestamp to error calls as in the example above, but that would affect help being displayed on stderr too. A bottleneck routine for non-help errors is probably easier for that.)

For me it looks wrong to write help to `stderr`, help must always go to `log` and errors to `error` and if it is needed I can override `log` to use `stderr`.

cravler:
For version output we can add:

```js
outputOverride({
  printVersion: (version) => console.log(version)
});
```

shadowspawn:
> For me it looks wrong to write help to stderr, help must always go to log and errors to error and if it is needed I can override log to use stderr.

In the normal case when the user requests the help, yes it goes to stdout. 

When the help is displayed in response to a usage error it is suggested to use stderr so the help does not get piped into another command or into a file as though it were valid command output. It is being displayed as part of a diagnostic message for the usage error:

- https://www.jstorimer.com/blogs/workingwithcode/7766119-when-to-use-stderr-instead-of-stdout
- https://unix.stackexchange.com/questions/8813/should-the-usage-message-go-to-stderr-or-stdout

shadowspawn:
Naming: or maybe configureOutput to match configureHelp.

cravler:
Sorry, maybe i explain wrongly. Maybe this will be more clear:

```
configureOutput({
  help: (str, error = true) => console[error ? 'error' : 'log'](str),
  error: (str) => console.error(str),
  version: (str) => console.log(str)
});

**Comment by shadowspawn** [comment]
Oh, I see what you mean now. I was thinking of log = console.log = stdout , and error = console.error = stderr. Splitting up the types of output is another way.

**Comment by shadowspawn** [comment]
My current experiment to allow overriding output and being careful about columns has:

```js
    this._outputConfiguration = {
      write: (arg) => process.stdout.write(arg),
      writeError: (arg) => process.stderr.write(arg),
      // columns is used for wrapping the help
      getColumns: () => process.stdout.isTTY ? process.stdout.columns : undefined,
      getErrorColumns: () => process.stderr.isTTY ? process.stderr.columns : undefined
    };
```

shadowspawn:
This PR was a good trigger for discussion and development. Closing in favour of #1387, with a key difference being weaker typing than using an explicit stream.

PR #1372, opened 2020-10-18
Include tests for supplying optional  param to .hideHelp

shadowspawn:
# Pull Request

## Problem

Was not testing the less common use of `.hideHelp()` with an explicit parameter.

## Solution

Add tests for `.hideHelp(true)` and `.hideHelp(false)`.


abetomo (approved):
👍

shadowspawn:
Ha, that was quick! 🏎️
Thanks @abetomo

PR #1380, opened 2020-10-24
Feature createOption

shadowspawn:
# Pull Request

This adds `.createOption()` to support Option subclassing, like `.createCommand()` and `.createHelp()`. Probably less likely to be used than those but follow the same pattern and make it possible!


abetomo (approved):
👍

shadowspawn:
Included in v7.0.0-1: https://github.com/tj/commander.js/releases/tag/v7.0.0-1

PR #1384, opened 2020-10-24
Give optional boolean parameters appropriate name, and coerce value to boolean

shadowspawn:
# Pull Request

One of the patterns we use is `.allowUnknownOption(param)` where param is optional. The parameters just had generic names, either `value` or `arg`.

- give parameters appropriate names
- coerce value to boolean so stored property value is always just a boolean

The coerce matches how I think about what the routine does, but does not make any practical difference in expected use. Happy to drop if you find odd!

### DIFF
--- a/index.js
+++ b/index.js
@@ -401,24 +401,24 @@ class Option {
   /**
    * Whether the option is mandatory and must have a value after parsing.
    *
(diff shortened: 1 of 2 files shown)


abetomo (approved):
It's easier to understand.

PR #1385, opened 2020-10-25
Update testing node versions, supported node versions, and archive 4.x CHANGELOG entries

shadowspawn:
# Pull Request

- move 4.x CHANGELOG entries to separate file
- update supported versions and dates in SECURITY
- update target versions for testing

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -9,7 +9,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        node-version: [10.x, 12.x, 14.x]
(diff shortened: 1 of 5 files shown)


.github/workflows/tests.yml
```diff
@@ -9,7 +9,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        node-version: [10.x, 12.x, 14.x]
+        node-version: [10.x, 12.x, 14.x, 15.x]
```
abetomo, line 12: 👍

abetomo (approved):
Thanks.

Issue #1388, opened 2020-11-03
Tests fail if directory has a space


shadowspawn:
```
 FAIL  tests/command.default.test.js
 FAIL  tests/command.executableSubcommand.lookup.test.js

Test Suites: 2 failed, 42 passed, 44 total
Tests:       15 failed, 356 passed, 371 total
```

Example of unsafe code in tests:

```js
const { stdout } = await execAsync(`node ${pm} --an-option`);
```

Uncovered when suggested adding Commander back into `citgm`:

- https://github.com/nodejs/citgm/issues/838

shadowspawn:
Fix released in 6.2.1

PR #1390, opened 2020-11-03
Switch exec to execFile to simply protect against spaces in path

shadowspawn:
# Pull Request

## Problem

Tests break if path has spaces.

Related issue: #1388

## Solution

The problem is in the test code, not in Commander itself.

Switch from `exec` to `execFile` so shell not processing arguments.


abetomo (approved):
👍

Issue #1207, opened 2020-03-03
Additional parameters for coercing functions


vonagam:
If you want to throw an error in a coercing function because a received string value does not satisfy some requirement you might want to have access to an option name (or a command name) to customise an error message.

How about changing coerce signature?
```ts
// from
(value: string, previous: T) => T
// to
(value: string, previous: T, option: Option, command: this) => T
```

One will be able to use `long`, `name()` or `attributeName()` from `Option` to customise a message. Also `negate` property is useful. None of those are accessible right now.

`command` can be used to get its name for an error or currently parsed options if a person wants to change coercing behaviour based on already present options. User already has access to command, so nothing new in terms of access, just convenience.

shadowspawn:
This is an interesting suggestion thanks, and aligns with the name shortcoming I noticed when I was writing a custom validator recently: #518 #1130 

Fair warning though, we are not going to rush a change. The previous changes to the parameters for the coercion function caused a lot of problems, in particular by breaking `parseInt`: #943 (The _next_ change should hopefully be less problematic!) 

The support for a RegEx to validate the option values has been removed from the documentation as underdelivering, and I commented there I thought the custom option processing was a better longterm approach: #967

Providing an array of allowed values for an option (#518) is one of the two most upvoted open issues.

Interested in seeing interest in this, and whether it looks a good approach for allowing the custom option processing to solve questions that come up.

shadowspawn:
There are more possibilities in Commander v7 with one coercing function moved down into the Option: #1331

This issue has not had any activity in over six months. It isn't likely to get acted on due to this report.

Feel free to open a new issue if it comes up again, with new information and renewed interest.
 
Thank you  for your contributions.

PR #1392, opened 2020-11-07
Support custom error messages for option argument coercion failures

shadowspawn:
# Pull Request

## Problem

It can be tricky to do nice error messages for custom option processing failures since the option name is not available.

Issues: #1207 https://github.com/tj/commander.js/issues/518#issuecomment-721932654

## Solution

We already added a custom exception to process the error message for the new `.choices()` routine. This takes that a little further to make the exception usable by custom option processing routines and not just Option methods.

For example:

```js
const program = commander.program;

program
  .option('-p, --port <port>', 'port number', (arg) => {
    const portNumber = parseInt(arg);
    if (isNaN(portNumber)) {
      throw new InvalidOptionArgumentError('Must be a number.');
    }
    return portNumber;
  });

program.parse();
```

```sh
options % node throw.js -p       
error: option '-p, --port <port>' argument missing
options % node throw.js -p 123   
options % node throw.js -p Madrid
error: option '-p, --port <port>' argument 'Madrid' is invalid. Must be a number.
```
## Release Note

- Added: throw`InvalidOptionArgumentError` from custom option processing for detailed error message.

## To Do

- [x] README
- [x] tests
- [x] example
- [x] TypeScript


shadowspawn:
FYI: @lydell @vonagam

lydell:
Looks great! Exactly what I need. Though maybe a little more low-level than I anticipated.

Two thoughts:

- I think this should be mentioned in the readme docs.

- In my project, I would define the following helper function:

```js
function invalidArgument(message) {
  return new commander.CommanderError(1, 'commander.optionArgumentRejected', message);
}

// Later, somewhere:
throw invalidArgument('Must be a number.')
```

Do we want a shorter way to do this out of the box? I’m thinking you can’t customize the exit code for other errors, such as “missing option”, right? So maybe we don’t need that ability here. Is there a use case for using any other code than `'commander.optionArgumentRejected'`?

shadowspawn:
> Though maybe a little more low-level than I anticipated.

Fair comment, and good feedback thanks. I was trying a quick repurpose of the existing internals. 😊

shadowspawn:
Added somewhat higher level wrapper:

```js
  .option('-p, --port <port>', 'port number', (arg) => {
    const portNumber = parseInt(arg);
    if (isNaN(portNumber)) {
      throw new commander.InvalidOptionArgumentError('Must be a number.');
    }
    return portNumber;
  })

```

lydell:
Looks good!

shadowspawn:
Finished missing pieces, ready for review.

typings/index.d.ts
```diff
@@ -14,6 +14,10 @@ declare namespace commander {
   }
   type CommanderErrorConstructor = new (exitCode: number, code: string, message: string) => CommanderError;
 
+  interface InvalidOptionArgumentError extends CommanderError {
+  }
+  type InvalidOptionArgumentErrorConstructor = new (message: string) => InvalidOptionArgumentError;
```
vonagam, line 19: Question: why in commander typings there are separations of class types into interfaces and constructor types?

So here instead of:
```ts
class InvalidOptionArgumentError extends CommanderError {
  constructor(message: string);
}
```
There is:
```ts
interface InvalidOptionArgumentError extends CommanderError {
}
type InvalidOptionArgumentErrorConstructor = new (message: string) => InvalidOptionArgumentError;
```

(To get a constructor type from a class variant - `typeof InvalidOptionArgumentError`.)

shadowspawn: The pattern is to allow export of both a global object and some types. See #1081

vonagam: So, i cloned [fson/commander-ts-repro](https://github.com/fson/commander-ts-repro) and could not reproduce the issue. Then i changed typescript dependency there from `^3.6.3` to simply `3.6.3` and i saw the expected error. With `^3.6.3` typescript was resolving to `3.9.7`.

Is it possible that the issue was fixed on typescript end and this design is no longer needed?

shadowspawn: Yes it is possible has been fixed or improved in TypeScript, I have not checked again since the original problem! I am planning to rework the exports and typings when ESM support is out of prototype in node.

abetomo (approved):
I appreciate it.

PR #1395, opened 2020-11-19
Rework wrap to have simpler usage pattern

shadowspawn:
# Pull Request

The `wrap` routine is now exposed and should make sense alone.

## Problem

The wrap routine was a bit hard to understand in isolation or use for new calls. It does produce
a column of `width` characters excluding the indent, but the column looks misaligned in the result. 

Conceptually:

```
      1234567890
term  d e s c r i p t i o n
123456


wrap('d e s c r i p t i o n', 10, 6):
d e s c r
     i p t i o n

put back together to form

term  d e s c r i
      p t i o n

```

## Solution

Pass in the full text. The width is now the total line length to wrap at, so closely related to the terminal width. The resulting string looks wrapped.

```
12345678901234567890
term  d e s c r i p t i o n

wrap('term  d e s c r i p t i o n', 16, 6):
term  d e s c r
      i p t i o n
```

### DIFF
--- a/index.js
+++ b/index.js
@@ -245,12 +245,11 @@ class Help {
     const termWidth = helper.padWidth(cmd, helper);
     const columns = helper.columns || 80;
     const itemIndentWidth = 2;
-    const itemSeparatorWidth = 2;
-    // itemIndent term itemSeparator description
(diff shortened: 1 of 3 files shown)


abetomo (approved):
👍

PR #1396, opened 2020-11-20
Rename columns to helpWidth

shadowspawn:
# Pull Request

## Problem

I named the Help property which determines help wrapping `columns`, since the default value is `process.stdout.columns` or `process.stderr.columns`. But this does not describe what it is used for in Help or what custom implementations should be supplying. Naming it based on the purpose could make it clearer for supplying custom values.

## Solution

Rename:
- `columns` -> `helpWidth`
- `getOutColumns()` -> `getOutHelpWidth()`
- `getErrColumns()` -> `getErrHelpWidth()`

## Release Note

This is a breaking change from the first prerelease, but does not need to appear in the final release notes.

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -632,7 +632,7 @@ You can configure the Help behaviour by modifying data properties and methods us
 
 The data properties are:
 
-- `columns`: specify the wrap width, useful for unit tests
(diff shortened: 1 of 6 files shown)


abetomo (approved):
That's nice!

PR #1397, opened 2020-11-20
Add missing param in TypeScript for helpInformation

shadowspawn:
# Pull Request

## Problem

The optional parameter for `.helpInformation()` was not in the TypeScript definition.

## Solution

Add definition.


abetomo (approved):
👍

Issue #1268, opened 2020-05-28
Extra arguments : document and add configuration to fail if they exist


Lectem:
By default it seems that any extra argument will be available in an array after the command object.

The issue comes from the fact that it is not a documented behavior, and if the callback does not have this additional array parameter, it will silently be executed.

While this can be useful if you want to parse those extra arguments yourself, I think the default should be to error out when some arguments are not parsed (or at least, give it as an config option).

Right now, one must do the following:

```js
.command('somecommand <filePath>')
.action((filePath, cmdObj, additionalArgs) => {
        if (additionalArgs && additionalArgs.length > 0)
        {
            cmdObj.help();
        }
        // ...
});
```

It would be nice if something like this could have the same effect:

```js
.command('somecommand <filePath>')
.noExtraArgs(true)
.action((filePath) => {
        // ...
});
```

shadowspawn:
Related issue, which prompted some improvements: #749

I want the error detection myself in one of my own programs. 

The arguments are (now) available through `cmdObj.args`. Are the additional arguments passed to the action handler worthwhile as a documented pattern, or a bit clumsy and not needed as there is an alternative? (I wasn't entirely happy with the pattern at the time I added the extra parameters and left them undocumented to see whether they would prove useful.)

Lectem:
> Are the additional arguments passed to the action handler worthwhile as a documented pattern, or a bit clumsy and not needed as there is an alternative?

I do not consider `cmdObj.args` as a direct alternative since it contains all arguments, not only the extra arguments.
I'm not really in favor of one method or another as I don't have the need for those extra arguments except to validate there aren't any (the `noExtraArgs` config). 
I think it kind of makes sense to have an additional parameter to the callback, but I would probably return an empty array instead of `undefined` when there are no extra elements. And/or a method `cmdObj.extraArgs` could wrap `cmdObj.args.slice()` I suppose?

shadowspawn:
Related: #259

lydell:
I created a helper function for it:

```js
function handleTooManyArgs(action) {
  return (...args) => {
    if (args.length < 2) {
      action(...args);
    } else {
      // The arguments to Commander actions are:
      // expectedCliArg1, expectedCliArg2, expectedCliArgN, Cmd, restCliArgs
      const rest = args[args.length - 1];
      if (rest.length > 0) {
        const expected = args.length - 2;
        const s = expected === 1 ? '' : 's';
        console.error(
[...]

PR #1399, opened 2020-11-21
WIP: Reject excess arguments by default

shadowspawn:
# Pull Request

## Problem

By default, excess arguments are silently ignored. Checking manually is a bit tricky.

See:  #259 #749 #1000 #1268 

## Solution

- add a check for excess arguments before calling action handler
- `.allowExcessArguments()` is like `.allowUnknownOption()` to suppress the new error
- `.command('*')` arguments are not checked to maintain legacy behaviour

Other changes:
- switch action handler from trailing parameters of "options/command" and undocumented "extraArgs", to "options/command" and "command". This will hopefully make it easier to switch to safer options stored separately.

## To Do

- `command:*` listener
- single commands with no action handler
- default command
- refine messaging (not happy with "excess")

TypeScript, tests, README et al

## ChangeLog

- excess arguments now cause an error (#1399)

### DIFF
--- a/index.js
+++ b/index.js
@@ -883,21 +883,20 @@ class Command extends EventEmitter {
 
   action(fn) {
     const listener = (args) => {
-      // The .action callback takes an extra parameter which is the command or options.
       const expectedArgsCount = this._args.length;
       const actionArgs = args.slice(0, expectedArgsCount);
+      if (args.length > expectedArgsCount && this._actionHandler) {
+        this.excessArguments(args.slice(expectedArgsCount));
+      }
+
+      // Add the options (which might be the command, with options as properties!)
       if (this._passCommandToAction) {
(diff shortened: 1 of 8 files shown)


shadowspawn:
I tried doing a couple of things at once here which I am thinking of as somewhat related. However, they are individually complicated and not certain to work out well. I'll close this and have another go at one thing at a time.

shadowspawn:
I am finding this is breaking quite a lot of my unit tests (due to lazily not declaring arguments) and not picking up actual problems. I think for Commander v7 it would be better to make this an opt-in behaviour that is inherited to subcommands (so can easily turn it on for whole program). Command 7 has breaking changes with options storage and action handler parameters, so plenty for users to deal with!

I'll think a bit more and do a PR.

PR #1400, opened 2020-11-23
WYSIWYG sort

shadowspawn:
# Pull Request

## Problem

There are lots of things we might consider when sorting the options in the help: short flags vs long flags, help flags, and how to sort negated flags. Using the `name` or `attributeName` is tempting but is not what is displayed in the help.

Having an easy to understand method is probably better than having a clever method!

## Solution

I looked at the help for `tar`, which sorts on short flags over long flags, and sorts `--no-foo` as `no-foo` rather than with `foo`.

This is straight forward and matches the way we display the flags in the help, with each line starting with short flags before long flags. Negated options and help options are not treated specially.

### DIFF
--- a/index.js
+++ b/index.js
@@ -68,11 +68,7 @@ class Help {
     }
     if (this.sortOptions) {
       visibleOptions.sort((a, b) => {
-        const compare = a.attributeName().localeCompare(b.attributeName());
-        if (compare === 0) {
-          return (a.negate) ? +1 : -1;
-        }
-        return compare;
+        return a.name().localeCompare(b.name());
       });
     }
     return visibleOptions;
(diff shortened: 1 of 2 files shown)


abetomo (approved):
LGTM

shadowspawn:
For interest, a test file I used:

```js
program
  .configureHelp({ sortOptions: true })
  .option('-a')
  .option('-c,--zebra')
  .option('-d,--yellow')
  .option('--Bad')
  .option('-c')
  .option('-C')
  .option('-1')
  .option('--no-alone')
  .option('--foo')
  .option('--no-foo')
  .option('--lamba')
  .option('-A')
  .option('--beta');
```

```sh
$ node sort.js -h
Usage: sort [options]

Options:
  -1
  -a
  -A
  --Bad
  --beta
  -c,--zebra
  -c
  -C
  -d,--yellow
  --foo
  --lamba
  --no-alone
  --no-foo
```

Issue #1402, opened 2020-11-26
`command:<name>` event doesn't bubble to the parent


grabbou:
Hey,

Not sure if I am using this correctly, but it does seem to be a good use case and not far away from working. 

My program is as follows:
```
const config = program.commad('config');
const start = program.command('start');

program.on('command:*', () => /* handle unexpected command */);

program.on('command:config', () => /* never gets called */);
```

In the above example, when `config` command is executed, the `command:config` listener never gets executed. 

After console logging around the Commander source code, I have found out that the reason this doesn't work is that listener is registered on `parent` (in this context, main program) whereas the `emit` function is called on a subcommand (`config`). 

Changing `this.emit` to `this.parent.emit` fixes it https://github.com/tj/commander.js/blob/master/index.js#L1003

Is there any way to bubble events up? It seems to be quite natural to listen to `command:<name>` on the main object, otherwise, there's no difference from attaching a listener to `config` vs tweaking action handler.

shadowspawn:
In your example, the emit for `command:config` should indeed be called on `program`. After some investigation, I think this is a bug introduced in Commander v5. The refactor of the code changed where the emit is called.

shadowspawn:
Fix included in Commander 7.0.0.

PR #1403, opened 2020-11-29
Improve backwards compatibility for command events

shadowspawn:
# Pull Request

## Problem

The custom command event (e.g. `'command:example'`) was being emit-ed from the wrong object to be useful and for backwards compatibility with Commander v4 and earlier.

See: #1402

## Solution

The command events used to be how Commander triggered the action handler, and custom listeners were somewhat of a free bonus. Now the parsing is rather different and it requires extra code to approximate the old behaviour.

There have never been explicit tests for the events. I have added some tests so at least the behaviour is more stable going forward. 

(From code inspection, there may be some minor differences in edge cases from historical behaviour which I won't be attempting to match, like what event name is generated for aliases).

## ChangeLog

- improve backwards compatibility for custom command event listeners (#1403)

### DIFF
--- a/index.js
+++ b/index.js
@@ -1490,6 +1490,7 @@ Read more on https://git.io/JJc0W`);
         this.unknownOption(parsed.unknown[0]);
       }
 
+      const commandEvent = `command:${this.name()}`;
       if (this._actionHandler) {
         const args = this.args.slice();
         this._args.forEach((arg, i) => {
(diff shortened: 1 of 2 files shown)


abetomo (approved):
LGTM

shadowspawn:
Thanks @abetomo 👓

PR #1404, opened 2020-11-29
WIP: CommandStrict

shadowspawn:
# Pull Request

## Problem

1) We are slow to adopt new best practices which break backwards compatibility. However, it isn't easy for new programs to start with best practices. Likewise it is not easy to switch existing programs to follow what we consider current best practices as they are now and change in future.

2) We get weak typings in TypeScript because the Command object may have options stored as properties, meaning TypeScript is not able to warn about typos etc.

The big example of a new best practice is `.storeOptionsAsProperties(false)`.

## Solution

Add a CommandStrict class. The implementation sets our best practice configuration. The TypeScript typings do not include options as properties.

```js
const { CommandStrict } = require('commander);
const program = new CommandStrict();
```

In the future this class might also error if deprecated code is used, but not looking at that now.

Alternatives considered: I did try adding a `.strict()`  method in an earlier attempted solution, but that did not help with the TypeScript typings.


cravler:
Better to start using `strict` mode by default with version 7. Mostly `major` updates are non-compatible, meaning consumers can not upgrade without changing their software. And just create example for BC.

shadowspawn:
Hmm, interesting feedback @cravler , thanks! I have been putting off actually changing the defaults. I'll think about that.

shadowspawn:
On hold, looking at changing default behaviour: #1409

PR #1406, opened 2020-12-01
Correctly display default value Infinity in optionHelp instead of null

fnogatz:
# Pull Request

## Problem

If an option has a default value of `Infinity`, it is printed as `null` in the help, because `JSON.stringify(Infinity) === null`.

Test program:

```js
program.option('-m, --max <int>', 'max value', parseInt, Infinity);
program.parse(process.argv);
```

Current output:

```sh
> ./test.js --help
Usage: test [options]

Options:
  -m, --max <int>  max value (default: null)
  -h, --help       display help for command
```

## Solution

This PR adds a wrapper function `stringify` for the used `JSON.stringify` which checks for the values `Infinity` and `-Infinity`.

Corrected output:

```sh
> ./test.js --help
Usage: test [options]

Options:
  -m, --max <int>  max value (default: Infinity)
  -h, --help       display help for command
```

## ChangeLog

Fixed: correctly print default value `Infinity` in option help text

### DIFF
--- a/index.js
+++ b/index.js
@@ -1509,7 +1509,7 @@ Read more on https://git.io/JJc0W`);
     // Explicit options (including version)
     const help = this.options.map((option) => {
       const fullDesc = option.description +
-        ((!option.negate && option.defaultValue !== undefined) ? ' (default: ' + JSON.stringify(option.defaultValue) + ')' : '');
+        ((!option.negate && option.defaultValue !== undefined) ? ' (default: ' + stringify(option.defaultValue) + ')' : '');
       return padOptionDetails(option.flags, fullDesc);
     });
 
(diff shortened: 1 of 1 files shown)


shadowspawn:
Good description, thanks.

Commander v7 will allow custom display of a default value, see #1331. This was prompted by discussion and investigation around improving certain classes where the JSON stringify was not as human readable as direct toString (#1283), quite like your case. The attempt to generalise it did not work out and instead added a custom solution to give user full control.

I think the new `.default()` should cover your use case?

fnogatz:
I was not aware of the v7 branch :innocent:

However, if there is no `defaultValueDescription` set (via the second argument of `.default()`), it still uses `JSON.stringify()` internally. Theat means though my problem would be resolved by explicitly calling `.default(Infinity, 'Infinity')` in Commander v7, I would expect `.default(Infinity)` to handle this correctly. If you also think so, I could target my PR against the `release/7.x` branch instead, it's just the same change for the new `optionDescription` function.

shadowspawn:
Yes, I wasn't suggesting `Option.default()` automatically supported Infinity, but does support supplying a custom description to be displayed as you saw. I don't want to add special case handling of unusual values to Commander itself.

PR #1407, opened 2020-12-02
Add .allowExcessArguments() and error message

shadowspawn:
# Pull Request

## Problem

By default, excess arguments are silently ignored. Checking manually is a bit tricky.

See:  #259 #749 #1000 #1268 

## Solution

- add a check for excess arguments before calling action handler
- `.allowExcessArguments()` is like `.allowUnknownOption()` to enable/disable the new error
- off by default (for now), but will be on by default for `CommandStrict`, see #1404

Like missing arguments, only checked if there is an action handler.

## ChangeLog

- `.allowExcessArguments()` to control whether excess arguments are allowed for command (#1407)

## Later

I have not added to the README yet and will do that separately. (Hopefully in combination with `CommandStrict` if that works out.)


shadowspawn:
Example

```js

program
  .arguments('<file')
  .allowExcessArguments(false)
  .action(() => {
    console.log('Called main action handler')
  });

program
  .command('sub <first> <second>')
  .allowExcessArguments(false)
  .action(() => {
    console.log('Called subcommand action handler')
  });

program.parse();
```

```sh
$ node foo.js alpha beta
$ node foo.js sub alpha beta
Called subcommand action handler
$ node foo.js sub alpha beta gamma
error: too many arguments for 'sub'. Expected 2 arguments but got 3.
```

index.js
```diff
@@ -1170,6 +1171,17 @@ Read more on https://git.io/JJc0W`);
     return this;
   };
 
+  /**
+   * Allow excess arguments on the command line.
+   *
+   * @param {Boolean} [allowExcess] - if `true` or omitted, no error will be thrown
+   * for excess arguments.
+   */
+  allowExcessArguments(allowExcess) {
```
abetomo, line 1180: Are the default parameters available?

```js
allowExcessArguments(allowExcess = true) 
```

shadowspawn: Good question, I'll look into that. We have a few methods with this sort of pattern that could use it. Likely can because we have moved forward our minimum version of node.

shadowspawn: Defaults from node 6, which is our minimum supported version: https://node.green/#ES2015-syntax-default-function-parameters

shadowspawn: Added default parameter instead of manual test to multiple routines.

abetomo: Thank you for fixing it.

abetomo (approved):
Great!

PR #1409, opened 2020-12-03
Change to safe storage of options by default, and change action parameters

shadowspawn:
# Pull Request

## Problem

1) The default behaviour is to store option values as properties on the command. This is a frequent cause of problems with clashes with existing properties of Command (and EventEmitter).

- https://github.com/tj/commander.js/issues/933#issue-422111380 collected older issues
- https://github.com/tj/commander.js/issues/933#issuecomment-617410045 collected recent issues

2) In Commander v6 the action handler may get passed various combinations of parameters based on settings and unexpected args:
- expected args + command with options as properties (most common case)
- expected args + options
- expected args + command without options as properties
- any of above + array of extra args

This has not worked out as nicely as hoped for moving to safe options incrementally!  The various combinations make implementations inconsistent, make it hard to write examples that work whatever the settings, and make it harder to write write a generic action handler. 

## Solution

These are BREAKING changes for many existing programs. But hopefully the code changes required are small and easy!

1) Store the options values safely by default!! Effectively `.storeOptionsAsProperties(false)`. 

This affects especially use of program options.

2) This PR includes the proposal from #1405 to change the action parameters to consistently be options+command.

The array of extra args is gone, but one of the common uses was to show an error for excess/unexpected command-arguments. There is now support for Commander to display an error for excess command-arguments, and this has been turned on by default in this PR. See #1407

The method `.passCommandToAction()` is gone. No longer needed with this pattern, the options and command are always available as parameters.

-----

Lots of files changed:
- examples updated
- example output removed from several example files (replaced with "Try the following")
- updated tests for new behaviours and deleted routines

### Changed

- *Breaking:* options are stored safely by default, not as properties on the command
    - this especially affects accessing options on program, use `program.opts()`
    - revert behaviour with `.storeOptionsAsProperties()`
- *Breaking:* action handlers are passed options and command separately

### Added

- *Breaking:* error message if there are too many command-arguments on command line for the action handler
    - if should be allowed then declare extra arguments, or use `.allowExcessArguments()`

### Deleted

- *Breaking:* `.passCommandToAction()`
    - no longer needed as action handler is passed options and command
- *Breaking:* "extra arguments" parameter to action handler
    - if being used to detect excess arguments, there is now an error displayed by default

### Migration Tips

 The biggest change is the parsed option values. Previously the options were stored by default as properties on the command object, and now the options are stored separately.

 If you wish to restore the old behaviour and get running quickly you can call `.storeOptionsAsProperties()`. 
 To allow you to move to the new code patterns incrementally, the action handler will be passed the command _twice_,
 to match the new "options" and "command" parameters (see below).

 **program options**

 Use the `.opts()` method to access the options. This is available on any command but is used most with the program.

 ```js
 program.option('-d, --debug');
 program.parse();
 // Old code before Commander 7
 if (program.debug) console.log(`Program name is ${program.name()}`);
 ```

 ```js
 // New code
 const options = program.opts();
 if (options.debug) console.log(`Program name is ${program.name()}`);
 ```

 **action handler**

 The action handler gets passed a parameter for each command-argument you declared. Previously by default the next parameter was the command object with the options as properties. Now the next two parameters are instead the options and the command. If you
 only accessed the options there may be no code changes required.

 ```js
 program
   .command('compress <filename>')
   .option('-t, --trace')
   // Old code before Commander 7
   .action((filename, cmd)) => {
     if (cmd.trace) console.log(`Command name is ${cmd.name()}`);
   });
 ```

 ```js
   // New code
   .action((filename, options, command)) => {
     if (options.trace) console.log(`Command name is ${command.name()}`);
   });
 ```

 If you already set `.storeOptionsAsProperties(false)` you may still need to adjust your code.

 ```js
 program
   .command('compress <filename>')
   .storeOptionsAsProperties(false)
   .option('-t, --trace')
   // Old code before Commander 7
   .action((filename, command)) => {
     if (command.opts().trace) console.log(`Command name is ${command.name()}`);
   });
 ```

 ```js
    // New code
    .action((filename, options, command)) => {
       if (command.opts().trace) console.log(`Command name is ${command.name()}`);
    });
 ```

 **excess command-arguments**

 There is now an error if there are too many command-arguments on the command line (only checked if there is an action handler).
 If the extra arguments are supported by your command then you can either declare the expected arguments, or allow excess arguments.

 ```js
 // Old code before Commander 7
 program
   .action(() => {});
 program.parse(['a', 'b', 'c'], { from: 'user' }); // now causes an error
 ```

 ```js
 // New code, declare arguments
 program
   .arguments('[args...]')
   .action(() => {});
 ```

 ```js
 // New code, allow excess arguments
 program
   .allowExcessArguments()
   .action(() => {});
 ```


shadowspawn:
Finished first pass through code and README and examples, including migration tips in the CHANGELOG. I intend to update the Examples section in the README next.

shadowspawn:
Wondering whether the `config` option in `deploy` example is worthwhile. Global options are definitely usable, but a few rough edges.

shadowspawn:
The code changes are quite small, but PR touches a LOT of files with the new improved behaviour!

shadowspawn:
I have added a copy of the Migration Tips to the pull request text as it covers in detail the breaking changes.

abetomo (approved):
LGTM.
awesome!

shadowspawn:
Thanks! I'll do a new prerelease with this.

shadowspawn:
Fixed, thanks.

PR #1412, opened 2020-12-06
Update dependencies in 7.x

shadowspawn:
# Pull Request

Update the dependencies. 

Linting: `standard` has got simpler, but running matching lint in TypeScript with `eslint-config-standard-with-typescript` requires lots of dependencies and one currently needs to be an old version:

- https://github.com/standard/eslint-config-standard-with-typescript#readme

(I like the lint advice but we could drop some of the extra lint methods if it gets annoying.)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5,28 +5,28 @@
   "requires": true,
   "dependencies": {
     "@babel/code-frame": {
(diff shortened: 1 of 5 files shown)


abetomo (approved):
📦

PR #1413, opened 2020-12-06
Bump required version of node to oldest LTS which is 10

shadowspawn:
# Pull Request

For releasing a new version, bump required node to the oldest LTS version to match what we support and test.


abetomo (approved):
1️⃣ 0️⃣

PR #1419, opened 2020-12-20
Update supported details for 7.x

shadowspawn:
# Pull Request

Update SUPPORTED and README to node 10 (as already done in `node.engines`!) 

I wondered about node 8, but it is not supported by latest test dependencies.

I added a comment to SUPPORTED saying we will consider merge requests for older versions too, partly because  TideLift analytics show most dependencies are for 2.x.

## ChangeLog

- Commander 7 requires a minimum of node 10.

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -865,8 +865,8 @@ More samples can be found in the [examples](https://github.com/tj/commander.js/t
 
 ## Support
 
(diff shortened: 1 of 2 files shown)


abetomo (approved):
7️⃣

PR #1421, opened 2020-12-21
Use custom command example to show revert to Commander v6 behaviours

shadowspawn:
# Pull Request

## Problem

With the changes in Commander 7, some people are going to want the old behaviours to upgrade with minimal code changes. This is easy for `.storeOptionsAsProperties()` which is inherited, but `.allowExcessArguments()` would need to be called on each subcommand.

## Solution

Change the example custom class to be `Command6` with the old behaviours. Can reference it as required.

(Also delete the custom-class-function example as not very interesting, although at the time I wanted to make sure we could support custom behaviour without a subclass. Our subclass support is better now.)


abetomo (approved):
LGTM

PR #1424, opened 2020-12-25
Modify TypeScript return type depending on storeOptionsAsProperties parameter

shadowspawn:
# Pull Request

## Problem

Now that options are stored safely by default and this is reflected in the TypeScript interface for Command, it is an error to access an option using legacy code pattern.

```ts
// Typescript
const program = new Command().storeOptionsAsProperties();
const value = program.myOption; // Error: Property 'myOption' does not exist on type 'Command'.
```

## Solution

Modify the return type when  `.storeOptionsAsProperties()` is called to add arbitrary properties for the options.

```ts
// Typescript
const program = new Command().storeOptionsAsProperties();
const value = program.myOption; // ok
```


abetomo (approved):
👍

PR #1425, opened 2020-12-27
Simplify eslint dependencies for TypeScript

shadowspawn:
# Pull Request

Optional change to address a pain point updating our direct dependencies.

## Problem

`eslint-config-standard-with-typescript` has a lot of  peer dependencies we have to install and manage, and it can be a pain working out which can be updated and which need to stay on older versions.

- https://github.com/standard/eslint-config-standard-with-typescript#readme

(This may be easier in future if eslint changes how plugins can manage their dependencies.)

## Solution

Extend the `eslint` recommended settings instead.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -15,7 +15,6 @@ const javascriptSettings = {
 
 const typescriptSettings = {
   files: ['*.ts'],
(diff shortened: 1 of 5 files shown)


abetomo (approved):
LGTM!

Issue #1426, opened 2020-12-29
Options shared by a root command and its subCommands


electrotype:
I had an issue that took me a long time to figure out and fix. It is about an `option` that I add to _all_ the commands of my application: `-s` for "silent". When this option is set, a command will disable all outputs when running.

Since this is a global option that I add to all commands, I created a `addGlobalOptions(command)` function and I call it for every command.

The issue is that I was also calling `addGlobalOptions()` _on the root Command_! 

My program looks a bit like:
```
const commander = new Command();
addGlobalOptions(commander);
addSubCommands(commander);
```
In `addSubCommands`, each command has the global options added too, using `addGlobalOptions`.

By calling `addGlobalOptions` on the root "commander" Command, a subCommand would always receive `undefined` for the `-s` argument, even when specified by the user! By removing the global options on the root "commander" Command, `true` is received correctly.

I did not find anything in the documentation about this? What is the expected behavior when an option is shared between a root Command and one of its subCommand? Or maybe I'm doing something wrong?

shadowspawn:
The root command parses its options from anywhere on the command line, so it consumes the option first before the subcommand. The relevant line in the README is:

> Options on the command line are not positional, and can be specified before or after other arguments.

You can add and access a common option on the root command, or add it to all your subcommands. A visible difference to the user is where in the help the option appears. Do you think of it as a "program" option on the root command, or a common option which every subcommand has locally?

There is an open issue with collected reports and research: #1229
and the related issues to your problem are: #598 #1033 #1307

electrotype:
Thank you for the explanation.

shadowspawn:
Pull Request opened to add `.enablePositionalOptions()` and `.passThroughOptions()`: #1427

shadowspawn:
Opened poll on possible enhancements to global options: #1551

shadowspawn:
Opened a draft PR for `.addCommonOption()` #1670

Issue #1307, opened 2020-07-20
subcommands cannot have the same option as top level commands


danielr1996:
Consider the following piece of code
```javascript
import commander from 'commander';

const program = new commander.Command();
program.passCommandToAction(false);

program
    .option('-f, --file <config-file>', ``, 'default1')
    .action(console.log)

program.command('init')
    .option('-f, --file <config-file>', ``, `default2`)
    .action(console.log)

program.parseAsync(process.argv);
```

# expected behaviour
Running `node bug.mjs -f config.js` should output `{file: 'config.js'}` because the `-f` flag was specified
Running `node bug.mjs init -f config.js` should output `{file: 'config.js'}` because the `-f` flag was specified

# actual behaviour
Running `node bug.mjs -f config.js` outputs `{file: 'config.js'}` which is expected
Running `node bug.mjs init -f config.js` outputs `{file: default2'}` which is not expected because the `-f` flag was specified but the default value is used

After reading #1229  and its linked issues I'm not sure if what I described is an actual bug or expeted behaviour. 

For the case that it is a bug I'd be happy to try and fix the bug. 

For the case that it is an expected behaviour I found a workarround:
You just don't use top level commands at all and use the `{isDefault: true}` option for the subcommand that should get executed when there is no sub command specified, effectively making it a top level command.

```
program
    .command('bump', {isDefault: true})
    .option('-f, --file <config-file>', ``, 'default')
    .action(o=>console.log('bump'))
```

shadowspawn:
Short version: expected behaviour.

Long version

Well researched!

This is currently expected behaviour as the program options are considered first and are not positional, and grab the option before the subcommand is considered. As you found, #1229 is the high level issue covering related issues.

As you suggested, a good work-around is to define a default command rather than attach the options to the top-level command (program).

danielr1996:
Thanks for the clarification.

Should we add this behaviour and the workaround to the documentation if somebody comes across the same problem as me?

shadowspawn:
> Should we add this behaviour and the workaround to the documentation if somebody comes across the same problem as me?

This hasn't come up much, so I don't currently want to try extending the README coverage. 


The open wider issue is #1229 and I have added this issue to it.

danielr1996:
Ok, then I'm closing the issue. Thanks again for your feedback.

shadowspawn:
Pull Request opened to add `.enablePositionalOptions()` and `.passThroughOptions()`: #1427

Issue #1293, opened 2020-07-04
A clean way to get pass through arguments?


teststaybaka:
Hi, I'm writing a simple CLI tool that triggers the execution of another JS file which also takes command line arguments, but I couldn't figure out a clean way to get pass-through arguments. A short snipet written in TypeScript is shown below. 

```TypeScript
import { Command } from "commander";
let program = new Command();
program
  .command("run <file>")
  .action(
    (file, options): void => {
      new Builder().build();
      console.log(file);
      console.log(options);
      spawnSync("node", [file, ...options.args], {
        stdio: "inherit",
      });
    });
program.parse();
```

Then when I execute the CLI as `cli run executable.js -- --test-flag`, it doesnt' work because `options.args` contains `[ 'executable.js', '--test-flag' ]`, which seems to match the document though:
> leaving any args not consumed by the program options in the program.args array.

Is there a better way to get an array of pass-through arguments, which excludes `<file>` and other potential positional arguments? Thanks!

shadowspawn:
1) I think what you have might work if you just remove `file` from the spawn arguments since it is also included in `args` like:

```
spawnSync("node", options.args, {
        stdio: "inherit",
      });
```

shadowspawn:
2) The extra arguments actually get passed to the action handler too.

```
const { Command } = require("commander");
let program = new Command();
program
  .command("run <file>")
  .action(
    (file, options, extraArgs) => {
      console.log(file);
      console.log(extraArgs);
    });
program.parse();
```

shadowspawn:
3) But my preferred approach is to make it explicit as part of the command so it appears in the help too, like:

```
const { Command } = require("commander");
let program = new Command();
program
  .command("run <file> [args...]")
  .action(
    (file, args, options) => {
      console.log(file);
      console.log(args);
    });
program.parse();
```

teststaybaka:
Thanks for the quick response.

1) I actually want to do some processing to the file name and I guess people could define more positional arguments.
2) I feel this is the best solution.
3) `[args...]` doesn't work if I need to pass through undefined options. I.e., `cli run executable.js --test-flag` complains "error: unknown option '--test-flag'". I want to pass through any unknown options, so I couldn't define them ahead of time. I guess we could configure somehow to ignore unknown options but I'm not sure it'd be better than 2).

shadowspawn:
>cli run executable.js --test-flag complains "error: unknown option '--test-flag'"

[...]

Issue #1229, opened 2020-03-26
Collect: global option parsing and values


shadowspawn:
Commander looks for the program options anywhere on the command line, including after operands (quite common) and after subcommands (less common). When a program option is defined the value is only available from the program options.  

The scenarios people have asked about are:
1) only looking for program option before subcommand, so say subcommand could use same option name
2) treating the option as "global" so value available from the subcommand, or even appear in the subcommand help
3) stop parsing at the first operand and and subsequent "options" are included as ordinary arguments, and can be used for calling another program (without needing to use `.allowUnknownOptions` or `--`).

(This discussion uses program and subcommand for simplicity as though there were only two levels, but there could be more levels.)

Related issues:
- #243 Accessing parent options from subcommand
- #598 main options intertwine with command options
- #797 sub-command not executed if using --version option
- #1033 Conflict with global options and command specific options
- #1078 Global options not displayed in sub-command help
- #1127 How to stop parsing for options after first argument?
- #1155 Add way to collect inherited/global options from parents
- #1293 A clean way to get pass through arguments?
- #1307 subcommands cannot have the same option as top level commands
- #1426 Options shared by a root command and its subCommands

## Standards

POSIX and GNU differ on how options are parsed relative to operands, but do not mention subcommands as such.

### POSIX (Open Group)

POSIX has options strictly before operands.

https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap12.html

> Guideline 9:
> All options should precede operands on the command line.

### GNU

GNU relaxes the order.

https://www.gnu.org/prep/standards/standards.html#Command_002dLine-Interfaces

> Note that the GNU version of getopt will normally permit options anywhere among the arguments unless the special argument ‘--’ is used. This is not what POSIX specifies; it is a GNU extension.

https://www.gnu.org/software/coreutils/manual/html_node/Common-options.html 

> Normally options and operands can appear in any order, and programs act as if all the options appear before any operands. For example, ‘sort -r passwd -t :’ acts like ‘sort -r -t : passwd’, since ‘:’ is an option-argument of -t. However, if the POSIXLY_CORRECT environment variable is set, options must appear before operands, unless otherwise specified for a particular command.

## Examples from Other Implementations

`Yargs` has `halt-at-non-option` which defaults to false.

> Should parsing stop at the first positional argument?

`Cobra` has Flags and PersistentFlags for whether flag [value] available to children, 
and TraverseChildren for whether to look for local flags of parent in children arguments.

[...]

PR #1427, opened 2021-01-01
Positional options

shadowspawn:
# Pull Request

Collected issues and background research in #1229

## Problem

Two quite similar problems in how they look on command line, but different use cases.

1) A program option can not be reused in a subcommand, as the program options are recognised anywhere on the command line.

Related: #598 #797 #1033 #1307 #1426

2) It is not easy to pass through options which are intended for another program. The options are recognised before or after command-arguments.

Related: #1127 #1293 

## Solution

Add two new routines to alter the parsing of options to be positional. The default behaviour is unchanged.

`.enablePositionalOptions()`

Enable positional options. Positional means global options are specified before subcommands which lets subcommands reuse the same option names, and also enables subcommands to turn on passThroughOptions.

The default behaviour is non-positional (as before) and global options may appear anywhere on the command line.

Positional options makes it different whether an option comes before or after a subcommand. e.g.

```js
program -p subcommand
program subcommand -p
program -p subcommand -p
```

`.passThroughOptions()`

Pass through options that come after command-arguments rather than treat them as command-options, so actual command-options come before command-arguments. Turning this on for a subcommand requires positional options to have been enabled on the program (parent commands).

The default behaviour is non-positional (as before) and options may appear before or after command-arguments.

Pass through options makes it different whether an option comes before or after a command-argument. e.g.

```js
program --verbose arg
program arg --verbose
program --verbose arg --verbose
```

## ChangeLog

- `.enablePositionalOptions()` to let program and subcommand reuse same option
- `.passThroughOptions()` to pass options through to other programs without needing `--`

## To Do:

- [x] TypeScript typings
- [x] tests
- [x] README
- [x] examples

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -694,7 +694,7 @@ an option for a different purpose in subcommands.
 
 Example file: [positional-options.js](./examples/positional-options.js)
 
-With positional options, these two lines are different:
+With positional options, the `-b` is a program option in the first line and a subcommand option in the second:
 
 ```sh
 program -b subcommand
@@ -704,11 +704,11 @@ program subcommand -b
 By default options are recognised before and after command-arguments. To only process options that come
 before the command-arguments, use `.passThroughOptions()`. This lets you pass the  arguments and following options through to another program
 without needing to use `--` to end the option processing. 
-To use pass through options in a subcommand, the programs need to enable positional options.
+To use pass through options in a subcommand, the program needs to enable positional options.
 
 Example file: [pass-through-options.js](./examples/pass-through-options.js)
 
-With pass through options, these two lines are different:
+With pass through options, the `--port=80` is a program option in the line and passed through as a command-argument in the second:
 
 ```sh
 program --port=80 arg
--- a/Readme.md
+++ b/Readme.md
@@ -35,6 +35,7 @@ Read this in other languages: English | [简体中文](./Readme_zh-CN.md)
   - [Custom event listeners](#custom-event-listeners)
   - [Bits and pieces](#bits-and-pieces)
     - [.parse() and .parseAsync()](#parse-and-parseasync)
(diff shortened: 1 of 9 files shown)


shadowspawn:
This has been implemented as a non-breaking change, so does not need to go in 7.0.0 if there are concerns.

abetomo (approved):
How nice!

shadowspawn:
Thanks @abetomo 🎈

CMCDragonkai:
Does this feature allow me to add global options before adding sub commands?

```
  const polykey = new Command();
  polykey.version('0.0.1');
  polykey.exitOverride();
  polykey.addCommand(commandKeys);
```

Is it then possible to do `polykey -v 2 subcommand`?

shadowspawn:
@CMCDragonkai 
You can use global options before subcommands, whether or not positional-arguments is enabled.

CMCDragonkai:
But how do I access these global options inside a subcommand action handler?

shadowspawn:
@CMCDragonkai 
The short answer is: you can access the global options on the `program` command with `program.opts()` or `cmd.parent.opts()`, they are not passed into the action handler direction. If you are not using positional options as is the default behaviour, then global options can come before or after subcommands.

Please open a new issue if you want a fuller explanation. If I understood your questions correctly, they do not seem to be about this Pull Request but about global options in general.

ruscon:
@shadowspawn 
`command.parent` does not exists in types. Can you fix it?
v7.1

shadowspawn:
@ruscon 
Ah right, last time it came up we were refactoring and I was not sure if `parent` could be `null` or `undefined` so didn't add it at the time. The typings are stricter in v7 too, previously `parent` would have not generated a warning since it could have been an option value.

If you don't mind, would you please open a new issue to track this? (I have added it to my private backlog in any case.)

PR #1429, opened 2021-01-03
Make new excess arguments error opt-in

shadowspawn:
# Pull Request

#1407  added support for an error for excess arguments and associated configuration, landing in 7.0.

## Problem

Made the error the default behaviour in #1409. I am having second thoughts because:

- other significant breaking changes in 7.0.0 making it harder to upgrade
- get throws in unit tests where don't really care about correctly describing arguments 
- finding cause of unexpected throws in existing unit tests not always obvious

## Solution

Make new error opt-in for now. (See if there is demand for error by-default in a future version.)

## ToDo

- [x] rework README (might wait for other PR to land)
- [x] rework CHANGELOG (might wait for other PR to land)
- [x] rethink example for custom-command-class as `Command6` less interesting, which is a good sign

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,8 +12,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ### Added
 
-- *Breaking:* error message if there are too many command-arguments on command line for the action handler ([#1409])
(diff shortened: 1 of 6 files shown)


shadowspawn:
I have updated the docs and example, ready for review.

abetomo (approved):
LGTM!

shadowspawn:
Thanks @abetomo 🚀

PR #1430, opened 2021-01-06
ability to specify `minColumnWidth` for `Help.wrap`

cravler:
# Pull Request

### DIFF
--- a/index.js
+++ b/index.js
@@ -316,17 +316,17 @@ class Help {
    * @param {string} str
    * @param {number} width
    * @param {number} indent
+   * @param {number} [minColumnWidth=40]
    * @return {string}
    *
    */
 
-  wrap(str, width, indent) {
+  wrap(str, width, indent, minColumnWidth = 40) {
     // Detect manually wrapped and indented strings by searching for line breaks
     // followed by multiple spaces/tabs.
     if (str.match(/[\n]\s+/)) return str;
     // Do not wrap if not enough room for a wrapped column of text (as could end up with a word per line).
     const columnWidth = width - indent;
-    const minColumnWidth = 40;
     if (columnWidth < minColumnWidth) return str;
 
     const leadingStr = str.substr(0, indent);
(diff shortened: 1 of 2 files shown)


shadowspawn:
Do you have a situation when you want to be able to change minColumnWidth? Or suggesting it because it is currently a hard-coded value in the code?

For historical interest, this was actually a parameter in the original PR! I removed it rather than try and add documentation for the parameter, as I struggled to differentiate it from the "width" parameter. The current variable name is the best I came up with at the time.

cravler:
yes, i want set it to `minColumnWidth=0`, so it would be totally disabled

i played with different terminal sizes and app looks better when it disabled

`minColumnWidth=0`:
image

`minColumnWidth=40`
image

cravler:
maybe better to make like this?

```js
class Help {
    formatHelp(cmd, helper) {
        //...

        function wrap(str, width, indent) {
            // Do not wrap if not enough room for a wrapped column of text (as could end up with a word per line).
            const columnWidth = width - indent;
            const minColumnWidth = 40;
            if (columnWidth < minColumnWidth) return str;
            return return helper.wrap(str, width, indent);
        }

        function formatItem(term, description) {
          if (description) {
            const fullText = `${term.padEnd(termWidth + itemSeparatorWidth)}${description}`;
            return wrap(fullText, helpWidth - itemIndentWidth, termWidth + itemSeparatorWidth);
          }
          return term;
        };

        // ...
    }
}
```

shadowspawn:
The reason for skipping the wrap at small sizes is that in the worst case can get a word per line, and unlike the fallback unwrapped output, this does not reflow as the terminal window is made wider as it is hard wrapped. e.g.

```
  daemon              Daemon
                      to
                      listen
                      to
                      multiple
                      sockets
                      connections
  init [name] [path]  Created
                      a new
                      application
```

But being able to configure the cut-off is a reasonable request. Still thinking about that.

shadowspawn:
I do not think this will be be used much, but it isn't a front-facing routine and this hopefully makes it more usable for some custom overrides. Ok.

If you are still keen add the TypeScript, and I'll add some comments when I merge it if it passes review.

shadowspawn (approved):
LGTM. I'll add an explanation of the new parameter.

abetomo (approved):
👍

PR #1435, opened 2021-01-12
examples/custom-command-class: MyCommand constructor takes no args

glensc:
# Pull Request

## Problem

examples/custom-command-class.js example is mistyped.

## Solution

See the PR.


glensc:
Perhaps the problem is that TypeScript typing is incorrect

Issue #1284, opened 2020-06-19
Add esm


xiaoxiangmoe:
https://nodejs.org/api/esm.html#esm_conditional_exports

Add support for esm. 

If you think this proposal is reasonable, I'll create a PR for this.

shadowspawn:
What is the problem you want to solve? The ability to "import" Commander from code written as ES module?

While I am interested in the esm progress and how Commander might support esm, I don't think we would make it a default behaviour yet as the Node 14 version of the linked page lists this as "Stability: 1 - Experimental".

> Stability: 1 - Experimental. The feature is not subject to Semantic Versioning rules. Non-backward compatible changes or removal may occur in any future release. Use of the feature is not recommended in production environments.

Related: #950

xiaoxiangmoe:
In the current situation
```js
import { program } from 'commander';
```
This will report an error in mjs.

```js
import commander from 'commander';
const {program} = commander; // Here commander and program are the same object
```
But this will perform well.

Should we encourage everyone to use default import in mjs?

shadowspawn:
Yes, for now. The "default export" can be used in the same way as can be done with require:

```
// const commander = require('commander'); // cjs
import commander from 'commander';

const globalProgram = commander.program;
const myProgram = new commander.Command();
```

Side note 1: the exported global `program` is currently mainly for style, as it is the same as the default export.

Side note 2: I had actually been wondering about this style for the README for cjs, but decided the destructuring assignment was more widely used. e.g. `const { program } = require...`

shadowspawn:
I noticed instructions for esm and cjs here, I'll keep an eye out for more: https://github.com/uuidjs/uuid

shadowspawn:
Opened PR to add to README: #1298

shadowspawn:
README update released with Commander v6.0.0 : https://github.com/tj/commander.js/releases/tag/v6.0.0

I am interested in supporting esm more directly in future, when the support is stable.

shadowspawn:
I came across a related PR and comments about the conditional pattern: https://github.com/tapjs/node-tap/pull/668#discussion_r414982540

shadowspawn:
I was reading about esm again today and getting keen on adding a wrapper, when came across this reminder that using an experimental feature can have downsides. In this case end users getting a warning that (a dependent package) using experimental feature, until the warning gets removed:
https://github.com/nodejs/modules/issues/485#issuecomment-587246376

shadowspawn:
[...]

PR #1440, opened 2021-01-20
Add support for named imports in ESM

shadowspawn:
# Pull Request

## Problem

Importing commander as CommonJs into an ECMAScript module only allows using the default import, and can not use named imports.

```js
// import { program } from 'commander'; // ERROR
import commander from 'commander';
const program = commander.program;
```

Related issue: #1284

## Solution

Add esm wrapper and use as a deep import:

```
import { program } from 'commander/esm.mjs';
```

In the future we will add support so the deep import is not needed by using conditional exports and/or subpaths. Being conservative for now as the functionality is still marked as experimental in node 12. (There have been breaking changes in the support in minor versions of node 12 and 13.) 

- https://nodejs.org/docs/latest-v12.x/api/packages.html#packages_conditional_exports
- https://nodejs.org/docs/latest-v12.x/api/packages.html#packages_subpath_exports

The `--experimental-modules` option used in the `test-esm` run-script will get dropped by node eventually, but currently allows that test to run locally with node 10 through node 15.

## ChangeLog

- added support for named imports from ECMAScript modules


abetomo (approved):
LGTM

shadowspawn:
So fast! 🚀 
Thanks @abetomo

package.json
```diff
@@ -19,18 +19,21 @@
...
     "test-typings": "tsc -p tsconfig.json",
     "typescript-checkJS": "tsc --allowJS --checkJS index.js --noEmit",
-    "test-all": "npm run test && npm run lint && npm run typescript-lint && npm run typescript-checkJS"
+    "test-all": "npm run test && npm run lint && npm run typescript-lint && npm run typescript-checkJS && npm run test-esm"
   },
-  "main": "index",
+  "main": "./index.js",
   "files": [
     "index.js",
+    "wrapper.mjs",
```
cravler, line 33: looks like a typo, must be `esm.mjs`

shadowspawn: Oh, good catch, thanks! I changed the name half way through.

shadowspawn: (For interest, I thought `wrapper` was a suitable name for a wrapper used with conditional exports and not seen by normal client use. But when I changed to deep import and the name became visible, I wanted something more meaningful.)

shadowspawn: Fixed in #1443

PR #1448, opened 2021-01-28
Replace process.mainModule with require.main

shadowspawn:
# Pull Request

## Problem

`process.mainModule` is deprecated from node 14.

https://nodejs.org/docs/latest-v14.x/api/process.html#process_process_mainmodule

## Solution

Replace with recommended alternative, which is direct replacement and has been available since node v0.1.17:

- https://nodejs.org/docs/latest-v14.x/api/modules.html#modules_accessing_the_main_module
- https://nodejs.org/docs/latest-v14.x/api/modules.html#modules_require_main

### DIFF
--- a/index.js
+++ b/index.js
@@ -1270,10 +1270,8 @@ class Command extends EventEmitter {
       default:
         throw new Error(`unexpected parse option { from: '${parseOptions.from}' }`);
     }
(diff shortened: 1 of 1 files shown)


abetomo (approved):
LGTM

PR #1449, opened 2021-01-30
Add cjs to list of expected node script extensions

shadowspawn:
# Pull Request

## Problem

Node uses `.cjs` for specifically CommonJS javascript files, like `.mjs` for esm. Commander does not check for that extension when considering how to launch external stand-alone subcommands.

## Solution

Add `.cjs`.

## ChangeLog

- Add `.cjs` to expected script file extensions.

## To Do

- [x] Add tests for file suffixes. We have a test for `.ts` but not for `.mjs` or `.cjs`.


abetomo (approved):
👍

PR #1453, opened 2021-01-30
Convert typings tests to tsd

shadowspawn:
# Pull Request

## Problem

Using a TypeScript compile (`tsc`) to test the TypeScript typings was useful, but some minor annoyances:

- felt a bit strange writing lots of fake code
- declaring lots of unused variables to do return type checking 
- suppressing lint warnings from the dummy code, like the unused variables!
- type checking by assignment is not strict

## Solution

Use `tsd` for a fairly direct conversion to strict type tests. Instead of assigning to a typed variable, use `expectType`.

One related change: `tsd` complained about using `typings` property in `package.json`, and in fact `types` is more standard so changed to use that.

## ChangeLog

- make TypeScript typings tests stricter (#1453)


abetomo (approved):
Cool!

Issue #1447, opened 2021-01-26
Adding `.choices(['a', 'b', 'c])` to a variadic option removes its variadic-ness


MikeyBurkman:
Version: `7.0.0`

I was super happy about the recent addition of `choices()` to options, but when I add `choices` to a variadic option, the option changes to a single value:

```ts
program
  .command('services')
  .addOption(
    new Option(
      '-e, --env <names...>',
      'The environment(s) to look up'
    ).choices(['dev', 'stg', 'prod'])
  )
  .action((args) => console.log(args));
```
```sh
foo services -e dev -e stg
```
Without `.choices()`, `env` becomes an array of `['dev', 'stg']`. However, after adding choices, it becomes a single value `'stg'`.

shadowspawn:
Short version: yes, this is currently a known limitation.


Long version

I had realised there were some limitations like this when adding the variadic support.

`.choices()` is implemented the same way as custom option processing functions supplied by the user. Commander does not know whether a custom function is a simple coercion (like `.choices()`) or a complex function like something that counts the number of calls. All variadic does when there is a custom function is call it multiple times and leaves it up to the custom function to decide what to return. This works fairly well with the complex custom functions but not with the simple coerce functions.

To get `.choices()` working with variadic we can either extend the choices implementation to understand about variadic, or add a way to tell apart simple and complex custom option processing functions.

shadowspawn:
`.choices()` now plays well with variadic from Commander 7.1.0

PR #1454, opened 2021-01-31
Add support for variadic to choices

shadowspawn:
# Pull Request

I don't this will be very common, but it is a shortcoming and did get reported!

## Problem

Variadic and choices do not play well together.

Related: #1447

## Solution

Add support for variadic to choices using same code as plain variadic processing.

(This does not add public support for mixing custom option processing with variadic. Keep it private until clearer what people might want to do.)

## ChangeLog

- allow using option choices and variadic together (#1454)

### DIFF
--- a/index.js
+++ b/index.js
@@ -426,6 +426,18 @@ class Option {
   };
 
   /**
+   * @api private
+   */
+
+  _concatValue(value, previous) {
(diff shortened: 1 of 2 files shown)


abetomo (approved):
LGTM!

Issue #1450, opened 2021-01-30
Allow arguments description without command description


pvgn:
Hi there, and thanks for the great work!

I'm in a situation where I want to add a description for my arguments, but not for my command.

Currently (using v6.2.1), adding both descriptions is possible with :
```js
program
  .name('my-program')
  .arguments('<my-arg>')
  .description('my program desc', {
    'my-arg': 'my arg desc'
  })

// -> help is displaying all descriptions correctly ✅ 
```

However, if I try to omit the command description, while keeping the arguments description, it doesn't work.
```js
program
  .name('my-program')
  .arguments('<my-arg>')
  .description(undefined, { // or null or ''
    'my-arg': 'my arg desc'
  })

// -> help is not displaying any description ❌ 
```

A workaround is to do :
```js
program
  .name('my-program')
  .arguments('<my-arg>')
  .description('\x1B[2A', { // hack to skip command desc
    'my-arg': 'my arg desc'
  })

// -> help is not displaying command desc (kind of)
// and displaying args desc ✅  (well, it's a hack...)
// tested on macOS, and should also work on Unix-like OSes
```

It would be nice to have a proper way to do this ;)

shadowspawn:
In Commander 7 this is supported. You may pass an empty string for the command description.

(This came up once before in #1192. At the time the arguments description was not documented.)

Issue #1192, opened 2020-02-14
Decouple arguments description from command description


sergiz:
https://github.com/tj/commander.js/blob/516a14c659d3db4650020f0eee1669957cd8918e/index.js#L1398-L1416

Hello.

Is there a special reason why we cannot only describe command arguments without the command itself?

shadowspawn:
I think no special reason. That code matches the way they are specified, where the description is an expected parameter and the `argsDescription` is an optional parameter. 

```
   * @param {string} str
   * @param {Object} [argsDescription]
...
  description(str, argsDescription)
```

(The `argsDescription` parameter is currently undocumented.)

sergiz:
I see, but the current implementation makes it impossible to avoid adding command description, when you only want to describe arguments.

For example, when you have a self explanatory command, but want to provide some additional description for the arguments.

My proposal is to change that, so this._argsDescription will be displayed regardless of the this._description. Optionally the description() method could also be adjusted, os if the first argument (str) is an object then it would swap the arguments (so you don't need to explicitly make str null or ''.

sergiz:
I see, but the current implementation makes it impossible to avoid adding command description, when you only want to describe arguments.

For example, when you have a self explanatory command, but want to provide some additional description for the arguments.

My proposal is to change that, so this._argsDescription will be displayed regardless of the this._description. Optionally the description() method could also be adjusted, so if the first argument (str) is an object then it would swap the arguments (so you don't need to explicitly make str null or ''.

shadowspawn:
Given the arguments descriptions are not currently documented, I don't this this issue will affect many people. No further activity in a month. Closing.

> My proposal is to change that, so this._argsDescription will be displayed regardless of the this._description. 

PR for this is welcome if someone does want to fix it.

> Optionally the description() method could also be adjusted, so if the first argument (str) is an object then it would swap the arguments (so you don't need to explicitly make str null or ''.

No, I do not want to complicate the method signature for a rare case.
Feel free to open a new issue if it comes up again, with new information and renewed interest.

Thank you  for your contributions.

shadowspawn:
In Commander 7, you can pass an empty string for the command description and supply an arguments description to add just arguments description.

PR #1459, opened 2021-02-02
Test for help for arguments without description

shadowspawn:
# Pull Request

## Problem

A couple of people have asked about showing help for command-arguments, without a description for the command. We didn't have a test for this now supported case.

Uses: #1450 #1192

## Solution

Add test.


abetomo (approved):
👍

PR #1462, opened 2021-02-04
Missing subcommand in action for test

shadowspawn:
# Pull Request

Stumbled across this when it failed unexpectedly while I was experimenting with changes to option handling.

## Problem

`when specify unknown option with subcommand and action handler then error` was not testing subcommand

## Solution

Change call to match test description and intent.


abetomo (approved):
👍

Issue #1460, opened 2021-02-03
Options parsed before validating argument


vck3000:
Hi there,

This is a great project, thank you for maintaining this. 

We're in the process of upgrading from v4.0.1 to v7.0.0 with the new features. We have a small problem whereby options are being parsed before the arguments themselves are validated. 

We were hoping to be able to continue to listen to `command:*` to detect bad arguments, as we've currently hooked it up to our own suggestion engine to provide "did you mean" suggestions for the bad argument. Upgrading to v7.0.0 has broken this functionality for us. 

E.g. 
```
$ program bad_argument --help

# Outputs the help menu instead of complaining that bad_argument is a bad argument
```

Another example:
```
$ program lintt --fix

# Complains that --fix is not a valid option. Instead, we expected `lintt` to be
# detected as a bad argument so we can suggest `lint` to the user.
```

Minimal reproduction:
```
import commander from 'commander';

const program = new commander.Command();

program.on('command:*', (arg: string) => {
  console.log(`Detected bad arg: ${arg}`);
});

program.command('lint').action(() => {
  console.log('lint');
});

program.parse();
```

I dug into the source code a bit and think I found it. It appears to be parsing the options and throwing the error first, before reaching the "command:*" section. 

https://github.com/tj/commander.js/blob/034ad9f326ee2b26001b864adda5fea465ee148b/index.js#L1472

Is there a solution or workaround to this that I haven't found? Thanks :).

shadowspawn:
Thanks for clear reproduce steps. I didn't expect the behaviour to change from v4 in these areas, but there have been many parsing improvements. I'll do some digging.

shadowspawn:
Short version: reproduced, broken from v5. Drat.

Long version

The PR that changed the behaviour is #1149 which shipped in v5.0.0. It was BIG. 

There were no explicit unit tests on `command:*` in v4.0.0. There were some tests added for v5.0.0, but none of them included options. So the change in behaviour was quietly missed. This is a use case that I had not explicitly considered. Interesting situation, that of course the options don't make sense when the command is spelled wrong.

Which check gets called where is a bit fragile.

To some extent I consider `command:*` as legacy and may not be able to preserve (restore!) all the original behaviours. However, I am interested in supporting "did you mean" suggestions in some way so if can't work out a backwards compatible method will consider new approaches.

vck3000:
Thanks for your investigation into this :). 

[...]

PR #1464, opened 2021-02-05
Skip unknown options check if there is a better error to display

shadowspawn:
# Pull Request

I think this is working our sensibly. Making changes in a complicated piece of the code so going slowly.

## Problem

In Commander v5 we started checking for unknown options in more situations, which was an improvement in many cases. However, it unintentional blocked some "better" errors such as an unknown command, and blocked suggestions for the unknown commands (listener for `command:*`).

Related issue: #1460

## Solution

Work through cases and don't block a better error by checking for unknown options. Especially, when the issue is likely to be a misspelled command.

## ChangeLog

- fix regression for legacy `command('*')` and call when command line includes options (#1464)
- fix regression for `on('command:*', ...` and call when command line includes unknown options (#1464)
- fix display best error for combination of unknown command and unknown option (i.e. unknown command) (#1464)

### DIFF
--- a/index.js
+++ b/index.js
@@ -1476,6 +1476,7 @@ class Command extends EventEmitter {
       outputHelpIfRequested(this, parsed.unknown);
       this._checkForMissingMandatoryOptions();
 
+      // We do not always call this check to avoid masking a "better" error, like unknown command.
       const checkForUnknownOptions = () => {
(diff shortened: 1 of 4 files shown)


abetomo (approved):
Nice!

PR #1466, opened 2021-02-05
Test for help for arguments without description

shadowspawn:
# Pull Request

## Problem

We have added support for describing the command arguments without adding a description, but not a matching test to reduce the chance of a regression.

Related issues: #1192 #1450

## Solution

Added test.


abetomo (approved):
👍

PR #1467, opened 2021-02-06
added addArguments method

Niryo:
# Pull Request

## Problem
inconsistent arguments and options syntax

## Solution
introducing new addArgument function

## ChangeLog
New command method: `addArgument()` and `.argument()`


Niryo:
@shadowspawn This is a draft with the mvp for addArguments method. can you take a look and tell me what you think and what are the chances for merge, before I continue to tests, docs and qa?

shadowspawn:
Modified code less than I expected for first draft, which is nice for reading the diff. Sticking to the features which are already supported (argument name and argument description) keeps it simpler.

> what are the chances for a merge, before I continue to tests, docs and qa?

50%
Stick to the code for now, I'm going to add a few comments.

I am actually liking this more by leaving out the default value and custom coerce functions for now. There is room in the syntax to add them later if there is demand.

index.js
```diff
@@ -2198,3 +2212,29 @@ function incrementNodeInspectorPort(args) {
     return arg;
   });
 }
+
+function parseArg(arg) {
```
shadowspawn, line 2216: I was going to suggest moving this, but changed my mind when I looked deeper, so leave it here for now. I'm adding a comment partly to remind myself! 

In general I now prefer adding internal class-related code as a private method on the class (weakly private) rather than a separate private function (strongly private).

BUT I see I added `_parseOptionFlags()` as a private function. It was easier to call it from both the Option class and the Command class while refactoring. That is likely to be the case here too.

shadowspawn:
I wouldn't release `.addArgument()` without a matching convenience routine on the Command for `.argument()`, like `.option()` and `.addOption()`.

index.js
Niryo: Yep, i wanted to share the same code between the two classes. Personally I think that extracting logic into pure functions that don't contain state (this) is most of the time a good thing, but if you have a different approach I can align my code of course.

Niryo:
@shadowspawn do you think we should deprecate `.arguments()` in order to keep the api small? I think that the syntax of `.argument()` is more elegant and robust (no need to specify the description separately using hash and open the door for typos)

shadowspawn:
> @shadowspawn do you think we should deprecate .arguments() in order to keep the api small?

At least to some extent, yes. I am thinking the description hash would get dropped from the README, but maybe still mention briefly that can use `.arguments('<a> <b> [c...]')` as a shortcut for adding multiple arguments. The example code in the README would switch to using `.argument()`.

Niryo:
> I wouldn't release `.addArgument()` without a matching convenience routine on the Command for `.argument()`, like `.option()` and `.addOption()`.

    What about the tests? Can I convert the tests that use arguments to tests that use argement and add argument, in order to verify that everything is working?  It will be more convenient to change the tests instead of duplicate everything

shadowspawn:
The old functionality is not being removed (yet), so the existing tests are still relevant to prevent regressions.

Niryo:
@shadowspawn this pr is ready for review

shadowspawn:
Oh drat, I have been making comments but they were not visible yet as I had not started a review. I wondered why you were ignoring them!

shadowspawn:
FYI: opened an issue to poll reactions to API changes in #1471

index.js
```diff
@@ -755,6 +769,20 @@ class Command extends EventEmitter {
...
+  /**
+   * Define argument syntax for the command.
+   * @param {Argument} argument
+   */
+  addArgument(argument) {
+    this._args.push(argument.argDetails);
+    if (!this._argsDescription) {
+      this._argsDescription = {};
+    }
+    this._argsDescription[argument.argDetails.name] = argument.description;
```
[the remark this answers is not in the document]
Niryo: definitely, after full deprecation of old functionality there is a room for lot's of cleanup.

index.js
```diff
@@ -344,6 +344,20 @@ class Help {
...
+class Argument {
+  /**
+   * Initialize a new argument with description.
+   *
+   * @param {string} arg
+   * @param {object} [description]
+   */
+
+  constructor(arg, description) {
+    this.argDetails = parseArg(arg);
```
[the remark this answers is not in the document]
Niryo: done

index.js
```diff
@@ -755,6 +769,20 @@ class Command extends EventEmitter {
     return this._parseExpectedArgs(desc.split(/ +/));
   };
 
+  /**
+   * Define argument syntax for the command.
+   * @param {Argument} argument
+   */
+  addArgument(argument) {
+    this._args.push(argument.argDetails);
```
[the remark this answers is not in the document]
Niryo: done

Readme.md
```diff
@@ -449,6 +457,26 @@ program
...
+  .description('test command', {
+    username: 'user to login',
+    password: 'password for user, if required'
+  })
+  .action((username, password) => {
+    console.log('username:', username);
+    console.log('environment:', password || 'no password given');
+  });
+```
+
```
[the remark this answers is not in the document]
Niryo: done

tests/command.action.test.js
```diff
@@ -40,6 +42,8 @@ test('when .action on program with required argument and argument not supplied t
     .exitOverride()
     .configureOutput({ writeErr: () => {} })
     .arguments('<file>')
+    .argument('<second>')
+    .addArgument(new commander.Argument('<last>'))
```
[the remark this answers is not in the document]
Niryo: done

index.js
```diff
@@ -804,37 +846,20 @@ class Command extends EventEmitter {
...
-
-      if (argDetails.name.length > 3 && argDetails.name.slice(-3) === '...') {
-        argDetails.variadic = true;
-        argDetails.name = argDetails.name.slice(0, -3);
-      }
-      if (argDetails.name) {
-        this._args.push(argDetails);
-      }
+      const argDetails = parseArg(arg);
+      argDetails && this._args.push(argDetails);
```
shadowspawn, line 850: Still adding argDetails to `_args` here.

Niryo: can you explain me what is the case were argDetails is undefined? I see it's something with the arg name, in which case an argument cannot have a name?

shadowspawn: In the original `_parseExpectedArgs`, if an argument is not in the expected format then the argument is silently ignored because the name does not get set. For example: `{oops}` or `foo` or `<...>`. 

In the refactored code that leads to undefined being returned from `parseArg`.

Niryo: So I wonder what we should do in that case if we use the Argument class...
It's a bit weird that the initialization of the class fails silently. I would say that it's better to throw an error if the argument is invalid, but it's a breaking change, so i think we should do it later and I'll try to work around it for now.

shadowspawn:
One thing I have been wondering about but undecided is whether `.argument()` should update an existing argument with same name and signature. This would allow adding descriptions after using a shortcut/old method. For example:

```js
program
  .command('build <target>')
  .argument('target', 'source file for compilation')
```

Niryo:
It's quite redundant because if someone will add the `argument` expression in order to add description, he can just remove the arg from the `command` expression on the way.

index.js
Niryo: Ok, after thinking about it again I realized that it's not a breaking change because we can do it only for the new synax. So what do you say? Do you think we should throw on bad argument?

shadowspawn: Yes. Throw on bad programmer input is a direction we are slowly moving.

Niryo: @shadowspawn I have added a throw inside the class, but I can't solve this comment because we don't want to introduce a breaking change so I still need to use argDetails instead of the Argument class here.
when we deprecate `arguments` prop we could remove all usage of argDetails.

Niryo:
I think all comments are resolved

shadowspawn:
I had a go at consistently using Argument class for `_args` and removing the previous/intermediate untyped argDetails object. I'll submit a draft PR to your branch. It led me to wonder whether to be more lenient with what should be done with incorrectly formatted argument. e.g.

```
.argument('foo')
# could be treated same as
.argument('<foo>')
```

Niryo:
I think that the direction we should go to should be towards more explicit syntax, for example I would prefer using something like `new Argument('bla').required()`, instead of remembering which means required - `<>` or `[]`. 
Also, with the explicit syntax you get auto complete..

examples/argument.js
```diff
@@ -0,0 +1,24 @@
...
+// This example shows specifying the arguments using addArgument() and argument() function.
+
+// const { Command } = require('commander'); // (normal include)
+const { Command, Argument } = require('../'); // include commander in git clone of commander repo
+const program = new Command();
+
+program
+  .version('0.1.0')
+  .addArgument(new Argument('<username>', 'user to login'))
+  .argument('[password]', 'password')
```
shadowspawn, line 12: The description from `arguments.js` would be better, `'password for user, if required'`.

tests/args.badArg.test.js
```diff
@@ -0,0 +1,7 @@
+const commander = require('../');
+
+test('should throw on bad argument', () => {
```
shadowspawn, line 3: The test naming convention being used is:
- _when_ something _then_ result

shadowspawn:
>I think that the direction we should go to should be towards more explicit syntax, for example I would prefer using something like new Argument('bla').required(), instead of remembering which means required - <> or []. 

I am comfortable with `<foo>` as the compact form since it matches the help, but supporting long form is good too. There are some routines on the `Option` class, but not for `required` as the interaction of the possibilities is more complicated. Simpler with `Argument` so could support required, optional, and variadic. Also Command and Option both have `.name()`.

We are using two patterns, with either getter/setter depending on whether an argument is passed, or setter with no argument or optional argument. The pure setters sometimes have a verb like `hideHelp` or `allowUnknownOption`.

`Option` has `required` and `optional` and `variadic` so using those names for the data properties is consistent. But that means other names for the setters.

Niryo:
Anyways, it was just a general thought, it's not really important for this PR, we'll discuss it further in the future if I want to add aa `required` prop:)

Niryo:
@shadowspawn The PR is ready on my side, if you need something else add more comments

shadowspawn:
The two comments I had left are:
- https://github.com/tj/commander.js/pull/1467#discussion_r598060374
- https://github.com/tj/commander.js/pull/1467#discussion_r598059623

(But they are small and I can fix them up later.)

shadowspawn:
I am thinking of pulling this onto a branch and dong some more work on it, and including in Commander v8 in June. The changes to how non-conforming arguments are processed/rejected make it a breaking change for a small subset of users. (e.g. `<foo>` and `[bar]` stay the same but `what-did-you-expect` changes)

shadowspawn (approved):
Approving for merging onto a feature branch (for some further work).

PR #1473, opened 2021-02-14
Add runtime test that TypeScript imports are working

shadowspawn:
# Pull Request

## Problem

We have tests that the TypeScript definitions are as intended through static analysis, but not that the expected exports from Commander are available at runtime.

## Solution

Add a TypeScript test to the Jest tests that performs some simple checks that the expected imports are available at runtime. Mirrors tests for esm.


shadowspawn:
For interest, but can safely be ignored, tests prompted partly by experimenting with removing default export of global Command. For a future version:
- https://github.com/shadowspawn/commander.js/tree/feature/remove-default-export-from-ts
- https://github.com/shadowspawn/commander.js/tree/feature/remove-default-export-from-js

abetomo (approved):
LGTM!

PR #1474, opened 2021-02-17
Increase test coverage of executable spawn

shadowspawn:
# Pull Request

## Problem

We don't have much coverage of the executable launching details in the Jest coverage results, especially error cases.

Note: we do have multiple executable tests that aren't captured by the Jest coverage.

## Solution

Add tests which mock `spawn` to allow injecting errors in more controlled fashion. These test failure paths rather than happy paths, are not end-to-end tests, but provide some coverage of existing behaviours. 

(I am still interested to see if we can include more of the existing tests in the Jest coverage results too.)

Increases Jest coverage from ~98% to ~99%.


abetomo (approved):
great!

PR #1475, opened 2021-02-19
Add Command parent property to TypeScript

shadowspawn:
# Pull Request

## Problem

The `parent` property is not declared in the TypeScript typings. The common use case for it is accessing the global option values in an action handler, and I have shown this in multiple examples in comments. 

Related: #1184 https://github.com/tj/commander.js/pull/1427#issuecomment-781732703

Previously accessing `cmd.parent` would not have caused a Typescript error, but the typings are stricter in Commander 7 now we no longer store options values as properties by default (hurrah!). This might lead to requests for other properties to be declared too, but planning to only add them on demand rather than declare everything.

Still thinking about providing a higher level way of accessing global options (#1229 #1155) but don't have one yet!

## Solution

Declare `parent` property on `Command` object.

## ChangeLog

- add `parent` property to TypeScript definition for Command

### DIFF
--- a/typings/index.d.ts
+++ b/typings/index.d.ts
@@ -156,8 +156,8 @@ declare namespace commander {
 
   interface Command {
     args: string[];
-
     commands: Command[];
+    parent: Command | null;
 
     /**
      * Set the program version to `str`.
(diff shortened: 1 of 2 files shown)


abetomo (approved):
Thank you for your quick response.

Issue #476, opened 2015-12-09
Global Options


alex-phillips:
Is there a way to add global options to a program that is applied to all commands? I'm trying to be able to define a single `verbose` and `quiet` flag that is applied not only to the `program` but to all commands added as well.

I can post sample code if necessary, but I think it's fairly self-explanatory.

leesei:
```sh
Program
  .version(Pkg.version)
  .description(Pkg.description)
  .options('-v, --verbose', 'be verbose')
  .option('-q, --quiet', 'be quiet')


Program
  .command('search <query>')
  .description('do search')
  .option('-r, --regex', 'use regex')
  .action(function search (query, options) {
    // Program.quite
    // options.regex
  });
```

donaldpipowitch:
Can I write a default action for an option? It is quite nice that I can create a "global option" which is passed to every command, but if I have many commands I don't want to call `loggingFramework.setLogLevel(options.quite)` everywhere.

I can't do `commander.parse(process.argv); commander.opts();`, because then the commands action is called _before_ I can do something with `opts`. But it looks I need to call `parse` before I can use options correctly.

thetutlage:
@donaldpipowitch Going through the source code and found this line. https://github.com/tj/commander.js/blob/master/index.js#L393

It seems to be possible as

```js
Program.on('verbose', function () {
  process.env.VERBOSE = this.verbose
})

Program.on('quiet', function () {
  process.env.QUIET = this.quiet
})
```

Also for options has accept values, the value is passed to the callback function. For example

```js
Program.on('env', function (env) {
  process.env.NODE_ENV = env
})
```

```bash
some-command --env=production
```

donaldpipowitch:
Cool, thanks.

daldridge-cs:
So global options work in the scenario above, but they do *not* get propagated to sub-commands.  [This line](https://github.com/tj/commander.js/blob/master/index.js#L551) in `Command.prototype.executeSubCommand` only takes sub-command args.

If that line were changed from
```
args = args.slice(1);
```
to
```
args = argv.slice(2, argv.length - args.length).concat(args.slice(1));
```
then sub-commands would be forwarded those global options.

shadowspawn:
The original question got answered, and I don't think there will be action due to the subsequent comments.

Feel free to open a new issue if it comes up again, with new information and renewed interest.
 
Thank you  for your contributions.

shadowspawn:
[...]

Issue #243, opened 2014-08-27
Accessing parent options from subcommand


marc0s:
Are parent's options available to subcommands? In my tests, they aren't.

It would be nice to have them available. In my case is for configuring the remote server my CLI app needs to connect to, so I have not to define the same options in every subcommand.

``` javascript
# ./cmd
#!/usr/bin/env node

var program = require("commander");
program.version("0.0.1")
  .option("-H, --host <host>", "server host")
  .command("sub", "subcommand");
program.parse(process.argv);
console.log(program.host);

# ./cmd-sub
#!/usr/bin/env node

var program = require("commander");
program.version("0.0.1");
program.parse(process.argv);
console.log(program.host);
```

And if you run them:

```
$ ./cmd --host localhost 
localhost
$ ./cmd-sub --host localhost 
error: unknown option `--host'
```

Any trick to get them from the subcommand or is just not possible?

Thanks.

aynik:
Resurrecting this issue from the dead just to explain I used a little filthy hack to inherit options from the top-most command (because I'm into that):

On the top command you can export `program` with:

```javascript
if (!module.parent) {
  program.parse(process.argv)
} else {
  program.commands = []
  program._execs = []
  module.exports = program
}
```

And then from any subcommand you can replace:

```javascript
var program = require('commander')
```
For:

```javascript
var program = require('./<top-command>')
````
This way just the options will be inherited.

**Comment by b4dnewz** [comment]
hi @marc0s, can you share some code or example of how you did? just for the records, thankx

**Comment by pburgmer** [comment]
@b4dnewz I did it like this

```
const command = commander.command('merge');

command
      .description('foo')
      .option('--bar', 'really?', false)

addLoggingOption(command);

command
      .action(async (args: any) => { });
```

and

```
function addLoggingOption(command: Command): void {
  command
    .option('--log-level <level>', 'print log messages of given level and above only', 'info')
    .option('--verbose', 'set log level to debug')
    .option('--quiet', 'set log level to warn');
}
```

shadowspawn:
Opened poll on possible enhancements to global options: #1551

shadowspawn:
Opened a draft PR for `.addCommonOption()` #1670

PR #1478, opened 2021-02-28
WIP: Add globalOpts()

shadowspawn:
# Pull Request

## Problem

People wonder how to get the global options from a subcommand. 

The current approaches are calling say `program.opts()` if have a variable available for the program, or calling `cmd.parent.opts()` from in the action handler. And looping if there are multiple levels of subcommands with options.

Related: #243 #476 https://github.com/tj/commander.js/pull/1024#issuecomment-568403615 #1155 #1229

The main expected use of the parent property is to get global options: #1475

## Solution

Add `.globalOpts()` which returns the combined global options from global/parent commands.

I did consider adding a parameter to `.opts()` to return combined local and global options, but I think explicit routine is simpler. Likewise, I wondered adding a configuration option to have the combined options passed into the action handler, but again explicit routine is simpler to understand.

## To Do

- [x] typings
- [ ] tests
- [ ] README
- [ ] CHANGELOG

### DIFF
--- a/index.js
+++ b/index.js
@@ -1720,7 +1720,7 @@ class Command extends EventEmitter {
   };
 
   /**
-   * Return an object containing global options as key-value pairs
+   * Return an object containing global options as key-value pairs, from parent command(s).
    *
    * @return {Object}
    */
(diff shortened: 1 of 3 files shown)


shadowspawn:
Not sure what people actually want with global options. Closing this for now.

PR #1487, opened 2021-03-20
Add test for last variation of executable failures

shadowspawn:
Add a test for a tricky case which requires mocking a few layers to intercept the behaviour.

(Increases coverage a little, over 99% now.)


abetomo (approved):
👍

PR #1488, opened 2021-03-20
Fix syntactically incorrect command name in test

shadowspawn:
# Pull Request

## Problem

With more aggressive syntax checking in the arguments PR, found some of the existing tests had invalid syntax when they failed.

## Solution

Give the commands in the test a valid single name.


abetomo (approved):
👍

Issue #1471, opened 2021-02-13
Poll: change api for adding command arguments?


shadowspawn:
The syntax for adding command arguments which have help descriptions is quite different from the rest of the API, and arguments and options are added quite differently although they could be similar. (Prompted by discussion about offering `.choices()` for arguments in #1458.)

A couple of questions for feedback via reactions or comments.

shadowspawn:
Should `.argument(usage, description)` replace the use of `.description()` for supplying optional help for the command-arguments? 👍 👎

shadowspawn:
Should `.argument()` replace the use of `.arguments()` for adding command-arguments? 👍 👎

shadowspawn:
Not huge numbers, but unanimous positive response after a month. (Thanks for feedback.)

1. 8 👍 
2. 5 👍

NicolasThierion:
I found I can coerce options with a function as a third parameter.
Is there any plan to allow argument corecion the same way with this new syntax ?

shadowspawn:
Yes, planned and done. `.argument()` supports coercion (custom processing) and default value, following same pattern as `.option()`.

https://github.com/tj/commander.js/pull/1508

Issue #1458, opened 2021-01-31
add choices to arguments


Niryo:
is it possible to have choices for arguments?
```javascript
createCommand('pickColor').arguments('<color>'); //color choices: ['red', 'blue', 'green']
```

shadowspawn:
No, choices for command-arguments are not supported. 

(But an alternative is to make `color` an option instead of an argument.)

Niryo:
Is it intentionally not supported or is it in the backlog? If it's in the backlog we can discuss about the api and maybe i could create a PR

shadowspawn:
At this stage it is a new suggestion.

I see `.choices()` as just one of the things that option-arguments support that command-arguments do not: default value, custom option processing (coerce), and choices.

Do any other CLI frameworks support any of these for command-arguments?

shadowspawn:
I had a quick look at argparse, and it configures command-arguments and option-arguments the same way:  https://docs.python.org/3/library/argparse.html#the-add-argument-method

(This is very different to how Commander currently works!)

shadowspawn:
Previous related issue: #971

Niryo:
My first intuition was that arguments should be treated the same as options in terms of validation and api, so it will be quite natural to have `addArguments(new Argument().choices())`, same as options, don't you think?

shadowspawn:
I think following the option syntax may work pretty well, allowing description in a more consistent way, and defaults and coercion, and the extensible syntax to allow choices. Specifying the arguments with `.arguments()` or along with the command name in `.command()` still works fine as a shortcut for the simple cases. Using examples from the README:

```js
// arguments and description
oldProgram
  .arguments('<username> [password]')
  .description('test command', {
    username: 'user to login',
    password: 'password for user, if required'
  });
newProgram
  .description('test command')
  .argument('<username>', 'user to login')
  .argument('[password]', 'password for user, if required');

// default
cmdUsingOpt
  .option('-c, --cheese <type>', 'add the specified type of cheese', 'blue');
cmdUsingArg
  .argument('[cheese-type]', 'add the specified type of cheese', 'blue')

// coerce
cmdUsingOpt
  .option('-f, --float <number>', 'float argument', parseFloat)
cmdUsingArg
  .argument('<number>', 'float argument', parseFloat)

// choices
cmdUsingOpt
  .addOption(new Option('-d, --drink <size>', 'drink size').choices(['small', 'medium', 'large']));
cmdUsingArg
[...]

PR #1490, opened 2021-03-27
More consistent way of adding command arguments

shadowspawn:
# Pull Request

## Problem

The syntax for adding command arguments which have help descriptions is quite different from the rest of the API, and arguments and options are added quite differently although they could be similar. (Prompted by discussion about offering .choices() for arguments in #1458)

Related: #1458 #1467 #1471

## Solution

This builds on PR #1467. With positive feedback about the new method in  #1471., switching over "normal" way of adding an argument in README and examples and tests from `.arguments()` to `.argument()`.

- `.argument(name, description)` method
    - c.f. `.option(flags, description)`
- `addArgument(arg)` method for future expansion
    - c.f. `.addOption(opt)`
- deprecate `cmd.description(cmdDescription, argsDescription)`

## ChangeLog

- added: `cmd.argument(name, description)` for adding command-arguments
- changed: deprecate second parameter of `cmd.description(desc, argDescriptions)` for adding argument descriptions (removed from README)
- changed: Help `.visibleArguments()` returns array of `Argument`

### DIFF
--- a/index.js
+++ b/index.js
@@ -817,7 +817,37 @@ class Command extends EventEmitter {
   };
 
   /**
-   * Define argument syntax for the command.
+   * Define argument syntax for command.
+   *
+   * The default is that the argument is required, and you can explicitly
+   * indicate this with <> around the name. Put [] around the name for an optional argument.
+   *
+   * @example
+   *
+   *     program.argument('<input-file>');
+   *     program.argument('[output-file]');
+   *
+   * @param {string} name
+   * @param {string} [description]
+   * @return {Command} `this` command for chaining
+   */
+  argument(name, description) {
+    const argument = new Argument(name, description);
+    this.addArgument(argument);
+    return this;
+  }
+
+  /**
+   * Define argument syntax for command, adding multiple at once (without descriptions).
+   *
+   * See also .argument().
+   *
+   * @example
+   *
+   *     program.arguments('<cmd> [env]');
+   *
+   * @param {string} names
+   * @return {Command} `this` command for chaining
    */
 
   arguments(names) {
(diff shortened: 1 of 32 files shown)


shadowspawn:
Allowing no brackets around argument name and defaulting to required option, but not documenting for now.

i.e. `foo` treated same as `<foo>`

shadowspawn:
Not documenting `.addArgument()` in README yet, as not interesting until has some extra behaviours (like .`choices()`!).

shadowspawn:
Might be done! I am going to have a look at the Argument data property names, but they are consistent with Option and Command. Having a plain data property name like `required` means we can't have a method called `required()` in the future (well, without it being a breaking change).

shadowspawn:
Also I will have another look at the README introduction of the arguments, as the "inline" form with "command-name <arg1> [arg2]" is the only one that can be used when declaring stand-alone executables from the parent command. `.argument()` does not replace this. (Not all the patterns work for program vs subcommand, and action handler vs stand-alone executable.)

index.js
```diff
@@ -344,6 +372,52 @@ class Help {
...
+        this.required = true;
+        this._name = name;
+        break;
+    }
+
+    if (this._name.length > 3 && this._name.slice(-3) === '...') {
+      this.variadic = true;
+      this._name = this._name.slice(0, -3);
+    }
+  }
```
cravler, line 408: looks like this is needed as BC
```js
this.term = this._name; // legacy, BC
```

shadowspawn: That will be due to recent change to `visibleArguments()`, so returns `Argument` rather than plain object with `{term, description}`.

https://github.com/tj/commander.js/pull/1490/commits/f54d1edce8c7b77afd779b13c2ce2d38d331085f

cravler: yes

shadowspawn:
I have had another go at the README to try and tidily cover the multiple approaches to specifying command-arguments.

abetomo (approved):
8️⃣

index.js
shadowspawn: Not planning to patch a `term` property into Argument, as following the Help pattern there is now full `argumentTerm()`  method.

PR #1491, opened 2021-03-28
Remove placeholder test for old issue

shadowspawn:
# Pull Request

Noticed a test file for rechecking open issues, but the only test is for an issue which has been fixed. Delete file.


abetomo (approved):
👍

PR #1497, opened 2021-04-07
Factory routine to create a new unattached argument

cravler:

### DIFF
--- a/index.js
+++ b/index.js
@@ -837,6 +837,21 @@ class Command extends EventEmitter {
   };
 
   /**
+   * Factory routine to create a new unattached argument.
+   *
+   * See .argument() for creating an attached argument, which uses this routine to
+   * create the argument. You can override createArgument to return a custom argument.
+   *
+   * @param {string} name
+   * @param {string} [description]
+   * @return {Argument} new argument
+   */
+
+  createArgument(name, description) {
+    return new Argument(name, description);
+  };
+
+  /**
    * Define argument syntax for command.
    *
    * The default is that the argument is required, and you can explicitly
@@ -852,7 +867,7 @@ class Command extends EventEmitter {
    * @return {Command} `this` command for chaining
    */
   argument(name, description) {
-    const argument = new Argument(name, description);
+    const argument = this.createArgument(name, description);
     this.addArgument(argument);
     return this;
   }
--- a/esm.mjs
+++ b/esm.mjs
@@ -1,4 +1,15 @@
 import commander from './index.js';
 
 // wrapper to provide named exports for ESM.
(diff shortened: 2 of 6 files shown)


shadowspawn:
Thanks. We have a rich pattern for Command then Option to follow! Not much to customise yet, but I do want to follow the same patterns.

shadowspawn:
(I saw from a deleted commit that we missed the esm export too.)

shadowspawn:
(Sorry, I should have waited to see if the ESM tweaks were coming back! 😊 )

shadowspawn (approved):
LGTM

shadowspawn:
(This will land on `release/8.x`.)

abetomo:
It would be better if you also update `typings/index.d.ts`.

shadowspawn:
(I'll add some tests too, but I can do the tests afterwards.)

shadowspawn:
(Added TypeScript and tests in a PR to the PR.)

abetomo (approved):
Thanks!

index.js
```diff
@@ -836,6 +836,21 @@ class Command extends EventEmitter {
...
+   * See .argument() for creating an attached argument, which uses this routine to
+   * create the argument. You can override createArgument to return a custom argument.
+   *
+   * @param {string} name
+   * @param {string} [description]
+   * @return {Argument} new argument
+   */
+
+  createArgument(name, description) {
+    return new Argument(name, description);
```
Niryo, line 851: What are the benefits of adding this factory when you can use `new Argument()` directly?

shadowspawn: The main purpose is so you can override the method and return a custom subclass of Argument, which will be used to create the arguments added by Command methods `.argument()` and `.arguments()` and `.command('sub <file>`).

This factory pattern was first offered with `.createCommand()` where it is more likely to be used, but offering the same pattern for Option and now Argument.

Niryo: But why will you want to override the createArgument method when you can just pass any custom Argument class to addArgument?
also, in order create custom argument, you need to somewhat know the internals of Argument class (in both the factory solution or by using it directly), so it sounds like a bad practice and a doc bloater, for probably a really rare use case

shadowspawn: `.createCommand()` was the first one added in #1191 to make Command more subclass friendly. `.createHelp()` was added in major help refactor in #1365.

`.createOption()` was added in #1380 to follow the pattern. We had established a tidy pattern for `Command` which also worked for `Help`, so following through and offering same support for `Option`.

> in order to create custom argument, you need to somewhat know the internals of the Argument class

Yes, that is implicit in writing a subclass.

I did consider whether to directly support subclassing at all for `Command`, but it a useful approach for extending functionality especially in custom cases which are not common use cases and would not be integrated into Commander itself.

PR #1501, opened 2021-04-10
Check command-arguments even without action handler

### LINKED ISSUE #1493: Required non-option arguments only checked with .action()?
It appears that `commander@7.2.0` only checks `.arguments()` for the top-level command if `.action()` is given.  For example:

```js
#!/usr/bin/env node
const command = new Command();
command
  .arguments('<source>')
  .allowExcessArguments(false)
  .description('do something with a single source file')
  .action(() => {});
command.parse();
console.log('Called with arguments:', command.args);
```

Has the following behavior

    $ ./example.js
    error: missing required argument 'source'
    $ ./example.js one
    Called with arguments: [ 'one' ]
    $ ./example.js one two

If `.action(() => {});` is commented out, it has the following behavior:

    $ ./example.js 
    Called with arguments: []
    $ ./example.js one
    Called with arguments: [ 'one' ]
    $ ./example.js one two
    Called with arguments: [ 'one', 'two' ]

Which was surprising to me.  If this is intentional, perhaps a comment could be added to the [Specify the argument syntax section of `Readme.md`](https://github.com/tj/commander.js/blob/master/Readme.md#specify-the-argument-syntax) to warn users?  If not, perhaps it could be fixed?

Thanks for considering,
Kevin

shadowspawn:
# Pull Request

## Problem

The number of command-arguments is only checked if there is an action handler. This means no checking for simple programs which fall-through to processing the results of the parsing in following code.

Fixes #1493

## Solution

Follow in the footsteps of the refactored tests for options. Both the option and command-argument tests were originally tied to the implementation of the action handler. They have now been separated out to be called where appropriate.

Check for missing required command-arguments, or excess command-arguments if enabled, in programs without an action handler.

## ChangeLog

- the number of command-arguments is now checked for programs without an action handler


shadowspawn:
(Wrong base branch, will redo and make new PR.)

shadowspawn:
(Wrong base branch, will redo and make new PR.)

PR #1502, opened 2021-04-10
Check command-arguments even without action handler

### LINKED ISSUE #1493: Required non-option arguments only checked with .action()?

shadowspawn:
# Pull Request

## Problem

The number of command-arguments is only checked if there is an action handler. This means no checking for simple programs which fall-through to processing the results of the parsing in following code.

Fixes #1493

## Solution

Follow in the footsteps of the refactored tests for options. Both the option and command-argument tests were originally tied to the implementation of the action handler. They have now been separated out to be called where appropriate.

Check for missing required command-arguments, or excess command-arguments if enabled, in programs without an action handler.

## ChangeLog

- the number of command-arguments is now checked for programs without an action handler

### DIFF
--- a/index.js
+++ b/index.js
@@ -1559,11 +1559,8 @@ class Command extends EventEmitter {
           this.unknownOption(parsed.unknown[0]);
         }
       };
-
-      const commandEvent = `command:${this.name()}`;
-      if (this._actionHandler) {
-        checkForUnknownOptions();
-        // Check expected arguments and collect variadic together.
+      // Check for missing or extra arguments, and refactor the arguments for passing to action handler.
+      const getCheckedArguments = () => {
         const args = this.args.slice();
         this._args.forEach((arg, i) => {
           if (arg.required && args[i] == null) {
@@ -1576,11 +1573,17 @@ class Command extends EventEmitter {
         if (args.length > this._args.length) {
           this._excessArguments(args);
         }
+        return args;
+      };
 
-        this._actionHandler(args);
+      const commandEvent = `command:${this.name()}`;
+      if (this._actionHandler) {
+        checkForUnknownOptions();
+        this._actionHandler(getCheckedArguments());
         if (this.parent) this.parent.emit(commandEvent, operands, unknown); // legacy
       } else if (this.parent && this.parent.listenerCount(commandEvent)) {
         checkForUnknownOptions();
+        getCheckedArguments();
         this.parent.emit(commandEvent, operands, unknown); // legacy
       } else if (operands.length) {
         if (this._findCommand('*')) { // legacy default command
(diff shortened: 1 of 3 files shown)


index.js
```diff
@@ -1576,11 +1573,17 @@ class Command extends EventEmitter {
...
 
-        this._actionHandler(args);
+      const commandEvent = `command:${this.name()}`;
+      if (this._actionHandler) {
+        checkForUnknownOptions();
+        this._actionHandler(getCheckedArguments());
         if (this.parent) this.parent.emit(commandEvent, operands, unknown); // legacy
       } else if (this.parent && this.parent.listenerCount(commandEvent)) {
         checkForUnknownOptions();
+        getCheckedArguments();
```
abetomo, line 1586: The name `getXXX` is a function to get something.

`this._actionHandler(getCheckedArguments());`.
For this code, it makes sense.

However, the code here does not use a return value.
I thought it was hard to understand what the 'get' was for.

abetomo (approved):
I commented on the naming, but the fix is LGTM.

index.js
shadowspawn: Thanks. I'll take another look. I started without it and added it to make the "get" usage make more sense. I'll try separating out the check and the variadic support since that is only needed in one case.

shadowspawn: Reworked. Longer but hopefully clearer.

abetomo: Thanks.

shadowspawn:
Note for future readers. Missing command-arguments will now cause an error whether or not there is an action handler. This changes the behaviour for simple single command programs.

```js
program.argument('<file>');
program.parse();
```

```
$ pizza
error: missing required argument 'file'
```

If you want to just display the help if the user does not supply any arguments, you could check before calling parse like:

```js
if (process.argv.length === 2) program.help();
program.parse();
```

Also, see `.showHelpAfterError()` to change the default behaviour to include extra help.

PR #1508, opened 2021-04-21
Extend .argument() to add default value and custom parse for command-argument

shadowspawn:
# Pull Request

## Problem

Follow through with the idea that `.argument()` follow same pattern as `.option()`, with support for default value (for optional arguments) and custom command-argument parsing.

## Solution

Support for default value (for optional arguments) and custom command-argument parsing.

Implementation notes:
- not emitting events for the arguments
- only processing the arguments for the action handler, not for `program.args`
- custom processing for arguments uses same style function as for options

## ChangeLog

- add default value for optional command-arguments
- add custom processing  function for command-arguments
- deprecate `InvalidOptionArgumentError` (replaced by `InvalidArgumentError`)
- `CommanderError` code `commander.invalidOptionArgument` renamed `commander.invalidArgument`

### DIFF
--- a/index.js
+++ b/index.js
@@ -385,6 +385,9 @@ class Argument {
   constructor(name, description) {
     this.description = description || '';
     this.variadic = false;
+    this.parseArg = undefined;
+    this.defaultValue = undefined;
+    this.defaultValueDescription = undefined;
 
     switch (name[0]) {
       case '<': // e.g. <required>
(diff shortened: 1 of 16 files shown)


index.js
```diff
@@ -258,6 +258,13 @@ class Help {
    */
 
   argumentDescription(argument) {
+    const extraInfo = [];
```
abetomo, line 261: `extraInfo` appears to have only 0 or 1 elements.
Does it need to be an array?

shadowspawn: Fair question. No, does not need to be an array currently. However, I left in the array as planning to add choices, like `optionDescription` which I copied the code from.

abetomo (approved):
LGTM!

shadowspawn:
Thanks @abetomo

PR #1511, opened 2021-05-02
Update example file extensions

hasantezcan:
## Problem
Example file's extensions is not valid. So I fixed that.

## Solution
Fixed the example files extensions like `.js`


shadowspawn:
1) The files without file extension are following a style which can be used on Mac and Windows, where the file is made executable and the script includes a special line at the start to tell the shell how to run it. If the files is put in the search path for command, then for example the `pizza` script can be run by typing `pizza`, rather than needing to type `pizza.js`.

- https://en.wikipedia.org/wiki/Shebang_(Unix)

Does adding a file extension help you significantly?

2) There are two files included in the PR with style changes which are not consistent with the style used in Commander.

hasantezcan:
The reason that I do this contribution is instability between examples file extensions. I thought that might be forgotten while created. And actually, they look cool after this touch. Now the all examples look symmetrical.

Q: Is that any problem with adding shebang the don't exist ones? @shadowspawn

shadowspawn:
I don't think I want to add extension to all the example files, but going the other way `pm-list.js` should probably not have a suffix to match the other `pm` files and be consistent in that group.

To acknowledge the files without extension I suggest we could add below the "Try the following:" an example of using it directly like:

```
// Try the following:
//  node custom-help --help
//
// On Mac and Linux the script can also be run directly like:
//  ./custom-help --help
```

shadowspawn:
Lack of file extension is not an error as such.

PR #1513, opened 2021-05-05
Refactor async support to return (possible) promise directly

shadowspawn:
# Pull Request

## Problem

The promise returned by an async action handler was being stored in root command and retrieved after parsing. This pattern does not extend to adding other async callbacks, such as discussed in #1197.

## Solution

Return the action handler result directly (rather than storing).

Factor out the preparation code from `.parse()` so can be called symmetrically from `.parseAsync()`. Use simple async and await pattern in `.parseAsync().


abetomo (approved):
It’s nice.

Issue #1457, opened 2021-01-31
Support event listener for option fallback to default value


Kerruba:
Hello,

I would like to set a global variable based on the value associated with an option. But right now there's no support for the default value of an option. My plan was to do something like this
```
let program = new commander.Command();
program.option('-t, --test <value>', 'a test option', 'my-default')
program.on('option:test', function(value) => {
   process.env.MY_ENV_VAR = value 
})
```

but this only works if the option is passed, not if the default value is used.

A possible workaround right now would be to set the env variable outside of the program and override it only if the option is passed, but check this other issue instead https://github.com/tj/commander.js/issues/1456

shadowspawn:
Another approach is to set the environment variable after the parsing when the option has the final value, so for example at the start of the action handler. 

Your work-around of setting the environment variable before parsing would mean less repeated code for doing this with a global option.

shadowspawn:
There is some overlap with #1197, from the point of view of a hook after the options are parsed but before the action is run.

shadowspawn:
I am not planning to emit events when falls back to default value. Some work-arounds discussed. No further activity in a month. Closing.

Feel free to open a new issue if it comes up again, with new information and renewed interest.

chriscalo:
I have a need for something similar. I'm trying to find a way of setting environment variables _inside_ the commander script and the suggested workaround doesn't seem to address the need.

> Another approach is to set the environment variable after the parsing when the option has the final value, so for example at the start of the action handler.

The problem with this is that I need to set environment variables **before** any action is run. As I understand it, as soon as you call `program.parse()`, command actions get called, making it too late to set environment vars.

I basically want to do something like the following:

``` sh
my-script --env=prod cmd foo
# OR
my-script cmd foo # defaults to --env=dev
```

and have some function called before any actions are called to handle the value of the `--env` option.

I don't care too much what the solution is, but I've looked deeply over the docs and searched for other answers and can't find any way of doing what I'm looking for.

One fix is what this bug describes, that a default value for `--env` will also trigger the `option:env` event. Then I could do something like the following, though this doesn't work today:

``` js
#!/usr/bin/env node
const { Command } = require("commander");

const program = new Command();

program
  .option("-e, --env <name>", "set env vars from <name>.env file", "dev")
  .on("option:env", function (name) {
    // I want to set env vars here
    // unfortunately, this doesn't get called for the default value
    // how do I force this function to always get called?
  });

program
  .command("foo")
  .action(function () {
[...]

Issue #1158, opened 2020-01-31
Add possibility to set program.on("close") event or event for all commands


prokopsimek:
I continue in our discussion in closed PR here: https://github.com/tj/commander.js/pull/996

My intent is to display a notification about the new package version in CLI at the end of each run of any command.

I have prepared my intended implementation here: https://github.com/DXHeroes/dx-scanner/pull/232#pullrequestreview-351337584 (the PR is invalid)

shadowspawn:
There isn't currently a generic event after any command runs.

How about checking for the notification after calling `parse`? It seems this might have a similar outcome that the notification is displayed if the parse did not detect errors or display the help, but I may be missing some of your workflow cases.

```
    await cmder.parseAsync(process.argv);
    this.notifyUpdate);
```

prokopsimek:
Thanks, I'll try that and give feedback.

prokopsimek:
@shadowspawn Wow, it works! 🎉
It works if I run a subcommand (e.g. `./bin/run practices`) but doesn't work if I run a root command that displays help (just ``./bin/run`). The `.help` function terminates with `process.exist(0)`?

```typescript
   if (!process.argv.slice(2).length) {
      cmder.help();
    }
```
ref.: https://github.com/DXHeroes/dx-scanner/pull/231/files#diff-f41e9d04a45c83f3b6f6e630f10117feR77-R79

shadowspawn:
`.outputHelp()` displays the help without calling exit.

shadowspawn:
This was a specific case which might have been resolved, and no recent activity. Closing in favour of #1197.

Feel free to open a new issue if it comes up again, with new information and renewed interest.
 
Thank you  for your contributions.

prokopsimek:
@shadowspawn Thanks for your help! It was a great experience. 🙂 👍

PR #1514, opened 2021-05-07
Add hooks for life cycle events: using .hook()

shadowspawn:
# Pull Request

## Problem

People ask for a way to run code before the action handler with access to the parsed options. And also some requests for a hook after the action.

Main issue: #1197
Related issues: #76 #229 #936 #996 #1158 #1457

## Solution

Add `.hook(event, callback)` method
- supports async callbacks
- supports multiple hooks for same event on single command

Supported hooks: `preAction`, `postAction`. Triggered by action handler for hooked command and its subcommands.

Callback is passed the hooked command and the command running the action handler.

## ChangeLog

- add `.hook()` with support for `"preAction"` and `"postAction"` callbacks

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -574,6 +574,8 @@ program
   });
 ```
 
+The callback hook can be `async`, in which case you call `.parseAsync` rather than `.parse`. You can add multiple hooks per event.
+
 The supported events are:
 
 - `beforeAction`: called before action handler for this command and its subcommands
--- a/Readme.md
+++ b/Readme.md
@@ -564,9 +564,8 @@ Example file: [hook.js](./examples/hook.js)
 ```js
 program
   .option('-t, --trace', 'display trace statements for commands')
-  .hook('beforeAction', (context) => {
-    if (context.hookedCommand.opts().trace) {
-      const actionCommand = context.command;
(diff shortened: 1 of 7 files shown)


shadowspawn:
Wondering about changing signature to named method and parameters, say:

```js
program
   .preAction((thisCommand, actionCommand) => {});
   .postAction((thisCommand, actionCommand) => {});
```

shadowspawn:
Why not preHandler or afterHandler to also cover executable? The stand-alone executable seems quite different, and I have been avoiding tackling! Initial concerns without deep digging:
- the command which is about to be "executed" is a dummy command
- the command object which is about to be executed does not currently have `.args` set
- the program gets terminated when the subprocess finishes (?), but could perhaps run afterHandler before exit? But what about async?

shadowspawn:
As noted in https://github.com/tj/commander.js/pull/1519#issuecomment-841617643, the `.hook()` api signature is similar to that used by Fastify:

https://www.fastify.io/docs/latest/Hooks/

abetomo (approved):
LGTM

shadowspawn:
Thanks @abetomo

I'll take another look at the code refactor next, and then focus on a prerelease of Commander v8.

PR #1515, opened 2021-05-08
Add test to cover exception from custom argument processing

shadowspawn:
# Pull Request

## Problem

New argument support reduced test coverage, didn't cover a failure path.

## Solution

Add test that not repackaging ordinary exceptions into CommandError.

(Statement test coverage changed from 99.06% to 99.17%.)


abetomo (approved):
👍

PR #1516, opened 2021-05-08
Simplify addHelpText TypeScript signature and improve typings test

shadowspawn:
# Pull Request

## Problem

Declaring that `.addHelpText()` callback may return `undefined` makes for a messy signature. The intent was to allow a falsey value returned to skip display, by just calling `return` in callback.

## Solution

Restrict to strings. User can easily return empty string to get the falsey behaviour of skipping display.

Improve the typings test of the callback.

## ChangeLog

- TypeScript definition for `.addTextHelp()` callback no longer allows result of undefined, now just string.


abetomo (approved):
👍

Issue #229, opened 2014-06-25
I've forked/added features to commander, which ones are worth doing pull requests for?


joshacheson:
I have my own fork of this project for a cli I'm putting together.

In my fork I currently have a few new features. They are:

1) Variadic arguments. Pretty self explanatory... I added argument parsing for variadic arguments using the syntax "command <somearg> [VARIADIC...]" where the word VARIADIC can be replaced by any string, of course. An array of values are then passed to an action callback in the 2nd to last position, right before the options/self/this reference.

2) A bootstrap function. This is basically a callback you can optionally add to any instance of command that will run before any sub command's action callback. The purpose of this feature was to have access to parsed args and be able to perform logic with them before the program proceeds further downstream to eventual .action() callbacks. In my case, I used this to create a bottleneck where I can configure my logger before any of the actions for my CLI are performed.

and soon I'll be 

3) slightly altering automatically generated help output to do string padding because i've found misaligned help output at times

If anyone believes it's worth actually putting together pull requests for any of these features  please let me know and I will do so as soon as I've got the time.

Cheers

PR #1519, opened 2021-05-13
Add hooks for life cycle events: using .preAction(), .postAction()

shadowspawn:
# Pull Request

## Problem

People ask for a way to run code before the action handler with access to the parsed options. And also some requests for a hook after the action.

Main issue: #1197
Related issues: #76 #229 #936 #996 #1158 #1457

## Solution

Add `.preAction()` and `.postAction()` methods:

- supports async callbacks
- supports multiple hooks for same event on single command

Triggered by action handler for hooked command and its subcommands.

Callback passed two Command parameters:

- `thisCommand`: the command which the hook is added to, useful for working with global properties. 
- `actionCommand`: the command with the action handler

## ChangeLog

- add `.preAction()` and `.postAction()` hooks

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -557,16 +557,15 @@ If the program is designed to be installed globally, make sure the executables h
 
 ### Life cycle hooks
 
-You can add callback hooks to a command for life cycle events.
+You can use `.beforeAction()` and `.afterAction()` to add callbacks before and after the action handler.
 
 Example file: [hook.js](./examples/hook.js)
 
 ```js
 program
   .option('-t, --trace', 'display trace statements for commands')
-  .hook('beforeAction', (context) => {
-    if (context.hookedCommand.opts().trace) {
-      const actionCommand = context.command;
+  .beforeAction((thisCommand, actionCommand) => {
+    if (thisCommand.opts().trace) {
       console.log(`About to call action handler for subcommand: ${actionCommand.name()}`);
       console.log('arguments: %O', actionCommand.args);
       console.log('options: %o', actionCommand.opts());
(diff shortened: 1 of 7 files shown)


shadowspawn:
# Pull Request

## Problem

People ask for a way to run code before the action handler with access to the parsed options. And also some requests for a hook after the action.

Main issue: #1197
Related issues: #76 #229 #936 #996 #1158 #1457

## Solution

Add `.preAction()` and `.postAction()` methods:

- supports async callbacks
- supports multiple hooks for same event on single command

Triggered by action handler for hooked command and its subcommands.

Callback passed two Command parameters:

- `thisCommand`: the command which the hook is added to, useful for working with global properties. 
- `actionCommand`: the command with the action handler (like `command` property passed to `.addHelpText()` callback)

## ChangeLog

- add .preAction()` and `.postAction()` hooks

shadowspawn:
For an example of a single hook routine which is passed an event name rather than separately named methods for the hooks, see:

https://www.fastify.io/docs/latest/Hooks/

A key reason for the single hook approach is if there are going to be lots of hooks. The pattern is similar to eventEmitter and `.on(event, ....)`.

shadowspawn:
For interest, also noticed in the fastify docs:

> Hook functions are invoked with `this` bound to the associated Fastify instance.

Commander does the same with the `action` callback. Because it does not work with arrow functions, I have not been treating this as the main way of passing in command context.

shadowspawn:
I am quite liking how `fastify` looks, and I like including "hook" in the wording as it is how this sort of pattern is often described.

Thinking of switching back to hook! This time with parameters:

```js
program
   .hook('preAction', (thisCommand, actionCommand) => {});
   .hook('postAction', (thisCommand, actionCommand) => {});
```

PR #1520, opened 2021-05-15
Simplify typings

### LINKED ISSUE #1465: Command' is not assignable to type CommanderStatic.Command
I try to create an init command using TypeScript, as follows：

```ts
import { Command } from "commander";
export const init = ():Command => {
  const init = new Command("init");

  init
    .arguments("[name]")
    .option("--disable-lint", "disable lint", false)
    .action(async (name = "", options) => {
      //  my action
    });

  return init;
};
```

But at last return init, typescript report error:

TS2322: Type 'import("/Users/qishaoxuan/work/leo/node_modules/commander/typings/index").Command' is not assignable to type 'import("/Users/qishaoxuan/work/leo/node_modules/commander/typings/index").CommanderStatic.Command'.   The types returned by 'version(...)' are incompatible between these types.     Type 'Command' is not assignable to type 'this'.       'Command' is assignable to the constraint of type 'this', but 'this' could be instantiated with a different subtype of constraint 'Command'.

shadowspawn:
# Pull Request

## Problem

The TypeScript typings are pretty complicated to describe default export of a Command object in addition to
Classes. This was appropriate at the time as a faithful description of the javascript implementation.

Related issues: #1037 #1465 and mention in https://github.com/tj/commander.js/pull/1392#discussion_r525995991

#1037 was previously fixed by describing the class constructors as separate functions. (Yuck.)

## Solution

Remove the default export of the program object and use named exports. Way simpler and tidier!

The default import of global Command object was removed from README in Commander v5. Deprecated from Commander v7. Deleted in this PR.

Fixes #1465 (and #1037 fixed too, in new way)

(Hide whitespace changes to make the diff look much simpler.) 

## ChangeLog

- remove default export of global Command object from TypeScript declarations
- refactor and simplify TypeScript declarations

### DIFF
--- a/docs/deprecated.md
+++ b/docs/deprecated.md
@@ -42,7 +42,9 @@ const { Command } = require('commander');
 comnst program = new Command()
 ```
 
-Removed from README in Commander v5. Deprecated from Commander v7. 
+- Removed from README in Commander v5.
+- Deprecated from Commander v7. 
+- Removed from TypeScript declarations in Commander v8.
 
 ## Callback to .help() and .outputHelp()
 
--- a/typings/index.d.ts
+++ b/typings/index.d.ts
@@ -75,7 +75,7 @@ export class Option {
   /**
    * Set the default value, and optionally supply the description to be displayed in the help.
    */
-  default(value: any, description?: string): this;
+  default(value: unknown, description?: string): this;
 
   /**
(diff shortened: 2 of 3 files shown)


abetomo (approved):
LGTM!

PR #1521, opened 2021-05-16
Add getOptionValue and setOptionValue

shadowspawn:
# Pull Request

(Note: this PR was branched from #1520 before that was merged, and first three commits are from that branch and actually already merged.)

## Problem

Now we are not storing options as properties on command, there is not a public
way of setting option values. This may become more common with the new hooks.

## Solution

Make the existing internal routines public and add tests et al

## ChangeLog

- add `.getOptionValue()` and `.setOptionValue()`


abetomo (approved):
👍

Issue #1496, opened 2021-04-05
Refactor index.js


Niryo:
Are we intentionally keeping this god `index.js` file as is, or are there plans for breaking it down to more manageable pieces? 
I am interested in contributing more stuff to this lib, and I thought that breaking up this huge file could be a good start if there are no objections...

shadowspawn:
My main concern with splitting up the monolithic `index.js` is if there are any common client use cases that would be broken by changing to multiple files. (No?)

A minor downside is refactoring the file makes it a harder to track back through the history to find why certain lines of code and logic were introduced. But the benefits to future work outweigh this.

For interest, the file was split up in #661 in a branch to modernise the codebase for v3.

Niryo:
About the first concern- I am not sure I can't think of a use case were client could be broken because of internal arrangement as long as the api stays the same. The api is the contract, everything else can change.

About the downside- yep, this is an annoying limitation of git, but as you said, the benefits to future work outweigh the this. 
And if you really need to track back through the history, you can still do it, nothing is lost, it just became less convenient.

For me it was super annoying to orient myself inside this huge index.js file, and a split is a no-brainer decision.

shadowspawn:
I had a play around with refactoring, but naively splitting Help and Command introduced a circular dependency. Keeping Help and Command in the same file is still the majority of the code. Not a trivial exercise. e.g.

https://github.com/shadowspawn/commander.js/tree/feature/refactor/src

shadowspawn:
I'll have another go at this when Pull Requests are quiet.

Niryo:
few notes:
1. I am not sure what are the implications of cyclic dependencies here, maybe when can live with it if we can't find any workarounds. But my hunch is that we can probably find a way to structure things in such a way that remove the cyclic dependency.
2. Even if keep those classes in the same file, we will still benefit from extracting small classes, even minor ones.
3. We can also extract pure functions. currently we have many big methods that sits inside the classes themselves, and some of them can be transformed to pure functions sittings outside the classes (where it makes sense of course, not all methods should be extracted this way).

shadowspawn:
Released in Commander v8.

PR #1522, opened 2021-05-18
Refactor classes into files

shadowspawn:
# Pull Request

## Problem

The single-file `index.js` is a bit intimidating to work in, specially for new contributors. We now have some classes which can stand alone.

Issue: #1496

## Solution

Short version: factor out the classes into separate files!

Implementation detail: the Help class takes parameter types of multiple classes, but this is a type dependency and not a code dependency. Using a TypeScript style JSDoc import allows the strong typing checks without introducing a circular `require` dependency between `help.js` and `command.js`. I am using Visual Studio Code which understands this for IntelliSense, and TypeScript uses this for `npm run typescript-checkJS`. If we have other editors people want to use that have their own imports, we can add them as required.

I like that the exports are now simple and obvious in `index.tab `.

## ChangeLog

- refactor `index.tab` into a file per class


shadowspawn:
Future reference: if we do need to do "real" imports then @cravler demonstrated how to simply break the cycle by having Command implement `createHelp` in a superclass in `index.tab`. Skeleton:

```js
class Command extends BaseCommand {
  createCommand(name) {
    return new Command(name);
  };

  createHelp() {
    return Object.assign(new Help(), this.configureHelp());
  };
}
```

shadowspawn:
I moved back to draft, I just want to check can install from a pack and it works before I call it ready.

abetomo (approved):
Great work!

PR #1524, opened 2021-05-20
Update dependencies

shadowspawn:
# Pull Request

## Problem

Update dependencies prior to major release.

## Solution

Update to latest of everything, apart from `@types/node` where used latest 14 version (lts) rather than 15 version.

Delete and rebuild `package-lock.json` and `node_modules` so also update indirect dependencies.


abetomo (approved):
📦

Issue #971, opened 2019-06-04
create commands that only accept fixed list of arguments


danielo515:
Hello,

I was wondering if it is possible to create a command on commander that looks like this

```
toggle (on|off) <args...>
```

Where `(on|off)` means that you must choose one or the other, mandatory.

Ideally, commander should take care of throwing the appropriate error and error message.

shadowspawn:
There is not support in Commander for this pattern. You could achieve something functionally similar by adding on and off commands, and throwing the appropriate error message yourself if not not supplied.

danielo515:
Thanks @shadowspawn , it is good to know.
Any plan to implement such feature?

Regards

shadowspawn:
No plan to implement. Something similar has been asked for options (#518) but I do not recall being asked for arguments. You are welcome to leave issue open if you want to see if there is interest from other people, or close when satisfied with answers. (I leave issues open for at least month before closing.)

(A small clarification, I was a bit hasty and misinterpreted your original question. I assumed `toggle` was the name of your actual program. My suggestion for "adding on and off commands" does not match your intended use.)

Yes `.usage` will override the usage displayed for the command itself. e.g.

```
# Before
$ node index.js toggle --help
Usage: toggle [options] <action> <FEATURE_NAME...>
...
# After
$ node index.js toggle --help
Usage: toggle (on|off) <FEATURE_NAME...>
...
```

shadowspawn:
An answer was provided, and no further activity in a month. Closing this as resolved.

Feel free to open a new issue if it comes up again, with new information and renewed interest.

Thank you  for your contributions.

shadowspawn:
This is supported from Commander version 8.

```mjs
import { Command, Argument } from 'commander';
const program = new Command();

program.command('toggle')
  .addArgument(new Argument('<state>', 'toggle state').choices(['on', 'off']))
  .argument('<args...>', 'further arguments')
  .action((state, args) => {
    console.log('state:', state);
    console.log('args:', args);
  });

program.parse();
```

```console
$ node index.mjs toggle purple 1 2 3
error: command-argument value 'purple' is invalid for argument 'state'. Allowed choices are on, off.
$ node index.mjs toggle on 1 2 3    
state: on
args: [ '1', '2', '3' ]
$ node index.mjs toggle --help
Usage: index toggle [options] <state> <args...>

Arguments:
  state       toggle state (choices: "on", "off")
  args        further arguments

Options:
  -h, --help  display help for command
```

PR #1525, opened 2021-05-21
Choices for arguments

shadowspawn:
# Pull Request

## Problem

We have `.choices()` for options but not for arguments.

Issue: #1458
Past issue: #971

## Solution

Follow same pattern as Option with the new Argument class, and add `.choices()`.

```js
program
  .addArgument(new Argument('color').choices(['red', 'blue']))
```

```sh
$ node example.js purple
error: command-argument value 'purple' is invalid for argument 'color'. Allowed choices are red, blue.
```

Includes some tests and typings for `.default()` and `.argParser()` that got missed.

## ChangeLog

- `.addArgument()` (#1490)
    - Argument supports `.choices()` (#1525)

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -22,8 +22,9 @@ Read this in other languages: English | [简体中文](./Readme_zh-CN.md)
     - [More configuration](#more-configuration)
     - [Custom option processing](#custom-option-processing)
   - [Commands](#commands)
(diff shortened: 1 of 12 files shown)


shadowspawn:
Rebased onto `release/8.x`.

abetomo (approved):
LGTM!

PR #1526, opened 2021-05-22
Support argument processing without action handler

shadowspawn:
# Pull Request

## Problem

The new processing of arguments is useful in a single command program without an action handler, but
argument defaults and custom processing and choices were only being applied if there was an action handler.

Prompted by `.choices()` not working without action handler.

## Solution

Refactor and process command-arguments whether or not action handler. 

When adding the functionality in #1508 I had decided that I preferred not to modify `this.args` with processed results.
Save processed arguments as `this.processedArgs` so accessible for simple program without action handler, and from the new hook support too.

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -479,7 +479,9 @@ program
 
 ### Custom argument processing
 
-You may specify a function to do custom processing of command-arguments before they are passed to the action handler.
+You may specify a function to do custom processing of command-arguments. The processed argument values are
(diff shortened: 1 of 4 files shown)


shadowspawn:
Ready for review _except_ I don't want to land anything until after #1522, so marking as draft.

(Started working on this on the wrong branch! 🤭 )

shadowspawn:
Redo Pull Request after code refactor. New PR is #1529

PR #1529, opened 2021-05-25
Support argument processing without action handler

shadowspawn:
# Pull Request

## Problem

The new processing of arguments is also useful in a single command program without an action handler, but
argument defaults and custom processing and choices were only being applied if there was an action handler.

Prompted by realising `.choices()` would not work without action handler.

## Solution

Refactor and process command-arguments whether or not action handler. 

When adding the functionality in #1508 I had decided that I preferred not to modify `this.args` with processed results.
Save processed arguments as `this.processedArgs` so accessible for simple program without action handler, and from the new hook support too.

This PR includes some minor other improvements:
- added .arg to Command constructor
- initialise .arg to empty array (matches existing TypeScript declaration)
- add JSDoc types for some key Command properties

## Changelog

- Changed: `Command` property `.arg` initialised to empty array (was previously undefined)

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -488,9 +488,11 @@ program
 
 #### Custom argument processing
 
-You may specify a function to do custom processing of command-arguments before they are passed to the action handler.
+You may specify a function to do custom processing of command-arguments (like for option-arguments).
 The callback function receives two parameters, the user specified command-argument and the previous value for the argument.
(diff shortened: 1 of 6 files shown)


abetomo (approved):
👍

shadowspawn:
Thanks @abetomo

Issue #1531, opened 2021-05-25
Is the documentation regarding options out of date?


jjs105:
Hi,

I have just started using commander and have been trying your examples (from the readme and example files) and not getting the expected results.

I think this is because the examples are out of date as follows (considering the options-common.js):

* It seems that the inclusion of boolean options as 'unassigned' in the result of the '.opts()' call doesn't happen, rather the properties simply do not exist
* After inspecting the code it seems that enabling the legacy option using '.storeOptionsAsProperties()' makes the example work as described

From the example, the following should occur:
```
$ pizza-options -d
{ debug: true, small: undefined, pizzaType: undefined }
```
However you actually get (unless setting legacy options to true):
```
$ pizza-options -d
{ debug: true }
```

TLDR; the new and legacy options objects act differently - this might be a bug or alternatively an documentation issue.

shadowspawn:
Yes, the README is out of date and shows the old behaviour. Thanks.

The treatment of undefined options in `.opts()` does change depending on `. storeOptionsAsProperties()`. 

`.storeOptionsAsProperties()` was added in Commander v4.1, but until Commander v7.0 the default behaviour was to store the options as properties so the README was still correct. With 7.0 the README became out of date.

shadowspawn:
Released in Commander v8.

PR #1532, opened 2021-05-26
Remove incorrect example output

shadowspawn:
# Pull Request

## Problem

Still showing old format for `.opts()` from before Commander 7 when `.storeOptionsAsProperties()` was default.

Issue: #1531

## Solution

Just remove the example, there is another "debug" example anyway, which as a bonus will display the same way independent of `.storeOptionsAsProperties()` setting as none undefined.

Also simplified the debug example by not combining the short flags.

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -131,12 +131,9 @@ if (options.pizzaType) console.log(`- ${options.pizzaType}`);
 ```
 
 ```bash
-$ pizza-options -d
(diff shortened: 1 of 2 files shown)


abetomo (approved):
👍

shadowspawn:
Nice easy review. 😄 Thanks @abetomo

PR #1537, opened 2021-05-29
feat: add generics to OptionValues

tyankatsu0105:
# Pull Request

## Problem

close #1536 

## Solution

TypeScript typings
- add generics to OptionValues
- add default generics to storeOptionsAsProperties and opts method

Issue #1536, opened 2021-05-29
[TypeScript typings]Accept generics value object for opts()


tyankatsu0105:
## What
In TypeScript, I want to use generics for opt methods.

```ts
type OptionValues = {
  cheese: string;
};

const options = program.opts<OptionValues>();

options.cheese
        ^^^^^^
// => (property) OptionValues.cheese: string
```

## Why

Now, interface OptionValues is record type that value is any.
https://github.com/tj/commander.js/blob/327a3ddd552f90d1dbd37f55fa7a88b9553468ae/typings/index.d.ts#L159-L161

I think generics help developers with TS if override the OptionValues type.

shadowspawn:
Released in Commander v8.

shadowspawn:
For some experimental strong inferred typing see: https://github.com/commander-js/extra-typings

PR #1539, opened 2021-05-30
Add strong typing of .opts() using generics

shadowspawn:
# Pull Request

## Problem

Unable to specify `.opts()` return type in TypeScript client code.

Issue: #1536

Original Pull Request: #1537 (manually moved code between branches, with co-author credit on commit)

## Solution

Add generics so can call like:

```js
interface MyCheeseOption {
  cheese: string;
}
const myCheeseOption = program.opts<MyCheeseOption>();
```

## ChangeLog

- add client typing of `.opts()` return type using TypeScript generics

### DIFF
--- a/typings/index.d.ts
+++ b/typings/index.d.ts
@@ -483,8 +483,8 @@ export class Command {
    *
    * @returns `this` command for chaining
    */
-  storeOptionsAsProperties(): this & OptionValues;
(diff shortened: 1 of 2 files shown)


abetomo (approved):
👍

shadowspawn:
For some experimental strong inferred typing see: https://github.com/commander-js/extra-typings

PR #1544, opened 2021-06-05
Improve example

shadowspawn:
Tiny change to make example more consistent. Want argument to be a technique, not an ingredient.


abetomo (approved):
🍳

PR #1547, opened 2021-06-12
Add tips about showing help instead of just missing arguments error

shadowspawn:
# Pull Request

## Problem

The behaviour is changing in Commander 8 for very simple programs which previously could decide after parsing what to do (if anything) about missing command-arguments.

## Solution

Add a couple of simple approaches to Migration Tips in CHANGELOG.

I am not sure whether this come up much, but I'm thinking of the fairly casual users of Commander.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,21 +15,21 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 If you have a simple program without an action handler, you will now get an error if
 there are missing command-arguments.
 
-```
+```js
 program
-   .option('-d, --debug')
-   .arguments('<file>');
+  .option('-d, --debug')
+  .arguments('<file>');
 program.parse();
 ```
 
-```
+```sh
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -8,6 +8,39 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
...
 <!-- markdownlint-disable MD004 -->
 
+## [Unreleased] (date goes here)
+
+### Migration Tips
+
+If you have a simple program without an action handler, you will now get an error if
+there are missing command-arguments.
+
+```
```
abetomo, line 18: I think an indent width of 2 is better.
Because other code examples are 2.

It would be nice to have a highlight setting as well. ` ```js `

abetomo (approved):
👍

CHANGELOG.md
shadowspawn: 👍

PR #1550, opened 2021-06-22
Note Chinese translations are stale

shadowspawn:
# Pull Request

## Problem

The changes to files translated into Chinese have not been updated.

## Solution

Make the simple code change in translated `options-taking-varying-arguments.md`.

Add a note that document is still describing Commander v7 in translated `Readme.md` and `deprecated.md`.


abetomo (approved):
🆗

PR #1553, opened 2021-06-29
Add dependabot and cache dependencies in workflow 

mmorel-35:


shadowspawn:
I have not been too concerned with the dependencies since Commander does not have any production dependencies. But something we can consider.

Documentation:
-  https://docs.github.com/en/code-security/supply-chain-security/keeping-your-dependencies-updated-automatically/configuration-options-for-dependency-updates

shadowspawn:
Actions cache:

- https://github.com/actions/cache

Possible complication with caching across operating systems:

- https://github.com/actions/cache/issues/591#issuecomment-845121153

mmorel-35:
That's what is defined by the dependabot.yml file. What would you like to add to this configuration? Change the frequency of updates from daily to weekly or monthly? Anything else?

shadowspawn:
I had a look at the logs for actions for this PR with caching and previous PR without. It seems to make a small improvement for this project, but less than 4%. (I totalled caching and install and main tests across the 12 node-vs-platform targets.)

I was interested to see the files for `dependabot` and caching, but I am not keen on either.

Happy to consider further if @abetomo would like to try either.

abetomo:
@shadowspawn I'm not enthusiastic about it either, but how about running it for a while?
If there is a problem, you can revert.

shadowspawn:
I might give dependabot a try in one of my own repos.

PR #1557, opened 2021-07-03
Add copyInheritedSettings

shadowspawn:
# Pull Request

## Problem

Adding a subcommand using `.command()` implicitly copies a bunch of inherited settings from the parent command to the subcommand. There is not currently any public way to transfer inherited settings, say if using `.addCommand()`, and many are not accessible using public routines or properties.

Related: #1186 https://github.com/tj/commander.js/pull/1185#issuecomment-584559274

## Solution

Add `.copyInheritedSettings()` so available for explicit use, particularly if using `.addCommand()'.

Adding for completeness, and not planning to add this to README for now. Has not come up in issues so far.

## ChangeLog

- add `.copyInheritedSettings()`

### DIFF
--- a/lib/command.js
+++ b/lib/command.js
@@ -76,13 +76,12 @@ class Command extends EventEmitter {
   /**
    * Copy settings that are useful to have in common across root command and subcommands.
    *
-   * (Used internally when adding a command using `.command()` to copy settings
-   * from the parent command to the subcommand.)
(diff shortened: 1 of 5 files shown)


abetomo (approved):
👍

shadowspawn:
Released in Commander v8.1

Issue #1359, opened 2020-09-24
Add Option/Command from hash?


shadowspawn:
Are authors interested in a factory method to construct commands or options from a hash of the properties? This would be an alternative approach rather than a replacement for the current chaining interface. The use case is adding options and commands from meta information supplied from elsewhere.

e.g.

```
const my = readOption();
const myOption = Option.fromHash({ flags: my.flags, description: my.description, mandatory: my.mandatory, hidden: my.hidden });
program.addOption(myOption)
```

shadowspawn:
Related:
- https://github.com/tj/commander.js/issues/1277#issuecomment-641849117

mshima:
I don't think this is required anymore with addOption been exposed:
```
const my = readOption();
program.addOption(
  new Option(my.flags, my.description)
    .makeOptionMandatory(!!my.mandatory)
    .hideHelp(!!my.hidden));
```

But would be nice to expose every option:
```
const my = readOption();
program.addOption(
  new Option(my.long, my.description)
    .short(my.short)
    .required(!!my.required)
    .optional(!!my.optional));
```

shadowspawn:
> But would be nice to expose every config

I did wonder about adding chaining setters for the lower level properties, but chose to start by keeping the existing Option constructor taking the flags parameter, and have the flags and flags related properties consistent when set this way. 

The properties are public.

shadowspawn:
This issue didn't get much interest. It isn't likely to get acted on due to this report.

Feel free to open a new issue if it comes up again, with new information and renewed interest.

Issue #665, opened 2017-07-20
Can configure option from JSON for more readability


luatnd:
Hi,
I feel that this commander.js is the best cli parser I found. I choose it to develop my small cli app.
But when I start to use it while reading its docs, it take me 2 times to find the `<>` `[]` syntax, but still missing the `-no-` syntax explanation.
So I've made a `configure()` that port to `option()` for an easy work:

JSON config --> configure() --> option()
It look like this, please tell me if you think it's good, I'll submit this code, hope this help:

![image](https://user-images.githubusercontent.com/1859127/28436887-cfeaa162-6dc2-11e7-94b7-f70c2c5809ef.png)

And it still have some issue:
• #601 Adding support for required arguments
• #230 Required arguments for options and commands are not enforced
• #574 Defining boolean options via Command#option() defaults to undefined, not false

----
For further refactoring, In my opinion, I think these features might good:
• Validation value
• Change parseOptions() behavior: It check through the `argv`'s option only, so every option defined but not in `argv` was not be checked for default value.
• Typescript

Thanks.

vanesyan:
Hello there! Really thanks for your interesting in project. While your feature is good, we'd prefer to keep commander's API small and consistent and sure easily to create high-level wrappers on top of it, means we are beside of opinionated decisions like this :-)

PR #1558, opened 2021-07-03
Extend Option with argRequired and argOptional

shadowspawn:
# Pull Request

Fill out the Option methods to allow explicit configuration as alternative to implicit syntax from string.

This is not likely to be used much, and some downsides to having two ways to do something! Trying out an  idea. 

Method naming is intended to work with Argument too.

## Problem

The syntax for required (`<arg>`) and optional (`[arg]`) option-arguments needs to be learned. (However, this is consistent with the syntax shown in the help so is useful to learn!)

Related: #665 https://github.com/tj/commander.js/issues/1359#issuecomment-699580203 

## Solution

Add `.argRequired()` and `.argOptional()` to Option.

Adding for completeness, and not planning to add this to README for now. 

## ChangeLog

- add Option methods for `.argRequired()` and `.argOptional()` as alternative way of specifying option-argument

### DIFF
--- a/lib/command.js
+++ b/lib/command.js
@@ -278,9 +278,8 @@ class Command extends EventEmitter {
    * indicate this with <> around the name. Put [] around the name for an optional argument.
    *
    * @example
-   *
-   *     program.argument('<input-file>');
-   *     program.argument('[output-file]');
+   *   program.argument('<input-file>');
+   *   program.argument('[output-file]');
    *
    * @param {string} name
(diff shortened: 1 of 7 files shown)


shadowspawn:
Happy with this as a proof of concept, but not sure it is worthwhile unless some demand.

Going to try similar pattern for Argument, where it should be simpler. (e.g. keep argument name separate)

Closing this for now.

PR #1561, opened 2021-07-07
Markdown autodocs

dineshsonachalam:
## Problem

To make your repo more appealing and useful you need to provide example code snippets in your README.md. Manually copy and pasting each code snippet in their respective places in your README would be inefficient and time-consuming.

## Solution

This problem can be solved using markdown-autodocs a GitHub Action that automatically generates & updates markdown content (like your README.md) from external or remote files. You need to add markers in your README.md that will tell markdown-autodocs where to insert the code snippet.

Github Actions Marketplace: https://github.com/dineshsonachalam/markdown-autodocs

## 🚀 markdown-autodocs github action used by:

* [iro.js - 🎨 Modular color picker widget for JavaScript, with support for a bunch of color formats](https://github.com/jaames/iro.js)
* [LucidDynamodb - A minimalistic wrapper to AWS DynamoDB](https://github.com/dineshsonachalam/Lucid-Dynamodb)

Please let me know if any changes are required.

cc: @tj

### DIFF
--- a/.github/workflows/markdown-autodocs.yml
+++ b/.github/workflows/markdown-autodocs.yml
@@ -1,0 +1,14 @@
+name: markdown-autodocs
+
+on: [push, pull_request]
+
+jobs:        
+  auto-update-readme:
+      runs-on: ubuntu-latest
+      steps:
+        - uses: actions/checkout@v2
+        - name: Markdown autodocs
+          uses: dineshsonachalam/markdown-autodocs@v1.0.2
+          with:
+            # Optional output file paths, defaults to '[./README.md]'.
+            output_file_paths: '[./README.md,./Readme_zh-CN.md]'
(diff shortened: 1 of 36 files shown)


shadowspawn:
I am not seeing enough value for how I have been working with the README.

1) I only show selected text from the full worked examples, to keep the quoted example concise and the README smaller. This PR (and `markdown-autodocs`) pull in the whole example file. The whole file is just a click away for readers, as we have a link if the quoted code is from a full example file. (Possibly `markdown-autodocs` could take a selection from full example file, but it would be extra markup which would detect from the example file being used directly.)

2) I do not see the advantage in creating separate files for small self-contained code or shell snippets. I feel this loses the context of the surrounding text which the code-block is supporting. Easier for me to just edit in place?

3) A possible benefit for the shell snippets would be if they could actually be run, and update the output generated by the commands. However, I don't think that is what `markdown-autodocs` can do.

PR #1566, opened 2021-07-11
Use getCommandAndParents for array of parents

shadowspawn:
Small code saving, use `getCommandAndParents` for building list of commands.


abetomo (approved):
👍

PR #1567, opened 2021-07-11
Feature/argument arg explicit

shadowspawn:
# Pull Request

Fill out the Argument methods to allow explicit configuration of required/optional with a fluent method. This is an alternative to implicit syntax using brackets in name string.

Related: https://github.com/tj/commander.js/issues/1359#issuecomment-699580203

## Problem

The syntax for required (`<arg>`) and optional (`[arg]`) command-arguments needs to be learned. (However, this is consistent with the syntax shown in the help so is actually useful to learn!)

## Solution

Offer an explicit method to set whether option is required or optional.

```js
program.addArgument(new Argument('foo').argRequired());
program.addArgument(new Argument('bar').argOptional());
```

The method naming is intended to be the same for adding matching methods to Option in future.

## ChangeLog

- add Argument methods for `.argRequired()` and `.argOptional()`.

### DIFF
--- a/lib/argument.js
+++ b/lib/argument.js
@@ -109,6 +109,22 @@ class Argument {
     };
     return this;
   };
+
+  /**
+   * Make option-argument required.
+   */
+  argRequired() {
+    this.required = true;
+    return this;
+  }
+
+  /**
+   * Make option-argument optional.
+   */
+  argOptional() {
+    this.required = false;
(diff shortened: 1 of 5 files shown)


abetomo (approved):
LGTM

shadowspawn:
Released in Commander v8.1

Issue #1569, opened 2021-07-16
Implicit program "name", deep dive


shadowspawn:
The user can set the program name. If they don't, Commander has a go at finding a name to use. This turns out to be quite hard and less consistent than you might expect! (In practice a user will only be using a subset of all the possibilities, but still some potential for getting caught out.)

## constructor

```js
const program = new Command('constructor');
program.parse();
```

```
% node example -h
Usage: constructor [options]
```

## .name()

```js
const program = new Command();
program.name('name');
program.parse();
```

```
% node example -h
Usage: name [options]
```

## Implicit, from args

A commonly used method for command-line scripts to find the name of the currently running scripts is to examine the passed arguments. The Node.js documentation for `process.argv` says:

> The process.argv property returns an array containing the command-line arguments passed when the Node.js process was launched. The first element will be process.execPath. See process.argv0 if access to the original value of argv[0] is needed. The second element will be the path to the JavaScript file being executed. The remaining elements will be any additional command-line arguments.

Commander uses the parsed arguments to determine the default name if they (should) include the script name. The problem is there are so many ways to launch a node application and not all pass the script name, and platform differences as well.

```
#!/usr/local/bin/node
const program = new Command();
program.parse(process.args);
```

Passing script directly:

```
% node script.js -h
Usage: script [options]
```

Running script as an executable:
```
$ ./executable-file -h
Usage: executable-file [options]
```

For running via `package.json` in directory `package-dir`:

```json
  "main": "main-file.js",
  "bin": {
    "bin-executable": "executable-file.js"
  },
```

Running a package returns the name of the package directory (not the script name):
```
% node . -h
Usage: package-dir [options]
```

Running an installed binary on Mac and Linux gets the bin name (not the script name):
```
% npm link
...
% bin-executable -h
Usage: bin-executable [options]
```

Running an installed binary on Windows gets the script name (not the bin name):
```
> npm link
...
> bin-executable -h
Usage: executable-file [options]
```

## Implicit, not from args

Since Commander v5, we support parsing arguments which do not include the script name. In this case we fallback to using the main module, from Commander 8 using `require.main.filename`. This does give consistent behaviour for all the launch methods with CommonJs.

```js
const program = new Command();
program.parse(process.argv.slice(2), { from: 'user' });
```

```
% node script.js -h
Usage: script [options]
```

```
% ./executable-file -h
Usage: executable-file [options]
```

```
% node . -h
Usage: main-file [options]
```

```
% npm link
...
[...]

Issue #1439, opened 2021-01-18
addCommand() -- how to?


btmurrell:
I see the discussion in #1186 that acknowledges that `addCommand` lacks doc... i am getting the error below when trying to use it.  the object I am passing to it is of type `const cmd = new Command();  cmd.command('foo'); `    Other than giving the command a name in this way ('foo'), how else would one specify a name?

```
/home/merl/dev/EDS/dev-cli/node_modules/commander/index.js:225
    if (!cmd._name) throw new Error('Command passed to .addCommand() must have a name');
                    ^

Error: Command passed to .addCommand() must have a name
    at Command.addCommand (/home/merl/dev/EDS/dev-cli/node_modules/commander/index.js:225:27)
    at Object.<anonymous> (/home/merl/dev/EDS/dev-cli/src/commands/eds.js:14:9)
```

shadowspawn:
To give your new command a name, you can supply the name with the constructor or by calling a method:

```js
const cmd1 = new Command('foo');

const cmd2 = new Command();
cmd2.name('bar');
```

btmurrell:
hmmmm. thanks @shadowspawn . that seems simple enough, and that worked, so I am unblocked... but there's no doc or pattern for `new Command('foo')`, and just one mention of using `cmd.name('foo')` to affect the usage info.  If I understood it more, I would put in a doc PR, but I think there's more there.  in the mean time, this issue might help others. thanks!

Chuwiey:
I might be doing an anti pattern, but I am running into a similar `if (!cmd._name) throw new Error('Command passed to .addCommand() must have a name');` error.

In my `main.ts`:
```
import { testCommand } from './commands/test'
const { Command } = require('commander')

const program = new Command()

program
    .version('0.0.1')
    .description('Fake package manager')
program.addCommand(testCommand())
program.parse(process.argv)
```

in `commands/test.ts`:
```
export function testCommand() {
        const heat = program.command('heat')
        heat
            .command('jug')
            .action(() => {
                console.log('heat jug')
            })
        heat
            .command('pot')
            .action(() => {
                console.log('heat pot')
            })
        return heat
}
```

A console.log on install() return the correct Command structure and I can see the relevant `_name` properties. However, the `addCommand` function sees them as undefined.

Issue #1417, opened 2020-12-17
Registering git style subcommands leveraging standalone executable files


brandongregoryscott:
_Context:_
I am unsure if there is something incorrect with my setup/proposed use-case, or if this is a feature that isn't supported yet by the API. May be related to #714. I am working off of version 6.0.0.

_Use-case:_
I would like to be able to register commands (parent and nested) that exist as standalone executable files from the entrypoint or 'top-level' command file.

_Problem:_
The `command()` overload that takes the `{ executableFile?: string }` object only exists on the `commander.CommanderStatic` instance and not the `addCommand()` function which is used to add prepared subcommands.

I have an [example repository here](https://github.com/brandongregoryscott/commanderjs-git-style-subcommand-example), which has a workaround solution in the `main` branch (ie, I register the subcommand `cli-dotnet-version.js` in the standalone file for the parent, `cli-dotnet.js`), whereas in the `issue` branch, I would really like to be able to register both `dotnet` and its subcommand `version` in `cli.js` so I can run `./cli.js dotnet version`

In the `issue` branch, the output from running the cli has both commands at the entrypoint level and `version` does not get nested under `dotnet`:

```SH
./cli.js
Usage: cli [options] [command]

Options:
  -h, --help      display help for command

Commands:
  dotnet          Parent command housing operations around the dotnet cli
  version         Displays dotnet installation info
  help [command]  display help for command
  ```
  ```SH
  ./cli.js dotnet
Usage: cli-dotnet [options]

This is my custom version of the dotnet command

Options:
  -h, --help  display help for command
  ```
  
  whereas the output from the `main` branch is what I'm looking for:
  
  ```SH
  ./cli.js
Usage: cli [options] [command]

Options:
  -h, --help      display help for command

Commands:
  dotnet          Parent command housing operations around the dotnet cli
  help [command]  display help for command
```
```SH
./cli.js dotnet
Usage: cli-dotnet [options] [command]

This is my custom version of the dotnet command

Options:
  -h, --help      display help for command

Commands:
  version         Displays dotnet info
  help [command]  display help for command
  ```

shadowspawn:
I am not seeing the problem on the `main` branch. You have a hierarchy of external subcommands. Each level only knows about its direct subcommand. The top level program does not know that `dotnet` has a `version` subcommand, but it does not need to as it passes the arguments on to that subcommand to process.

It works the way I expect when I try:

```sh
$ ./cli.js
# displays help from cli.js, including command dotnet

[...]

Issue #1044, opened 2019-09-14
Sub-commands do not work with package.json main


samherrmann:
Consider the following setup:

lib/index.js
```js
const program = require('commander');
 
program
  .command('update', 'update installed packages', { executableFile: './myUpdateSubCommand' })
  .parse(process.argv);
```

lib/myUpdateSubCommand.js
```js
const program = require('commander');

program.command('foo').action(() => console.log('hello from foo'))
program.parse(process.argv);

if (program.args.length === 0) {
  program.help();
}
```

package.json
```json
{
  "main": "lib/index.js"
}
```

# What works:
```sh
$ node dist/index.js update
Usage: myUpdateSubCommand [options] [command]

Options:
  -h, --help  output usage information

Commands:
  foo
```

# What does not work:
```sh
$ node . update
error: ./myUpdateSubCommand(1) does not exist, try --help
```

# Expected Behavior
It's expected that executing `node . update` yields the same result as executing `node dist/index.js update`.

# Where the problem is:
The path to the sub-command is determined from [argv[1]](https://github.com/tj/commander.js/blob/v3.0.1/index.js#L526), which does not take into consideration the path assigned to the `main` property in `package.json`.

# Version
"commander": "^3.0.1",

shadowspawn:
Yes, for resolving subcommand executables Commander expects `argv[1]` is either the source file, or the installed binary command (i.e. a symbolic link). There is not currently support for coping with a package being passed to node (like when you call `node .`).

The approach I often use is a setup closer to the eventual install. So with a `package.json` like:

```
{
  "name": "my-command",
  "bin": "bin/index.js"
}
```

Instead of `node .` I use:

```
npm link
my-command update
```

(You could work-around the current design in your own code by supplying `argv[1]` yourself. I did a bit of reading and wonder how reliable using `require.main.filename` might be. )

shadowspawn:
Related issue, when calling from Electron app: #512

shadowspawn:
Closing as covered by #512

PR #1571, opened 2021-07-18
Rework default program name, and searching for subcommands

shadowspawn:
# Pull Request

The classic approach for scripts to find how they are called is to check the passed arguments, which often start with the interpreter and the script name. e.g. `['/usr/local/bin/node', '/Users/developer/test.js', '--print']`. However, this is not always available or consistent for Commander use cases.

The implied script path found in `.parse()` affects the default program name, and searching for subcommands if stand-alone executable commands are used.

## Problem

- default program name gives different results depending on how script launched
- default stand-alone executable subcommand search gives different results for npm global commands on Windows and non-Windows
- naming the program (or any command) is under-documented in the README
- searching for stand-alone executable subcommands is under-documented in the README
- looking for stand-alone executable subcommand using script name is a bit subtle, and using command name would be a simpler and more consistent model
- adding the `require.main.filename` fallback has introduced different (arguably better) behaviours than script name, but can't be made primary behaviour because not available in es6
- if script name not available, must use `executableFile` per command

Note: we do have `.name()` and `executableFile`, so there is technically a way for user to get things working.

Deep dive: #1569

Related (including some issues for which `executableFile` is a solution): #481 https://github.com/tj/commander.js/issues/521#issuecomment-450038996 #532 #714 #785 #786 #826 #830 #875 #890 #912 #1044 #1417 #1439

## Solution

Node.js itself does not offer universal solutions for finding a reference script name or a directory which we can use from inside Commander. There are some approaches that client can use. e.g. `__filename`, `__dirname`, `process.argv[1]`, `require.main.filename`, `import.meta.url`.

Keep it simple. The script name is the classic mechanism, and Commander has always used it as the primary implicit context. Stick with it, and improve the documentation, and improve custom control.

Expand documentation about naming commands, and especially program. Rework documentation about finding subcommands. 

Remove fallback to `require.main.filename` since it adds yet more behaviours, and only a partial solution as not available in es6.

Use program (command) name rather than script name as prefix for looking for standalone executable subcommands. Include a backwards compatible fallback subcommand lookup using the script name.

Add `.executableDir()` for simple custom configuration, including when script name not supplied. (c.f. `.executableFile`)

(Includes some additional minor markdown and example code fixes.)

## ChangeLog

- changed: use command name as prefix for subcommand stand-alone executable name (with fallback to script name for backwards compatibility)
- added `.executableDir()` for custom search for subcommands
- allow absolute path with `executableFile`
- removed internal fallback to `require.main.filename` when script not known from arguments passed to `.parse()` (can supply details using `.name()`, and `.executableDir()` or `executableFile`)
- removed restriction that nested subcommands must specify `executableFile`


abetomo (approved):
LGTM 👍

shadowspawn:

PR #1574, opened 2021-07-26
Update github-actions

mmorel-35:
Updates github-actions with their most recent version and enable caching of the npm dependencies


.github/workflows/tests.yml
```diff
@@ -13,10 +13,11 @@ jobs:
         os: [ubuntu-latest, windows-latest, macos-latest]
 
     steps:
-    - uses: actions/checkout@v1
+    - uses: actions/checkout@v2.3.4
     - name: Use Node.js ${{ matrix.node-version }}
-      uses: actions/setup-node@v1
+      uses: actions/setup-node@v2.3.0
```
abetomo, line 18: I think `v2` is better than `v2.x.x`.

abetomo (approved):
Thanks!

.github/workflows/tests.yml
mmorel-35: It's done, the workflow is working fine on my repository.

shadowspawn (approved):
LGTM

PR #1580, opened 2021-08-10
style: use `||=`

abetomo:
# Pull Request

## Solution

I think it's better to use a `||=`, what do you think?


abetomo:
Ah, it was a syntax error in Node.js 14 and earlier.

PR #1589, opened 2021-08-25
Conditional exports to simply ECMAScript import

shadowspawn:
# Pull Request

## Problem

We are currently using a subpath for ECMAScript imports.

```
import { Command } from 'commander/esm.mjs';
```

Node.js has added support for conditional imports to mask the differences between CommonJs and ECMAScript:
- https://nodejs.org/dist/latest-v14.x/docs/api/packages.html#packages_conditional_exports

but the combination of ECMAScript without a warning and the conditional export support are only available from Node.js v12.20.0.

Node.js v12 is still LTS until 2022-04-30:

- https://nodejs.org/en/about/releases/

Related: #1284

## Solution

Add conditional exports to `package.json` for unqualified import. Still support the old subpath import. Update required Node.js version to v12.20.0.

v12.20.0	was released 2020-11-24, so by the time Commander v9 is released it will have been available for over a year.

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -65,30 +65,26 @@ Commander exports a global object which is convenient for quick programs.
 This is used in the examples in this README for brevity.
 
 ```js
+// CommonJS (.cjs)
 const { program } = require('commander');
-program.version('0.0.1');
 ```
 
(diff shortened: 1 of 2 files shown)


shadowspawn:

Issue #660, opened 2017-07-17
program.on behaviour seems to be broken in v2.11.0


deepu105:
The below snippet which used to work with v2.10.0 seems to be broken in v2.11.0 it doesn't do anything now
```
program.on('*', (name) => {
        console.log(name);
        process.exit(1);
    });
```

Actual usage in JHipster CLI
```
program.on('*', (name) => {
        logger.error(`${chalk.yellow(name)} is not a known command. See '${chalk.white(`${cliName} --help`)}'.`);

        const d = didYouMean(name.toString(), program.commands, '_name');

        if (d) {
            logger.info(`Did you mean: ${chalk.yellow(d)}?`);
        }

        process.exit(1);
    });
```

vanesyan:
Yeah, we break compatibility in v2.11. `*` event renamed to `command:*`. Sorry for that.

vanesyan:
see #659 for more details

deepu105:
Thanks for the update.

madjam002:
This should not be closed. Can this not be fixed for 2.10 and the breaking change be moved to a version 3 release?

vanesyan:
@madjam002 this is a dup of #659.

vanesyan:
Unfortunately we cannot push things back as 2.11 already released, and we cannot delete versions from npm. Plus things already done, no way to re-inverting wheels, sorry for that :)

Issue #432, opened 2015-09-07
running an unknown subcommand does not fail


francoislaberge:
This may be by design, but I would have expected calling:

``` bash
command <unknown-sub-command> 
```

...to fail. But instead it just exits immediately, this can lead to confusion when tired as you might think your command isn't working, but instead it's just spelt incorrectly.

Is there a flag or something to enable this error. Or should I just make a white list of known commands and fail when the first parameter isn't a known sub command?

CodeMan99:
You should fail in your parent command explicitly. Remember that some programs may do processing in the parent command while still having subcommands.

francoislaberge:
Yeah, this is what I ended up doing. Added some git style:
```bash
Unknown command. Did you you mean:
  push
  clone
  ...
```

CodeMan99:
It would be nice to have it fail if your "program chain" contains only `.command` and no `.option`. Or have the ability to turn on failing when there isn't a matching subcommand.

That said, this is definitely a feature request. Perhaps we could work together to create a pull request?

gangstead:
Here's a couple of lines of code to add to your top level command to fail for unknown subcommands:

```js
const _ = require('lodash');


const subCmd = _.head(program.args);
const cmds = _.map(program.commands, '_name');

if (!_.includes(cmds, subCmd)) {
  program.help();
}
```

francoislaberge:
@gangstead Yeah, I ended up doing something like what git does. Where it detects if there isn't a match and then does a fuzzy match and asks you if you meant to type: <list of subcommands with a close hamming distance)

gangstead:
That's much nicer than my quick and lazy solution.  Is that code anywhere I can check it out?

francoislaberge:
It's for our internal cli. Here's a gist though with the relevant parts isolated:
https://gist.github.com/francoislaberge/80ca6cbc33232c438680

shadowspawn:
By default, unknown subcommands show an error message in the published pre-release of Commander v5.0.0 (#1163)

shadowspawn:
Commander v5.0.0 has been released.

https://github.com/tj/commander.js/releases/tag/v5.0.0

PR #1590, opened 2021-08-28
Suggestion for unknown command and unknown option

shadowspawn:
# Pull Request

## Problem

It is common for big utilities to suggest matches for command typos:

```
% git loe
git: 'loe' is not a git command. See 'git --help'.

The most similar commands are
	clone
	log

% npm publih

Usage: npm <command>

...short-help...

Did you mean one of these?
    publish
    unpublish

% hg loe
hg: unknown command 'loe'
(did you mean one of clone, locate, log?)
```

CLI parsing:

- https://yargs.js.org/docs/#api-reference-recommendcommands opt-in, single suggestion
- https://github.com/spf13/cobra/blob/master/user_guide.md#suggestions-when-unknown-command-happens opt-out

Resolves issue: #1015

Note. There have not been a lot of requests for this but I have been noticing it in programs I use. It is already possible for user to add their own suggestions for commands, and there has been interest in that:

- https://github.com/tj/commander.js/issues/660#issue-243400277
- #1176
- https://github.com/tj/commander.js/issues/432#issuecomment-175785162
- https://github.com/tj/commander.js/issues/1015#issuecomment-521103051

## Solution

1. Check for possible matches to add to the unknown-command error message. Take into account the visible command alias, the implicit help command, and ignore hidden commands. Use the Damerau–Levenshtein distance to calculate the edit distance, which tries insertion, deletion, substitution, and transposition.

```
% node index.js rat
error: unknown command 'rat'
(Did you mean cat?)
```

This does not affect users who already have their own `command:*` listener.

2. Check for possible matches to add to the unknown-option error message. (None of git or hg or npm do this.) Take into account global options, the implicit help option, and ignore hidden options.

```
% node index.js sub --debgu
error: unknown option '--debgu'
(Did you mean --debug?)
```

3. Offer multiple suggestions (if more than one at closest edit distance).

```sh
% node index.js fake
error: unknown command 'fake'
(Did you mean one of bake, make, rake?)
```

4. Make the suggestions off by default to minimise breakage in Commander 8, and probably switch to on by default in Commander 9.

## ChangeLog

- added: `.showSuggestionAfterError()` to show suggestions after unknown command or unknown option

And fixed a bug discovered when testing:

- changed: show error for unknown global option before subcommand (rather than just help)


abetomo (approved):
awesome!

shadowspawn:
Thanks Abetomo. 👍

shadowspawn:
Shipped in Commander v8.2.0

nicojs:
Thanks for adding this feature. Easy to use and works like a charm. Brings a lot of value! 💚

Issue #1588, opened 2021-08-22
Add a quick start section in README


ayoubmehd:
For new Users the current README way is confusing.
it should be a quick start section that explains the basic usage.

shadowspawn:
Do the examples help as a quick introduction?

https://github.com/tj/commander.js#examples

(I reworked the examples and had intended to link to them from the top of the document as a quick idea of what using Commander was like, but didn't add the link.)

ayoubmehd:
@shadowspawn Yes, that's Good

Amoghtech:
Can i take this issue?

shadowspawn:
Short version: yes.


Long version

Thanks for asking and offering.

I am interested in improving onboarding for a new user, and it is years since I was one.

(I may make changes or do something different in the end, as I have rewritten much of the README in the last couple of years and mostly in "my" style now.)

ayoubmehd:
÷ I opened a pull request #1591

shadowspawn:
I was inspired to have another go at this today: #1665

shadowspawn:
Commander v9 has been released.

PR #1591, opened 2021-08-28
docs(README): add quick start section with a simple code example

ayoubmehd:
# add quick start section 

## Problem
Add a quick start section #1588 

## Solution

README

Add a quick start section with a link at the end to the Example section.

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -11,6 +11,7 @@ Read this in other languages: English | [简体中文](./Readme_zh-CN.md)
 
 - [Commander.js](#commanderjs)
   - [Installation](#installation)
+  - [Quick Start](#quickstart)
   - [Declaring _program_ variable](#declaring-program-variable)
   - [Options](#options)
     - [Common option types, boolean and value](#common-option-types-boolean-and-value)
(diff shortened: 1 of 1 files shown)


Readme.md
```diff
@@ -59,6 +60,21 @@ For information about terms used in this document see: [terminology](./docs/term
 npm install commander
 ```
 
+## Quick Start
+
+```js
+const { program } = require("commander");
+
+program
+    .command('list') // The name of your command
```
shadowspawn, line 69: I see potential for confusion in the comment for new user between giving the program a name, and adding a new subcommand?

Perhaps: add a subcommand named "list"

shadowspawn (changes requested):
Cosmetic: be consistent with whether starting comments with lowercase or uppercase letter

The big two elements are commands and options, and programs with only options are a real use case. I would like an example of an option too (on `list` command).

Readme.md
ayoubmehd: Yeah I see

shadowspawn:
Do you think having both a small example at the top and the longer example at the bottom is useful? Or should the example be pulled up to the top of the file?

ayoubmehd:
@shadowspawn The Quick Start examples are for new users, the bottom examples will be just fine there, those are more 'advanced'

shadowspawn:
I was wondering about showing some example usage, but think focus on simple code could be good.

(Still reading and thinking abut Quick Starts.)

shadowspawn:
This has been a useful exercise for me to see a candidate Quick Start, but it isn't adding enough value in current form. I'll close the PR and keen the issue open, and see if any other issues come up that would be helped by a Quick Start.

PR #1599, opened 2021-09-09
Remove TypeScript declaration for Option.argumentRejected

### LINKED ISSUE #1598: Option.argumentRejected declared but not implemented.
Hi in your typing file your write:

````typescript
  /**
   * Validation of option argument failed.
   * Intended for use from custom argument processing functions.
   */
  argumentRejected(messsage: string): never;
````

but `argumentRejected` do not exist anywhere.

shadowspawn:
# Pull Request

## Problem

Have left over TypeScript declaration for `Option.argumentRejected ` which never shipped.

Resolves: #1598

See also: #1392

## Solution

Delete the declaration!

Normally deleting a method would be a breaking change, but in this case there is no implementation anyway so I think it is ok to ship in a minor version.

## ChangeLog

- remove stale TypeScript declaration of `Option.argumentRejected`

### DIFF
--- a/typings/index.d.ts
+++ b/typings/index.d.ts
@@ -126,12 +126,6 @@ export class Option {
   hideHelp(hide?: boolean): this;
 
   /**
-   * Validation of option argument failed.
-   * Intended for use from custom argument processing functions.
-   */
-  argumentRejected(messsage: string): never;
-
-  /**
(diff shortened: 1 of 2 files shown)


abetomo (approved):
👍

Issue #1603, opened 2021-09-14
Does not work on node 12


paladox:
We're using the node package from debian which is 12.22.5 and also npm which is 7.5.2.

We cannot get the commands to work... I've tried the demo where I could reproduce (e.g. https://github.com/tj/commander.js/blob/master/examples/defaults)

Returns

```
root@mw8:/srv/mathoid/commander.js# examples/defaults
  - undefined template engine
  - undefined cheese
  - undefined
```

shadowspawn:
Oh, that example is out of date. But convenient to show the problem!

Option values used to be stored on the command. The new default, since Commander v7.0, is to store them separately. The option values can be accessed with `.opts()` whichever way they are stored. The relevant setting is `storeOptionsAsProperties()`.

See the migration tips in the v7.0.0 release notes: https://github.com/tj/commander.js/releases/tag/v7.0.0

paladox:
Aha! Thanks!. I see there's storeOptionsAsProperties which fixed it!

PR #1604, opened 2021-09-15
Tidy up some old examples

shadowspawn:
# Pull Request

Prompted by mention in #1603 of a broken example, cleaned up some of the old examples.

`defaults` had multiple problems, and there is a newer `options-defaults.js` used in the README , so deleted!


abetomo (approved):
👍

Issue #1609, opened 2021-09-21
Listen to any command


kuncevic:
I look to the docs and found that [example](https://github.com/tj/commander.js#custom-event-listeners):

```
program.on('command:*', function (operands) {
  console.error(`error: unknown command '${operands[0]}'`);
  ...
});
```

However, it does not work unless you specify a particular command.

I can see that with [life cycle hooks](https://github.com/tj/commander.js#life-cycle-hooks) you can achieve something similar to what I was expected from the above example. 

The question is, does the above example suppose to handle any command (`command:*`)?

shadowspawn:
The `'command:*'` dates back to the original implementation of having a default command. It is sent when there is an argument that does _not_ match one of the known subcommands. It is now only sent for backwards compatibility. As a side-affect it allowed handling unknown commands as in the example, but there is a built-in error now and also `.showSuggestionAfterError()`. (Which makes me think `'command:*'` should finally be deprecated!)

> The question is, does the above example suppose to handle any command (command:*)?

No. For that, look at a life cycle hook as you found.

PR #1612, opened 2021-09-24
Deprecate command:* command and event

shadowspawn:
# Pull Request

## Problem

The command events are an original implementation detail which has changed. They have been shown in the README to supply some missing functionality, so we keep them working.

The `.command('*')` and `.on('command:*')` events in particular are no longer needed as there are newer more explicit ways of doing what they were used for. 

This PR prompt by confusion in: #1609

I have not tracked down the commits that removed them, but back in v4 the README had these two code fragments:

```js
// error on unknown commands
program.on('command:*', function () {
  console.error('Invalid command: %s\nSee --help for a list of available commands.', program.args.join(' '));
  process.exit(1);
});
```

```js
program
  .command('*')
  .action(function(env){
    console.log('deploying "%s"', env);
  });
```

## Solution

Remove from README and add to deprecated.

## ChangeLog

- deprecated: `.command('*')`
- deprecated: `on('command:*')`

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -778,13 +778,6 @@ You can execute custom actions by listening to command and option events.
 program.on('option:verbose', function () {
   process.env.VERBOSE = this.opts().verbose;
 });
-
-program.on('command:*', function (operands) {
-  console.error(`error: unknown command '${operands[0]}'`);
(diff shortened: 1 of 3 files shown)


abetomo (approved):
👍

Issue #1584, opened 2021-08-14
Add support for config files


Mingun:
# Problem

Adding support of config files when you have an option with a default value and you want the following priorities, is complicated.

1. Value from the command line
2. Otherwise value from the config file
3. Otherwise the default value

Because commander has no built-in support for config files, you can't use all its power, because options object that you get by calling `opts()` contains mixed default parameters (point 1 from above) and cli parameters (3) and you should manually inspect each option with default to correctly apply the config file (2) parameters. I would like if commander could do this for me.

The current workaround looks like this:
```js
import { Command, Option } from "commander";

/// All options with defaults should be declared there
const DEFAULTS = {
  optionWithDefault: "default value",
};
let options = new Command()
  .option("--config <json>")
  .addOption(
    new Option("--option-with-default <value>")
      .default(DEFAULTS.optionWithDefault)
  )
  .parse()
  .opts();

// Remove all default options from the parsed options
for (const key of Object.keys(DEFAULTS)) {
  if (DEFAULTS[key] === options[key]) {
    delete options[key];
  }
}

const cliOptions = options;
const config = options.config;
delete options.config;

options = Object.assign({}, DEFAULTS);
if (config) {
  options = Object.assign(options, readFileAsJSON(config));
}
```

# Proposal
The proposed API could looks like:
```js
import path from "path";

let options = new Command()
  .option("--option")
  .addOption(
    new Option("--default <value>")
      .default("default")
  )
  .addOption(
    new Option("--config <file>")
      // Optional function should return Object with properties with values
      // from the config file
      // By default uses the specified function that could process both .json
      // and .js (as CommonJS modules, like eslint) config files
      .config(val => require(path.resolve(val)))
      // If config not specified and have a default, this file tried to read
      // using config read function from `.config()`
      .default(".config.js")
  )
```
The examples for the CLI definition below:
## `.config.js` does not exist
- Command: `prog`
  Result:
  ```js
  { default: "default" }
  ```
- Command: `prog --default a`
  Result:
  ```js
  { default: "a" }
  ```
- Command: `prog --option b`
  Result:
  ```js
  { default: "default", option: "b" }
  ```
## `.config.js` contains `option`
```js
module.exports = {
  option: "config",
};
```
- Command: `prog`
  Result:
  ```js
  { default: "default", option: "config" }
  ```
- Command: `prog --default a`
  Result:
  ```js
  { default: "a", option: "config" }
  ```
- Command: `prog --option b`
  Result:
  ```js
[...]

PR #1607, opened 2021-09-19
Add support for merging config into option values, and get+set option value source

shadowspawn:
# Pull Request

Work in progress.

## Problem

People would like to load configuration, from files or elsewhere.

Resolves: #1584

Config file/hash: #151 #328 #665 https://github.com/tj/commander.js/issues/1197#issuecomment-673505653

At a lower-level:
- people ask about how to tell whether options were specified on the command-line. This is not easy when option takes optional value or has default value.
- people ask about managing priority of default/config/env/cli for rolling their own solutions

Related: #1394 #1541 #1595

## Solution

There are many different potential config file formats and file locations and patterns of usage. The goal here is to provide the tools to support author implementing their own desired config support. 

1) low level: expose the "source" of option values. i.e. default/config/env/cli

2) medium level: add method to merge configuration values. The priority order is default < config < env < cli. The merge is quite subtle to support full Commander functionality including option types and custom parsing.

The actual reading of a configuration file is left to author to support JSON/js/es/yaml/ini/TOML or `package.json` property or whatever they wish to offer.

Issue #1595, opened 2021-09-05
Implicit value for command options 


darrenrahnemoon:
Say I have the following command option:

```
program.addOption(new Option('--watch', 'some description...'));
```
(The style in which the option doesn't take an argument itself)

If I don't specify the `--watch` option when running the command, it returns `option: false` whereas a better value IMO would've been `watch: undefined` which is falsy enough but also gives us a hint as to whether or not this option was explicitly false or implicitly false. 

I understand that this philosophy wouldn't work for negatable options since in that case `undefined === implicit true` which would be odd since undefined is falsy. 

Is there a way to tell whether or not an option was explicitly set or implicitly?

---
Bit of context:

I have a command that takes in the option mentioned in the example above, but also can check for the same "flag" in the `process.env`. 

Now if the option flag has been explicitly set, I'd ignore the environment variables, since it's a direct override from the shell. However if the option is not explicitly set I'd like to use the environment variables as backup. 

Problem is: by not explicitly adding watch returns `false` and there's nowhere to find out whether it was explicitly set or implicitly other than maybe traversing through the process args by hand and seeing if the flag is present.

shadowspawn:
> If I don't specify the --watch option when running the command, it returns option: false whereas a better value IMO would've been watch: undefined 

I think you have something else in your code, as that is the the expected behaviour.

```js
program.addOption(new Option('--watch', 'some description...'));
program.parse();
console.log(program.opts());
```

```

darrenrahnemoon:
> This might be what you are waiting for: #1587
Wow either the universe has aligned or that feature was really high in demand 🗡️ 
Closing this issue then. 

Side note: What's the ETA on that entire round getting merged to master?

shadowspawn:
> Side note: What's the ETA on that entire round getting merged to master?

#1587 will go in a feature release and does not require a major release. Probably fairly soon. Like weeks, not months.

Issue #1541, opened 2021-06-01
Check if parameter was specified or is default


jakobrosenberg:
Is there a way to check whether a parameter has been manually set?

I would like to be able to prompt users for input on arguments that they haven't passed.

shadowspawn:
Unless you include a default value in code, options which have not been specified on the command line will be undefined.

(For some special cases you can use listeners to option events, but I wouldn't start there.)

jakobrosenberg:
I ended up putting program inside program, so now I can run program while I run program. 😎

```javascript
const createProgram = (defaults, optionsWithDefaults) => program
    .command('create [a-value]')
    .option('-a, <value>', 'this is a', defaults.a)
    .option('-b, <value>', 'this is b', defaults.b)
    ...
    .option('-z, <value>', 'this is z', defaults.z)
    .option('-d --use-defaults', 'use defaults when available')
    .action((aValue, options, command) => {

        if (!optionsWithDefaults) {
            // 1st iteration - get defaults and rerun createProgram. For 2nd run move
            // defaults to the second parameter so commanderjs doesn't receive them
            createProgram({}, options).parse(process.argv);
        } else {
            // 2nd iteration - commanderjs was created without defaults this time, so options
            // only include user specified values
            myApp.run({
                ...optionsWithDefaults,
                userSpecifiedOptions: options, // I won't prompt for these values
            });
        }
    });

createProgram(defaultsForCreate);
```

It feels a bit hackish, but gets the job done. It lets me prompt users for unspecified arguments when `--use-defaults` isn't used.

I might do `myApp.run(optionsWithDefaults.useDefaults ? optionsWithDefaults : options)` in the future.

shadowspawn:
Some approaches suggested (including from OP), and no further activity in a month.

Feel free to open a new issue if it comes up again, with new information and renewed interest.

cirosantilli:
A simpler approach is to leave ungiven values undefined:

```
.option('-a, <value>', 'this is a')
```

and at the end merge with:

```
options = lodash.merge({}, defaults, commander.program.opts())
```

which skips undefined values when merging.

The only problem is that it doesn't work for boolean options, which are always true or false, and I can't see how to check if they were passed without hacks.

shadowspawn:
> Is there a way to check whether a parameter has been manually set?

There is now. The `.getOptionValueSource()` method was added in Commander v8.3 . The sources of interest are `default`, `env`, and `cli`.

[...]

PR #1613, opened 2021-09-25
Add setOptionValueWithSource and getOptionValueSource

shadowspawn:
# Pull Request

## Problem

- It is difficult for clients to detect whether options with default values have been specified on the command-line.
- It is difficult for clients to manage priorities when merging in option values from different sources.

Resolves: https://github.com/tj/commander.js/issues/1584#issuecomment-917535941
Related: #151 #328 #1394 #1541 #1595 #1607

## Solution

The work to support `.env()` had to solve similar problems, and I had config and cli-detection in mind.

Make `.setOptionValueWithSource()` public. Add matching `.getOptionValueSource()`.

(These routines are quite low-level with long names, but my thinking is they are tools for allowing clients to roll their own custom behaviours and won't be used by the majority of users.)

## ChangeLog

- Added: `.getOptionValueSource()` and `.setOptionValueWithSource()`, where source is one of 'default', 'env', 'config', 'cli'

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -98,7 +98,8 @@ const program = new Command();
 Options are defined with the `.option()` method, also serving as documentation for the options. Each option can have a short flag (single character) and a long name, separated by a comma or space or vertical bar ('|').
 
 The parsed options can be accessed by calling `.opts()` on a `Command` object, and are passed to the action handler.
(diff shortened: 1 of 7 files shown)


abetomo (approved):
LGTM!

PR #1614, opened 2021-09-26
Minor doc improvement for .action()

shadowspawn:
# Pull Request

## Problem

The example code in the JSDoc and TSDoc is to add a command named "help", but that is a poor example since usually can use the built-in help command.

## Solution

Change to adding command named "serve".


abetomo (approved):
👍

PR #1637, opened 2021-11-06
Remove mention of Commander 2

shadowspawn:
# Pull Request

Minor tweak to description of supported node versions.

The mention of Commander v2 seemed useful for Command v3 (which required node 4) and Commander v4 (required node 6, supported from node 8). Does not seem useful now.


abetomo (approved):
👍

PR #1638, opened 2021-11-08
Wrap emit to preserve callback order when async

shadowspawn:
# Pull Request

## Problem

(Noticed while reading code.)

The emit for a subcommand listener was not being chained, so could happen before the action handler if there were async hooks or action.

## Solution

Wrap emit in a function and chain too.

## ChangeLog

- Emit subcommand event in correct order when using async callbacks.


abetomo (approved):
LGTM!

PR #1642, opened 2021-11-19
New Option Constructor using config file

the-unfactoring-guru:
# Pull Request

## Problem
Adding a new form of creating an option object. This is to simplify work when creating multiple options

## Solution
Use option object to specify the new option with its fields. Later on we can implement a function that can create multiple options from one config object array `[ config1, config2, config3] and save time when creating a new program option.

`index.d.ts`
```typescript
export class Option {
//...
    constructor(config: { flags: string; description?: string });
//...
}
```

`option.js`
```javascript
constructor(config) { // Config object that contains flags and description?
    this.flags = config['flags']; //Adding flags
    this.description = config['description'] || ''; //Adding description

    this.required = flags.includes('<'); // A value must be supplied when the option is specified.
    this.optional = flags.includes('['); // A value is optional when the option is specified.
    // variadic test ignores <value,...> et al which might be used to describe custom splitting of single argument
    this.variadic = /\w\.\.\.[>\]]$/.test(flags); // The option can take multiple values.
    this.mandatory = false; // The option must have a value after parsing, which usually means it must be specified on command line.
    const optionFlags = splitOptionFlags(flags);
    this.short = optionFlags.shortFlag;
    this.long = optionFlags.longFlag;
    this.negate = false;
    if (this.long) {
      this.negate = this.long.startsWith('--no-');
    }
    this.defaultValue = undefined;
    this.defaultValueDescription = undefined;
    this.envVar = undefined;
    this.parseArg = undefined;
    this.hidden = false;
    this.argChoices = undefined;
  }
```

### DIFF
--- a/lib/option.js
+++ b/lib/option.js
@@ -34,6 +34,30 @@ class Option {
     this.argChoices = undefined;
   }
 
+  constructor(config) { // Config object that contains flags and description?
+    this.flags = config['flags'];
+    this.description = config['description'] || '';
+
+    this.required = flags.includes('<'); // A value must be supplied when the option is specified.
+    this.optional = flags.includes('['); // A value is optional when the option is specified.
+    // variadic test ignores <value,...> et al which might be used to describe custom splitting of single argument
+    this.variadic = /\w\.\.\.[>\]]$/.test(flags); // The option can take multiple values.
+    this.mandatory = false; // The option must have a value after parsing, which usually means it must be specified on command line.
+    const optionFlags = splitOptionFlags(flags);
+    this.short = optionFlags.shortFlag;
+    this.long = optionFlags.longFlag;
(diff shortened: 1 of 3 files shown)


shadowspawn:
Related: #1359

shadowspawn:
1) This is not valid JavaScript! Maybe you modified the existing constructor when you were playing about, and then added it to the original without testing again.

2) Some code formatter has modified lots of the source code, changing quote marks, reformatting et al.

Already closed by OP.

the-unfactoring-guru:
Okay thank you, I will try to make corrections and see what I did wrong, sorry for the formatting, forgot it was on in my editor.
<!-- You could be a bit more careful with your fucking words... idiot --!>

PR #1643, opened 2021-11-20
Created new OptionConfig class (child of Option) to include config ob…

the-unfactoring-guru:
# Pull Request

## Problem
Adding a new form of creating an option object. This is to simplify work when creating multiple options

## Solution
Create an OptionConfig class that extends Option class, the only difference is that this new class recieves a config object with necessary fields.

Since TS and JS are really peeky with multiple constructors in one class, I chose this option.

`index.d.ts`
```typescript
//...
 export class OptionConfig extends Option{
    constructor(config: {flags: string; description?: string});
}
//...
```

`option.js`
```javascript
class OptionConfig extends Option {
  /**
   * Initialize a new `Option` with the given `config` object
   *
   * @param {{flags: string; description?: string}} config
   */
  constructor(config) {
    super(config.flags, config.description);
  }
}
```

### DIFF
--- a/index.js
+++ b/index.js
@@ -2,7 +2,7 @@ const { Argument } = require('./lib/argument.js');
 const { Command } = require('./lib/command.js');
 const { CommanderError, InvalidArgumentError } = require('./lib/error.js');
 const { Help } = require('./lib/help.js');
-const { Option } = require('./lib/option.js');
+const { Option, OptionConfig } = require('./lib/option.js');
 
 // @ts-check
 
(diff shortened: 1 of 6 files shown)


the-unfactoring-guru:
References and improves on #1642

shadowspawn:
Related: #665 #1359

shadowspawn:
> Adding a new form of creating an option object. This is to simplify work when creating multiple options

I am not sure I understand the problem you have in mind. Can you give an example of how OptionConfig will make it easier to create multiple options?

shadowspawn:
I think a factory function might be a more appropriate approach than a subclass for the implementation.  e.g.
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/from

For a fuller implementation, issues have mentioned short and long, and an alternative approach to setup to avoid the difference between `<required>` and `[optional]`.

As it is, this PR can easily be added in client code.

Issue #1648, opened 2021-12-07
Refactor option value processing to improve maintainability and fix bugs


shadowspawn:
There is some subtle code in `addOption` which is hard to make sense of, and has some bugs. If is solid on options used once on command line, but from code inspection I was able to reproduce some logic errors affecting repeated options for boolean with a non-boolean default value (bit obscure!) and optional options.

The problems have been present for a long time. I was able to reproduce with v2.20.0 and traced code back to at least v1.1.0. The issues have not been generating reports, and the motivation for changing the code is as much to make the code easier to work on as to improve the behaviour.

https://github.com/tj/commander.js/blob/43f4743864e2f670db5eebcf88c92aa4612c54f1/lib/command.js#L515

shadowspawn:
Example bug 1

```
program.option('-f, --flag', 'boolean option with a non-boolean default', 'flag seen')
program.parse(process.argv);
console.log(program.opts());
```

```
% node bug1.js -f
{ flag: 'flag seen' }
defaults % node bug1.js -f -f
{ flag: undefined }
```

shadowspawn:
Example bug 2

```
program.option('-o, --optional [value]')
program.parse(process.argv);
console.log(program.opts());
```

```
defaults % node bug2 -o
{ optional: true }
defaults % node bug2 -o -o foo
{ optional: 'foo' }
defaults % node bug2 -o -o foo -o
{ optional: 'foo' }
```

shadowspawn:
Commander v9 has been released.

Issue #1394, opened 2020-11-10
Default value for an option is always there even if flag isn't used 


Simonwep:
First of all, great library!

Consider this code:
```js
const {program} = require('commander');

program
    .option('-p, --pizza [type]', 'Pizza name', 'pepperoni')
    .parse(process.argv)


console.log({pizza: program.pizza});
```

Usage:
```sh
$ script.js 
# Output: { pizza: 'pepperoni' }

$ script.js --pizza
# Output: { pizza: 'pepperoni' }

$ script.js --pizza salami
# Output: { pizza: 'salami' }
```

Is there a way to see if the `--pizza` flag is actually used? Currently the default-value is always present and used, even if the flag itself is not even present in the arguments. This makes it impossible to pre-process values **and** making the flag optional as at this point the flag is neither optional or mandatory, it is _always_ there, even if not specified.

Related: #440 #1036 

Do I miss something or how can I achieve that using commanderjs?

shadowspawn:
Related: #1135

shadowspawn:
You didn't miss something. If you need to be able to tell if the option was on the command-line, don't specify a default value for the option. You can then handle the value for "missing" and "optional" option in code.

I am not sure what you have in mind with "pre-process values". If you can't see how to achieve what you want, please explain a little more about your desired workflow.

Simonwep:
Thank you for the quick reply! I ended up doing the following as suggested:
```ts
program
    [...]
    .option('-p, --prettify [number|tab]', 'Prettify files (default: 4 spaces)', parseIndentation)
    .option('--duplicates [strict|loose]', 'Find duplicates (default: loose)', parseMode('--duplicates'))
    .option('--diff [strict|loose]', 'Find differences and conflicts (default: strict)', parseMode('--diff'))
    .action((args, cmd) => {

        // Inject fallback values, true will be an option without a value specified
        cmd.prettify = cmd.prettify === true ? 4 : cmd.prettify;
        cmd.duplicates = cmd.duplicates === true ? 'loose' : cmd.duplicates;
        cmd.diff = cmd.diff === true ? 'strict' : cmd.diff;

        ...
    })
    .parse();
```
`parseMode` and `parseIndentation` are rather booring, this solution just seems rather repetitive to me and I also had to manually add a default-value-hint for the user to the option-description.

Now, let's take `--diff` as example, we get the following results:
1. `script`: `cmd.diff` is undefined, not set / used. (**This would be `strict` with `strict` as default value**)
[...]

Issue #1328, opened 2020-08-09
Sets the optional parameters of the option


AloneScar:
Set the default value. The required parameter should be changed to [Type]

shadowspawn:
The example works fine with a required value of `<type>`?

```js
program
  .option('-c, --cheese <type>', 'add the specified type of cheese', 'blue');
```

https://github.com/tj/commander.js/blob/master/examples/options-defaults.js

AloneScar:
But I used the 
```javascript
.option('-c, --cheese <type>', 'add the specified type of cheese', 'blue');
```
return error

When I change `<type>` to `[type]`, it works

AloneScar:
I'm sorry That I didn't fully understand the role of <type>

There is nothing wrong with this program. It may be that my requirements are related to the optional parameters that cause unnecessary errors

I think when I use this option I can pass no arguments and use the default values

Excuse me. Have a nice day

shadowspawn:
This option has a required value: `-c, --cheese <type>`. If the option is used then you must supply the value.

```
example
example -c blue
example -cblue
example --cheese blue
example --cheese=blue
example --cheese   <-- error, value not specified
```

This option has an optional value, it can be used just as a boolean flag or with a value: `-c, --cheese [type]`

```
example
example --cheese
example --cheese blue
```

shadowspawn:
Closing this as resolved.

Feel free to open a new issue if it comes up again, with new information and renewed interest.

Issue #1201, opened 2020-02-26
Optional type and Custom option processing do not work together


prodigitalson:
Opening a new issue exactly the same as #984.

I have a use case for a where I would like a flag with an optional value, however instead of the flag being `true` I want to evaluate it to a default value.

Ideally the following would happpen:

```javascript
program
  .option('-w, --with-cheese [value]', 'Use something special', 'cheddar');

program.parse(process.argv);

if (program.withCheese) {
  console.log(`You ordered a sandwich with ${program.withCheese} cheese`)
} else {
   console.log(`You ordered a sandwich`);
}
```
```
$ sandwich
You ordered a sandwich
$ sandwich --with-cheese
You ordered a sandwich with cheddar cheese
$ sandwhich --with-cheese provolone
You ordered a sandwich with provolone cheese
```

That seems like it could lead to unexpected and ambiguous results though. So the next best thing, and how i immediately thought to handle it, was with custom option processing:

```javascript
const defaultIfTrue= (defaultValue) => {
  return (curr, prev) => curr === true ? defaultValue : curr;
}
program
  .option('-w, --with-cheese [value]', 'Use something special', defaultIfTrue('cheddar'));

program.parse(process.argv);

if (program.withCheese) {
  console.log(`You ordered a sandwich with ${program.withCheese} cheese`)
} else {
   console.log(`You ordered a sandwich`);
}
```

However given the details in #984, this doesn't work either. I'm investigating the event listener, but I can't/don't want to store something outside of the normal options, meaning I need to update the command so that `program.withCheese` AND `program.opts().withCheese` reflect the value appropriately. I have yet to figure that out...

shadowspawn:
I expect you are looking for a more sophisticated solution, but I'll point out there is a simple manual approach to handling the default value in the README:

```
if (program.cheese === undefined) console.log('no cheese');
else if (program.cheese === true) console.log('add cheese');
else console.log(`add cheese type ${program.cheese}`);
```

shadowspawn:
(On a historical note the default value in this case is being treated the same way as for an option with a required value, rightly or wrongly, and has been that way since the beginning: 

- https://github.com/tj/commander.js/blob/d5186ba4b9b64a72cd685fccbb9ec5d0ec0c430d/index.js#L420
- https://github.com/tj/commander.js/blob/2a2165584e5a149ad9949ffc999abab1b19980f9/index.js#L317
)

shadowspawn:
[...]

PR #1652, opened 2021-12-12
Rework option defaults and add preset

shadowspawn:
# Pull Request

## Problem

1) Using "default" value with boolean options and optional options is a bit fragile and buggy.
- non-boolean default used with boolean option is intended to replace "true" when option used, but different behaviour than normal default, never documented in README, and bugs when option repeated
- default used with optional does set the default, but breaks the use of optional without an option argument which is just ignored
- behaviour varies depending on previous value

2) No way to  set value for optional used without argument .

3) Optional used without argument does not trigger custom processing, complicating some uses.

4) This area of code is a problem for adding features to the code with some difficult to understand behaviours and intentions.

Matching issue: #1648

Related: #116 #440 #928 #984 #1135 #1201 #1328 #1394

## Solution

1) Make default consistently set just the default (initial) value for boolean, required, and optional options. Fix use of optional with a default value.

2) Add `Option.preset()` for explicitly setting the preset value to use instead of `true` with a boolean option, or optional without an option argument. 

3) Using preset does trigger custom processing.

## ChangeLog

- fixed: option with optional argument not supplied on command line now works when option already has a value, whether from default value or from previous arguments
- changed: default value specified for boolean option now always used as default value (see `.preset()` to match some previous behaviours)
- changed: default value for boolean option only shown in help if true/false
- added: `Option.preset()` allows specifying value/arg for option when used without option-argument (especially optional, but also boolean option)

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -313,7 +313,7 @@ program
   .addOption(new Option('-t, --timeout <delay>', 'timeout in seconds').default(60, 'one minute'))
   .addOption(new Option('-d, --drink <size>', 'drink size').choices(['small', 'medium', 'large']))
   .addOption(new Option('-p, --port <number>', 'port number').env('PORT'))
-  .addOption(new Option('--donate [amount]').preset('20').argParser(parseFloat));
+  .addOption(new Option('--donate [amount]', 'optional donation in dollars').preset('20').argParser(parseFloat));
 ```
 
 ```bash
--- a/Readme.md
+++ b/Readme.md
@@ -330,8 +330,8 @@ Options:
 $ extra --drink huge
 error: option '-d, --drink <size>' argument 'huge' is invalid. Allowed choices are small, medium, large.
 
-$ PORT=80 extra 
-Options:  { timeout: 60, port: '80' }
+$ PORT=80 extra --donate
+Options:  { timeout: 60, donate: 20, port: '80' }
 ```
 
 ### Custom option processing
--- a/Readme.md
+++ b/Readme.md
@@ -146,7 +146,7 @@ pizza details:
 
 ### Default option value
 
-You can specify a default value for an option which takes a value.
+You can specify a default value for an option.
 
 Example file: [options-defaults.js](./examples/options-defaults.js)
 
@@ -172,7 +172,7 @@ You can define a boolean option long name with a leading `no-` to set the option
 Defined alone this also makes the option true by default.
 
 If you define `--foo` first, adding `--no-foo` does not change the default value from what it would
-otherwise be. You can specify a default boolean value for a boolean option and it can be overridden on command line.
+otherwise be.
 
 Example file: [options-negatable.js](./examples/options-negatable.js)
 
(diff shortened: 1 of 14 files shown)


shadowspawn:
The core of the processing dates all the way back to at least v1.1.0. For interest:

v1.1.0 code setting default value (note, not including boolean):
```
// preassign default value only for --no-*, [optional], or <required>
if (false == option.bool || option.optional || option.required) {
    // when --no-* we make sure default is true
    if (false == option.bool) defaultValue = true;
    // preassign only if we have a default
    if (undefined !== defaultValue) self[name] = defaultValue;
 }
```

v1.1.0 code setting new option value. Note: null and undefined val. Default value being used in multiple places. Choosing behaviour based on previous value. Tricky!

```
    if (null !== val && fn) val = fn(val, undefined === self[name]
      ? defaultValue
      : self[name]);

    // unassigned or bool
    if ('boolean' == typeof self[name] || 'undefined' == typeof self[name]) {
      // if no value, bool true, and we have a default, then use it!
      if (null == val) {
        self[name] = option.bool
          ? defaultValue || true
          : false;
      } else {
        self[name] = val;
      }
    } else if (null !== val) {
      // reassign
      self[name] = val;
    }
    /// else ignoring optional so value unchanged
  });
```

shadowspawn:
This PR changes the boolean behaviour with defaults. Long version of

-----
commander@2

```js
program
  .option('--plain', 'option without a default')
  .option('--default-false', 'option with a boolean default', false)
  .option('--default-string', 'option with a non-boolean default', 'stringy');

program.parse(process.argv);
console.log(program.opts());
```

Notes:
1) "default" is not used as default
2) double option with non-boolean default returns to undefined
```
% node  bug1
{ plain: undefined, defaultFalse: undefined, defaultString: undefined }
% node  bug1 --plain --default-false --default-string  
{ plain: true, defaultFalse: true, defaultString: 'stringy' }
% node bug1 --plain --plain --default-false --default-false --default-string --default-string
{ plain: true, defaultFalse: true, defaultString: undefined }
```

shadowspawn:
I should probably add the preset to the usage in the help?

```
  -t, --timeout <delay>  timeout in seconds (default: one minute)
  -d, --drink <size>     drink cup size (choices: "small", "medium", "large")
  -p, --port <number>    port number (env: PORT)
  --donate [amount]      (preset: 20)
```

shadowspawn:
This PR fixes the behaviour of optional used without an option-argument, so it works when option already has a value whether from default or from previous arguments on command line.

```js
program
  .option('-o, --optional [value]')
  .option('-p [value]', 'description', 'default');

// program.P = 'oops';

program.parse(process.argv);
console.log(program.opts());
```

Commander 8
```
% node bug2.js           
{ p: 'default' }
% node bug2.js -o -p 
{ p: 'default', optional: true } # BUG
% node bug2.js -o cli -p cli
{ p: 'cli', optional: 'cli' }
% node bug2.js -o cli -p cli -o -p
{ p: 'cli', optional: 'cli' } #BUG
```

Behaviour with this PR
```
% node bug2.js                    
{ p: 'default' }
% node bug2.js -o -p              
{ p: true, optional: true }
% node bug2.js -o cli -p cli      
{ p: 'cli', optional: 'cli' }
% node bug2.js -o cli -p cli -o -p
{ p: true, optional: true }
```

abetomo (approved):
Great work!

shadowspawn:

PR #1656, opened 2021-12-21
Combine back into single CHANGELOG to simplify searches

shadowspawn:
# Pull Request

## Problem

My motivation for splitting up the CHANGELOG in #1115 was largely to reduce the size of the published package. However, the CHANGELOG is no longer included in the package if published using npm 8 or npm 9. Hurrah!

I have found two annoyances with split CHANGELOG
- it is more complicated searching the CHANGELOGs to see what version a change shipped in
- splitting out the CHANGELOG when a version drops out of support is a tedious amount of manual work to move the issue and PR references 

## Solution

Put them back together again!


abetomo (approved):
👍

PR #1657, opened 2021-12-21
Make showSuggestionAfterError true by default

shadowspawn:
# Pull Request

In https://github.com/tj/commander.js/pull/1590#issue-981808526 I said:

> Make the suggestions off by default to minimise breakage in Commander 8, and probably switch to on by default in Commander 9.

I am still happy with the suggestions. I think this adds value for most users.

If accepted, this PR turns it on by default.

## ChangeLog

- changed: showSuggestionAfterError is now on by default


abetomo (approved):
LGTM!

shadowspawn:
Thanks @abetomo , very quick!

Issue #1658, opened 2021-12-21
Update package-lock.json format?


shadowspawn:
Node.js v14 includes npm 6.14.15 which writes `package-lock.json ` with ` lockfileVersion@1` format.

Node.js v16 includes npm 8.1.2 which writes `package-lock.json ` with `lockfileVersion@2` format.

Changing versions is noisy. Node.js v14 is still in LTS but now in maintenance mode. 

I am leaning towards switching to `lockfileVersion@2` for Commander v9 so more convenient to develop Commander with Node.js v16 installed?

-----

Installing using old lock file and new npm:

```
my-fork % n lts
   installed : v16.13.1 (with npm 8.1.2)
my-fork % npm install
npm WARN old lockfile 
npm WARN old lockfile The package-lock.json file was created with an old version of npm,
npm WARN old lockfile so supplemental metadata must be fetched from the registry.
npm WARN old lockfile 
npm WARN old lockfile This is a one-time fix-up, please be patient...
npm WARN old lockfile 

changed 1 package, and audited 696 packages in 10s

94 packages are looking for funding
  run `npm fund` for details

7 moderate severity vulnerabilities

To address all issues, run:
  npm audit fix

Run `npm audit` for details.
```

Installing using new lock file and old npm:

```
my-fork % n 14
   installed : v14.18.2 (with npm 6.14.15)
my-fork % npm install
npm WARN read-shrinkwrap This version of npm is compatible with lockfileVersion@1, but package-lock.json was generated for lockfileVersion@2. I'll try to do my best with it!
updated 1 package and audited 696 packages in 3.862s

94 packages are looking for funding
  run `npm fund` for details

found 18 moderate severity vulnerabilities
  run `npm audit fix` to fix them, or `npm audit` for details
```

shadowspawn:
Commander v9 has been released.

PR #1659, opened 2021-12-21
Upgrade package-lock.json to lockfile@2

shadowspawn:
# Pull Request

## Problem

npm 6 and 8 use different formats for `package-lock.json `. Problems switching. Which shall we use?

Related: #1658

## Solution

Go for the version that works nicely with Node.js v16 so convenient for developers (maintainers!) working on new versions.

Not upgrading TypeScript yet, as generates warnings from eslint.
Not upgrading eslint or any of the eslint related dependencies yet, waiting for eslint 8 compatible versions.

## ChangeLog

- changed: Update `package-lock.json` to `lockfile@2` format.
- changed: update dependencies

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9,7 +9,7 @@
       "version": "8.3.0",
       "license": "MIT",
       "devDependencies": {
-        "@types/jest": "^26.0.23",
+        "@types/jest": "^27.0.3",
         "@types/node": "^14.17.3",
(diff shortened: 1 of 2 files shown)


shadowspawn:
I will do a prerelease of Commander v9 soon, but no hurry on this PR. It does not need to be in it.

abetomo (approved):
2️⃣

PR #1661, opened 2021-12-22
Update version and CHANGELOG for 9.0.0-0

shadowspawn:
- bumped  version number
- collected CHANGELOG entries from PR

A little whitespace noise, replacing some non-visible unicode whitespace with plain spaces.


abetomo (approved):
9️⃣

PR #1663, opened 2021-12-22
chore: add Dependabot for NPM and Action updates

nschonni:
# Pull Request

## Problem

Keep the devDependencies in sync for the `develop` branch, along with the GitHub Actions

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,15 @@
+version: 2
+updates:
+  - package-ecosystem: github-actions
+    directory: /
+    target-branch: develop
+    schedule:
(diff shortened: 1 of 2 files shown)


nschonni:
I did this against the `master` branch since I believe it needs to be on the default branch. It might just make sense to change `develop` to the default branch, and drop `master, since you seem to use the release branches for the stable code anyway.
If that happens, I can update this one

shadowspawn:
Previous PR: #1553

We don't aggressively update the dependencies, as have no production dependencies (as you noted). I am concerned about the noise and not seeing significant advantages at this time.

Issue #1639, opened 2021-11-10
Alternative order of arguments supplied to action handler


edmundmaruhn:
I have a rather simple question (and suggestion):
Why is the order of arguments passed in to the action handler function the way it is?

Usually I would expect all fixed parameters at the beginning while all the others follow afterwards.

So currently I have to do something like this in order to pick the options and the command from the arguments:
```javascript
const actionHandler = (...args) => {
	const command = args[args.length - 1]
	const options = args[args.length - 2]
}
```
Or I could do something like:
```javascript
const actionHandler = (...args) => {
	const a = args.reverse()
	const command = args[0]
	const options = args[1]
}
```

Both feel quite odd and in the latter one I would have to re-reverse everything from index 2 in order to get back the correct order of arguments.

So the more natural, thus straightforward alternative would be:
```javascript
const actionHandler = (command, options, ...args) => {
	// ...
}
```

The only reason I can see for having it like it is right now, is that one requires to access the options and command less frequent than the arguments, which makes sense then. However, still in this case there would be a better alternative, by just passing a single object like so:

```javascript
const actionHandler = (actionArguments) => {
	actionArguments.command
	actionArguments.options
	actionArguments.args
}
```

So my suggestion would be to change this in one or the other way. However, I'm aware that this would be a breaking change, but probably there's also a way to make that backward compatible.

What do you think?

Source location:
https://github.com/tj/commander.js/blob/43f4743864e2f670db5eebcf88c92aa4612c54f1/lib/command.js#L476-L492

shadowspawn:
> Why is the order of arguments passed in to the action handler function the way it is?

History! Initially the action handler was only passed the command-arguments. To better support subcommands, the "options" (actually, the command object) were added on the end of the arguments in this commit: #28

> So my suggestion would be to change this in one or the other way. 

The alternative I have been wondering about is passing just the command. It already `.opts()`, `.args`, and `.processedArgs`.

> However, I'm aware that this would be a breaking change, but probably there's also a way to make that backward compatible.

Very breaking! 😢 

Thinking out loud, not deeply considered:
- I like how `.parse(args, { from: 'user')})` works, but adding a parameter after the callback might be a bit clumsy
- could offer another callback method which is passed the different arguments, say `.run()`
- I don't like making the signature a configuration setting

Related: #1277

shadowspawn:
[...]

Issue #1277, opened 2020-06-09
Variable action parameters makes really hard to create a generic action handler.


mshima:
When implementing a program with options and actions that are created at runtime, its really hard to create generic action handlers.

See:
https://github.com/jhipster/generator-jhipster/blob/b24652dc0382ebc3219da0c13add9b65b3d2a1dc/cli/cli.js#L95-L134

Would be really nice to have an option to receive a object like `{args, options, unknown, command}` instead of an array.

Would be nice to add options as an object `{name, description, defaultValue}` instead of `name, description, defaultValue` too.

shadowspawn:
> Would be nice to add options as an object {name, description, defaultValue} instead of name, description, defaultValue too.

Comment heard, but add a separate issue if you would like some visibility and opportunity for community interest on that. Off topic for the title of this issue. 🙂

shadowspawn:
Related past issue: #1055

shadowspawn:
With current behaviour, I had a go at repackaging the action parameters for more generic processing:

```js
const { program } = require('commander');

program
  .arguments('<name>')
  .option('-e,--example')
  .action((...everything) => {
    const last = everything.pop();
    let extraArgs;
    let cmd;
    if (Array.isArray(last)) {
      extraArgs = last;
      cmd = everything.pop();
    } else {
      cmd = last;
    }
    const options = cmd.opts();
    const args = everything;

    console.log('args: ', args);
    console.log('options: ', options);
    console.log('extraArgs: ', extraArgs);
  });

program.parse();
```

```
$ node index.js -e foo  
args:  [ 'foo' ]
options:  { example: true }
extraArgs:  undefined
$ node index.js -e foo bar
args:  [ 'foo' ]
options:  { example: true }
extraArgs:  [ 'bar' ]
```

mshima:
Looks much better but I have to deal with unknown args too.
```
const { program } = require('commander');

program
  .allowUnknownOption(true)
  .arguments('[first]')
  .arguments('[name...]')
  .option('-e,--example')
  .action((...everything) => {
    const last = everything.pop();
    let extraArgs = [];
    let cmd;
    if (Array.isArray(last)) {
      extraArgs = last || [];
      cmd = everything.pop();
    } else {
      cmd = last;
    }

    // move unknown options back to extraArgs
    const splitUnknown = argsToSplit => {
      const args = [];
      const unknown = [];
      argsToSplit.find((item, index) => {
        if (item === undefined) {
          return;
        }
[...]

PR #1664, opened 2021-12-25
Add EventEmitter style use of this for action handler to README

shadowspawn:
# Pull Request

## Problem

1) Some use cases make declaring the various command-arguments for the action handler inconvenient or unnecessary. We have an alternative approach available since we were originally using EventEmitter and event listener. (Which I keep forgetting as being available! Use arrow functions so much.)

2) The code pattern for a simple program without an action handler feels quite different to putting the same code in an action handler. Change from working with `program.opts()` and `program.args` to naming command-arguments and passed options.

Related: #1277 #1639

## Solution

Document using `this` in a action handler.

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -535,6 +535,20 @@ program
   });
 ```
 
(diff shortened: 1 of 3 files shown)


abetomo (approved):
👍

PR #1665, opened 2021-12-26
Refactor README examples into a QuickStart

shadowspawn:
# Pull Request

## Problem

We have a comprehensive README but would like new users to get a quick idea of what Commander does, whether they like the look of it, and how to start writing code.

Related: #1588

## Solution

Write a brief overview of Commander supported by new single-command and multiple command examples. 

The single-command example deliberately leaves out the option descriptions to show a minimal program with just the two main option types declared. The example usage shows Commander being strict, with a suggestion as a bonus!

The second example shows features common to a more complete program without going into details: help descriptions, name, version, subcommand, command-argument, default value for an option. The example usage shows the help output which is an important feature.

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -103,7 +103,7 @@ const program = new Command();
 
 program
   .name('string-util')
-  .description('CLI to some JavaScript string utilities)
+  .description('CLI to some JavaScript string utilities')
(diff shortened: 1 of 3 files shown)


abetomo (approved):
👍

PR #1667, opened 2021-12-30
Allow readonly array as parameter of .choices()

ouuan:
## Problem

Without `readonly`, readonly arrays cannot be passed as parameters.

Example: https://www.typescriptlang.org/play?#code/GYVwdgxgLglg9mABMAjACgIYCcsC5EDOUWMYA5gNoC6AlIgN4C+AUKJLAsgEyY75YBTDABMEAGwCehYqUq0GLZhARFE2LCkQBeRBQDkGPVQDcSlVDU4u23QaNqCiZWCKnW6dShqngPT9-deLC4A3yCQ4yA

## Solution

Add `readonly` in the TypeScript typings.

## ChangeLog

Allow readonly arrays as parameters of `.choices()`.

### DIFF
--- a/lib/argument.js
+++ b/lib/argument.js
@@ -92,7 +92,7 @@ class Argument {
   /**
    * Only allow argument value to be one of choices.
    *
-   * @param {string[]} values
+   * @param {readonly string[]} values
    * @return {Argument}
    */
 
--- a/lib/argument.js
+++ b/lib/argument.js
@@ -99,8 +99,8 @@ class Argument {
   choices(values) {
     this.argChoices = values.slice();
     this.parseArg = (arg, previous) => {
-      if (!values.includes(arg)) {
-        throw new InvalidArgumentError(`Allowed choices are ${values.join(', ')}.`);
+      if (!this.argChoices.includes(arg)) {
+        throw new InvalidArgumentError(`Allowed choices are ${this.argChoices.join(', ')}.`);
       }
       if (this.variadic) {
         return this._concatValue(arg, previous);
--- a/lib/argument.js
+++ b/lib/argument.js
@@ -92,12 +92,12 @@ class Argument {
   /**
    * Only allow argument value to be one of choices.
    *
-   * @param {readonly string[]} values
+   * @param {string[]} values
    * @return {Argument}
    */
 
   choices(values) {
-    this.argChoices = values;
+    this.argChoices = values.slice();
     this.parseArg = (arg, previous) => {
       if (!values.includes(arg)) {
         throw new InvalidArgumentError(`Allowed choices are ${values.join(', ')}.`);
--- a/lib/option.js
+++ b/lib/option.js
@@ -119,8 +119,8 @@ class Option {
   choices(values) {
     this.argChoices = values.slice();
     this.parseArg = (arg, previous) => {
-      if (!values.includes(arg)) {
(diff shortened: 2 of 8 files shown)


shadowspawn:
I have been doing a bit of reading and still thinking about this. Just changing the type does allow passing a ReadonlyArray as parameter, and reflects the intent that the call itself will not change the array, and lets the caller prevent casual changes to their reference to the array. However, the Commander code effectively removes the readonly in the implementation, so it is weak!

The array is stored in `argChoices`:

```
argChoices?: string[];
```

```js
choices(values) {
  this.argChoices = values;
  ...
  }
```

The readonly could be honoured by taking a defensive copy of the parameter:

```js
choices(values) {
  this.argChoices = Array.isArray(values) ? values.slice() : values;
  ...
  }
```

Reference:
- https://www.typescriptlang.org/docs/handbook/2/objects.html#the-readonlyarray-type

Related:
- `aliases()`, but in this case the implementation does take a copy

shadowspawn:
I don't think readonly can be used for parameters in JSDoc?

- https://jsdoc.app/tags-type.html
- https://jsdoc.app/tags-readonly.html

ouuan:
> ```js
> this.argChoices = Array.isArray(values) ? values.slice() : values;
> ```

Could you elaborate on why `Array.isArray(values)` can be false?

shadowspawn:
> Could you elaborate on why Array.isArray(values) can be false?

Hmm, good question. Probably not needed! I was being paranoid, and had in mind perhaps passing `undefined` or `null` to remove the choices, although that is not supported either in the code or in the typings.

shadowspawn:
For interest, what is your use case? (Why do you have a readonly array of choices?)

ouuan:
> For interest, what is your use case? (Why do you have a readonly array of choices?)

I use a library that provides a union type for valid options. The library doesn't provide an array for the valid options, so I created a readonly array in my own code that contains these options and type-checked the array against the union type provided by the library.

lib/argument.js
```diff
@@ -97,7 +97,7 @@ class Argument {
    */
 
   choices(values) {
-    this.argChoices = values;
+    this.argChoices = values.slice();
     this.parseArg = (arg, previous) => {
       if (!values.includes(arg)) {
```
shadowspawn, line 102: The uses of `values` in the parse function should be replaced with `this.argChoices` now they may be different.

lib/option.js
```diff
@@ -117,7 +117,7 @@ class Option {
    */
 
   choices(values) {
-    this.argChoices = values;
+    this.argChoices = values.slice();
     this.parseArg = (arg, previous) => {
       if (!values.includes(arg)) {
```
shadowspawn, line 122: The uses of `values` in the parse function should be replaced with `this.argChoices` now they may be different.

shadowspawn:
I offered some tests in a PR back to the branch to check the parameter is actually treated as readonly at runtime.

shadowspawn (approved):
I don't think this will be widely needed and requires taking a defensive copy of param, but do I like making the TypeScript API more usable (readonly) and safer (defensive copy).

Thumbs up from me. 👍

abetomo (approved):
👍

shadowspawn:

Sammiegirl411 (commented):
Let's do this

PR #1668, opened 2022-01-02
Reconfigure standard and eslint

shadowspawn:
# Pull Request

## Problem

We were installing `standard`, but then running `eslint` and trying to access its dependency configuration files rather than depending on their packages explicitly. This is fragile.

`standard` cli not usable because of custom eslint rules.

The TypeScript config was not using `standard` base configuration due to previous annoyance with the multiple dependencies.

## Solution

Install `standard` ESLint configurations and not the whole package. Install necessary peer dependencies, so works with npm below npm 8 which require explicitly installing peer dependencies. Basically same dependencies for `eslint-config-standard` and `eslint-config-standard-with-typescript` so go back to using `standard` for TypeScript too.

Used `eslint --fix` to fix most of the problems picked up. Nice, don't need `standard` CLI to get `--fix`!

The downside is we still have to juggle dependencies when some `standard` dependencies lag behind versions of ESLint or TypeScript, like currently still waiting for one last package (`eslint-plugin-node`) to get ESLint 8 support.

Considered another approach but rejected: switch to using `semistandard` and `ts-standard` instead of `eslint`. Reasons:
- we prefer semicolons, and while there is `semistandard` there is no `ts-semistandard`
- there is VSCode extension for `standard`, but not sure if can work with both `semistandard` and `ts-standard`
- we have some custom rules, and the point of `standard` is no configuration
- (I like ESLint, and ESLint plugin to VSCode, and being able to add custom rules!)

## ChangeLog

- refactor `standard` related `devDependencies` and usage

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -7,7 +7,8 @@ const javascriptSettings = {
   rules: {
     'no-else-return': ['error', { allowElseIf: false }],
     'space-before-function-paren': ['error', 'never'],
-    semi: ['error', 'always']
+    semi: ['error', 'always'],
+    'no-extra-semi': 'error'
   }
 };
 
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -6,8 +6,6 @@ const javascriptSettings = {
   ],
   rules: {
     'no-else-return': ['error', { allowElseIf: false }],
-    'no-var': 'warn',
-    'one-var': 'off',
     'space-before-function-paren': ['error', 'never'],
     semi: ['error', 'always']
   }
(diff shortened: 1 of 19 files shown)


shadowspawn:
Reminder to self on possible dependency update approach with less trial and error.

standard related: check `peerDependencies` of [eslint-config-standard](https://github.com/standard/eslint-config-standard/blob/master/package.json) and [eslint-config-standard-with-typescript](https://github.com/standard/eslint-config-standard-with-typescript/blob/master/package.json) for supported ranges. e.g. as of 2022-01-03 requires ESLint 7.

```
  "peerDependencies": {
    "eslint": "^7.12.1",
    "eslint-plugin-import": "^2.22.1",
    "eslint-plugin-node": "^11.1.0",
    "eslint-plugin-promise": "^4.2.1 || ^5.0.0"
  },
```

abetomo (approved):
great job!

PR #1669, opened 2022-01-02
Allow readonly array as parameter to .aliases(), .parse(), .parseAsync()

shadowspawn:
# Pull Request

## Problem

`readonly` array cannot be passed directly as parameter to `.aliases()`, `.parse()`, or `.parseAsync()`.

Work-around: make a modifiable copy, like by calling `.slice()`.

Following pattern used in #1667 for `.choices()`.

## Solution

No implementation changes needed, just declaration and tests.

- add readonly to TypeScript declaration file
- add readonly test to TypeScript tests
- add readonly tests of JavaScript implementation

## ChangeLog

- changed: allow passing readonly string array to `.parse()`
- changed: allow passing readonly string array to `.parseAsync()`
- changed: allow passing readonly string array to `.aliases()`

### DIFF
--- a/tests/command.alias.test.js
+++ b/tests/command.alias.test.js
@@ -96,3 +96,29 @@ test('when set alias on executable then can get alias', () => {
     .alias(alias);
   expect(program.commands[0].alias()).toEqual(alias);
 });
+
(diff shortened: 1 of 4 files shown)


abetomo (approved):
👍

Issue #1631, opened 2021-10-24
Adding options after commands have been created


mikemilano:
Is there a way to add options to commands after they have been created?

In this app, I'm separating commands into their own files and dynamically requiring them. I'd like to go back through and add global options to each of the commands that exists. Is this possible?

I've tried simply appending the `options` array of each command with a `new Option` but the option doesn't show when I run help on the command.

mikemilano:
Thanks a lot John, that did it!

Issue #1551, opened 2021-06-27
Poll: Enhancements to global options


shadowspawn:
Commander supports global options by declaring them on the program. This is flexible, but Commander does not offer any additional support to make them easier to use as a programmer or find them as a user. Example program with current behaviour:

```js
program
  .option('-g, --global');

program.command('sub')
  .option('-l, --local')
  .action((options, sub) => {
    console.log('subcommand options: %o', options);
    console.log('global options: %o', program.opts());
  });

program.parse();
```

```sh
$ node example.js --global sub --local
subcommand options: { local: true }
global options: { global: true }
```

```sh
$  node example.js help sub
Usage: example sub [options]

Options:
  -l, --local
  -h, --help   display help for command
```

Related: #243 #476 #905 https://github.com/tj/commander.js/pull/1024#issuecomment-568403615 #1078 #1426 #1631

What additional support would you like? (I am wondering what is most popular and useful, not planning to add everything!)

shadowspawn:
1) An explicit way to access combined global and local options from action handler, without needing to use program variable.

This could be a parameter to `.opts()`, or a separate routine.

```js
console.log('combined options: %o', sub.opts({ includeGlobals: true }));
```

shadowspawn:
2) Configuration so `.opts()` and the options passed to the action handler always include combined global and local options.

```js
program.includeGlobalsInOptions();
```

shadowspawn:
3) A way to show the global options in the help for a subcommand.

```js
program.includeGlobalOptionsInHelp();
```

```sh
$  node example.js help sub
Usage: example sub [options]

Options:
  -l, --local
  -h, --help   display help for command

Global Options:
  -g, --global
```

shadowspawn:
4) A way to add a shared option that is actually added to (all) the subcommands, rather than the program.

```js
program.sharedOption('-l, --local');
```

binyamin:
@shadowspawn what is the time span for this poll?

shadowspawn:
Counts at 2021-12-21.

1) An explicit way to access combined global and local options from action handler, without needing to use program variable.
    8 👍 

2) Configuration so .opts() and the options passed to the action handler always include combined global and local options.
    3 👍 

3) A way to show the global options in the help for a subcommand.
    4 👍 5 🚀 (6 people total)

4) A way to declare a common option that is actually added directly to (all) the subcommands, rather than to the program. 
    10 👍 

Thanks for the feedback.

[...]

PR #1670, opened 2022-01-03
addCommonOption for options common to subcommands

shadowspawn:
# Pull Request

## Problem

Sometimes people would like to be able to add common (global) options to all the subcommands, rather than to the program. This was the most popular enhancement in poll https://github.com/tj/commander.js/issues/1551#issuecomment-869093337

Related:
- #243 _so they're all already populated with the common options_
- #476  _add global options to a program that is applied to all commands_
- https://github.com/tj/commander.js/issues/1229#issuecomment-623613637  _a set of common options for all my subcommands_
- #1426 _an option that I add to all the commands of my application_
- #1631 _add global options to each of the commands_

## Solution

- add `.addCommonOption()`
- implemented using `.copyCommonOptions()`
- following same pattern as `.copyInheritedSettings()`, and common settings are automatic for `.command()` but manual for `.addCommand()`

This is implemented for one level deep. The common options are added to direct subcommands and not sub sub commands, as don't have a good way of identifying leaf vs non-leaf commands at the time they are added.

Why not `.commonOption()` matching `.option()`? Because would need more variations to cover `.requiredOption()` and `.addOption()`.

## Example Usage

```js

program
  .addCommonOption(new Option('-d, --debug'))
  .addCommonOption(new Option('--quiet'));

program
  .command('sub')
  .option('-l, --local')
  .action(() => {
    console.log('Thanks for calling sub');
  });

program.parse();
```

```
$ node index.js help sub
Usage: index sub [options]

Options:
  -d, --debug
  --quiet
  -l, --local
```

### DIFF
--- a/lib/command.js
+++ b/lib/command.js
@@ -24,6 +24,8 @@ class Command extends EventEmitter {
     this.commands = [];
     /** @type {Option[]} */
     this.options = [];
+    /** @type {Option[]} */
+    this.commonOptions = []; // for subcommands
     this.parent = null;
     this._allowUnknownOption = false;
     this._allowExcessArguments = true;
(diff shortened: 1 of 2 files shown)


shadowspawn:
There is not a load of code in this PR which makes me realise this could be done manually fairly simply. Some of the linked issues use subclasses, here is a direct approach:

```
// Add common options after setting up program and subcommands
program.commands.forEach((cmd) => {
   cmd.option('-d, --debug');
   cmd.option('--quiet');
});
```

shadowspawn:
This implementation was more limited and less useful than I expected. Lot of scenarios. Considerations are:
- `.option()` vs `.requiredOption()` vs `.addOption()`
- listeners for option events
- separately created commands (added using `.addCommand()`) should get some control
- besides `.command()`, do common options get added for `.addCommand()` or `.createCommand()` ?

Current plan is write some examples on how to do it yourself with a simple loop, and maybe a simple subclass.

PR #1671, opened 2022-01-04
Add support for getting merged options including globals

shadowspawn:
# Pull Request

## Problem

Currently no direct support for getting merged options including globals. The work-around is to access the global options on the program or command parent. 

This was the second most popular enhancement in poll https://github.com/tj/commander.js/issues/1551#issuecomment-869093146

Related issues: #243 https://github.com/tj/commander.js/issues/476#issuecomment-941042505 #1024 #1229

Previous consideration by me: https://github.com/tj/commander.js/issues/1229#issuecomment-743944952 #1155 #1478

## Solution

<del>
Add mergeOptions parameter to `.opts()`.

```js
const options = cmd.opts();
const mergedOptions = cmd.opts({ includeGlobals: true });
```
</del>

Add `command.optsWithGlobals()`.

## Example Usage

```js

program
  .option('-g, --global');

program
  .command('sub')
  .option('-l, --local')
  .action((options, cmd) => {
    console.log({
      opts: cmd.opts(),
      optsWithGlobals: cmd.optsWithGlobals())
    });
  });

program.parse();
```

```
$ node index.js sub --global --local
{
  opts: { local: true },
  optsWithGlobals: { local: true, global: true }
}
```

## ChangeLog

- add `.optsWithGlobals()` to return merged local and global options

### DIFF
--- a/lib/command.js
+++ b/lib/command.js
@@ -1445,7 +1445,7 @@ Expecting one of '${allowedValues.join("', '")}'`);
    *
    * @return {Object}
    */
-  opts(mergeOptions) {
+  opts() {
     if (this._storeOptionsAsProperties) {
       // Preserve original behaviour so backwards compatible when still using properties
       const result = {};
@@ -1466,7 +1466,7 @@ Expecting one of '${allowedValues.join("', '")}'`);
    *
    * @return {Object}
    */
-  optsWithGlobals(mergeOptions) {
+  optsWithGlobals() {
     // globals overwrite locals
     return getCommandAndParents(this).reduce(
       (combinedOptions, cmd) => Object.assign(combinedOptions, cmd.opts()),
--- a/lib/command.js
+++ b/lib/command.js
@@ -1441,11 +1441,25 @@ Expecting one of '${allowedValues.join("', '")}'`);
   }
 
   /**
-   * Return an object containing options as key-value pairs
+   * Return an object containing options as key-value pairs.
    *
+   * @example
+   * const options = cmd.opts();
+   * const mergedOptions = cmd.opts({ includeGlobals: true });
+   *
+   * @param {Object} [mergeOptions]
+   * @param {boolean} mergeOptions.includeGlobals
    * @return {Object}
    */
-  opts() {
+  opts(mergeOptions) {
+    if (mergeOptions && mergeOptions.includeGlobals) {
+      // globals overwrite locals
+      return getCommandAndParents(this).reduce(
+        (combinedOptions, cmd) => Object.assign(combinedOptions, cmd.opts()),
+        {}
+      );
+    }
+
     if (this._storeOptionsAsProperties) {
       // Preserve original behaviour so backwards compatible when still using properties
       const result = {};
--- a/lib/command.js
+++ b/lib/command.js
@@ -1441,7 +1441,7 @@ Expecting one of '${allowedValues.join("', '")}'`);
   }
 
   /**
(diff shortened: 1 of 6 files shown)


kuncevic:
Oh BTW, why wouldn't it `{ includeGlobals: true }` become a default ?

kuncevic:
Oh BTW, why wouldn't `{ includeGlobals: true }` becomes a default ?

kuncevic:
Oh, BTW, why wouldn't `{ includeGlobals: true }` became a default behavior?

shadowspawn:
Option 2 in the poll was a configuration to always include the globals in the options, but it didn't get a lot of likes.

kuncevic:
Yeah i c, in that case you had to call that `[program.includeGlobalsInOptions();](https://github.com/tj/commander.js/issues/1551#issuecomment-869093223)` however you may want to pass more params to `.opts()` right?

shadowspawn:
Yes. Commander has some routines where the boolean arg is optional because the method name is clear about what will happen:

```
program
  .allowUnknownOption()
  .storeOptionsAsProperties();
```

I don't like adding a plain boolean parameter which has no obvious meaning, like say:

```
program.opts(true);
```

Adding an options parameter makes effect clearer by naming the value, especially when not commonly used, and also allows future expansion.

```
program.parse(['--debug'], { from: 'user' });
program.help({ error: true });
program.command('foo', 'foo description', { isDefault: true });
```

kuncevic:
Yeah, plain boolean is definitely no go. However  speaking`program.includeGlobalsInOptions();` vs `.opts({ includeGlobals: true })` as you already have things like `.allowUnknownOption()` having `.includeGlobalsInOptions();` seems quite natural.

shadowspawn:
Yes I agree `.includeGlobalsInOptions()` flows better than `.opts({ includeGlobals: true })`.

Using a separate method would look like:

```
//const mergedOptions = cmd.cmd.opts({ includeGlobals: true })
const mergedOptions = cmd.optsWithGlobals();
```

Do you like that better? 

(Or are you now thinking option 2 is better!?)

kuncevic:
Yeah, `cmd.optsWithGlobals();` looks good, it seems more natural and really flows better along with `.allowUnknownOption()`, etc.

shadowspawn:
Switched to `.optsWithGlobals` in PR description and code.

kuncevic (commented):
That is great. Looking forward to see it pushed in the next release 🔥

lib/command.js
```diff
@@ -1441,11 +1441,11 @@ Expecting one of '${allowedValues.join("', '")}'`);
   }
 
   /**
-   * Return an object containing options as key-value pairs
+   * Return an object containing local option values as key-value pairs.
    *
    * @return {Object}
    */
-  opts() {
+  opts(mergeOptions) {
```
abetomo, line 1448: Is the `mergeOptions` argument unnecessary?

shadowspawn: Oops, I think that is left over from first try. Thanks.

shadowspawn: Removed. And the copy-and-paste copy of it too. 😊

abetomo (approved):
LGTM!

shadowspawn:
Thanks @kuncevic and @abetomo

shadowspawn:
`.optsWithGlobals()` is included in prerelease Commander v9.0.0-1.

shadowspawn:

PR #1673, opened 2022-01-04
Fix JSDoc comment for copyInheritedSettings

shadowspawn:
Fix bogus comment, probably left over from copy-paste when adding method.


abetomo (approved):
+1

PR #1675, opened 2022-01-08
Add error() for displaying errors from client code

shadowspawn:
# Pull Request

## Problem

People adding extra error checks do not have a way to take advantage of the error configuration built into Commander such as: `.configureOutput()`, `.exitOverride()`, `.showHelpAfterError()`.

There have not been many requests for this, but we do now have quite rich error handling which clients can use for their own errors too.

Related: #1358 #1632

## Solution

Provide a low level routine that is also used by Commander itself. Refactor private `._displayError()` into public `.error()`. Change the signature to make the error id `code` and `exitCode` optional so can be called with just a message.

```
if (cmd.opts().good && cmd.opts().bad)
  cmd.error('error: specify just one of "--good" and "--bad"');
```

## ChangeLog

- add `error()` for generating errors from client code just like Commander generated errors, with support for  `.configureOutput()`, `.exitOverride()`, and `.showHelpAfterError()`

PR #1676, opened 2022-01-14
Fix errata in example for error

shadowspawn:
Oops, all this talk of error and exit, used the wrong one in the example (which I did not run...).

https://github.com/tj/commander.js/pull/1675#discussion_r785171583


abetomo (approved):
👍

Issue #1358, opened 2020-09-18
Implied Options


brannonh:
I just started using Commander.js yesterday, and I am very happy with it. That being said, I noticed there is no way to mark that a boolean option implies another boolean option. My thought process is that if such a boolean option is provided to the program, the _implied_ option(s) would _also_ be set to `true`.

Having this available would allow for less verification code within the program itself. You could always check a single option, rather than checking multiple options.

As an example, here's a(n edited) snippet of my current code:

```javascript
program
  .option('-d, --dump', 'dump data on error and end of execution')
  .option('-t, --trace', 'log everything (implies -d)');
```

Currently, I would either have to check for both options or check for `-t` and set `-d`. With an implied option, I could do something like this (rough example):

```javascript
program
  .option('-d, --dump', 'dump data on error and end of execution')
  .option('-t, --trace', 'log everything')
  .implies('trace', 'dump');
```

The automated help would also automatically add something like `(implies -d)` to the description for the `trace` option.

I would think that a negatable boolean option would take precedence over the implication (i.e., `program -t --no-dump` would result in `trace` being `true` and `dump` being `false`).

Anyways, that's all. I think this would be a nice addition to Commander.js' current functionality. If it's possible now and I just missed it, please let me know. Otherwise, I'm also always open to feedback. Thanks!

shadowspawn:
Thanks. Another other sort of configuration sometimes offered is mutually exclusive options. 

I see these as fairly low value as only a few lines of code to do the exact custom logic required. Happy to leave issue open though, and see if it is popular. 👍

brannonh:
Thanks for the feedback. I understand that completely. I guess the value would be higher in cases where the custom logic is required in multiple places. In that case adding/removing a flag implication would require updating the custom logic in all of those places, which always opens the door for missing one.

shadowspawn:
Related: #1391 #1408

shadowspawn:
For potential interest to people following this issue, there is a PR open for similar-but-different `conflicts`. See #1678

shadowspawn:
I am experimenting with calling signature `Option.implies(optionBag)` that specifies the implied options and their values, like:

```js
program
  .addOption(new Option('--quiet').implies({ logLevel: 'off' }))
  .addOption(new Option('--log-level <level>').choices(['info', 'warning', 'error', 'off']).default('info'));
```

shadowspawn:
Commander v9.3.0 adds `Options.implies()`

Issue #1408, opened 2020-12-02
Is there a way to group options? (one of)


Niryo:
is there any way to group options together like in argparse? 
let's say I have two options: `-a` and `-b`, is there a way to group them under "one of" so that the user will get an error if he passes them both?

shadowspawn:
No grouping support, no.

The related open issue is #1358

shadowspawn:
I looked up the argparse support:
- https://docs.python.org/3/library/argparse.html#mutual-exclusion

shadowspawn:
An answer was provided, and no further activity in a month. Closing this as resolved.

Feel free to open a new issue if it comes up again, with new information and renewed interest.

shadowspawn:
For interest to people following this issue, there is a PR open for `'conflicts()`. See https://github.com/tj/commander.js/pull/1678

PR #1678, opened 2022-01-15
feat(option): allow to set options as conflicting

erezrokah:
# Pull Request

Hello 👋 Thank you for this lovely package, this PR implements exclusive options (see details below).
Happy to rename it to `conflicts` or anything else you find suiting, and of course open to any kind of feedback (especially if you think this should be user code and not package code).
I searched previous issues and PRs and could only find this related comment: https://github.com/tj/commander.js/issues/1358#issuecomment-695149256 

## Problem

I would like to configure options that are mutually exclusive. For example `--json` (to set the output type), conflicts with `--silent` (to suppress output).

Related to https://github.com/tj/commander.js/issues/1358#issuecomment-695149256

## Solution

Added a new option method `conflicts` that accepts an array of strings (or a single string) with options that conflicts with the configured option.

- [x] TypeScript typings
- [x] JSDoc documentation in code
- [x] tests
- [x] README
- [x] examples/

## ChangeLog

feat(option): allow to configure conflicting options

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -398,7 +398,7 @@ program
   .addOption(new Option('-d, --drink <size>', 'drink size').choices(['small', 'medium', 'large']))
   .addOption(new Option('-p, --port <number>', 'port number').env('PORT'))
   .addOption(new Option('--donate [amount]', 'optional donation in dollars').preset('20').argParser(parseFloat))
-  .addOption(new Option('-d, --disable-server', 'disables the server').exclusive(['port']));
+  .addOption(new Option('-d, --disable-server', 'disables the server').conflicts(['port']));
 ```
 
 ```bash
--- a/Readme.md
+++ b/Readme.md
@@ -410,6 +410,7 @@ Options:
   -d, --drink <size>     drink cup size (choices: "small", "medium", "large")
   -p, --port <number>    port number (env: PORT)
   --donate [amount]      optional donation in dollars (preset: 20)
+  -d, --disable-server   disables the server
   -h, --help             display help for command
 
 $ extra --drink huge
--- a/Readme.md
+++ b/Readme.md
@@ -398,7 +398,7 @@ program
   .addOption(new Option('-d, --drink <size>', 'drink size').choices(['small', 'medium', 'large']))
   .addOption(new Option('-p, --port <number>', 'port number').env('PORT'))
   .addOption(new Option('--donate [amount]', 'optional donation in dollars').preset('20').argParser(parseFloat))
-  .addOption(new Option('-d, --disable-server', 'disables the server').conflicts(['port']));
+  .addOption(new Option('--disable-server', 'disables the server').conflicts(['port']));
 ```
 
 ```bash
@@ -410,7 +410,7 @@ Options:
   -d, --drink <size>     drink cup size (choices: "small", "medium", "large")
   -p, --port <number>    port number (env: PORT)
   --donate [amount]      optional donation in dollars (preset: 20)
-  -d, --disable-server   disables the server
+  --disable-server       disables the server
   -h, --help             display help for command
 
 $ extra --drink huge
@@ -418,6 +418,9 @@ error: option '-d, --drink <size>' argument 'huge' is invalid. Allowed choices a
 
 $ PORT=80 extra --donate
 Options:  { timeout: 60, donate: 20, port: '80' }
+
+$ extra --disable-server --port 8000
+error: option '--disable-server' cannot be used with option '-p, --port <number>'
 ```
 
 ### Custom option processing
--- a/Readme.md
+++ b/Readme.md
@@ -397,7 +397,8 @@ program
   .addOption(new Option('-t, --timeout <delay>', 'timeout in seconds').default(60, 'one minute'))
   .addOption(new Option('-d, --drink <size>', 'drink size').choices(['small', 'medium', 'large']))
   .addOption(new Option('-p, --port <number>', 'port number').env('PORT'))
-  .addOption(new Option('--donate [amount]', 'optional donation in dollars').preset('20').argParser(parseFloat));
+  .addOption(new Option('--donate [amount]', 'optional donation in dollars').preset('20').argParser(parseFloat))
+  .addOption(new Option('-d, --disable-server', 'disables the server').exclusive(['port']));
 ```
 
 ```bash
--- a/Readme.md
+++ b/Readme.md
@@ -398,7 +398,7 @@ program
   .addOption(new Option('-d, --drink <size>', 'drink size').choices(['small', 'medium', 'large']))
   .addOption(new Option('-p, --port <number>', 'port number').env('PORT'))
   .addOption(new Option('--donate [amount]', 'optional donation in dollars').preset('20').argParser(parseFloat))
-  .addOption(new Option('--disable-server', 'disables the server').conflicts(['port']));
+  .addOption(new Option('--disable-server', 'disables the server').conflicts(['port'])); // or `.conflicts('port')` for a single conflict
 ```
 
 ```bash
--- a/lib/command.js
+++ b/lib/command.js
@@ -1315,16 +1315,22 @@ Expecting one of '${allowedValues.join("', '")}'`);
    * @api private
    */
   _checkForExclusiveOptions() {
-    const definedOptions = this.options.filter(
-      (option) => this.getOptionValue(option.attributeName()) !== undefined
+    const definedNonDefaultOptions = this.options.filter(
+      (option) => {
+        const optionKey = option.attributeName();
+        if (this.getOptionValue(optionKey) === undefined) {
+          return false;
+        }
+        return this.getOptionValueSource(optionKey) !== 'default';
+      }
     );
 
-    const optionsWithExclusiveNames = definedOptions.filter(
+    const optionsWithExclusiveNames = definedNonDefaultOptions.filter(
       (option) => option.exclusiveNames.length > 0
     );
 
     optionsWithExclusiveNames.forEach((option) => {
-      const exclusiveAndDefined = definedOptions.filter((defined) =>
+      const exclusiveAndDefined = definedNonDefaultOptions.filter((defined) =>
         option.exclusiveNames.includes(defined.attributeName())
       );
       if (exclusiveAndDefined.length > 0) {
@@ -1594,9 +1600,6 @@ Expecting one of '${allowedValues.join("', '")}'`);
     const getErrorMessage = (option) => {
       const optionKey = option.attributeName();
       const source = this.getOptionValueSource(optionKey);
-      if (source === 'default') {
-        return `option '${optionKey}' with default value`;
-      }
       if (source === 'env') {
         return `environment variable '${option.envVar}'`;
       }
--- a/lib/command.js
+++ b/lib/command.js
@@ -1216,7 +1216,7 @@ Expecting one of '${allowedValues.join("', '")}'`);
 
     outputHelpIfRequested(this, parsed.unknown);
     this._checkForMissingMandatoryOptions();
-    this._checkForExclusiveOptions();
+    this._checkForConflictingOptions();
 
     // We do not always call this check to avoid masking a "better" error, like unknown command.
     const checkForUnknownOptions = () => {
@@ -1310,11 +1310,11 @@ Expecting one of '${allowedValues.join("', '")}'`);
   }
 
   /**
-   * Display an error message if exclusive options are used together.
+   * Display an error message if conflicting options are used together.
    *
    * @api private
    */
-  _checkForExclusiveOptions() {
+  _checkForConflictingOptions() {
     const definedNonDefaultOptions = this.options.filter(
       (option) => {
         const optionKey = option.attributeName();
@@ -1325,16 +1325,16 @@ Expecting one of '${allowedValues.join("', '")}'`);
       }
     );
 
-    const optionsWithExclusiveNames = definedNonDefaultOptions.filter(
-      (option) => option.exclusiveNames.length > 0
+    const optionsWithConflicting = definedNonDefaultOptions.filter(
+      (option) => option.conflictsWith.length > 0
     );
 
-    optionsWithExclusiveNames.forEach((option) => {
-      const exclusiveAndDefined = definedNonDefaultOptions.find((defined) =>
-        option.exclusiveNames.includes(defined.attributeName())
+    optionsWithConflicting.forEach((option) => {
+      const conflictingAndDefined = definedNonDefaultOptions.find((defined) =>
+        option.conflictsWith.includes(defined.attributeName())
       );
-      if (exclusiveAndDefined) {
-        this._exclusiveOptionPresent(option, exclusiveAndDefined);
+      if (conflictingAndDefined) {
+        this._conflictingOptionPresent(option, conflictingAndDefined);
       }
     });
   }
@@ -1594,9 +1594,10 @@ Expecting one of '${allowedValues.join("', '")}'`);
    * `Option` conflicts with another option.
    *
    * @param {Option} option
+   * @param {Option} conflictingOption
    * @api private
    */
-  _exclusiveOptionPresent(option, exclusive) {
+  _conflictingOptionPresent(option, conflictingOption) {
     const getErrorMessage = (option) => {
       const optionKey = option.attributeName();
       const source = this.getOptionValueSource(optionKey);
@@ -1606,8 +1607,8 @@ Expecting one of '${allowedValues.join("', '")}'`);
       return `option '${option.flags}'`;
     };
 
-    const message = `error: ${getErrorMessage(option)} cannot be used with ${getErrorMessage(exclusive)}`;
-    this.error(message, { code: 'commander.exclusiveOptionPresent' });
+    const message = `error: ${getErrorMessage(option)} cannot be used with ${getErrorMessage(conflictingOption)}`;
+    this.error(message, { code: 'commander.conflictingOptionPresent' });
   }
 
   /**
--- a/lib/command.js
+++ b/lib/command.js
@@ -1331,7 +1331,7 @@ Expecting one of '${allowedValues.join("', '")}'`);
         this.exclusiveOptionPresent(option, exclusiveAndDefined[0]);
       }
     });
-  };
+  }
 
   /**
    * Parse options from `argv` removing known options,
@@ -1591,9 +1591,21 @@ Expecting one of '${allowedValues.join("', '")}'`);
    * @api private
    */
   exclusiveOptionPresent(option, exclusive) {
-    const message = `error: option '${option.flags}' cannot be used with '${exclusive.flags}'`;
-    this.error(message, { code:'commander.exclusiveOptionPresent' });
-  };
+    const getErrorMessage = (option) => {
+      const optionKey = option.attributeName();
+      const source = this.getOptionValueSource(optionKey);
+      if (source === 'default') {
+        return `option '${optionKey}' with default value`;
+      }
+      if (source === 'env') {
+        return `environment variable '${option.envVar}'`;
+      }
+      return `option '${option.flags}'`;
+    };
+
+    const message = `error: ${getErrorMessage(option)} cannot be used with ${getErrorMessage(exclusive)}`;
+    this.error(message, { code: 'commander.exclusiveOptionPresent' });
+  }
 
   /**
    * Unknown option `flag`.
--- a/lib/command.js
+++ b/lib/command.js
@@ -1334,7 +1334,7 @@ Expecting one of '${allowedValues.join("', '")}'`);
         option.conflictsWith.includes(defined.attributeName())
       );
       if (conflictingAndDefined) {
-        this._conflictingOptionPresent(option, conflictingAndDefined);
+        this._conflictingOption(option, conflictingAndDefined);
       }
     });
   }
@@ -1597,7 +1597,7 @@ Expecting one of '${allowedValues.join("', '")}'`);
    * @param {Option} conflictingOption
    * @api private
    */
-  _conflictingOptionPresent(option, conflictingOption) {
+  _conflictingOption(option, conflictingOption) {
     const getErrorMessage = (option) => {
       const optionKey = option.attributeName();
       const source = this.getOptionValueSource(optionKey);
@@ -1608,7 +1608,7 @@ Expecting one of '${allowedValues.join("', '")}'`);
     };
 
     const message = `error: ${getErrorMessage(option)} cannot be used with ${getErrorMessage(conflictingOption)}`;
-    this.error(message, { code: 'commander.conflictingOptionPresent' });
+    this.error(message, { code: 'commander.conflictingOption' });
   }
 
   /**
--- a/lib/command.js
+++ b/lib/command.js
@@ -1216,6 +1216,7 @@ Expecting one of '${allowedValues.join("', '")}'`);
 
     outputHelpIfRequested(this, parsed.unknown);
     this._checkForMissingMandatoryOptions();
+    this._checkForExclusiveOptions();
 
     // We do not always call this check to avoid masking a "better" error, like unknown command.
     const checkForUnknownOptions = () => {
@@ -1307,6 +1308,30 @@ Expecting one of '${allowedValues.join("', '")}'`);
       });
     }
   }
+
+  /**
+   * Display an error message if exclusive options are used together.
+   *
+   * @api private
+   */
+  _checkForExclusiveOptions() {
+    const definedOptions = this.options.filter(
+      (option) => this.getOptionValue(option.attributeName()) !== undefined
+    );
+
+    const optionsWithExclusiveNames = definedOptions.filter(
+      (option) => option.exclusiveNames.length > 0
+    );
+
+    optionsWithExclusiveNames.forEach((option) => {
+      const exclusiveAndDefined = definedOptions.filter((defined) =>
+        option.exclusiveNames.includes(defined.attributeName())
+      );
+      if (exclusiveAndDefined.length > 0) {
+        this.exclusiveOptionPresent(option, exclusiveAndDefined[0]);
+      }
+    });
+  };
 
   /**
    * Parse options from `argv` removing known options,
@@ -1558,6 +1583,17 @@ Expecting one of '${allowedValues.join("', '")}'`);
     const message = `error: required option '${option.flags}' not specified`;
     this.error(message, { code: 'commander.missingMandatoryOptionValue' });
   }
+
+  /**
+   * `Option` conflicts with another option.
+   *
+   * @param {Option} option
+   * @api private
+   */
+  exclusiveOptionPresent(option, exclusive) {
+    const message = `error: option '${option.flags}' cannot be used with '${exclusive.flags}'`;
+    this.error(message, { code:'commander.exclusiveOptionPresent' });
+  };
 
   /**
    * Unknown option `flag`.
--- a/lib/command.js
+++ b/lib/command.js
@@ -1330,11 +1330,11 @@ Expecting one of '${allowedValues.join("', '")}'`);
     );
 
     optionsWithExclusiveNames.forEach((option) => {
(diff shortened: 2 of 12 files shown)


shadowspawn:
Only one previous request I am aware of: #1408

shadowspawn:
Thinking about pattern of usage. This could be applied at the command level with a set which are mutually exclusive, or could be directional and per option with x excludes y. Looking at other packages I can see both, but x->y is more common that I expected. (Which is convenient, because it is less intrusive adding less common functionality down in `Option!)

Support in some other cli parsers:
- [argparse](https://docs.python.org/3/library/argparse.html#mutual-exclusion) has `add_mutually_exclusive_group` which is at the command level and applied to subsequently added options
 
- [oclif](https://oclif.io/docs/flags) has `exclusive` which is per flag (option)
- [yargs](https://github.com/yargs/yargs/blob/main/docs/api.md#conflictsx-y) has `conflicts` which is x->y

shadowspawn:
I was imagining these sorts of routines would identify the target option(s) using the cli flags, like `.exclusive('--json')`. There isn't other client facing code that works this way so I didn't particularly like that idea.

You have used the option name as the key. Interesting! There _is_ client facing code that works this way, like `.getOptionValue(key)`.

shadowspawn:
An interesting question is whether a conflict is if both options are specified, or both options have values. In other words, can an excluded option have a default value. Related issues in yargs:
- https://github.com/yargs/yargs/issues/1910
- https://github.com/yargs/yargs/issues/929

With `.requiredOption()` the test is whether the option has a value. (And predates being able to tell where a option value came from. Hmm, maybe `requiredOption` should ignore the default value these days too!)

shadowspawn:
No recommendations yet, just exploring the problem. 🤔 🔍

shadowspawn:
Should the exclusions be listed in the help? (i.e. `.optionDescription()`)

shadowspawn:
The value triggering the exclusive check can potentially come from an environment variable rather than the CLI. This might be difficult to reasonably capture in the error message.

shadowspawn:
One last complication to be aware of: negated options, like `--no-colour`

shadowspawn:
Thanks for a tidy description of PR. 👍 

Prompted lots of thoughts, I'll stop the stream of comments for now and think about it for a bit!

shadowspawn:
I was wondering what a hook check might look like. Using explicit test rather than generalised (and using `.error()` from Commander 9):

```js
const { Command, Option } = require('commander');

program
  .addOption(new Option('-p, --port <number>', 'port number').env('PORT'))
  .addOption(new Option('-d, --disable-server', 'disables the server')) // .exclusive(['port']));
  .hook("preAction", (thisCommand, actionCommand) => {
    const options = actionCommand.opts();
    if (options.port !== undefined && options.disableServer !== undefined)
      actionCommand.error("error: option '-program, --port <number>' cannot be used with '-d, --disable-server");
  })
  .action (() => {});

program.parse();
```

```
% node index.js --disable-server --port 8000
error: option '-program, --port <number>' cannot be used with '-d, --disable-server
% PORT=23 node index.js --disable-server    
error: option '-program, --port <number>' cannot be used with '-d, --disable-server
```

erezrokah:
> Thinking about pattern of usage. This could be applied at the command level with a set which are mutually exclusive, or could be directional and per option with x excludes y. Looking at other packages I can see both, but x->y is more common that I expected. (Which is convenient, because it is less intrusive adding less common functionality down in `Option`!)

Very good thought. My take is that you have more context on exclusive options when you add a new option, and not when creating a branch new command.
What was counterintuitive for me at first is that the usage is at the option level, and the implementation is at the command level (since we need to look at all the options to determine violations).

erezrokah:
> An interesting question is whether a conflict is if both options are specified, or both options have values. In other words, can an excluded option have a default value. Related issues in other parsers:

Great point. I didn't consider default values and required options.

I would say a conflict exists if both are specified, regardless where the value comes from (e.g. user, default or required).
Required options or options with a default value should not be able to be configured as exclusive, and we should print a relevant error message for it (I can add that logic to the PR if you agree).

The reasoning is that the purpose of this code if to save users from adding custom validation code. If we were to say that setting a default value is not conflicting, users will still need to do some custom validation to handle that specific case.
I believe this matches https://github.com/yargs/yargs/issues/929#issuecomment-525840660

> With `.requiredOption()` the test is whether the option has a value.

Can you elaborate on this? I'm understanding that testing for value, implies testing for existence too

erezrokah:
> Should the exclusions be listed in the help? (i.e. `.optionDescription()`)

Good point, let me experiment with this a bit

erezrokah:
> The value triggering the exclusive check can potentially come from an environment variable rather than the CLI. This might be difficult to reasonably capture in the error message.

True! Do we have a way to know if that was the case? I suppose we can add a `valueOrigin` property

erezrokah:
> One last complication to be aware of: negated options, like `--no-colour`

We can either have `exclusive: ['color', 'no-colour']` or have `colour` imply `no-colour`.
Not sure what's better actually, I lean towards implying `no-colour` for the same reasoning as in https://github.com/tj/commander.js/pull/1678#issuecomment-1013884175

erezrokah:
> I was wondering what a hook check might look like. Using explicit test rather than generalised (and using `.error()` from Commander 9):

Very cool, I assume I can create a utility hook and re-use it in various commands based on the logic in this PR.

shadowspawn:
> I would say a conflict exists if both are specified, regardless where the value comes from (e.g. user, default or required).
> Required options or options with a default value should not be able to be configured as exclusive, and we should print a relevant error message for it (I can add that logic to the PR if you agree).

We don't tend to warn for broken combinations if user will easily discover for themselves. And I think required+exclusive is fairly obvious. (But I am not against warning about broken combinations that are subtle.)

default+exclusive can work, as per other comments.

shadowspawn:
> > With `.requiredOption()` the test is whether the option has a value.
> 
> Can you elaborate on this? I'm understanding that testing for value, implies testing for existence too

An option value may come from the default, or from an environment variable, or from the command-line (or from custom code). The validation for `.requiredOption()` just checks that the option has a value after parsing. From any source.

shadowspawn:
> True! Do we have a way to know if that was the case? I suppose we can add a `valueOrigin` property

Conveniently, support for this was added recently. See `. getOptionValueSource()`.

shadowspawn:
> > One last complication to be aware of: negated options, like `--no-colour`
> 
> We can either have `exclusive: ['color', 'no-colour']` or have `colour` imply `no-colour`. Not sure what's better actually, I lean towards implying `no-colour` for the same reasoning as in [#1678 (comment)](https://github.com/tj/commander.js/pull/1678#issuecomment-1013884175)

In the case where there is a lone negated option, like `--no-colour`, the option key is `colour` and I think it might Just Work like other options (assuming default values are ignored, since there is an implicit value of true). It reads a confusing way (`.exclusive('colour')`) but arguably still means the option value for `colour` may not be specified if this option is specified. Not sure I have convinced myself!

In the case where there is a dual option with a positive (`--colour`) and negated (`--no-colour`) I don't think we currently have enough information to definitively tell which supplied the value. We could muck around and take a good guess, but I am tempted not to add special code and leave that in the too-hard basket for now.

erezrokah:
@shadowspawn, I rebased the PR and followed up with better error reporting and help information.
Please let me know if I missed anything

shadowspawn:
Thanks, I will take a deeper look and give it a try.

shadowspawn:
I like the clarity of this error involving an option coming from an environment variable:
```
% PORT=80 node index.js  -d
error: option '-d, --disable-server' cannot be used with environment variable 'PORT'
```

The author may want command-line arguments to "win" against environment variables, but in that case they can do it themselves and not use `exclusive`.

shadowspawn:
I wanted to experiment with adding to the help, thanks for doing that. Having seen an example and thought about it some more, I think I would prefer not to add to the help! 

The full-sentence way of referring to exclusive is  a bit different, and `exclusive` on its own does not scan as well. e.g.

> The options `--red` and `--blue` are mutually exclusive.

The help could use the short option (to save space) or the long option (as more meaningful) or both, but I don't think it can use the attribute name as that is an internal detail. In other words, the author knows that `--foo-bar` is option `fooBar` but the command-line user may not.

In many cases it will be reasonably obvious to the user that the options don't make sense used together.

The net result is I think it is better to leave it to the author to include the information themselves, if they wish.

-----

I checked `yargs` and `argParse` and they do not add info to the help for the option. Credit to `argParse` though, it does a full synopsis and does include the mutual exclusion as `[--red | --blue]`:

```
% node exclusive.js -h             
usage: PROG [-h] [--red | --blue]

optional arguments:
  -h, --help  show this help message and exit
  --red
  --blue
```

shadowspawn:
I am still thinking about interaction with defaults. I feel exclusive more naturally applies to the user not specifying mutually exclusive options than to interaction with the author supplied defaults. And conflicts with implicit or explicit defaults prompted issues against other libraries.

The situation that muddies the waters is an author reading values from a config files and putting them into the program setup as defaults. However, this is a special case and the author is being more sophisticated anyway, so I think acceptable if exclusive does not work for them out of the box.

erezrokah:
> I think I would prefer not to add to the help!

Thanks for the input! I undid the change to the help.

> I am still thinking about interaction with defaults. I feel exclusive more naturally applies to the user not specifying mutually exclusive options than to interaction with the author supplied defaults. And conflicts with implicit or explicit defaults prompted issues against other libraries.

Should we not consider options that their value source if default? I think the change will be update the filter function here https://github.com/tj/commander.js/pull/1678/files#diff-d8db36db6bfb32591336006f1dc1adee9e2caf7b021469995aa687c5c5723d0cR1318

shadowspawn:
> Should we not consider options that their value source if default? I think the change will be update the filter function here

Yes, I think ignoring defaults leads to less surprises. Other viewpoints welcome though!

erezrokah:
Please let me know if https://github.com/tj/commander.js/pull/1678/commits/92736388be97e706eb5342d26cfa03875a0ea6bc is what you had in mind.

> Other viewpoints welcome though!

Your suggestions make sense to me. Truly appreciate the feedback and the time you spend thinking about all the use cases.

shadowspawn:
I was concerned about the weak support for negated options. I had a look at the yargs support, and it is much the same as planned here! The exclusive/conflicts is based on the option name and the option having a value. No support for a difference between (say) `--colour` and `--no-colour`. That is somewhat reassuring that acceptable.

shadowspawn:
I think `conflicts` is a simpler word that scans better than `exclusive` for the method name, so recommend renaming as you offered.

shadowspawn:
I think the filtering in `_checkForExclusiveOptions` is a bit too keen. Each option with conflicts listed is a set of mutually exclusive options. Not a directional conflict. So there may be options with no value specified which list conflicts that still matter.

Consider this program:
```
program
  .addOption(new Option('--red'))
  .addOption(new Option('--blue'))
  .addOption(new Option('--green').exclusive(['red', 'blue']))
```

The second case should also be detected as a conflict. This is a b it sneaky, but avoids the author needing to list the conflicts against all of the options.

```
% node filter-bug.js --red --green
error: option '--green' cannot be used with option '--red'
% node filter-bug.js --red --blue 
{ red: true, blue: true }
```

shadowspawn:
Oh, or is that a feature and as intended? Hmmm... (Checking Yargs for comparison.)

lib/command.js
```diff
@@ -1559,6 +1590,26 @@ Expecting one of '${allowedValues.join("', '")}'`);
     this.error(message, { code: 'commander.missingMandatoryOptionValue' });
   }
 
+  /**
+   * `Option` conflicts with another option.
+   *
+   * @param {Option} option
+   * @api private
+   */
+  exclusiveOptionPresent(option, exclusive) {
```
shadowspawn, line 1599: This method can start with an underscore to show it is private. (The older routines don't have an underscore, but you can see the convention with more recent `_excessArguments`.)

lib/command.js
```diff
@@ -1308,6 +1309,36 @@ Expecting one of '${allowedValues.join("', '")}'`);
...
+        return this.getOptionValueSource(optionKey) !== 'default';
+      }
+    );
+
+    const optionsWithExclusiveNames = definedNonDefaultOptions.filter(
+      (option) => option.exclusiveNames.length > 0
+    );
+
+    optionsWithExclusiveNames.forEach((option) => {
+      const exclusiveAndDefined = definedNonDefaultOptions.filter((defined) =>
```
shadowspawn, line 1333: Observation rather than suggestion: this could be done using `find`, since we only care about one. (I don't mind the filter in part because of the previous filters, in that mindset!)

shadowspawn:
If you are keen, you can add simple tests for the new method to:
- `option.chain.test.js`
- `command.error.test.js`
- `index.test-d.ts`

(I don't mind adding those myself.)

lib/command.js
erezrokah: Great catch! Done in https://github.com/tj/commander.js/pull/1678/commits/adabd8750161322c38101cbfcfaf2ef3d8aa5026

lib/command.js
erezrokah: Right! At first I thought about reporting all conflicting options and saw it was too confusing (other tools also report the first one).

Fixed in https://github.com/tj/commander.js/pull/1678/commits/744a3396086477867d889a48eb13c41837e8a2f3

erezrokah:
Thanks for the input @shadowspawn!

I renamed the method to `conflicts` and added tests for `option.chain.test.js` and `index.test-d.ts`.
I'm not sure what I should be testing with `command.error.test.js` though, can you help me understand that?

shadowspawn:
> I'm not sure what I should be testing with `command.error.test.js` though, can you help me understand that?

Oops, wrong file, sorry! Have a look at `command.exitOverride.test.js`. This is the file that checks for each error what code is thrown and what text is in the error. Have a look at `commander.missingMandatoryOptionValue` as an example.

erezrokah:
Thanks for clarify @shadowspawn, I added the test in https://github.com/tj/commander.js/pull/1678/commits/81065dc6f0d8e1fcdf6b10a3601372c73c1984fe

lib/command.js
```diff
@@ -1559,6 +1590,27 @@ Expecting one of '${allowedValues.join("', '")}'`);
...
   }
 
+  /**
+   * `Option` conflicts with another option.
+   *
+   * @param {Option} option
+   * @param {Option} conflictingOption
+   * @api private
+   */
+  _conflictingOptionPresent(option, conflictingOption) {
```
shadowspawn, line 1600: Suggest drop the `Present`

lib/command.js
```diff
@@ -1559,6 +1590,27 @@ Expecting one of '${allowedValues.join("', '")}'`);
...
+      const optionKey = option.attributeName();
+      const source = this.getOptionValueSource(optionKey);
+      if (source === 'env') {
+        return `environment variable '${option.envVar}'`;
+      }
+      return `option '${option.flags}'`;
+    };
+
+    const message = `error: ${getErrorMessage(option)} cannot be used with ${getErrorMessage(conflictingOption)}`;
+    this.error(message, { code: 'commander.conflictingOptionPresent' });
```
shadowspawn, line 1611: Suggest drop the `Present`

shadowspawn:
Thanks @erezrokah, looking good. It might be a week or two before I can give this another tryout for final review, but I only have a cosmetic suggestion from a read through today.

lib/command.js
erezrokah: Done in https://github.com/tj/commander.js/pull/1678/commits/f7739debcc830d3017c98d4cbfb797814d1e815a

lib/command.js
erezrokah: Done in https://github.com/tj/commander.js/pull/1678/commits/f7739debcc830d3017c98d4cbfb797814d1e815a

erezrokah:
Sounds good to me @shadowspawn ❤️

tests/command.conflicts.test.js
```diff
@@ -0,0 +1,92 @@
+const commander = require('../');
+
+describe('command with conflicting options', () => {
+  function makeProgram() {
+    const actionMock = jest.fn();
+    const program = new commander.Command();
+    program
+      .exitOverride()
```
shadowspawn, line 8: To avoid the error messages polluting the test display, can suppress the output using:
```js
      .configureOutput({
        writeErr: () => {}
      })
```

Readme.md
```diff
@@ -397,7 +397,8 @@ program
   .addOption(new Option('-t, --timeout <delay>', 'timeout in seconds').default(60, 'one minute'))
   .addOption(new Option('-d, --drink <size>', 'drink size').choices(['small', 'medium', 'large']))
   .addOption(new Option('-p, --port <number>', 'port number').env('PORT'))
-  .addOption(new Option('--donate [amount]', 'optional donation in dollars').preset('20').argParser(parseFloat));
+  .addOption(new Option('--donate [amount]', 'optional donation in dollars').preset('20').argParser(parseFloat))
+  .addOption(new Option('-d, --disable-server', 'disables the server').conflicts(['port']));
```
shadowspawn, line 401: 1) This new option will be displayed in the help, so example help needs updating.

2) Suggest add an example use generating an error in the examples below.

shadowspawn (approved):
Three more minor suggestions, but I don't mind doing those myself after merge.

Nice work @erezrokah

shadowspawn:
A musing in case we get people asking about conflicts with dual positive/negative options. This is more than I feel the need to try and solve now, but capturing an idea I had while testing.

```
program
  .addOption(new Option('--red').conflicts(['color']))
  .addOption(new Option('--color'))
  .addOption(new Option('-N, --no-color'));
```

```
% node index.js --red -N
error: option '--red' cannot be used with option '--color'
```

We could perhaps extend the error message to cover both forms of option:

```
error: option '--red' cannot be used with options '--color' and '-N, --no-color'.

```

tests/command.conflicts.test.js
erezrokah: Great tip! Done in https://github.com/tj/commander.js/pull/1678/commits/351ef38144c0eae579ba2365b011117d3082067e

Readme.md
erezrokah: 1. Nice catch, done in https://github.com/tj/commander.js/pull/1678/commits/91ab6b11c0cababe6c6fbf3354693c126c6625fd

erezrokah:
> We could perhaps extend the error message to cover both forms of option:

I'll try to add this today

Readme.md
erezrokah: 2. Done in https://github.com/tj/commander.js/pull/1678/commits/79e6a7cec30712bf48bab0ad56d3ec2652ca7369
Also removed the shorthand version of `--disable-server` as it's already defined with `--drink`

erezrokah:
> We could perhaps extend the error message to cover both forms of option:

Done in https://github.com/tj/commander.js/pull/1678/commits/39eba86e2ac58ad22661482a12c76f0d2d4ab808.
The error message is not exactly the same as in your example for simplicity of implementation.
Please let me know what you think.

shadowspawn:
Options with positive and negated forms, hmm...

I was looking at the error generation code and realised it could show the wrong environment variable too. That somewhat breaks my idea of showing both forms of the option in the error, or at least confirms it is a hack and not a general solution.

```js
program
  .addOption(new Option('--red'))
  .addOption(new Option('--dual').env('DUAL').conflicts('red'))
  .addOption(new Option('--no-dual').env('NO_DUAL'))
```

```
% NO_DUAL= node index.js --red 
error: environment variable 'NO_DUAL' cannot be used with option '--red'
% DUAL= node index.js --red  
error: environment variable 'NO_DUAL' cannot be used with option '--red'
```

shadowspawn:
I don't want to store further information about the source of the option value for now, just to support `conflicts`, so still working with the limited information we have, and seeing what we can do. 

I commented previously: https://github.com/tj/commander.js/pull/1678#issuecomment-1014230704

> In the case where there is a dual option with a positive (`--colour`) and negated (`--no-colour`) I don't think we currently have enough information to definitively tell which supplied the value. We could muck around and take a good guess, but I am tempted not to add special code and leave that in the too-hard basket for now.

Yup, yup, limited information and messy. I'm interested in trying that "good guess" now to see if that is worth the effort to improve the messaging, having talked around it more! i.e. The negated option usually just produces `false`. I'll have a play around with that...

shadowspawn:
Here is a try at a finding the best option to display.

```js
  _conflictingOption(option, conflictingOption) {
    // The calling code does not know whether a negated option is the source of the
    // value, so do some work to take an educated guess.
    const findBestOptionFromValue = (option) => {
      const optionKey = option.attributeName();
      const optionValue = this.getOptionValue(optionKey);
      const negativeOption = this.options.find(target => target.negate && optionKey === target.attributeName());
      const positiveOption = this.options.find(target => !target.negate && optionKey === target.attributeName());
      if (negativeOption && (
        (negativeOption.presetArg === undefined && optionValue === false) ||
        (negativeOption.presetArg !== undefined && optionValue === negativeOption.presetArg)
      )) {
        return negativeOption;
      }
      return positiveOption || option;
    };

    const getErrorMessage = (option) => {
      option = findBestOptionFromValue(option);
      const optionKey = option.attributeName();
      const source = this.getOptionValueSource(optionKey);
      if (source === 'env') {
        return `environment variable '${option.envVar}'`;
      }
      return `option '${option.flags}'`;
    };

    const message = `error: ${getErrorMessage(option)} cannot be used with ${getErrorMessage(conflictingOption)}`;
    this.error(message, { code: 'commander.conflictingOption' });
  }
```

shadowspawn:
And some manual testing:

```js
program
  .addOption(new Option('--red'))
  .addOption(new Option('--blue').conflicts(['red']))
  .addOption(new Option('-N, --no-color').conflicts(['red']))
  .addOption(new Option('--dual').env('DUAL').conflicts('red'))
  .addOption(new Option('--no-dual').env('NO_DUAL'))
  .addOption(new Option('--dual2 <str>').conflicts('red'))
  .addOption(new Option('--no-dual2').preset('BAD'))
```

```
1678 % node index.js  --red --blue      
error: option '--blue' cannot be used with option '--red'
1678 % node index.js  --red -N    
error: option '-N, --no-color' cannot be used with option '--red'
1678 % node index.js  --red --dual
error: option '--dual' cannot be used with option '--red'
1678 % node index.js  --red --no-dual
error: option '--no-dual' cannot be used with option '--red'
1678 % DUAL= node index.js  --red    
error: environment variable 'DUAL' cannot be used with option '--red'
1678 % NO_DUAL= node index.js  --red 
error: environment variable 'NO_DUAL' cannot be used with option '--red'
1678 % node index.js  --red --dual2 foo
error: option '--dual2 <str>' cannot be used with option '--red'
1678 % node index.js  --red --no-dual2 
error: option '--no-dual2' cannot be used with option '--red'
```

erezrokah:
Thanks @shadowspawn, I followed up in https://github.com/tj/commander.js/pull/1678/commits/46bf4b03aa10479473937a2ed42bdd57bdfb00ee based on your suggestions.

You saved me going down a rabbit 🕳️ , as I was ready to look into storing additional information on the option.

shadowspawn:
I had wondered about whether to support a single string being passed to `.conflicts()` rather than an array. I wasn't going to raise it until I noticed I made that very mistake in one of my examples!

```
  .addOption(new Option('--dual2 <str>').conflicts('red'))
```

It does not fail at runtime, but is doing a string `includes` rather than an array `includes` in the conflict testing and not actually correct.

shadowspawn:
A possible implementation, which as a bonus makes repeated calls accumulate:

```
  conflicts(names) {
    if (!Array.isArray(names) && typeof names !== 'string') {
      throw new Error('conflicts() argument must be a string or an array of strings');
    }

    this.conflictsWith = this.conflictsWith.concat(names);
    return this;
  }
```

This will change the parameter type in the JSDoc and Typescript, and need a few more tests too!

(If you think this is sensible but don't want to do it yourself, I am happy to try adding it.)

erezrokah:
@shadowspawn great point as I didn't notice that error as well. Let me know if https://github.com/tj/commander.js/pull/1678/commits/0059715b6a9d501a038719f2c40a1151d012c658 is what you had in mind.

shadowspawn:
Yes, nice. 👍

shadowspawn (approved):

shadowspawn:
We have given other people a couple of weeks for commenting after mentioning this on the `.implies()` issue and previous conflicts issue.

Over to you @abetomo for a look, when you have time. 😄

abetomo (approved):
Thanks for your contribution!
LGTM.

shadowspawn:
Thanks for all the hard work @erezrokah. This can go out in a minor release, probably within the next few weeks.

erezrokah:
Thanks @abetomo and @shadowspawn for the review.
I truly enjoyed the interaction and thoughtful feedback ❤️

vassudanagunta:
@shadowspawn @erezrokah This is obviously late, but I came across this thread while evaluating Commander vs yargs. (Wouldn't it be great if there was a decision tree service, perhaps interactive, that would help people make such choices?)

Let's assume `PORT=80` is set in the environment somehow (I don't want to set it on the same line as the command as in the test examples in previous comments because that is visually misleading). Since CLI options override environment variables (I assume this to be Commander's rule):

```
% node index.js -p 8080   # no conflict error, port is 8080, env is ignored
```

wouldn't it also be logical that a related "conflicting" option also have precedence over the same environment variable and not result in a conflict error?  For example, default/ENV server-related values should simply  irrelevant when the server is disabled, not generate a conflict error:

```
% node index.js --disable-server'  # no conflict, server disabled, related PORT env var is ignored
```

To me a default is not the same as an explicit value. It is included for convenience/usability/safety only. I would think that it is a very common occurrence, to continue to server example, for the program to have a default port that takes effect *if and only if* the command results in starting a server, without requiring the command to do so. 

Yes, 
>  The author may want command-line arguments to "win" against environment variables, but in that case they can do it themselves and not use exclusive.

but if the more common and intuitive behavior is as I describe, shouldn't that be the default behavior and authors can do themselves if they want otherwise?

Just a thought.

vassudanagunta:
@shadowspawn, as I said I was only looking at this thread to help me decide which CLI library to adopt. What I was looking for is a sense of the maintainer(s), how they think, how they evaluate ideas, what kind of care and judgement they use. Just wanted to say I am highly impressed!

vassudanagunta:
If you think my point has merit, I can open a fresh issue if you would like.

shadowspawn:
The environment variables might get used for defaults, or they might be the primary way options are set. This second approach is common in container environments so you can modify the runtime without rebuilding the container.

There is some discussion around env in an earlier PR: #1266

We decided in this PR that conflicts would not consider default values. One big difference between simple defaults and env is that the author knows what the defaults are and why. The environment variables are coming from the end-user, who may be using them like defaults or like CLI options, and may or may not understand how they interact.

One work-around to get the behaviour you describe is to set the environment variables as the default value in code, and not use the `.env()` support.  https://github.com/tj/commander.js/pull/1266#issuecomment-633753085

vassudanagunta:
Makes sense. Another option might be to make explicit the two different roles that an environment variable might play, e.g.

```
.addOption(new Option('-p, --port <number>').env('PORT', {asDefault: true}))
```

or maybe just an boolean:

```
.addOption(new Option('-p, --port <number>').env('PORT', true))
```

The nice things about this solution are:
- backward compatible: if the extra argument is not supplied, it defaults to treating the env var as an explicit setting, the same as today.
- simple implementation: if `asDefault` is true, then treat the environment value using the existing logic for option defaults (which don't trigger conflicts).
- there may be other benefits, if there are other areas where the two different roles of environment variables should be treated differently by Commander.

A `role:  'default' | 'explicit' | 'fixed'` option would be more powerful and future proof:

```
//'fixed' means an env value can't be overridden even by a CLI arg, a nice safety feature.
.addOption(new Option('-p, --port <number>').env('PORT', {role: 'fixed'}))
```

Another option I considered doesn't seem as elegant or intuitive: 
```
.addOption(new Option('--disable-server')
        .conflicts('port', {includeEnv: false}))
```

Just some ideas :)

PR #1679, opened 2022-01-16
Allow readonly arrays to be used as choices

Pzixel:
Consider following usage:

```ts
const SupportedDrinks = ['small', 'medium', 'large'] as const;
type Drink = typeof SupportedDrinks[number];

const program = new Command()
      .addOption(new Option('-d, --drink <size>', 'drink size').choices(SupportedDrinks))
```

This allows user to get a strongly typed enum and reuse it to automatically populate supported drinks. 
This code doesn't compile now though since readonly array is not assignable to just `string[]`.

# Pull Request

### DIFF
--- a/typings/index.d.ts
+++ b/typings/index.d.ts
@@ -61,7 +61,7 @@ export class Argument {
   /**
    * Only allow argument value to be one of choices.
    */
-  choices(values: string[]): this;
+  choices(values: readonly string[]): this;
 
   /**
    * Make option-argument required.


Pzixel:
Sorry I didn't see there is a template for PRs, I've just sent it via regular "edit-and-create-PR" flow. Should I add any tests/more motivation here?

shadowspawn:
This will be supported in Commander 9. See #1679. That PR started out fairly small too, but we then made code changes to make the usage of the parameters more robustly readonly.

Two PRs in quick succession, so readonly is something people are hitting. 😄

Pzixel:
1679 is this PR, you probably meant to mention something else... 😅

shadowspawn:
Cough. Try this one: #1667

PR #1685, opened 2022-01-31
Follow best practice and require process rather than use global

shadowspawn:
# Pull Request

Making a minor change that I saw in another project and looked up the related documentation.

> The `process` object provides information about, and control over, the current Node.js process. While it is available as a global, it is recommended to explicitly access it via require or import:

https://nodejs.org/dist/latest-v16.x/docs/api/process.html#process


abetomo (approved):
👍

Issue #1150, opened 2020-01-14
Question: anybody generate Readme from commander.js annotates?


akaguny:
relative to #756

shadowspawn:
This issue has not had any activity in a month.

Feel free to open a new issue if it comes up again, with new information and renewed interest.
 
Thank you  for your contributions.

Issue #756, opened 2018-01-23
Auto-Generate Help Documentation (Maybe as a section within a readme document)?


mofosyne:
Is there a way to autogenerate a document from a base command that list out all the commands and command structure in a easily searchable markdown document?

This allows for self documenting cli interfaces.

------------

edit: okay mjavelinux if I get the time, I'll see if I can do it.

Notes:

```
# Help generation
Command.prototype.helpInformation() -- Gen help message string
--> Command.prototype.outputHelp() -- Send to std out
--> Command.prototype.help() -- do all above then exit
--> Command.prototype.addImplicitHelpCommand() -- add help command automagically

# Sub commads like `pm-install` is handled in executeSubCommand()

Command.prototype.executeSubCommand()
---> `var bin = basename(f, '.js') + '-' + args[0];`

```

Objective would be to use the above as a template, but be able to traverse the command() structure and print all commands under it in markdown format.

mojavelinux:
You can take ideas from gli, which as this feature. See https://github.com/davetron5000/gli/blob/gli-2/lib/gli/commands/doc.rb

shadowspawn:
This issue has not had any activity in over six months. It isn't likely to get acted on due to this report. (But this is an interesting idea.)

Feel free to open a new issue if it comes up again, with new information and renewed interest.
 
Thank you  for your contributions.

kuncevic:
I did this as a quick workaround

```
  program
    .addOption(new Option('-cmd2json')).action(name => {
    console.log(JSON.stringify((program.commands.map((x: any) => ({
      command:`${x.name()} ${x.alias() !== undefined ? x.alias() : ''}`,
      description: x.description()
    })))));
  });
```

Then pasted it [here](https://kdelmonte.github.io/json-to-markdown-table) and copy pasted the result in to my README.md  
You also can use [markdown-table](https://www.npmjs.com/package/markdown-table) to build table locally

kuncevic:
I did this as a quick workaround

```
  program
    .addOption(new Option('-cmd2json').hideHelp());

  program.on('option:-cmd2json', () => {
    stdout(JSON.stringify((program.commands.map((x: any) => ({
      command:`${x.name()} ${x.alias() !== undefined ? x.alias() : ''}`,
      description: x.description()
    })))));
    process.exit();
  });
```

Then pasted it [here](https://kdelmonte.github.io/json-to-markdown-table) and copy pasted the result in to my README.md  
[...]

PR #1688, opened 2022-02-05
feat: automate help documentation

tomasbjerre:
# Pull Request

Not complete, but wanted to open the PR early to see if this is something you may want to include? Totally ok if you don't.

## Problem

I have the output of `help` in my [README.md](https://github.com/tomasbjerre/serverless-server-spawner/blob/18fe5bfc2c5fb255447b434dddc270c9c4b766dc/README.md?plain=1#L112).
When I change the configuration of `Commander`, I often forget to update my `README.md`.

## Solution

Have the tool update a given file, when run with:

```shell
UPDATE_HELP_FILE=README.md program --help
```

It will look for:

```md
[//]: # Commander help

`` `
whatever
`` `
```

And replace `whatever` with the help-output.

TODO:

 - tests


shadowspawn:
My first impression is I think this is a fairly custom behaviour which does not need to be built into Commander.

A different way of invoking this could be as a hidden option, which would be easier to use on Windows.

This feels like a generic problem, populating some parts of a document by running commands, filling out a template. I wonder if there are higher level solutions available.

tomasbjerre:
Yes, it does not need to be built in. I can implement it in my user code:
https://github.com/tomasbjerre/serverless-server-spawner/blob/18fe5bfc2c5fb255447b434dddc270c9c4b766dc/src/spawner/updateReadme.ts

I just don't like having to implement it in every tool where I use Commander.

shadowspawn:
> A different way of invoking this could be as a hidden option, which would be easier to use on Windows.

Hmm... Hooking the help does mean it could work for invoking on subcommand. But that would needing a custom target for the substitution too.

shadowspawn:
From your user code (not PR code). I think you can simplify:

```
const helpText = new Help().formatHelp(command, new Help());
```
 to

```
const helpText = commander.helpInformation();
```

shadowspawn:
A heads-up, the easy ways of getting at the help text do not include custom text added by `addHelpText`.

There is extra work done by `outputHelp` to add these: https://github.com/tj/commander.js/blob/02a124c7d58dbae2ef11f9284b2c68ad94f6dc8b/lib/command.js#L1835

tomasbjerre:
I pushed test cases now. I split the code into `updateFileWithHelp` and `_updateFileWithHelp` and only wrote test cases on `_updateFileWithHelp`. Because `updateFileWithHelp` would require mocking and, in my opinion, those test cases would just duplicate production code.

From what I understand:
 * You are not really convinced this feature should be added or not.
 * I'm not really sure where/how the feature is best triggered.

shadowspawn:
> You are not really convinced this feature should be added or not.

I am leaning towards no. I understand the use case, and that you have implemented a solution which works for your programs and how you document them. 

I don't think it is a standard enough problem and robust enough solution to pull into Commander and maintain and support. It can be implemented by an add-on package, or using external tooling which would allow other templating.

From a security point of view, I have reservations about including file modification functionality into end-user packages.

I found one high-level tool in a quick search: https://nedbatchelder.com/code/cog/

Related Commander issues: #756 #1150

shadowspawn:
I'm happy to leave this open for a week to see if there is any further feedback from other people.

(Thanks @tomasbjerre for contributing, and thoughtful comments. 😄 )

PR #1692, opened 2022-02-19
Fix syntax errors in migration example

shadowspawn:
# Pull Request

Fix syntax errors in Commander 7 migration tips.

(Have made same changes in Release Notes for 7.0.0 on GitHub.)

Same code as mentioned in: https://github.com/tj/commander.js/pull/1409#issuecomment-1036348880


abetomo (approved):
👍

PR #1702, opened 2022-03-14
Update CHANGELOG and version for 9.1.0

shadowspawn:
Preparing for v9.1.0. I put Friday as date, but just when suits you @abetomo. Later or earlier. 📆


shadowspawn:
Another release done. Thanks! 😄

abetomo:
@shadowspawn Thank you for your meticulous work!
We have released it.

PR #1704, opened 2022-03-16
Add `"types"` to `"exports"`

### LINKED ISSUE #1703: Be able to import types when using `NodeNext` TypeScript module resolution
I'm making an npm package which uses `ora`. To use `ora`, I switched to `NodeNext` module resolution. After doing that, the type checking for `commander` stopped working. I think `package.json` needs to be updated to be compatible with `NodeNext` module resolution. 

Here is the guide to by the author of `ora`: https://gist.github.com/sindresorhus/a39789f98801d908bbc7ff3ecc99d99c.
Here is some related TypeScript documentation: https://devblogs.microsoft.com/typescript/announcing-typescript-4-5-beta/#esm-nodejs

ChocolateLoverRaj:
# Pull Request

## Problem
Fixes #1703 

## Solution
Add `"types"` to `"exports"`

### DIFF
--- a/package.json
+++ b/package.json
@@ -39,9 +39,9 @@
   "main": "./index.js",
   "exports": {
     ".": {
+      "types": "./typings/index.d.ts",
       "require": "./index.js",
-      "import": "./esm.mjs",
-      "types": "./typings/index.d.ts"
+      "import": "./esm.mjs"
     },
     "./esm.mjs": "./esm.mjs"
   },
--- a/package.json
+++ b/package.json
@@ -40,7 +40,8 @@
   "exports": {
     ".": {
       "require": "./index.js",
-      "import": "./esm.mjs"
+      "import": "./esm.mjs",
+      "types": "./typings/index.d.ts"
     },
     "./esm.mjs": "./esm.mjs"
   },


shadowspawn (changes requested):
Thanks. The example on this page: https://www.typescriptlang.org/docs/handbook/esm-node.html#packagejson-exports-imports-and-self-referencing

has a comment saying "types" must come first:

```
            // Entry-point for TypeScript resolution - must occur first!
            "types": "./types/index.d.ts",
```

(It did work for me last when I was experimenting, but I think we better follow the example.)

shadowspawn:
This PR is to work with a _future_ version of TypeScript. I have been doing a lot of reading to try and understand whether it is safe for us to ship now. And still not sure! Maybe I should ask...

1) The "types" value in "exports" is described in the Node.js 16 documentation so that part at least appears in something released.
https://nodejs.org/dist/latest-v16.x/docs/api/packages.html#community-conditions-definitions

2) I found these comments in long threads, two talking about updates by library authors:
- https://github.com/microsoft/TypeScript/issues/46452#issuecomment-1031713598
- https://github.com/microsoft/TypeScript/issues/46452#issuecomment-1031961407
- https://github.com/microsoft/TypeScript/issues/46334#issuecomment-943632787

Some interesting information about emit vs resolving with answer from an expert:
- https://stackoverflow.com/questions/71463698/why-we-need-nodenext-typescript-compiler-option-when-we-have-esnext

3) `rxjs` have merged a PR doing this just recently, and also wondering whether to go before TypeScript supports it!

- https://github.com/ReactiveX/rxjs/pull/6802

shadowspawn:
Asked if stable enough for release: https://github.com/microsoft/TypeScript/issues/46452#issuecomment-1070369344

shadowspawn:
And got a "100% yes".

https://github.com/microsoft/TypeScript/issues/46452#issuecomment-1070377450

shadowspawn:
Oh, but then there was a followup. I need to reread some of the earlier comments again!

https://github.com/microsoft/TypeScript/issues/46452#issuecomment-1070418207

ChocolateLoverRaj:
@shadowspawn How could adding a `"types"` entry in `"exports"` break things?

shadowspawn:
> How could adding a `"types"` entry in `"exports"` break things?

In general, by making the behaviour worse for a use case where something reads `"exports"` and recognises the `"types"` entry. Not just TypeScript, say ts-node or WebStorm. (Hypothetical examples, I have not idea what the do with `"exports"`!)

shadowspawn:
@abetomo I am going to test now one more time with `typescript@next` generating CommonJS and ESM, then call it good. Could make it into v9.1, but no worries if it misses.

shadowspawn (approved):
LGTM

abetomo (approved):
👍

abetomo:
I think it is fine to release it next time.

shadowspawn:
Ok. I won't merge until after release we have lined up.

PR #1708, opened 2022-03-20
Add example for "global" options

shadowspawn:
# Pull Request

## Problem

Sometimes people would like to be able to add common (global) options to all the subcommands, rather than to the program. This was the most popular enhancement in poll https://github.com/tj/commander.js/issues/1551#issuecomment-869093337

Related:
- #243 _so they're all already populated with the common options_
- #476  _add global options to a program that is applied to all commands_
- https://github.com/tj/commander.js/issues/1229#issuecomment-623613637  _a set of common options for all my subcommands_
- #1426 _an option that I add to all the commands of my application_
- #1631 _add global options to each of the commands_

I tried adding an explicit routine in #1670 but it turned out less effective than leaving it up to the user.

## Solution

Add an example file showing two ways to add global options to subcommands.

### DIFF
--- a/examples/common-options.js
+++ b/examples/common-options.js
@@ -1,0 +1,45 @@
+#!/usr/bin/env node
+
+// This example shows a couple of ways of adding a common option to all the subcommands.
+// We are using one level of subcommands. (Adding options to just the leaf subcommands
+// with deeper nesting would be a little more work.)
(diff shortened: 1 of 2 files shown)


abetomo (approved):
👍

PR #1709, opened 2022-03-20
Clarify option arguments

shadowspawn:
1) Options with a required option-argument are greedy and consume the next argument whatever the value. Options with an optional option-argument are choosy, and ignore arguments which start with a dash. We haven't written that down in the README before.

2) Added explicit example of the ways an option and its value can be specified.

3) Moved and slightly reworked the description of combining boolean short options, and trailing option taking a value.


abetomo (approved):
LGTM!

Issue #1710, opened 2022-03-24
9.1.0 - option.conflicts not working


rpearce63:
I'm trying to use the option conflicts configuration, but it doesn't appear to catch the conflicts.

```
const { program, Option } = require("commander");

const handleAction = (command) => {
  const options = program.opts();
  //options.exact && (options.wild || options.all) && console.log("conflict");
  console.log(
    `${command.name()} with options ${JSON.stringify(program.opts())}`
  );
};
//Global options
program
  .addOption(
    new Option("-a, --all", "all wallets").conflicts(["exact", "wild"])
  )
  .addOption(
    new Option("-x, --exact <name>", "exact wallet name").conflicts([
      "all",
      "wild",
    ])
  )
  .addOption(
    new Option("-w, --wild <partial>", "wild card").conflicts(["all", "exact"])
  )
  .option("-p, --password <password>");

//Sub commands
program.command("claim").action((options, command) => {
  handleAction(command);
});

program.parse();

```

When I run with:
node commands.js claim -a -x OG

The result is:
claim with options {"all":true,"exact":"OG"}

Have I configured it incorrectly? Also tried specifying just the short names. Same result.

Thanks.

shadowspawn:
Thanks for the example.

This is a bug. When a subcommand is used, the conflicts are only being checked on the subcommand and not the program. 

(In the code it looks right with `_checkForConflictingOptions` in same place as `_checkForMissingMandatoryOptions`, but that second call walks up the stack.)

rpearce63:
Any idea when this might be released?

shadowspawn:
Hi @rpearce63 , likely within a week. 

(Not a hard rule, but I prefer to wait a month between releases.)

shadowspawn:
Fixed in Commander v9.2.0

rpearce63:
The update worked.  Thanks.

```
 % node commands.js claim -a -x OG
error: option '-a, --all' cannot be used with option '-x, --exact <name>'
```

PR #1711, opened 2022-03-25
Check for conflicts in parent commands

shadowspawn:
# Pull Request

## Problem

Only testing for conflicts in leaf command.

See: #1710

## Solution

Follow same pattern as checking for required options:
- run up parents checking them as well
- call in dispatch to external command too

## ChangeLog

Fixes:
- detect conflicts in parent commands of called subcommand (#1710)


lib/command.js
```diff
@@ -1340,6 +1341,19 @@ Expecting one of '${allowedValues.join("', '")}'`);
     });
   }
 
+  /**
+   * Display an error message if conflicting options are used together.
+   * Called after checking for help flags in leaf subcommand.
+   *
+   * @api private
+   */
+  _checkForConflictingOptions(cmd) {
```
abetomo, line 1350: I don't think the `cmd` argument is necessary.

shadowspawn: Oops, left over from an earlier version of code. Thanks. I should have reread it!

shadowspawn: Fixed.

abetomo (approved):
LGTM!

PR #1718, opened 2022-04-18
Deprecate multi-character short flags

shadowspawn:
Deprecate multi-character short flags, like say `-ws, --work-space`. We never deliberately supported it, but I have seen it being used in the wild.

Add it to the deprecated list so if someone checks they can see should not be using it, and we may block or drop it in future.


abetomo (approved):
👍

PR #1719, opened 2022-04-18
Rework example to avoid graphql warnings

shadowspawn:
Update example for command-arguments:
- remove the stale file showing deprecated extended `.description()`
- rework example as graphql complains about logging variable named "password"!


examples/argument.js
```diff
@@ -1,24 +1,24 @@
...
-  .description('example program for argument')
-  .action((username, password) => {
-    console.log('username:', username);
-    console.log('password:', password);
+  .name('connect')
+  .argument('<server>', 'connect to the specified server')
+  .argument('[user]', 'user account for connection', 'guest')
+  .description('example program with argument descriptions')
+  .action((server, user) => {
+    console.log('username:', server);
```
abetomo, line 15: I think `'username'` would be better as `'server'` or something like that.

shadowspawn: Oops, missed that one! Thanks.

shadowspawn: Fixed

abetomo (approved):
👍

PR #1720, opened 2022-04-18
Fail on missing required option-arg when option is not last argument (resolves #1684)

### LINKED ISSUE #1684: Missing values for required options wrongly get value of next option name
``` 
calRTT [options]

       --times <num>        "Times desc"
       --verbose            "if verbose"
```
For example, my CLI name is calRTT and it has two arguments, `times` and `verbose`. If I run this program in terminal and input as
`calRTT --times --verbose`, it will return string "--verbose" if I call `options.times`. But what I want is to throw an exception as it missed value **num**, instead of return the name of next option name with quotes. How can I solve this?

currently my code is:
```
const program = new Command()
program
.name("calRTT")
.option('-t, --times', 'Times desc')
.option('-v, --verbose' 'if verbose')

program.parse(process.argv)
const options = program.opts()
```

rucker:
## Problem
Existing logic to catch missing required option-arguments works only in the case that the option is the last argument.
Example command:

```
program
  .option('--no-sauce', 'Remove sauce')
  .option('--cheese <flavour>', 'cheese flavour', 'mozzarella')
  .option('--no-cheese', 'plain with no cheese')
  .parse();
```
Calling this command with `--cheese` as the last/only argument (missing its required option-argument `<flavour>`) will produce the error: `error: option '--cheese <flavour>' argument missing`.

However, calling the same command with arguments `--cheese --no-sauce` (still missing the required `<flavour>` option-argument) will produce no error and the command will run with a value of `--no-sauce` assigned to the option `--cheese`. This issue was reported in #1684.

## Solution
When parsing options, check that required option-argument values do not match existing option names.

## ChangeLog
Check for missing required option-args when option is not last one passed.

### DIFF
--- a/lib/command.js
+++ b/lib/command.js
@@ -1404,7 +1404,7 @@ Expecting one of '${allowedValues.join("', '")}'`);
         if (option) {
           if (option.required) {
             const value = args.shift();
-            if (value === undefined) this.optionMissingArgument(option);
+            if (value === undefined || this._findOption(value)) this.optionMissingArgument(option);
             this.emit(`option:${option.name()}`, value);
           } else if (option.optional) {
             let value = null;
--- a/lib/command.js
+++ b/lib/command.js
@@ -1404,7 +1404,10 @@ Expecting one of '${allowedValues.join("', '")}'`);
         if (option) {
           if (option.required) {
             const value = args.shift();
-            if (value === undefined || this._findOption(value)) this.optionMissingArgument(option);
+            if (value === undefined || (this._findOption(value.split('=')[0])) ||
+              (maybeOption(value) && this._findOption(value.slice(0, 2)))) {
+              this.optionMissingArgument(option);
+            }
             this.emit(`option:${option.name()}`, value);
           } else if (option.optional) {
             let value = null;
--- a/tests/options.required.test.js
+++ b/tests/options.required.test.js
@@ -37,6 +37,66 @@ describe('option with required value, no default', () => {
     // Assert
     expect(mockOptionMissingArgument).toHaveBeenCalled();
   });
+
+  test('when option value not specified and other options follow then error', () => {
+    // Arrange. Mock error routine to allow interception.
+    const mockOptionMissingArgument = jest.fn(() => {
+      throw new Error('optionMissingArgument');
+    });
+    const program = new commander.Command();
+    program.optionMissingArgument = mockOptionMissingArgument;
+    program
+      .option('--cheese <type>', 'cheese type')
+      .option('--sauce <type>', 'sauce type');
+
+    // Act. The throw is due to the above mock, and not default behaviour.
+    expect(() => {
+      program.parse(['node', 'test', '--cheese', '--sauce', 'red']);
+    }).toThrow();
+
+    // Assert
+    expect(mockOptionMissingArgument).toHaveBeenCalled();
+  });
+
+  test('when option value not specified and other options follow using alternate long option syntax then error', () => {
+    // Arrange. Mock error routine to allow interception.
+    const mockOptionMissingArgument = jest.fn(() => {
+      throw new Error('optionMissingArgument');
+    });
+    const program = new commander.Command();
+    program.optionMissingArgument = mockOptionMissingArgument;
+    program
+      .option('--cheese <type>', 'cheese type')
+      .option('--sauce <type>', 'sauce type');
+
+    // Act. The throw is due to the above mock, and not default behaviour.
+    expect(() => {
+      program.parse(['node', 'test', '--cheese', '--sauce=red']);
+    }).toThrow();
+
+    // Assert
+    expect(mockOptionMissingArgument).toHaveBeenCalled();
+  });
+
+  test('when option value not specified and other options follow using alternate short option syntax then error', () => {
+    // Arrange. Mock error routine to allow interception.
+    const mockOptionMissingArgument = jest.fn(() => {
+      throw new Error('optionMissingArgument');
+    });
+    const program = new commander.Command();
+    program.optionMissingArgument = mockOptionMissingArgument;
+    program
+      .option('--cheese <type>', 'cheese type')
+      .option('-s --sauce <type>', 'sauce type');
+
+    // Act. The throw is due to the above mock, and not default behaviour.
+    expect(() => {
+      program.parse(['node', 'test', '--cheese', '-sred']);
+    }).toThrow();
+
+    // Assert
+    expect(mockOptionMissingArgument).toHaveBeenCalled();
+  });
 });
 
 // option with required value, with default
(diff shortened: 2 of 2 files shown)


shadowspawn:
This is a pretty complex issue. The current behaviour follows the POSIX [Utility Conventions](https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap12.html#tag_12_01) and [notes](https://pubs.opengroup.org/onlinepubs/9699919799/xrat/V4_xbd_chap12.html)

Guideline 7 states: _Option-arguments should not be optional_. This is clarified in the  notes to make it clear that the option is greedy:

> Guideline 7 allows any string to be an option-argument; an option-argument can begin with any character, can be - or --, and can be an empty string. For example, the commands [pr](https://pubs.opengroup.org/onlinepubs/9699919799/utilities/pr.html) -h -, [pr](https://pubs.opengroup.org/onlinepubs/9699919799/utilities/pr.html) -h --, [pr](https://pubs.opengroup.org/onlinepubs/9699919799/utilities/pr.html) -h -d, [pr](https://pubs.opengroup.org/onlinepubs/9699919799/utilities/pr.html) -h +2, and [pr](https://pubs.opengroup.org/onlinepubs/9699919799/utilities/pr.html) -h " contain the option-arguments -, --, -d, +2, and an empty string, respectively. Conversely, the command [pr](https://pubs.opengroup.org/onlinepubs/9699919799/utilities/pr.html) -h -- -d treats -d as an option, not as an argument, because the -- is an option-argument here, not a delimiter.

rucker:
Thanks for the references. I figured this would get a little complicated once we got down into the details. Erroring 
on next positional args starting with `-` seemed too strict so I figured the next-best sign of a missing option-argument would be the next argument being another (known) option entirely.

Ultimately my aim is to help produce meaningful errors when required option-args are omitted. Barring this, the command could result in any number of downstream errors that would probably be too far removed from the issue (missing input) to be meaningful to the user.

Regarding your examples:

```
% node index.js --cheese --with WWW
error: option '--cheese <flavour>' argument missing
```
This is the error I would expect.
```
% node index.js --cheese --with=WWW
cheese: --with=WWW
```
I overlooked the `option=value` syntax. I pushed an update that addresses this.
```
% node index.js --cheese -w WWW    
error: option '--cheese <flavour>' argument missing
```
This is the error I would expect, unless `-w` is a sub-option to `--cheese`. But I don't think Commander supports that.
```
% node index.js --cheese -wWWW 
cheese: -wWWW
```
I overlooked this short option syntax. I pushed an update that addresses this.
```
% node index.js --cheese -xy  
cheese: -xy
```
If `-x` is a short option, ditto the previous comment. Otherwise I believe this is expected behavior: in the case that the next arg is not an option, it is accepted as the option-argument.
```
% node index.js --cheese --chese
cheese: --chese
```
I believe this is expected behavior as well.

shadowspawn:
Quick update, I meant to post the code I used for the examples. `-xy` is a short option group, and I suspect covered by your updated code.

```
program
  .option('--cheese <flavour>', 'cheese flavour', 'mozzarella')
  .option('-w, --with <value>')
  .option('x')
  .option('y');

program.parse();
console.log(`cheese: ${program.opts().cheese}`);
```

shadowspawn:
A few suggestions:

1) To start with, I would like to only reject following argument matching long option (alone or with value). This case is "obviously" a missing argument, or in rare case that intended the user will agree that it looks like it might be. Rejecting anything starting with a short option is less clear.

2) I would like a custom error. I prefer the idea we are choosing to reject the argument because it is ambiguous, rather than because it is missing.

3) Fancy error message! Leave this until last, and ok if you aren't up for this part.

I worked on a message for this case recently in `parseArgs`. The error message suggestion leaves out the short variation unless the user was actually using the short option, since `-fVALUE` is not as well known as `--foo=VALUE`.

https://github.com/pkgjs/parseargs/blob/6ca85cdc55feee7bef6d8a5f2fb601e97ebb480a/index.js#L89

That version of the `parseArgs` code is using `-XYZ` to avoid possible quoting problems. I have some untried code to use the "real" value when simple. As is often the case with good error messages, more code doing a nice error than testing for the error case!

```
 // Only show simple values in messages, to avoid quoting complications.
 const dashed = RegExpPrototypeTest(/^-[A-Za-z0-9-_.]+$/, optionValue) ? optionValue : '-XYZ';
 // Only show short example if user used short option.
 const example = (shortOrLong.length === 2) ?
   `'--${longOption}=${dashed}' or '${shortOrLong}${dashed}'` :
   `'--${longOption}=${dashed}'`;
```

rucker:
>To start with, I would like to only reject following argument matching long option

I can see how, given the different short options, this might be the simplest thing to do. Did you have any examples in mind that are too tricky to cover?
>I would like a custom error. I prefer the idea we are choosing to reject the argument because it is ambiguous, rather than because it is missing.

I really like the example from `parseArgs` you provided. I think treating cases we can identify as _ambiguous_ is a good way to think about it.
>Fancy error message! Leave this until last, and ok if you aren't up for this part.

I will think about this. If a combination of the original "argument missing" error + an added "ambiguous argument" message aren't enough, I'm open to adding one.

shadowspawn:
> > To start with, I would like to only reject following argument matching long option
> 
> I can see how, given the various short option syntaxes, this might be the simplest thing to do. Did you have any examples in mind that are too tricky to cover?

I'll give a long answer to explain what I am concerned about.

Say we have a program:

```js
const { program} = require('commander');

program
  .option('-i, --identifier <id>')
  .option('-a, --append <value>')
  .argument('[files...]');

program.parse();
console.log(program.opts());
console.log(program.args);
```

Here is call with all arguments supplied with Commander v9.2.0:

```console
% node example.js --identifier -a75fe-353 --append AAA file1 file2
{ identifier: '-a75fe-353', append: 'AAA' }
[ 'file1', 'file2' ]
```

If the user forgets the identifier option-argument before the command-arguments then we can't detect a problem, with or without this PR:
```
% node example.js --identifier file1 file2
{ identifier: 'file1' }
[ 'file2' ]
```

With this PR, if the user forgets the identifier option-argument before a recognised option, then they get an error:
```
% node example.js --identifier --append AAA file1 file2 
error: option '-i, --identifier <id>' argument missing
```

But with this PR and aggressive short flag detection, they also get an error for what was an intended call:
```
% node example.js --identifier -a75fe-353 --append AAA file1 file2
error: option '-i, --identifier <id>' argument missing
```

There is a trade-off between detecting actual user errors, and false positives where the user did everything correctly but due to a _possible_ error we reject the option-argument. 

When the option-argument is a fully spelled out long option, like `--append`, it is fairly likely to be an error, and fairly obvious to the user why Commander might reject it for safety.

When the option-argument just starts with a short option, like `-a75fe-353`, I think it is less certain it is an error, and less obvious to the user why that might be rejected when actually intended. Basically we are rejecting (some) option-arguments that start with a dash.

shadowspawn:
(Full disclosure, I am still not sure about changing the current "greedy" behaviour. But considering the possibilities.)

rucker:
Thanks for the detailed example. I do see an issue: while I am using `_findOption` to check whether the option-argument is a known option to avoid false positives, but it's possible an option-argument beginning with `-` (which as you say _should_ be allowed) could also match another known option, e.g. given an option `-c --cheese`, passing `--cheese -cheddar` would trigger a false positive because the leading `-c` in the option-argument matches the short option.

I will continue to look at ways to be a little more "smart" without being too aggressive with option-argument inspection.

shadowspawn:
I will close this as inactive, but can reopen if needed.

PR #1721, opened 2022-04-20
Add string[] to options defaultValue type

everett1992:
# Pull Request

## Problem

The default value for a variadic option should be an array but the type requires a boolean or string.
If you pass a string as the default value to a variadic option the default value will be a string and you'll have to handle string or string[] when using the options. If you pass an array you only have to handle arrays, and you can pass a multi-valued default.

## Solution

I added string[] to option's defaultValue.

## ChangeLog

Add string[] to options defaultValue type.

### DIFF
--- a/typings/index.d.ts
+++ b/typings/index.d.ts
@@ -523,10 +523,10 @@ export class Command {
    *
    * @returns `this` command for chaining
    */
-  option(flags: string, description?: string, defaultValue?: string | boolean): this;
+  option(flags: string, description?: string, defaultValue?: string | boolean | string[]): this;
   option<T>(flags: string, description: string, fn: (value: string, previous: T) => T, defaultValue?: T): this;
   /** @deprecated since v7, instead use choices or a custom function */
-  option(flags: string, description: string, regexp: RegExp, defaultValue?: string | boolean): this;
+  option(flags: string, description: string, regexp: RegExp, defaultValue?: string | boolean | string[]): this;
 
   /**
    * Define a required option, which must have a value after parsing. This usually means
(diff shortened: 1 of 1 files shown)


shadowspawn:
The problem is that the `.option()` method is overloaded. By using `'unknown` for the first overload, it will match any argument and mask the next two overloads.

My recollection is in theory the default value can be anything, but in practice it is usually a `string` or `boolean`, or the type returned by the coerce function. However, we might have added variadic after that work, so maybe adding `string[]` does make sense. (I'll check the timing of those features.)

```
option(flags: string, description?: string, defaultValue?: unknown): this;
option<T>(flags: string, description: string, fn: (value: string, previous: T) => T, defaultValue?: T): this;
option(flags: string, description: string, regexp: RegExp, defaultValue?: unknown): this;
```

shadowspawn:
Related: https://stackoverflow.com/questions/71846275/node-js-commander-default-for-variadic-option

everett1992:
that makes sense, I didn't consider the overlaps between override functions. I updated the PR to add string[].

shadowspawn:
For historical interest, added variadic after the tighter typing was introduced:
- added typings in #1119 on  9 Jan 2020
- added variadic in #1250 on 1 Jun 2020

shadowspawn:
It might be next week before I have time to look into the constraints, but I think `string[]` is probably what we want.

shadowspawn:
Some extra information for future readers. The default value can technically be any type, but in practice is almost always:
- string for option with value
- boolean for boolean option
- array of strings for variadic option (added in this PR)
- same as type returned by custom option processing function (coercion function)

If we use `any` for the type for default argument then get no type checking, so we specify just the expected types to provide help for the normal cases.

If needed, the author can specify some other type without using an explicit coercion by using `Option.default()`:

```
program.addOption(new Option('-e, --example <value>').default({ very: 'tricky' });
```

shadowspawn (approved):

abetomo (approved):

shadowspawn:
Included in Commander v9.3.0. Thanks @everett1992

PR #1722, opened 2022-04-21
ci: add nodejs18 as it is released

abetomo:
# Pull Request

## Solution

* nodejs18 is released.
* Drop nodejs17 as it will be eol soon.


shadowspawn (approved):
👍

PR #1724, opened 2022-05-08
Add Option.implies()

shadowspawn:
# Pull Request

## Problem

See: #1358

## Solution

Add support for supplying a hash of implied option values. Add implementation support to have separate implied options for positive and negative dual options.

Details:
- a default value does not trigger implied values
- support all option types for source option, and for implied option
- the implied value overrides a default value, but not other option sources
- a dual positive and negative option have separate implied options
- not adding to help (leaving it up client to describe if desired)

Side notes:
- switched to `console` for code blocks with mixed commands and output for nicer display in README
- use `sh` rather than `bash` for code blocks which don't show output

## ChangeLog

- add `Option.implies()` to set other option values when the option is specified

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -57,7 +57,7 @@ For information about terms used in this document see: [terminology](./docs/term
 
 ## Installation
 
-```bash
+```sh
 npm install commander
 ```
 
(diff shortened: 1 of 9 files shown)


abetomo (approved):
LGTM

shadowspawn:
I have added extra code to support separate treatment of positive and negative dual options based on the stored option values. (In future, this might be used to expand the `.conflicts()` support in similar way.)

shadowspawn:
Back to draft. Improving examples, and looking into interaction with `.conflicts()` and `.requiredOption()`.

shadowspawn:
Sorted again!

abetomo (approved):
awesome!

PR #1725, opened 2022-05-13
change option default value type to unknown

wwwenjie:
## Problem
current the type of option default value is `string | boolean`, which will cause warning in IDE since it can be number or other.

## Solution
this PR change the type from `string | boolean` to `unknown` to make sure default value can accpet other types.

## ChangeLog
change option default value type to unknown


shadowspawn:
See also: #1721

shadowspawn:
You are right the default argumentment can technically be anything, _but_ the `.option()` method is overloaded. Changing the default argument type to `unknown` (or `any`) means basically no type checking at all after the flags and description.

The current definitions (after #1721) are:
```
option(flags: string, description?: string, defaultValue?: string | boolean): this;
option<T>(flags: string, description: string, fn: (value: string, previous: T) => T, defaultValue?: T): this;
/** @deprecated since v7, instead use choices or a custom function */
option(flags: string, description: string, regexp: RegExp, defaultValue?: string | boolean): this;
```

Example calls are:
```
.option('-a`, 'no default of coerce')
.option('-b <value>', 'some default', 'xyz')
.option('-c <value>', 'custom processing', parseFloat)
.option('-d <value>, 'custom processing and default', parseFloat, 1.23);
```

shadowspawn:
What is your use case that currently causes a type warning?

wwwenjie:
Thanks for pointing out. That make since. Cuurent I'm using a default value of number. I'm not sure if we need add the number type to default values, i think it depends on the uasge of uesrs.

```
.option('-d, --deep <length>', 'length of cards for selecting', 5)
```

warning: Argument type number is not assignable to parameter type string | boolean | undefined   Type number is not assignable to type string | boolean     Type number is not assignable to type boolean

wwwenjie:
maybe use custom processing is a better way...

```
.option('-d, --deep <length>', 'length of cards for selecting', parseInt, 5)
```

shadowspawn:
Yes. (I was typing up a longer answer when you suggested it yourself! 😄 )

Without custom processing, you would need to deal with mixed types in your code. 

```bash
cards # number 5 from default
cards --deep=5 # string '5' from CLI
```

With the custom processing function, Commander can warn you if the default value does not match the return type of the coerce function.

Note: You can not use `parseInt` directly because it takes 2 arguments and Commander passes two arguments to allow "reduce" type functions. You need to wrap the call and pass `parseInt` just the value. There is an example in the README.

When you digest all this, was the warning actually useful in this case?

shadowspawn:
Good. Thanks for confirming. 😄

Issue #1506, opened 2021-04-20
Show expanded help


YuriGal:
Is there a way to display short descriptions when global help command is called, e.g.
```
myprogram help

# Commands:
#  command1 <parameter>  description 1 goes here
#  command2 <parameter>  description 2 goes here   
```

  [1]: https://www.npmjs.com/package/commander
but also have ability to show expanded help when help for specific command is called:
```
myprogram help command1

# command1 is used for...
#
# ... detailed description
#
# ...

shadowspawn:
There is not currently support for setting different descriptions for the two use cases, but there is some support for providing a custom solution yourself. You can customise `commandDescription()` and `subcommandDescription()` using `.configureHelp()` or a Help subclass.

(You can also add extra text to the specific command help with `.addHelpText()`, but that is different than what you were asking.)

Related: #898 #1291

YuriGal:
Thanks for the prompt response and pointing me to the right direction!

shadowspawn:
Opened a PR to add `.summary()` for a short summary to use when listing subcommands.

Issue #1469, opened 2021-02-07
Where should I put long help text?


Niryo:
I want to add some long help text, that tells in details how to use some command. 
The `description` prop should be short (because it's being printed from out side the command), and I see that the "usage" support is not the correct place for such a long help message...
what prop should I use?

shadowspawn:
Have a look at the the README section on custom help which covers `.addHelpText()`.

- https://github.com/tj/commander.js#custom-help

Related issues: #898 #1291

Niryo:
thanks, it solved my problem. I think it worth considering adding an explicit prop for this, something like `.extraInfo()` or `.detailedInfo()`. The current api is not intuitive.

shadowspawn:
An answer was provided, and no further activity in a month. Closing this as resolved.

Feel free to open a new issue if it comes up again, with new information and renewed interest.

shadowspawn:
Opened a PR to add `.summary()` for a short summary to use when listing subcommands.

Niryo:
Amazing! Thanks

Issue #1291, opened 2020-07-03
Long command descriptions


Lakitna:
I'm in a situation where I need less technical people to use my code. A CLI works great for me here, but I think it would be good to add a bunch of documentation to it.

Which leads me to what looks to be a limitation of the current `commander`. I want the help text to do something like this:

## Expected

```plain
> my-app --help
Usage: my-app <command> [options]

Options:
  -V, --version           output the version number
  -h, --help              display help for command

Commands:
  execute [options]       Execute a very interesting command.
  help [command]          display help for command
```

```plain
> my-app execute --help
Usage: my-app execute [options]

Execute a very interesting command. Lorem ipsum dolor sit amet, consectetur adipiscing elit.
Curabitur tristique interdum ullamcorper. Mauris malesuada viverra mattis. Donec aliquet, ligula nec
placerat mollis, purus est rutrum leo, a dapibus mi tortor quis turpis. Cras non commodo tortor.
Etiam laoreet dui risus. Curabitur lacus purus, euismod vitae odio quis, tincidunt posuere elit.
Mauris gravida tempus mi, sit amet condimentum lectus eleifend vel. Nam posuere quam vel lorem
mattis, ut tristique eros pretium. Curabitur eget lobortis libero. Integer leo urna, luctus nec
turpis et, consequat imperdiet elit. Nunc eu quam id lectus vulputate ultricies non eu dolor.
Aliquam luctus vulputate feugiat. Proin consectetur at lorem at congue. Ut dolor quam, dapibus et
arcu eu, porttitor pharetra turpis.

Options:
  -V, --version           output the version number
  -h, --help              display help for command

Commands:
  help [command]          display help for command
```

## Current

When I simply add all the text to `.description()` it will also render the full text in `> my-app --help`. This makes it very hard to read.

```plain
> my-app --help
Usage: my-app <command> [options]

Options:
  -V, --version           output the version number
  -h, --help              display help for command

Commands:
  execute [options]       Execute a very interesting command. Lorem ipsum dolor sit amet, consectetur adipiscing elit.
Curabitur tristique interdum ullamcorper. Mauris malesuada viverra mattis. Donec aliquet, ligula nec
placerat mollis, purus est rutrum leo, a dapibus mi tortor quis turpis. Cras non commodo tortor.
Etiam laoreet dui risus. Curabitur lacus purus, euismod vitae odio quis, tincidunt posuere elit.
[...]

Issue #898, opened 2018-12-23
Add command short description for global help


jbardon:
Hi, 
Commands only have one description for now. 
This description is used for both automatic global and command help.

```
program
    .command('test')
    .description(`Description of the command can be very long but it's displayed entirely in the global help`)
    .action(() => { });
```
**Global help:**
In a terminal the command description wraps to a second line.
This help purpose is only to have a quick overview not an acurate description.

```
$ node index.js  -h
Usage: index [options] [command]

Options:
  -h, --help  output usage information

Commands:
  test        Description of the command can be very long but it's displayed entirely in the global help
```
**Command help:**

```
$ node index.js  test -h
Usage: test [options]

Description of the command can be very long but it's displayed entirely in the global help

Options:
  -h, --help  output usage information
```

I think it might be useful to provide a short description for the global help. The current description would be only used in the command help to provide many details.

What do you think?

shadowspawn:
There is support for custom command help so you can expand on the short description like this: 

```js
program
    .command('test')
    .description('summary description for global help')
    .action(() => { })
    .on("--help", () => {
        console.log();
        console.log("Description of the command can be very long");
        console.log("and just displayed in the command help");
    });

program.parse(process.argv);
```

```bash
$ node . test --help
Usage: test [options]

summary description for global help

Options:
  -h, --help  output usage information

Description of the command can be very long
and just displayed in the command help
```

jbardon:
Thanks for your answer. 
Using description and --help event does the trick.

The description is still displayed on both the global help and on top of the command help.
And the --help texts appear on the bottom of the command help to provide more details.

shadowspawn:
Opened a PR to add `.summary()` for a short summary to use when listing subcommands.

PR #1726, opened 2022-05-18
Add optional summary for short form of subcommand description.

shadowspawn:
# Pull Request

## Problem

The same description is used in the compact listing of subcommands, as in the full help for a subcommand. In the former case a one-liner makes sense, but may want multiple lines in the full command description.

Related:  #898 #1291 #1469  #1506 

A work-around is to add a long description after the built-in help in `.addHelpText('after', longDescription)`.

## Solution

Add optional `.summary()`.

For backwards compatibility, the description is used for the subcommand listing if the summary is missing. So by default there is no change in existing programs.

## ChangeLog

- Add `.summary()` for a short summary to use instead of description when listing subcommands in help


abetomo (approved):
👍

PR #1729, opened 2022-05-21
Add test for Option passed to wrong routines

shadowspawn:
# Pull Request

## Problem

Missing test coverage for throw if pass `Option` to `.option()` or `.requiredOption()`.

## Solution

Add tests.


abetomo (approved):
👍

PR #1730, opened 2022-05-22
Add more tests for implies

shadowspawn:
# Pull Request

## Problem

I remembered that `.conficts()` did not work on program when called subcommand, and no test for that case for `.implies()`.

## Solution

Added test for implies on program when subcommand called, and a couple of other behaviours too.


abetomo (approved):
Great addition to the tests!

PR #1732, opened 2022-05-24
chore: Included githubactions in the dependabot config

naveensrinivasan:
This should help with keeping the GitHub actions updated on new releases. This will also help with keeping it secure.

Dependabot helps in keeping the supply chain secure https://docs.github.com/en/code-security/dependabot

GitHub actions up to date https://docs.github.com/en/code-security/dependabot/working-with-dependabot/keeping-your-actions-up-to-date-with-dependabot

https://github.com/ossf/scorecard/blob/main/docs/checks.md#dependency-update-tool
Signed-off-by: naveen <172697+naveensrinivasan@users.noreply.github.com>

PR #1737, opened 2022-05-28
Add target-branch to dependabot config

shadowspawn:
#1732 immediately triggered three action update PR! Noisy since we haven't been updating, but probably good.  Got some reading of release notes to do...

However, we prefer PR against `develop`. Add config.


abetomo (approved):
👍

PR #1740, opened 2022-05-29
Add develop branch to codeql scanning for earlier warnings

shadowspawn:
I noticed the `branches` when looking at the version bumps, and we probably want warnings when things are targeting `develop` too for earlier warning about issues.


abetomo (approved):
That’s nice.

Issue #1753, opened 2022-06-14
Feature discussion: Inherit options in subcommands


arve0:
Hi! Thanks for commander 🙌
Is commander considered feature complete? Any chance for a *inherit options* feature? Or always populating options in subcommands when same is also defined in parent commands?

I find it hard to create a hierarchy of commands that share common options. I also find the "options are global behavior" unintuitive. Especially when you are unaware of any parent command using the same option.

I've read issues like #1750, #1739, #1616 and #1229.

Take this example, where `print` and `print pdf` would have very different action-implementations:
```js
#!/usr/bin/env node
const { Command } = require('commander');

class MyRootCommand extends Command {
  createCommand(name) {
    const cmd = new Command(name);
    cmd.option('-v, --verbose', 'use verbose logging');
    return cmd;
  }
}

const program = new MyRootCommand();

const print = program.command('print')
  .option('--a4', 'Use A4 sized paper')
  .action((options) => {
    console.log('print options: %O', options);
  });

print.command('pdf')
  .option('--a4', 'Use A4 sized paper')
  .action((options) => {
    console.log('print pdf options: %O', options);
  });

program.parse();
```

The help suggests that `--a4` is available to both "print" and "print pdf":
```sh
> ./common-options.js print --help
Usage: common-options print [options] [command]

Options:
  -v, --verbose  use verbose logging
  --a4           Use A4 sized paper
  -h, --help     display help for command

Commands:
  pdf [options]
```
```sh
> ./common-options.js print pdf --help
Usage: common-options print pdf [options]

Options:
  --a4        Use A4 sized paper
  -h, --help  display help for command
```

Though, it is not:
```sh
> ./common-options.js print --a4
print options: { a4: true }

> ./common-options.js print pdf --a4
print pdf options: {}
```

Solvable through `optsWithGlobals` or splitting into `print-printer` and `print-pdf` commands. Maybe also with splitting `print pdf` to separate executable?

Anyhow, would be nice if options was set in subcommands when same options are defined longer up in the hierarchy.

shadowspawn:
Thanks for the detailed description.

The feature which is a good match for most of your description is `.enablePositionalOptions()`. This turns off the "global" behaviour where the option can appear before or after a subcommand.

> Especially when you are unaware of any parent command using the same option.

With positional options, options after the subcommand belong to the subcommand.

> Take this example, where print and print pdf would have very different action-implementations

[...]

PR #1754, opened 2022-06-16
global options example: pass arguments to sub-subcommands too

arve0:
## Problem
Was hard to understand why sub-subcommand did not get option, even though `--help` shows the option. I read the example, others will probably too. Expand it to also add options to sub-subcommands.

See issue #1753.

## Solution
Added sub-subcommand to global-options example.

### DIFF
--- a/examples/global-options.js
+++ b/examples/global-options.js
@@ -9,28 +9,38 @@
 // (A different pattern for a "global" option is to add it to the root command, rather
 // than to the subcommand. That is not shown here.)
 
-// const { Command } = require('commander'); // (normal include)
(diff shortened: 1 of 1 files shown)


shadowspawn:
Thanks for the suggestion, but I don't want repeated ("inherited") options in nested subcommands in this example file. That is a complex pattern with some drawbacks.

I wrote this example so that options are only added in "leaf" commands, so options are not repeated at different levels of the hierarchy. To keep the code simple I limited the subcommands to just one level deep.

Note the name of the Command subclass (`MyRootCommand`), and also the comment on line 6:

```
// The code in this example assumes there is just one level of subcommands.
```

PR #1756, opened 2022-06-18
Export InvalidOptionArgumentError in esm

everett1992:
After upgrading a typescript project to native es modules  imports for
`InvalidOptionArgumentError` were undefined at runtime but 
defined in typescript as an alias to InvalidArgumentError.

```
# test.ts
import { InvalidOptionArgumentError } from 'commander'
//       ^ No type error here, typescript thinks InvalidOptionArgumentError should exist

assert(InvalidOptionArgumentError != null); // Yet will throw an error
```

A quick test showed that there was a difference between the commonjs and
ecmascript module. The issue wasn't limited to typescript.

```
# test.cjs
const { InvalidOptionArgumentError } = require('commander')
assert(InvalidOptionArgumentError != null);
```

```
# test.mjs
import { InvalidOptionArgumentError } from 'commander'
assert(InvalidOptionArgumentError != null); // This will throw an error
```

This commit fixes the issue by exporting InvalidOptionArgumentError from
the esm entry point. Alternatively we could remove
InvalidOptionArgumentError from typings, I'm happy as long as build and
runtime match.

### DIFF
--- a/esm.mjs
+++ b/esm.mjs
@@ -8,7 +8,7 @@ export const {
   createOption,
   CommanderError,
   InvalidArgumentError,
-  InvalidArgumentError: InvalidOptionArgumentError,
+InvalidOptionArgumentError, // deprecated old name
   Command,
   Argument,
   Option,
--- a/esm.mjs
+++ b/esm.mjs
@@ -8,7 +8,7 @@ export const {
   createOption,
   CommanderError,
   InvalidArgumentError,
-InvalidOptionArgumentError, // deprecated old name
+  InvalidOptionArgumentError, // deprecated old name
   Command,
   Argument,
   Option,
--- a/esm.mjs
+++ b/esm.mjs
@@ -8,6 +8,7 @@ export const {
   createOption,
   CommanderError,
   InvalidArgumentError,
+  InvalidArgumentError: InvalidOptionArgumentError,
   Command,
   Argument,
   Option,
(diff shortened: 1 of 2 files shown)


shadowspawn:
The esm export got dropped/renamed in #1508 when `InvalidOptionArgumentError` got deprecated. I might have thought it wouldn't matter in the "new" esm code, but as you discovered it is an issue when switching from cjs to esm runtime.

esm.mjs
```diff
@@ -8,6 +8,7 @@ export const {
   createOption,
   CommanderError,
   InvalidArgumentError,
+  InvalidArgumentError: InvalidOptionArgumentError,
```
shadowspawn, line 11:
```suggestion
InvalidOptionArgumentError, // deprecated old name
```

Some small suggestions. I would prefer the esm just uses the cjs export, so when the export is removed from the cjs it breaks the esm. And a comment that it is deprecated.

The TypeScript uses an alias because with the stronger typing it seemed more appropriate. I don't see that as making a difference in esm?

shadowspawn (changes requested):

shadowspawn:
Looks like there was an indentation boobytrap in the suggestion I made, sorry about that!

shadowspawn (approved):
LGTM (just an indentation issue).

shadowspawn:
(Pushed indentation fix.)

abetomo (approved):
👍

shadowspawn:
Thanks @everett1992

shadowspawn:
Released in Commander v9.4.0

Issue #1137, opened 2020-01-02
Pass arguments to executable sub-commands


gitoutthere:
I am implementing pass parameters by json file allowing users to provide -i and path to a json file which has options and their values. I have 10 sub-commands and I want this parameters from json file to be passed to sub-command file to be utilised.

I have currently achieved this by putting content of input file in process.env.inputFIle and within each sub-command file I unpack this var and shove it inside process.argv

funApp.js
```
program.version('2.1')
.command('add' [parameters], ' adding stuff')
.command('update' [parameters], ' update stuff')
.command('remove' [parameters], ' remove stuff')
[... etc]
.option('-i, --input <value>',' json param file')
[...]
program.on('option:input', function(inputFile) {
  const fileData = JSON.parse(fs.readFileSync(inputFile,'utf8'))
// Somehow pass these args to the funApp-add.js as args?!
})
```
That way my CLI could be `funApp add -i params.json --overrideArg value --newArg value`

shadowspawn:
The open issue is: #563

I had the same idea as you, passing context using environment variables: https://github.com/tj/commander.js/issues/563#issuecomment-520112985

You could perhaps treat environment variables as a first class source of options, like:
https://github.com/tj/commander.js/issues/1135#issuecomment-570016609

gitoutthere:
Thank you for the prompt reply. Funny enough, I reviewed all those links before I posted mine. I am not fan of using env var as default value on .options since I would have less control (if env is not set and other issues)

I tried to insert the file arguments into rawArgs and other vars I thought relevant, I was hoping they get passed into subcommand as argv same way CLI options values gets passed to sub-commands.

Perhaps I can ask my question differently, can I manipulate the arguments before it is passed to the sub-command file? In my above example, can do manipulate values of --overrideArg and --newArg before they are passed to FunApp-add.js as argv?

shadowspawn:
Good question. There is not currently support for modifying arguments before (for) passing to subcommands.

evg656e:
I ran into the somewhat similar problem.

The use case is to have common options (e.g. verbosity level, config path etc.) for the subcommands. I found some workarounds, e.g. to create option setter function and call this function for each subcommand, or to use `.on('option:` handler and store common options in global variable, but all of these is quite messy.

I would like that options set before first command to be pass through somehow:
```javascript
program
  // common options for all subcommands
  .option('-v, --verbose', 'show more info')
  .option('-c, --config <path>', 'path to config file')
[...]

Issue #563, opened 2016-08-22
How to get options from sub command exes


kahunacohen:
In my 'foo' sub command file (Eg. `pm-foo.js`) how do I get hold of the options passed?

`pm.js`:

```
...
program
  .version('0.0.1')
  .command('foo [name]', 'desc' {a:'b'})
  .parse(process.argv);
```

I tried passing parameters after the description parameter, but how do you get hold of them in the `pm-foo.js` file? I tried printing out `name` in `pm-foo.js` too. Any help is appreciated.

sgen:
Commander passes the remaining arguments to a child process running the sub command script. You can access them with `process.argv`

kahunacohen:
These are just the remaining arguments on the command-line. How do I get a hold of extra arguments passed after the `description` argument of `command()`? In `pm-foo.js` if I print out `process.argv` I get `['node', PATH_TO_PM_FOO]`. No `{'a': 'b'}`.

sgen:
Commander.command() doesn't actually call a subcommand. It just defines one.

```
program
  .version('0.0.1')
  .command('foo [name]', 'desc' {a:'b'})
  .parse(process.argv);
``` 

is defining a subcommand called `foo`. The `[name]` argument is just for the help message. You're also missing a comma in your example. As for the options the only options mentioned in the docs are `opts.noHelp` and `opts.isDefault`, [no other arguments are passed](https://github.com/tj/commander.js/blob/master/index.js#L158).

Your `pm-foo.js file` might look like this:

```
var subCommand = require('commander');

subCommand
  .option('-a, --an-option')
  .parse(process.argv);


var args = subCommand.args.slice(2);
var name = args.pop();

if (!name) // ...

if(!subCommand.anOption) // ...
```

b4dnewz:
the main options are not forwarded to the subcommands for example:

pm.js
```js
program
  .name('cli')
  .option('-k, --key <apikey>')
```

pm-foo.js
```js
program
  .command('list')
  ..action(options => {
    console.log(program);
    console.log(options);
    console.log(program.options);
    console.log(program.apikey);
})
```

`cli -k 'ABCD' list `

is possible to get the apikey from the "root" command?

PaulBeaudet:
[...]

PR #1763, opened 2022-06-27
feat: add preSubcommand hook

### LINKED ISSUE #1762: Need a hook or event to set process.env before `_parseOptionsEnv`
Hallo,
Is there any way to set environment variables before `_parseOptionsEnv` function? 
my use case is to read `.env` file via `dotenv` and set environment variables before all subcommands are executed.

The dotenv file can be specified, and I am use `enablePositionalOptions` to process options.
The issue is `preAction` hook are emit after `_parseOptionsEnv` function, this will cause `new Option().env('SOME_ENV')` not to perform as expected.

```typescript
program.addOption(new Option('--env <env>', 'env file').default('.env'))
  .hook('preAction', async () => {
    const { env } = program.opts();
    dotenv.config({ path: env, override: true });
  });
```

I also tried the `option:env` event, but it doesn't solve the problem perfectly either.

I searched relevant issues and stackoverflow as much as possible, but no solution found.

Thanks.

hungtcs:
# Pull Request

## Problem

Read `--env` options and process it before subcommand parsed.

Closes #1762

## Solution

Add `preSubcommand` hook to handle some operations that take effect for all subcommands.


lib/command.js
```diff
@@ -1054,11 +1054,16 @@ Expecting one of '${allowedValues.join("', '")}'`);
     const subCommand = this._findCommand(commandName);
     if (!subCommand) this.help({ error: true });
 
-    if (subCommand._executableHandler) {
-      this._executeSubCommand(subCommand, operands.concat(unknown));
-    } else {
-      return subCommand._parseCommand(operands, unknown);
-    }
+    let actionResult;
+    actionResult = this._chainOrCallHooks(actionResult, 'preSubCommand');
```
shadowspawn, line 1058: I think this will call the `preSubCommand` for all parent commands as well. So when there are nested commands, the same hook will get called multiple times.

shadowspawn: This could perhaps get handled by the callback looking at the parameters to tell what the context is, but my first thought is that the callback is only called once at each level by the current command before its subcommand is called. I'll need to think about that. The existing hooks only get called from one location and I haven't thought through this new pattern.

shadowspawn:
An example program for thinking about expected callbacks. This has deeply nested commands. Should the program callback get called just once before the direct subcommand, or multiple times from each level of subcommand?

```js
program
  .name('program')
  .hook('preSubCommand', (thisCommand, subcommand) => {
    console.log(`pre ${thisCommand.name()}/${subcommand.name()}}`)
  });

program
  .command('first')
  .command('second')
  .command('third');

program.parse(['first', 'second', 'third'], { from: 'user' });
```

Current behaviour at time of writing:
```console
 % node index.js
pre program/first}
pre program/second}
pre program/third}
```

hungtcs:
https://github.com/tj/commander.js/blob/3ae30a291a777ad1f748d86f49903579277f9ea0/lib/command.js#L1204

`operands` are consumed when parsing each level of menu, when `operands` is empty, this indicates that it was the final subcommand to be executed, I take this as the basis for judgment.

shadowspawn:
> `operands` are consumed when parsing each level of menu, when `operands` is empty, this indicates that it was the final subcommand to be executed, I take this as the basis for judgment.

That test won't work. The operands includes command-arguments too. Like: `program first second third test.txt`

shadowspawn:
I was wondering about only calling the hook for the immediate subcommands like:

```js
  _chainOrCallSubcommandHook(promise, subCommand, event) {
    let result = promise;
    if (this._lifeCycleHooks[event] !== undefined) {
      this._lifeCycleHooks[event].forEach((hook) => {
        result = this._chainOrCall(result, () => {
          return hook(this, subCommand);
        });
      });
    }
    return result;
  }
```

```
    hookResult = this._chainOrCallSubcommandHook(actionResult, subCommand, 'preSubCommand');
```

For the call:
```
program.parse(['first', 'second', 'third'], { from: 'user' });
```

The output would be:
```
 % node index.js
pre program/first
```

This feels quite natural for subcommand processing, but is quite different than the way the action hooks work!

hungtcs:
@shadowspawn  I feel like I'm in the wrong process, what you said makes sense.

hungtcs:
> That test won't work. The operands includes command-arguments too. Like: program first second third test.txt

I totally agree with https://github.com/tj/commander.js/pull/1763#issuecomment-1168612994

But before that, I additionally realized that I can use `subCommand.commands.length` to judge whether it is the actual command executed.

shadowspawn:
The hook event name should be `preSubcommand` rather than `preSubCommand`, as the README refers to "subcommands" rather than "sub-commands".

shadowspawn:
These two previous issues ask about passing program options into executable subcommands: #563 #1137

This PR is not a direct solution, but the preSubcommand hooks does allow modifying the environment variables before calling subcommand.

(I don't think the parsing is currently set up to allow directly modifying the remaining arguments to be parsed, which was also mentioned as a possible approach. Future middleware concept!)

typings/index.test-d.ts
```diff
@@ -90,6 +90,12 @@ expectType<commander.Command>(program.hook('preAction', (thisCommand, actionComm
   expectType<commander.Command>(thisCommand);
   expectType<commander.Command>(actionCommand);
 }));
+expectType<commander.Command>(program.hook('preSubcommand', () => {}));
+expectType<commander.Command>(program.hook('preSubcommand', (thisCommand, actionCommand) => {
```
shadowspawn, line 94: Second parameter here is named subcommand (rather than actionCommand).

shadowspawn:
How about a table?

---
The supported events are:

| event name | when hook called | callback parameters |
| :-- | :-- | :-- |
| `preAction`, `postAction` |  before/after action handler for this command and its nested subcommands |   `(thisCommand, actionCommand)` |
| `preSubcommand` | before parsing direct subcommand  | `(thisCommand, subcommand)` |

---
Raw markdown:
```
The supported events are:

| event name | when hook called | callback parameters |
| :-- | :-- | :-- |
| `preAction`, `postAction` |  before/after action handler for this command and its nested subcommands |   `(thisCommand, actionCommand)` |
| `preSubcommand` | before parsing direct subcommand  | `(thisCommand, subcommand)` |
```

hungtcs:
Yes, it looks much clearer

shadowspawn:
Here is a big rework of the existing hook example file to include all the events, with just one subcommand, and one example call each. (Using port idea from your example.)

```js
#!/usr/bin/env node

// const { Command, Option } = require('commander'); // (normal include)
const { Command, Option } = require('../'); // include commander in git clone of commander repo

// This example shows using some hooks for life cycle events.

const timeLabel = 'command duration';
program
  .option('--profile', 'show how long command takes')
  .hook('preAction', (thisCommand) => {
    if (thisCommand.opts().profile) {
      console.time(timeLabel);
    }
  })
  .hook('postAction', (thisCommand) => {
    if (thisCommand.opts().profile) {
      console.timeEnd(timeLabel);
    }
  });

program
  .option('--trace', 'display trace statements for commands')
  .hook('preAction', (thisCommand, actionCommand) => {
    if (thisCommand.opts().trace) {
      console.log('>>>>');
      console.log(`About to call action handler for subcommand: ${actionCommand.name()}`);
      console.log('arguments: %O', actionCommand.args);
      console.log('options: %o', actionCommand.opts());
      console.log('<<<<');
    }
  });

program
  .option('--env <filename>', 'specify environment file')
  .hook('preSubcommand', (thisCommand, subcommand) => {
    if (thisCommand.opts().env) {
      // One use case for this hook is modifying environment variables before
      // parsing the subcommand, say by reading .env file.
      console.log(`Reading ${thisCommand.opts().env}...`);
      process.env.PORT = 80;
      console.log(`About to call subcommand: ${subcommand.name()}`);
    }
  });

program.command('start')
  .argument('[script]', 'script name', 'server.js')
  .option('-d, --delay <seconds>', 'how long to delay before starting')
  .addOption(new Option('-p, --port <number>', 'port number').default(8080).env('PORT'))
  .action(async(script, options) => {
    if (options.delay) {
      await new Promise(resolve => setTimeout(resolve, parseInt(options.delay) * 1000));
    }
    console.log(`Starting ${script} on port ${options.port}`);
  });

// Some of the hooks or actions are async, so call parseAsync rather than parse.
program.parseAsync().then(() => {});

// Try the following:
//    node hook.js start
//    node hook.js --trace start --port 9000 test.js
//    node hook.js --profile start --delay 5
//    node hook.js --env=production.env start
```

Readme.md
```diff
@@ -701,7 +701,10 @@ If the program is designed to be installed globally, make sure the executables h
 
 You can add callback hooks to a command for life cycle events.
 
-Example file: [hook.js](./examples/hook.js)
+Example file:
```
shadowspawn, line 704: Back to one example file

Readme_zh-CN.md
```diff
@@ -678,7 +678,10 @@ program.parse(process.argv);
 
 可以在命令的生命周期事件上设置回调函数。
 
-示例代码：[hook.js](./examples/hook.js)
+示例代码：
```
shadowspawn, line 681: Back to one example file

shadowspawn (approved):
Just a couple of minor changes to README references to example file to do.

LGTM

hungtcs:
Thanks for your help 😃  @shadowspawn

abetomo (approved):
Awsome!

shadowspawn:
Released in Commander v9.4.0

PR #1767, opened 2022-07-12
Update dependencies

shadowspawn:
A long wait, but https://github.com/standard/eslint-config-standard-with-typescript has been updated for eslint@8. Updated all the related eslint dependencies, and jest, and tsd+typescript.

Only one small lint change to fix new lint warning.


abetomo (approved):
👍

PR #1771, opened 2022-07-29
Option-arguments after the double dash being ignored

matibat:
# Pull Request

### Overview

When we use a variadic option argument, it will not accept literal arguments starting with dash (`-`, e.g. `-abcdefg-`), that's expected because we consider those arguments to be option arguments. In order to pass a non-option argument (one starting with dash) we have to make use of double dash (`--`) before it (such as `... --variadic-option arg1 arg2 -- -non-option-argument arg4 ...`). See [Utility Syntax Guidelines, Guideline 10](https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap12.html).

## Problem

Option-arguments placed after the first double dash will not be read, but they should be.

So, given this example:

```sh
$ my-program my-command --my-variadic-option-argument "value1" "value number 2" -- "-my-non-option-argument" "one more argument"
```

The expected behavior is:

```ts
program.parse(<the given example argv>);
const myValues = program.opts().myVariadicOptionArgument;
// myValues = ['value1', 'value number 2', '-my-non-option-argument', 'one more argument']
```

But the current behavior is:

```ts
program.parse(<the given example argv>);
const myValues = program.opts().myVariadicOptionArgument;
// myValues = ['value1', 'value number 2']
```

## Solution

**TBD:** needs to be confirmed before proceeding.

## ChangeLog

- Adds a test case for the expected behavior of option-arguments that are placed after the first double dash in the argv.

**TODO**
- Make the parser to do not ignore option-arguments after the first double dash; only stop parsing options.

### DIFF
--- a/tests/options.variadic.test.js
+++ b/tests/options.variadic.test.js
@@ -167,4 +167,15 @@ describe('variadic special cases', () => {
 
     expect(program.options[0].variadic).toBeFalsy();
   });
+
+  test('when option flags arguments includes -- then some literals the argument values includes all of them except --', () => {
+    // This might be used to describe coercion for comma separated values, and is not variadic.
+    const program = new commander.Command();
+    program
+      .option('-a, --arguments [values...]');
+
+    program.parse(['--arguments', '--', 'one', 'two', '-non-option', 'four'], { from: 'user' });
+
+    expect(program.opts().arguments).toEqual(['one', 'two', '-non-option', 'four']);
+  });
 });


tests/options.variadic.test.js
```diff
@@ -167,4 +167,15 @@ describe('variadic special cases', () => {
 
     expect(program.options[0].variadic).toBeFalsy();
   });
+
+  test('when option flags arguments includes -- then some literals the argument values includes all of them except --', () => {
+    // This might be used to describe coercion for comma separated values, and is not variadic.
```
matibat, line 172: Copy/pasta! 🍝 
```suggestion
```

matibat:
Hey team!

I'm pretty sure that the described problem is a valid one, I'd only need some of you to confirm that it is a thing so I'd dig into a solution for it. 

Please feel free to give any feedback or ask questions, I will be happy to answer <3

shadowspawn:
I understand the problem, there is not a way to "escape" an optional option-argument to allow a dash. But I think it would lead to as many problems as it solves to use `--` in the way you suggest.

Currently, I can end the variadic options to add command-arguments:

```
$ my-program my-command --my-variadic-option-argument o1 o2 o3 -- arg1 arg2
```

Making the `--` after the options part of the syntax is a suggested approach to resolve ambiguities in https://github.com/tj/commander.js/blob/master/docs/options-taking-varying-arguments.md#alternative-make

For example:
```
Usage: utility_name clone [options] -- <source> [destination]
```

---

Guideline 10 of the POSIX Utility Conventions says that the arguments following the option-terminator are operands, and not option-arguments which are a separate term.

> Guideline 10:
> The first -- argument that is not an option-argument should be accepted as a delimiter indicating the end of options. Any following arguments should be treated as operands, even if they begin with the '-' character.

matibat:
Thanks so much for the quick response! And also thanks for clarifying that, it makes total sense.

Now,
> there is not a way to "escape" an optional option-argument to allow a dash

Would you say that implementing a custom rule to escape the dash for the specific option is a valid approach? In that case I would document that custom behavior in the help command.

shadowspawn:
Yes, you could have your own guard or escape character convention and document it in your help.

Note, you are currently able to add variadic option-arguments that start with a dash. But you need to specify the option-argument together with the option, or first after the option if it is a required option-argument. For example:

```js

program
  .option('-v, --variadic <values...>');

program.parse();
console.log({ options: program.opts(), args: program.args});
```

```
 $ node index.js --variadic 1 2 --variadic=-3 --variadic -4 5
{ options: { variadic: [ '1', '2', '-3', '-4', '5' ] }, args: [] }
```

matibat:
Those are really nice workarounds!

Thanks for your time, @shadowspawn! 🚀 It was really useful 🙌

PR #1788, opened 2022-08-31
fix option source value

mshima:
# Pull Request

## Problem

'config' option source looks incorrect:
https://github.com/tj/commander.js/blob/82fcb98cc27164a98e0c5f2c6f54621b5bbceef9/lib/command.js#L1618

### DIFF
--- a/lib/command.js
+++ b/lib/command.js
@@ -37,7 +37,7 @@ class Command extends EventEmitter {
     this._scriptPath = null;
     this._name = name || '';
     this._optionValues = {};
-    this._optionValueSources = {}; // default < config < env < cli
+    this._optionValueSources = {}; // default < implied < env < cli
     this._storeOptionsAsProperties = false;
     this._actionHandler = null;
     this._executableHandler = false;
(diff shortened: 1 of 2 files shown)


shadowspawn:
Yes. 

`config` appears in code as available for author use, although not set by Commander. #1584 #1613

`implied` is set by Commander, but wasn't added to the JSDoc or TSDoc listing the known values. #1724

mshima:
Is there any example on how to use setOptionValueWithSource with 'config'?
Something like this?
```
const options = program.option('--config', 'config.js', 'config.js').parse().opts();
if (options.config) {
  const config = require(options.config);
  for (const [opt, value] of Object.entries(config)) {
    program.setValueWithSource(opt, value, 'config');
  }
}
const updatedOptions = program.opts();
```

shadowspawn:
The use case I had in mind was one of the scenarios in #1584

1) configure program (which includes defaults)
2) read and store config
3) `program.parse()`

mshima:
> The use case I had in mind was one of the scenarios in #1584
> 
> 1. configure program (which includes defaults)
> 2. read and store config
> 3. `program.parse()`

The problem of this workflow is that custom config needs to be parsed manually.
Maybe an event like `on('options:parsed', loadConfigurationFiles)`.

shadowspawn:
A big disadvantage of events is the lack of support for async. The life cycle hooks could be better for something which might include async file handling.

PR #1792, opened 2022-09-01
fix(types): update `defaultValue` types

vansergen:
# Description

Update the `defaultValue` type.  Example, 

```typescript
program
  .requiredOption("-p --port <number>", "port number", 8080)
  .option("-n --number <number>", "other number", 123);
```

### DIFF
--- a/typings/index.d.ts
+++ b/typings/index.d.ts
@@ -535,10 +535,10 @@ export class Command {
    *
    * @returns `this` command for chaining
    */
-  option(flags: string, description?: string, defaultValue?: string | boolean | string[]): this;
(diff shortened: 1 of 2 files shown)


shadowspawn:
It is technically possible to have `any` type for a default as far as the implementation is concerned. However, it usually makes more sense for the default type to match the option values coming from the command-line. In your examples the option value will be a number if it comes from the default and a string if it comes from the command-line arguments, so extra care required in the processing.

The TypeScript typings encourage consistent typing whatever the source of the option value. In particular, the typing allow a default of the same type as a custom processing function. For example:

```js
program
  .option("-f --float <number>", "other number", parseFloat, 23);
```

PR #1794, opened 2022-09-04
Add implied to option value sources

shadowspawn:
# Pull Request

## Problem

`implied` is missing from `OptionValueSource` but is set by the `.implies()` feature.

## Solution

Add `implied` to TypeScript and method descriptions.

Add a mention of `config` source in the env description, which is the only place it is used in the code. Reword description to not state a priority ordering, and instead more accurately reflect actual behaviour.

Add undefined to return type of `.getOptionValueSource()`.

## ChangeLog

- fix: TypeScript : add `implied` to `OptionValueSource` for option values set by using `.implies()`
- fix: TypeScript : add undefined to return type of `. getOptionValueSource()`

### DIFF
--- a/lib/command.js
+++ b/lib/command.js
@@ -789,11 +789,11 @@ Expecting one of '${allowedValues.join("', '")}'`);
   }
 
   /**
-   * Store option value and where the value came from.
+    * Store option value and where the value came from.
     *
     * @param {string} key
     * @param {Object} value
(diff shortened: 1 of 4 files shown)


abetomo (approved):
👍

shadowspawn:
Released in Commander v9.4.1

PR #1795, opened 2022-09-04
setOptionValue should clear option source

shadowspawn:
# Pull Request

## Problem

Calling `.setOptionValue()` would leave behind any previous value for option source.

## Solution

Move the implementation of `.setOptionValue()` into `.setOptionValueWithSource()` and call it.

## ChangeLog

- fix: `.setOptionValue()` clears option source


abetomo (approved):
+1

shadowspawn:
Released in Commander v9.4.1

Issue #1779, opened 2022-08-27
requiredOption with choices


firmart:
I already had 
```js
.addOption(new Option("-s, --service <service>", "Service provider to use.")
        .choices(["a", "b"])
        .default("a", "A"))
```
I want to make this option required. There is `.requiredOption` but not as powerful as `.addOption` (i.e. cannot specify `choices`). AFAIK there is no way to specify an option `choices` and make it required at the same time.
An `.addRequiredOption` would seem redundant. Would it be possible to add a `.required()` method to `Option` ?

shadowspawn:
Use `.makeOptionMandatory()`.

Background: the reason for the name change is that "required" inside the Option class has always meant the option has a required option-argument, as opposed to an optional option-argument, or no option-argument for a boolean option. When we added `.requiredOption()` the naming made sense at the Command level, but I felt it was too confusing to use "required" in two different ways in the Option class and used "mandatory" for the new functionality.

shadowspawn:
Note: the runtime test for a required option is that it has a value after parsing. When you make your option required (mandatory) you probably want to drop the default value to make the user specify a value.

shadowspawn:
Related issues: #1495 #1564

firmart:
@shadowspawn Thanks for the quick reply and the remarks. I have checked the code source, but completely missed that method as I kept searching for the word "required" and only found, as you said, required argument. Will keep this issue opened as it is needed to be documented.

shadowspawn:
Added to README in #1797

shadowspawn:
Added mention to README in Commander v9.4.1

Issue #1564, opened 2021-07-09
Add support for required options with choices


Thomas-Smyth:
## Background: Required Options
Currently, to make an option be required one may use the `requiredOption` method.

For example:
```js
program.option('-f, --favourite-cake <name>', 'your favourite cake');
```
Would become:
```js
program.requiredOption('-f, --favourite-cake <name>', 'your favourite cake');
```

## Background: Choices
Currently, to make an option have a set of valid values one may use the snippet `.addOption(new Option(...).choices([...]))`.

For example:
```js
program.option('-f, --favourite-cake <name>', 'your favourite cake');
```
Would become:
```js
program.addOption(
  new Option(
    '-f, --favourite-cake <name>', 
    'your favourite cake'
  ).choices([
    'Chocolate Fudge Cake',
    'Victoria Sponge',
    'Carrot Cake'
  ])
);
```

## Problem
To my knowledge and as far as the documentation is aware, there's currently no way to make an option required and specify a set of valid values.

I propose a new method be added to cater for this:
```js
program.addRequiredOption(
  new Option(
    '-f, --favourite-cake <name>', 
    'your favourite cake'
  ).choices([
    'Chocolate Fudge Cake',
    'Victoria Sponge',
    'Carrot Cake'
  ])
);
```

Alternatively, a new option/method be added to the `Option`  class.

shadowspawn:
It is not included in the README, but there is a method on Option for this: `.makeOptionMandatory()`.

```
program.addOption(
  new Option(
    '-f, --favourite-cake <name>', 
    'your favourite cake'
  ).choices([
    'Chocolate Fudge Cake',
    'Victoria Sponge',
    'Carrot Cake'
  ])
  .makeOptionMandatory()
);
```

We talk about required options, and options with a required argument. When adding the support for the higher level  ".requiredOption()", `Option` already had a property called `required` for the required-argument case. I went with "mandatory" for the terminology in Option to make it distinct.

shadowspawn:
An answer was provided, and no further activity in a month. Closing this as resolved.

Feel free to open a new issue if it comes up again, with new information and renewed interest.

Issue #1495, opened 2021-04-05
add required option


Niryo:
is there any api like `addOption` for adding a required option?

shadowspawn:
No there is not a separate API method, as you use `.addOption()` itself.

```
program
  .addOption(new Option('-r,--required <value>', 'required option').makeOptionMandatory());
```

shadowspawn:
An answer was provided, and no further activity in a month. Closing this as resolved.

Feel free to open a new issue if it comes up again, with new information and renewed interest.

PR #1797, opened 2022-09-08
Add mention of makeOptionMandatory in README

shadowspawn:
# Pull Request

The Option methods are mostly covered as an example.

## Problem

Some people have not found that Option `.makeOptionMandatory()` is the equivalent of Command `.requiredOption()`.

See: #1495, #1564, #1779

## Solution

Add mention in README.

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -330,7 +330,7 @@ For information about possible ambiguous cases, see [options taking varying argu
 
 ### Required option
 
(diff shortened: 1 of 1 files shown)


abetomo (approved):
👍

shadowspawn:
Added mention to README in Commander v9.4.1

Issue #1785, opened 2022-08-30
Best way to wrap description text by terminal width?


cobbzilla:
**this is of minor priority**

When I put in long lines in my description, they get chopped mid-word on "max 80-char width" displays; this looks bad, it's hard to read.

So I limit my description text to max 80 chars per line, but it's weird then to do `-h` and see the options take advantage of the much wider terminal, while the description is hard-coded to "max 80".

How can I display descriptions according to terminal width? Ideally I only put newlines where they make semantic sense, and let another layer handle the wrapping (on words, though, not chars).

I'm hoping there is already a way to do this that I haven't found, but I haven't found it :)

shadowspawn:
The description is not currently being word-wrapped. (I thought it was!)
I think it probably could be.

(The wrapping got added in #956 and was focussed on option and subcommand and argument descriptions.)

shadowspawn:
This is doable now by subclassing the Help class. Perhaps more than you are willing to do, but I wanted to check it was possible!

A trap to be aware of is that no wrapping is done if the text includes a line-feed followed by whitespace, in case the text was manually formatted already. A single line-feed followed by text is fine.

```js
const { Command, Help } = require('commander');

class MyHelp extends Help {
  commandDescription(cmd) {
    const description = cmd.description();
    if (!description) return '';
    const helpWidth = this.helpWidth || 80;
    return this.wrap(description, helpWidth, 0);
  }
}

class MyCommand extends Command {
  createCommand(name) {
    return new MyCommand(name);
  }
  createHelp() {
    return new MyHelp;
  }
}

const program = new MyCommand();

program
  .description(`. . . . . . . . . . . . . . . . . . . . . . . . . . . . . . . . . . . . . . . . . . . . . . . . 
this long line will be wrapped without breaking words across lines if there is enough room to make it worthwhile
and this text is on another line`);

program.parse();
```

```console
$ node index.js -h
Usage: index [options]

. . . . . . . . . . . . . . . . . . . . . . . . . . . . . . . .
. . . . . . . . . . . . . . . .
this long line will be wrapped without breaking words across
lines if there is enough room to make it worthwhile
and this text is on another line

Options:
  -h, --help  display help for command
```

shadowspawn:
Adding wrapping for (full) command description in #1804

shadowspawn:
Released in Commander v10.0.0

PR #1804, opened 2022-09-25
Wrap command description in help

shadowspawn:
# Pull Request

## Problem

The "full" command description is not being wrapped in the help.

See: #1785

## Solution

Add wrapping of full description.

To work nicely with complex multiline descriptions, some subtle changes to the patterns:
- don't block wrapping just because text contains consecutive empty lines (look for pre-indented lines)
- preserve empty lines in description

Being conservative and marked as semver major because changing help may break client unit tests that depend on the exact output, although only a cosmetic change.

## ChangeLog

- add wrapping of command description in help

### DIFF
--- a/lib/help.js
+++ b/lib/help.js
@@ -381,23 +381,27 @@ class Help {
    */
 
   wrap(str, width, indent, minColumnWidth = 40) {
-    // Detect manually wrapped and indented strings by searching for line breaks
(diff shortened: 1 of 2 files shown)


abetomo (approved):
👍

shadowspawn:
Released in Commander v10.0.0

Issue #1801, opened 2022-09-15
Add localization method to Command


gylove1994:
Even though we can use `.configureOutput()` to configure the output to change the help header and error, but it is not convenient for user to change it entirely ( include error message and help header and etc. ) for localization.

I want to add a method to Command like `program.localization('zh_CN')` to change the output template words and error message could be  Simplified Chinese entirely.

If it is passable to add , I will try to make a PR then.

Related Issues: #128 #774 #1608

shadowspawn:
PR related to #128: #131

shadowspawn:
I am interested in localisation, but I think this is a hard problem that will require a lot of code changes and make the code more difficult to write and maintain. So only a maybe, and worth discussing before investing a lot of time in a PR. Somewhat to my surprise, #128 did not get many likes or comments despite being open for six years. (If any readers are interested, please 👍  the opening comment here so we know there is interest!)

I do not have experience with localisation in JavaScript.

1) I have looked at it casually in the past, and wasn't sure how to handle building error messages. The code uses the convenient template literals a lot, like:

```
    const message = `error: too many arguments${forSubcommand}. Expected ${expected} argument${s} but got ${receivedArgs.length}.`;
```

Have you got an idea of how to handle these? (The parameters, not the plural. That is a separate issue on its own.)

In other languages I have used formatting support that allows numbered parameters so translations may change the order of the substituted parameters, like say "too many arguments %1. Expected %2". But I have not found anything native in Node.js that does this.

2) If we add localisation, then we want people to be able to share them!

We did add an organisation recently, so now have somewhere that we could host localisation projects: https://github.com/commander-js

gylove1994:
Sorry for my slow reply.

It does need to change a lot of code to achieve it, so I will be waiting to handle the next step after the discussion result come out.

For the issue which how to handle error messages, I found some discussions here: [javascript-equivalent-to-printf-string-format](https://stackoverflow.com/questions/610406/javascript-equivalent-to-printf-string-format)

shadowspawn:
The keys for the localised strings could be the original text, or an identifier. 

```js
{ 
   "Hello, world!": "¡Hola Mundo!" // english original as key
}
```
```js
{
   helloWorld: "¡Hola Mundo!" // identifier as key
}
```

I slightly prefer the english original as the key, so can see the text and parameters all at once when used. (Can change later.)

shadowspawn:
[...]

Issue #1608, opened 2021-09-20
Headers are hard-coded in Help.formatHelp()


edukisto:
Is there a cleaner way to internationalize the headers, which are hard-coded in `Help.formatHelp()`, i. e. [Arguments](https://github.com/tj/commander.js/blob/8571a7590b3a802523a719da82a49a88da687812/lib/help.js#L323), [Commands](https://github.com/tj/commander.js/blob/8571a7590b3a802523a719da82a49a88da687812/lib/help.js#L339), [Options](https://github.com/tj/commander.js/blob/8571a7590b3a802523a719da82a49a88da687812/lib/help.js#L331), [Usage](https://github.com/tj/commander.js/blob/8571a7590b3a802523a719da82a49a88da687812/lib/help.js#L310)?

```javascript
import process from 'process';
import { Command, Help } from 'commander';

class MyHelp extends Help {
  formatHelp(cmd, helper) {
    const result = super.formatHelp(cmd, helper);
    return result.replace('Usage:', 'Verwendung:'); // Translate!
  }
}

class MyCommand extends Command {
  createHelp() {
    return Object.assign(new MyHelp(), this.configureHelp());
  }
}

const program = new MyCommand();
program.parse(process.argv);
```

shadowspawn:
There is not support to internationalize strings so that approach is reasonable. You can either reimplement the routine, or modify the string it returns as you did.

You can also achieve the same thing without making a subclass using `.configureHelp()`, but same idea.

Related: #128 #774

edukisto:
`Command.configureHelp()`? Did you mean `Command.configureOutput()`?

```javascript
import process from 'process';
import { Command } from 'commander';

const program = new Command();

program.configureOutput({
  writeErr: (str) => process.stderr.write(
    str
      .replace('error:', 'ошибка:')
      // Proceed to replace other tokens from lib/command.js.
  ),

  writeOut: (str) => process.stdout.write(
    str
      .replace('Arguments:', 'Аргументы:')
      .replace('Commands:', 'Команды:')
      .replace('Options:', 'Опции:')
      .replace('Usage:', 'Использование:')
  ),
});

program.parse(process.argv);
```

shadowspawn:
`configureOutput` works too! I did mean `configureHelp` .

```ts
import * as process from 'process';
import { Command, Help } from 'commander';

const program = new Command();

program.configureHelp({
  formatHelp: (cmd, helper) => {
    const originalHelp = new Help();
[...]

Issue #1498, opened 2021-04-07
Editing error


DiaxManPl:
Is there any way to edit this: 
![image](https://user-images.githubusercontent.com/63294962/113859950-c1b20700-97a5-11eb-82d5-1bbae5f83500.png)
I want to make custom "unknown option" text.

shadowspawn:
There is not direct support for changing the error messages.

You could perhaps override `outputError` and intercept the message, or if you want additional text after the current message is displayed could catch the matching error and add the text.

https://github.com/tj/commander.js#override-exit-and-output-handling

DiaxManPl:
Thanks. `outputError` doesn't work, but custom error handling does.

DiaxManPl:
Oh. Custom error handling does not replace the message. It just adds custom text to the message.

shadowspawn:
This is how I thought `outputError` could be used.

```js
const { program } = require('commander');

program.option('-p, --port <number>');

program.configureOutput({
  outputError: (str, write) => {
    if (str.startsWith('error: unknown option'))
      write('custom unknown option error\n');
    else
      write(str);
  }
});

program.parse();
```

```
$ node index.js --port
error: option '-p, --port <number>' argument missing
$ node index.js --surprise
custom unknown option error
```

DiaxManPl:
Thanks a lot! It works!

shadowspawn:
There is an  open issue about adding support for modifying and/or localising error messages:  #1801

PR #1807, opened 2022-09-28
Add support for built-in translations and author supplied strings

shadowspawn:
# Pull Request

## Problem

People want to modify or translate the built-in error messages and help section titles. 

See: #128 #486 #774 #1498 #1608 #1801

We do already have individual support for modifying the help option, and help command name and description, and version option.

## Solution

Add translation support for the error messages and help related strings. Use tagged string templates as suggested in #1801, which preserves the existing message generation.

Add:
- `program.configureStrings(strings)` for author to supply their own strings
- `program.locale(localeName)` to load built-in translation
-  locale for `zh-CN` thanks to @gylove1994 

Currently includes some partial translations using strings from yargs (`locales/*-x-partial`), partly to encourage full translations, partly as an experiment  to fill out the locales folder and get a feel for what it could be like.

Technical note: node 12 does not include the full ICU and the `Intl.ListFormat` support falls back to default for non-English on at least some platforms. However, the next version of Commander will be dropping node 12 and later versions do have full ICU and get consistent behaviour with node 14, 16, and 18. (The list formatting is not critical, but nice to do it an official way.)

Implementation notes:
- Leaving `I18n` private for now so can easily make changes. Adding the functionality on `Command` rather than exposing the lower-level building blocks.
- only offering manual selection of locale for now (not automatic)
    - authors supporting multiple languages with their own strings can do their own auto-detection
    - not sure how to do good auto-detection on Windows


examples/translations.js
```diff
@@ -0,0 +1,111 @@
...
+process.env.PORT = 'nan';
+run(['serve'], '$ PORT=nan program server');
+delete process.env.PORT;
+
+// commander.excessArguments
+run(['1', '2', '3']);
+run(['serve', '1', '2', '3']);
+
+// Try the following:
+//    node locale-ever-message en
```
gylove1994, line 110: Maybe we need to change it to 
```js
// node translations.js en
// node translations.js zh-CN
```
or change the file name to match `locale-ever-message`.

locales/en.js
```diff
@@ -0,0 +1,55 @@
...
+const translations = {
+  // errors
+  // commander.missingArgument
+  "error: missing required argument '{0}'": "error: missing required argument '{0}'",
+  // commander.optionMissingArgument
+  "error: option '{0}' argument missing": "error: option '{0}' argument missing",
+  // commander.missingMandatoryOptionValue
+  "error: required option '{0}' not specified": "error: required option '{0}' not specified",
+  // commander.conflictingOption
+  'error: {0} cannot be used with {1}': 'error: {0} cannot be used with {1}',
```
gylove1994, line 12: Can change the template strings and substitutions for it, or provide a way to modify it by translater?
The word `option` seems to be hard-code in substitution, which will course some translations problem:
```js
// en
// $ program answer --yes --no
// error: option '--no' cannot be used with option '--yes'
// zh-CN
// $ program answer --yes --no
// 错误: 选项 option '--no' 不能与 option '--yes' 共同使用
//            ^                   ^

File: examples/translations.js (hunk 1 above)
Comment (shadowspawn) [reply examples/translations.js:110]: Oops, I changed the filename and not the instructions.

File: locales/en.js (hunk 2 above)
Comment (shadowspawn) [reply locales/en.js:12]: I missed that one. Message being constructed with two values that may come from CLI option arguments or may come from environment variable. In `getErrorMessage`.

https://github.com/tj/commander.js/blob/0b4198dde02cc8fc571c1265afa5bde447f6fca8/lib/command.js#L1682

```
      if (source === 'env') {
        return `environment variable '${bestOption.envVar}'`;
      }
      return `option '${bestOption.flags}'`;
```

Comment (shadowspawn) [reply locales/en.js]: I have reworked the error message to remove the hard-coded `option` and `environment variable` strings in the code. The environment variable is now listed after the option, using the same translation as in the description of the option in the help.

I did a simple update to `zh-CN` but it may need further changes. It displays error messages like:

```
$ program answer --right --wrong
错误: 选项 '--wrong' 不能与 '--right' 共同使用

-----

$ RIGHT=1 program answer --wrong
错误: 选项 '--wrong' 不能与 '--right' (所需环境变量: RIGHT) 共同使用
```

gylove1994: It looks great now.

shadowspawn: 👍

shadowspawn:
json file format is more common than js for storing translation. I think the main advantage of js is able to add comments, but they aren't so important once the translation is done. Might switch...

shadowspawn:
Started switch to json files, just did the full translations first: `zh-CN` and `en`.

shadowspawn:
This has not generated any likes or feedback in six months and touches a lot of code. It makes it more work for maintainers to add and modify strings displayed by Commander. Translations would tend to drift out of sync.

Much as I like the idea of supporting customising strings and translations, there needs to be more demand to make this worth doing. Closing for now!

Issue #1774, opened 2022-08-17
Strong TypeScript typing for options and action handler


shadowspawn:
Currently the options returned by`.opts()` and the parameters to `.action()` are basically untyped.

TypeScript is now capable of inferring the options and command-arguments from the method calls, like `.option()` and .argument()`. But it is pretty hard-core! There is a separate package with some experimental strong typings for Commander to develop further. These may stay stand-alone, or eventually be integrated into Commander if they work well.

https://github.com/commander-js/extra-typings

shadowspawn:
Added to README.

PR #1814, opened 2022-10-30
Add extra-typings to README

shadowspawn:
We have experimental "extra-typings" available with strong inferred types. The type definition file is much more complex and comes with some downsides, so not a simple switch. Likely to stay as a separate project at least until we find out more.

Add mention to the README so has some visibility.

See #1774


abetomo (approved):
👍

PR #1815, opened 2022-11-05
Update minimum version of node in v10 README

shadowspawn:
I bumped the required version of node in `package.json` for Commander v10, but missed the mention in the README.


abetomo (approved):
👍

PR #1826, opened 2022-11-27
Update dependencies

shadowspawn:
Update dependencies.

The new version of `tsd` no longer fights with TypeScript over the `tsc` binary, so don't need to keep the version of `tsd` and TypeScript carefully coordinated and can update them independently.

Minor whitespace change to satisfy new `ts-lint`.


abetomo (approved):
👍

Issue #1689, opened 2022-02-07
Is there a way to show global options in subcommand help?


nitzano:
```typescript
import { Command } from 'commander';

const subCommand = new Command('sub').option('-s1 --sub1', 'sub option')

const globalCommand = new Command('global');

globalCommand 
	.description('global command')
	.option('-g1 --global1', 'global option')
	.addCommand(subCommand)
```

Is there a way to show `-g1` exists when running `global sub --help`? (which shows only `-s1` )

I know I can use `optsWithGlobals()` inside the action but was wondering if there is an option to also show
in the subcommand's help the existance of it's global options from the parent command(s)?

Thanks :)

nitzano:
Thanks @shadowspawn, if there is a way to vote for this feature again I'd be happy to give my +1

shadowspawn:
Added `showGlobalOptions` in Commander 9.5.0.

Issue #1078, opened 2019-10-15
Global options not displayed in sub-command help


jjs105:
I may be missing something but is there any way to include the global options information in the sub-command help.

Test Code:
```
program
  .description('Test program/command.')
  .option('-g, --global', 'a global option')
  .version(VERSION);

program
  .command('sub')
  .description('Sub-command!')
  .option('-l, --local', 'a sub-command local option')
  .action(() => {});

program.parse(process.argv);
```

Program Help Output:
```
$> node test.js -h
Usage: test [options] [command]

Test program/command.

Options:
  -g, --global   a global option
  -V, --version  output the version number
  -h, --help     output usage information

Commands:
  sub [options]  Sub-command!
```

Sub-Command Help Output:
```
$> node test.js sub -h
Usage: test sub [options]

Sub-command!

Options:
  -l, --local  a sub-command local option
  -h, --help   output usage information
```

What I would want/expect:
<pre>
$> node test.js sub -h
Usage: test sub [options]

Sub-command!

Options:
  <b>-g, --global   a global option
  -V, --version  output the version number</b>
  -l, --local  a sub-command local option
  -h, --help   output usage information
</pre>

shadowspawn:
You didn't miss anything, there is not a way to include the global options information in the sub-command help.

shadowspawn:
Related: #905

shadowspawn:
An answer was provided, and no further activity or likes in a few months.

Feel free to open a new issue if it comes up again, with new information and renewed interest.

shadowspawn:
Opened poll on possible enhancements to global options: #1551

shadowspawn:
Added `showGlobalOptions` in Commander 9.5.0.

PR #1828, opened 2022-12-03
Support showing global options in help

shadowspawn:
# Pull Request

Commander has long supported options at both the program level and in subcommands, but using both at once was a little subtle for both the author and the end-user.

## Problem

In the Help you can't see all the options at once for a subcommand when the program has global options.

See: #1078 #1689 

The global options poll has continued accumulating votes for displaying global options after the poll was closed. https://github.com/tj/commander.js/issues/1551#issuecomment-869093297

## Solution

Add help configuration to show the global options. This is off by default so can go in a feature version without changing behaviour of executing programs.

The global options are included in the padding calculations and are optional sorted. 

The global options are shown in a separate section. This is the layout that was shown in https://github.com/tj/commander.js/issues/1551#issuecomment-869093297).

Using the updated example file:
```console
$ node examples/global-options-nested.js sub --help
Usage: global-options-nested sub [options]

Options:
  -l, --local
  -h, --help    display help for command

Global Options:
  -g, --global
```

## ChangeLog

- added: `showGlobalOptions` to `.configureHelp{}`


abetomo (approved):
LGTM!

shadowspawn:

Issue #1713, opened 2022-04-07
getOptionValueSource() in a command context returns `undefined` for global options


sinedied:
Hello,

I need to distinguish in my code which options comes from the user (via CLI options), and which one are just the default values.
For that, I use `getOptionValueSource()` but I have an issue with that, as it always return undefined when trying to get the source of a global option in a command context.

I retrieve all options using `command.optsWithGlobals()`, but then `command.getOptionValueSource()` only works for command-level options, and if I use `program.getOptionValueSource()` it only works with global-level options.

Should I merge all options manually instead of using `optsWithGlobals()` to separate the two contexts when trying to determine the source of the options? It would seems a bit complicated for a simple use case.

Would you consider adding a `getOptionValueSourceWithGlobals` to match the behavior of `optsWithGlobals`, so that we could retrieve the option's source if it comes either from command or global level?

shadowspawn:
> Should I merge all options manually instead of using optsWithGlobals() to separate the two contexts when trying to determine the source of the options? It would seems a bit complicated for a simple use case.

I am not sure of best approach for your case. You could possibly write your own `getOptionValueSourceWithGlobals` which searches up the command chain looking for a hit. (I haven't tried this and not sure if there are complications.)

If you have global options and local options, and whatever you are doing with them you need to tell where they came from using `.getOptionValueSource()`, I don't think you get to call it simple. 😄 

> I need to distinguish in my code which options comes from the user (via CLI options), and which one are just the default values.

For interest, can you give an insight into why you need to distinguish?

> Would you consider adding a getOptionValueSourceWithGlobals to match the behavior of optsWithGlobals, so that we could retrieve the option's source if it comes either from command or global level?

My first impression is it is not a common enough case to build into Commander. But happy to leave issue open and see if other people have similar need.

shadowspawn:
I'll add that `.optsWithGlobals()` was added in 9.0 and `.getOptionValueSource()` in 8.3, so both fairly new and may take a while to see how they are being used and could be improved. Interesting to hear you are using both. 👍

sinedied:
> For interest, can you give an insight into why you need to distinguish?

I'm working on a CLI that can use options from a config file, that I load when the CLI is invoked. But I need to be able to override any options from the config file using CLI options, hence this need. I would have thought that the use case is pretty common 😄 

[...]

PR #1832, opened 2022-12-17
Add getOptionValueSourceWithGlobals

shadowspawn:
# Pull Request

## Problem

We added `.optWithGlobals()` to make working with globals easier. And `.getOptionValueSource()` for more complex programs. But how do you get the option value source including globals?

See #1713

## Solution

Added `.getOptionValueSourceWithGlobals()`.

Sticking with the obvious and _long_ name. Not planning to add to README.

## ChangeLog

- Added: `.getOptionValueSourceWithGlobals()`

### DIFF
--- a/lib/command.js
+++ b/lib/command.js
@@ -815,6 +815,25 @@ Expecting one of '${allowedValues.join("', '")}'`);
   }
 
   /**
+    * Get source of option value. See also .optsWithGlobals().
+    * Expected values are default | config | env | cli | implied
+    *
+    * @param {string} key
+    * @return {string}
+    */
+
+  getOptionValueSourceWithGlobals(key) {
+    // global overwrites local, like optsWithGlobals
+    let source;
+    getCommandAndParents(this).forEach((cmd) => {
+      if (cmd.getOptionValueSource(key) !== undefined) {
+        source = cmd.getOptionValueSource(key);
+      }
+    });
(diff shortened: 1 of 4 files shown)


lib/command.js
```diff
@@ -814,6 +814,25 @@ Expecting one of '${allowedValues.join("', '")}'`);
...
+    * Get source of option value. See also .optsWithGlobals().
+    * Expected values are default | config | env | cli | implied
+    *
+    * @param {string} key
+    * @return {string}
+    */
+
+  getOptionValueSourceWithGlobals(key) {
+    // global overwrites local, like optsWithGlobals
+    let source;
```
abetomo, line 827: For reference :
You can also write the following using `findLast`.

```js
    return getCommandAndParents(this)
      .findLast((cmd) => cmd.getOptionValueSource(key) !== undefined)
      .getOptionValueSource(key);
```

abetomo (approved):
👍

lib/command.js
abetomo: You can use Node.js 18 or later.
My apologies.

shadowspawn: I started out planning to target Commander v10.x just so I could use `??`, but switched to old-school Node 12 compatible code for now. Eventually!

shadowspawn:

Issue #1831, opened 2022-12-13
[docs] `.alias()` support is not documented in the Readme


viglucci:
I only found that commander support aliasing via searching and finding the below issues/PRs:

- https://github.com/tj/commander.js/pull/1236
- https://github.com/tj/commander.js/issues/531

Is it correct that this is not documented, or did I miss it somewhere?

shadowspawn:
You are right, it is not mentioned in the README and should be.

It was used in one example in the README until v9. It got lost when the examples got reworked in #1665. (I didn't want it in the starting example, but didn't add it elsewhere.)

shadowspawn:
Added mention in Commander v9.5.0

viglucci:
@shadowspawn excellent, thanks!

PR #1833, opened 2022-12-19
Add mention of .alias() to README.

shadowspawn:
# Problem

There is no longer any mention of `.alias()` in the README.

See: #1831

## Solution

Add a brief mention and example file.


abetomo (approved):
👍

shadowspawn:

PR #1835, opened 2022-12-26
*: update dev deps and make tweaks

VoltrexKeyva:
- Update the developer dependencies to their latest versions.
- Remove usage of the `--experimental-modules` CLI flag as it's a no-op now.
- Make changes to some of the typings files to make the linters happy.

### DIFF
--- a/typings/index.d.ts
+++ b/typings/index.d.ts
@@ -277,7 +277,8 @@ export interface OptionValues {
 export class Command {
   args: string[];
   processedArgs: any[];
-  commands: Command[];
+  readonly commands: readonly Command[];
+  readonly options: readonly Option[];
   parent: Command | null;
 
   constructor(name?: string);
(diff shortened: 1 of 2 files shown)


shadowspawn:
Commander 9 supports node v12.20.0 and up, so can't upgrade to latest Jest on mainline yet as it requires node 14. The `release/10.x` branch has minimum of node 14 for the next major version.

VoltrexKeyva:
I see, although I'm not sure what happened here while changing the target branch to `release/10.x`, this is weird.

shadowspawn:
Yes, commits disappeared?!

I made some updates in #1826, but not including the `record` change suggested by (I assume) new version of eslint.

VoltrexKeyva:
Hmm, I guess I'll just open another PR 😅

PR #1836, opened 2022-12-26
*: update dev deps and make tweaks

VoltrexKeyva:
- Update the developer dependencies to their latest versions.
- Remove usage of the `--experimental-modules` CLI flag as it's a no-op now.
- Change the typings file to make the linters happy.


shadowspawn (approved):
LGTM

abetomo (approved):
👍

Issue #1687, opened 2022-02-03
Variadic option with env


trhinehart-godaddy:
```js
// cli.js
const choices = ['a', 'b'];
program
  .addOption(new Option('--test <ex...>')
    .choices(choices)
    .default(choices)
    .env('CLI_TEST')
  )
```

Works fine if I set `CLI_TEST` to `a` or `b`, but unable to do with `a b` or a serialize array.

shadowspawn:
There is not built-in support for encoding an array in an environment variable.

One possible approach is to handle the splitting a different way that works both from the command-line and from and environment variable.

```js
const { Command, Option } = require('commander');
const program = new Command();

function commaSeparatedList(value, dummyPrevious) {
  return value.split(',');
}

program
  .addOption(new Option('-c <value>').argParser(commaSeparatedList).env('COMMA'));

program.parse();
console.log(program.opts())
```

```
% COMMA=a,b node index.js 
{ c: [ 'a', 'b' ] }
```

trhinehart-godaddy:
I'm not looking to encode, I'm looking for env to handle `a b` the same way using an option works.

ie:
```js
cross-env CLI_TEST="a b" ./cli.js
```
should be the same as
```js
node ./cli.js --test a b
```

The workaround you provided doesn't work with `<value...>` which I need since I'm using `choices`

shadowspawn:
The shell passes node an array of arguments, which Commander gets from `process.argv` by default. It is not Commander that is splitting up `a b` into `a` `b`. Your example might help make this clearer:

```
cross-env CLI_TEST="a b" ./cli.js
```

is the same as

```
node ./cli.js --test "a b"
```

shadowspawn:
To roll-your-own solution, you could implement the validation (choices) yourself in your custom option processing function.

trhinehart-godaddy:
I think I understand the problem, env needs to be processed just like argv. Here's a hack to work around that, however security might be a concern.

https://github.com/tj/commander.js/compare/master...trhinehart-godaddy:patch-1

trhinehart-godaddy:
[...]

PR #1846, opened 2023-02-08
Support of multiple values from env for variadic options

dhyegocalota:
# Pull Request

## Problem

We already support multiple values parsed from CLI for variadic options but we didn't have that same feature for values parsed from env.

```js
BAR=firstItem secondItem thirdItem ./cli
// options.bar=['firstItem', 'secondItem', 'thirdItem']
```

## Solution

We solved the problem by splitting the env value with blank spaces just the way we parse values from CLI.

## ChangeLog

Add support of multiple values from env for variadic options

### DIFF
--- a/lib/command.js
+++ b/lib/command.js
@@ -545,7 +545,7 @@ Expecting one of '${allowedValues.join("', '")}'`);
           throw err;
         }
       } else if (val !== null && option.variadic) {
-        val = option._concatValue(val, oldValue);
+        val = option._concatValue(val.split(/ +/), oldValue || []);
       }
 
       // Fill-in appropriate missing values. Long winded but easy to follow.
--- a/tests/options.env.test.js
+++ b/tests/options.env.test.js
@@ -206,12 +206,27 @@ describe('custom argParser', () => {
 });
 
 describe('variadic', () => {
-  test('when env defined and no cli then array from env', () => {
+  test('when env defined and no cli then array from env with single value', () => {
     const program = new commander.Command();
     process.env.BAR = 'env';
     program.addOption(new commander.Option('-f, --foo <required...>').env('BAR'));
     program.parse([], { from: 'user' });
     expect(program.opts().foo).toEqual(['env']);
+    delete process.env.BAR;
+  });
+
+  test('when env defined and no cli then array from env with multiple values', () => {
+    const program = new commander.Command();
+    process.env.BAR = 'firstItem secondItem thirdItem';
+    program.addOption(
+      new commander.Option('-f, --foo <required...>').env('BAR')
+    );
+    program.parse([], { from: 'user' });
+    expect(program.opts().foo).toEqual([
+      'firstItem',
+      'secondItem',
+      'thirdItem'
+    ]);
     delete process.env.BAR;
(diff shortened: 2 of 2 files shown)


shadowspawn:
This has came up once before: #1687

My concern with splitting on spaces is it is making an opinionated decision about how all environment variables should be split. From the command-line, the shell decides how the line is split into separate arguments, which may include spaces:

```
$ util --bar a b "c and C"
```

Are there other command-line libraries that automatically split environment variables?

-----

I suggested a roll-your-own approach in https://github.com/tj/commander.js/issues/1687#issuecomment-1029343834

Would that work for you? Would it be useful to make an example file of this approach?

dhyegocalota:
Got it, your concern makes sense! I'm not sure if there other libs that have this feature. My motivation came from the need to use custom arg parsers multiple times to achieve this behavior in my CLI app.

This would definitely work for me. Yes, it would be useful to have an example file.

The most useful use-case for me is to pass multiple options like IDs like this:

```js
program
  .option('-i, --ids <ids...>', 'specify IDs').env('IDS');

program.parse();
```

```bash
$ collect -i 1 2 3
Options:  { ids: [ '1', '2', '3' ] }
Remaining arguments:  []
$ IDS=1 2 3 collect
Options:  { ids: [ '1', '2', '3' ] }
Remaining arguments:  []
```

Regarding enforcing this feature to all users, we might be able to restrict it using some feature-flag. 

A feature-flag approach would be to add the same three-dots to the end of the env var name like this —

```js
program
  .option('-i, --ids <ids...>', 'specify IDs').env('IDS...');
```

shadowspawn:
For interest, similar open issue reported for Yargs with some discussion: https://github.com/yargs/yargs/issues/821

shadowspawn:
Minor, but I think your example is not valid syntax:
```console
$ IDS=1 2 3 collect
bash: 2: command not found

% IDS=1 2 3 collect
zsh: command not found: 2
```

So might be:
```console
$ IDS="1 2 3" collect
```

shadowspawn:
I think I have seen some uses which use multiple variables, like `FOO_1=27 FOO_2=46 collect`, but that is a bit clunky.

(A quick flurry of comments after doing some light digging this morning!)

dhyegocalota:
Interesting that other libs have similar scenarios! Well, thanks for correcting my command sample. It was indeed incorrect.

We could make this feature optional and flexible enough so we don't enforce its usage. We also could possibly provide the ability to change the split regex rule.

shadowspawn:
To be safe, I think this either needs to be opt-in (e.g. your `IDS...` suggestion) or have configuration, which makes it less attractive. The supported general-purpose approach is to use a custom option processor, as you mentioned.

A different thing that has come up is how to represent boolean options in an environment variable. This is the same sort of problem, how to get a single string (i.e. environment variable value) to represent option use on the command-line.

A high level approach might be to add support for a custom env processor at the command level (rather than per option). This would reduce the configuration from every affected option to one.

This has not come up often to date, and this solution isn't an easy win, so I'm closing this for now.

Thanks for the suggestion and discussion.

PR #1853, opened 2023-03-11
Add custom usage compatibility to help menu

hiimjustin000:
# Pull Request

## Problem
The help menu does not display the custom command usage. If there are custom syntax changes that are critical for users to understand the program, this can be problematic.

## Solution

Currently, this can be fixed by reconfiguring the helper, which is fine, but it would be nice to have this functionality by default. So, I slightly altered how `subcommandTerm` outputs command usage. If that is not set by the user yet, it will default to the usual processing.

## Final Thoughts
Let me know if this was intentionally implemented, as it appeared that  `subcommandTerm` was stated to be a "legacy" function despite still being used to list subcommands in the help menu..

### DIFF
--- a/lib/help.js
+++ b/lib/help.js
@@ -141,10 +141,10 @@ class Help {
    */
 
   subcommandTerm(cmd) {
-    const term = cmd._name + (cmd._aliases[0] ? '|' + cmd._aliases[0] : '');
-    if (cmd._usage) return term + ' ' + cmd._usage;
+    const cmdNames = cmd._name + (cmd._aliases[0] ? '|' + cmd._aliases[0] : '');
+    if (cmd._usage) return cmdNames + ' ' + cmd._usage;
     const args = cmd._args.map(humanReadableArgName).join(' ');
-    return term +
+    return cmdNames +
       (cmd.options.length ? ' [options]' : '') + // simplistic check for non-help option
       (args ? ' ' + args : '');
   }
--- a/lib/help.js
+++ b/lib/help.js
@@ -141,10 +141,10 @@ class Help {
    */
 
   subcommandTerm(cmd) {
-    // Legacy. Ignores custom usage string, and nested commands.
-    const args = cmd._args.map(arg => humanReadableArgName(arg)).join(' ');
-    return cmd._name +
-      (cmd._aliases[0] ? '|' + cmd._aliases[0] : '') +
+    const term = cmd._name + (cmd._aliases[0] ? '|' + cmd._aliases[0] : '');
+    if (cmd._usage) return term + ' ' + cmd._usage;
+    const args = cmd._args.map(humanReadableArgName).join(' ');
+    return term +
       (cmd.options.length ? ' [options]' : '') + // simplistic check for non-help option
       (args ? ' ' + args : '');
   }
--- a/tests/help.commandTerm.test.js
+++ b/tests/help.commandTerm.test.js
@@ -40,4 +40,14 @@ describe('subcommandTerm', () => {
     const helper = new commander.Help();
     expect(helper.subcommandTerm(command)).toEqual('program|alias [options] <argument>');
   });
+
+  test('when command has usage then returns name|alias (usage)', () => {
+    const command = new commander.Command('program')
+      .usage('<argument> [options]')
+      .alias('alias')
+      .option('-a,--all')
+      .argument('<argument>');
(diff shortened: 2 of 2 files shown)


shadowspawn:
(The "legacy" comment was that it was fairly simple and different to what is displayed in the command help, but had displayed that way for a long time and I didn't have a reason to change it when I rewrote the help.)

lib/help.js
```diff
@@ -141,10 +141,10 @@ class Help {
    */
 
   subcommandTerm(cmd) {
-    // Legacy. Ignores custom usage string, and nested commands.
-    const args = cmd._args.map(arg => humanReadableArgName(arg)).join(' ');
-    return cmd._name +
-      (cmd._aliases[0] ? '|' + cmd._aliases[0] : '') +
+    const term = cmd._name + (cmd._aliases[0] ? '|' + cmd._aliases[0] : '');
```
shadowspawn, line 144: Minor: "term" is used to describe the whole thing, so I would like a different variable name here. Maybe cmdNames?

shadowspawn:
Some thoughts, no conclusions yet!

High level, this is doing same as displaying `cmd.usage()` if there is a custom usage. So why don't we just call `cmd.usage()` all the time?

A high level concern is that in the display of the help for the command, the usage and the description are displayed on separate lines.

```
% node plain.js busy --help
Usage: plain busy|bz [options] <file>

bells and whistle
...
```

When listing the subcommands, we add the description directly after the usage:
```
% node plain.js --help 
...
Commands:
  simple                      no arguments no options
  busy|bz [options] <file>    bells and whistle
...
```

(Side note: by default every command gets `[options]`in its full help "usage" because of the help option. Which does cause complaints anyway, so maybe we could change that small part!)

The usage generated by Commander is pretty simplistic. Unix commands often show at least all the single letter options, and sometimes multiple lines of different use cases in the synopsis. That isn't the thing we are trying to solve here, but it does suggest that custom usage lines may be longer than the simple default one.

For example `man ls` includes:

```
SYNOPSIS
     ls [-@ABCFGHILOPRSTUWabcdefghiklmnopqrstuvwxy1%,] [--color=when] [-D format] [file ...]
```

We had a similar problem with the description, that a long description is good for the full help but may suck when displaying in a list of subcommands. To solve that we added `.summary()` to optionally supply a different (likely shorter) version of the description.

shadowspawn:
> Currently, this can be fixed by reconfiguring the helper, which is fine, but it would be nice to have this functionality by default. 

If is fairly easy to display full usage, but I acknowledge that if it is extra work.

```js
function fullUsage(cmd) {
  const cmdNames = cmd.name() + (cmd.alias() ? '|' + cmd.alias() : '');
  return cmdNames + ' ' + cmd.usage();
  }

program.configureHelp({ subcommandTerm: fullUsage });
```

shadowspawn:
(Good detail in the Pull Request description, thanks.)

shadowspawn:
I am guessing this is something you wanted in one of your programs. Have you got some examples of what you put in custom command usage?

hiimjustin000:
```
  base|radix <radix> [options] -- <input...>  Encodes the specified input data with the specified base/radix
  base32|b32 [options] -- <input...>          Encodes the specified input data with base 32
  base36|b36 [options] -- <input...>          Encodes the specified input data with base 36
  base64|b64 [options] -- <input...>          Encodes the specified input data with base 64
  binary|bin [options] -- <input...>          Encodes the specified input data with base 2
  decimal|dec [options] -- <input...>         Encodes the specified input data with base 10
  hexadecimal|hex [options] -- <input...>     Encodes the specified input data with base 16
  octal|oct [options] -- <input...>           Encodes the specified input data with base 8
  help [command]                              Outputs help for command
```

shadowspawn:
I am interested in improving the usage displayed for subcommands. I don't think using the custom usage is the answer though. In the general case I think the usage in the list of subcommands benefits from being terse.

The approach we took with the subcommand description was adding an optional `summary`, so there could be a `usageSummary`. That does not help with the original issue in this PR, and is a separate idea.

Since the goal of using custom usage can be achieved with a reasonably simple override (https://github.com/tj/commander.js/pull/1853#issuecomment-1465034399), closing this PR.

shadowspawn:
Tracking similar issues (in case need to reevaluate):
-  #1885

Issue #1852, opened 2023-03-09
Add check for `implies` argument type


canopy-js-user:
Hi there, twice I've accidentally written a line of code like this:

```
.addOption(new Option(
    '--port <number>', 
    'Which port to run the server on for sync mode'
).default(undefined).implies('sync'))
```

Ie, I want the `--port` option to imply the boolean flag `sync`. However, `sync` is expecting an object of implied key-value pairs. However, as is, this fails silently and converts the string/array `'sync'` into an object, and merges it with the rest of the options:

```
{
  '0': 's',
  '1': 'y',
  '2': 'n',
  '3': 'c',
  editor: true,
  logging: true,
  open: true,
  backup: true,
  sync: true,
  port: '4004'
}
```

I'm proposing a check that the argument to `implies` is an object and not a string, and if it is a string, error. (Or, interpret the string as the name of a boolean option that the programmer is asserting should be true.)

Thanks!

canopy-js-user:
(Another reason I think this an easy mistake is that many of the option methods _do_ take strings, eg `.conflics`. Thanks!)

shadowspawn:
Commander does not have a lot of runtime checks on the parameters, but I am happy to consider adding checks or improving behaviour for routines that are easy to get wrong.

(A past one that was added was #1655.)

canopy-js-user:
Makes sense, thanks!

shadowspawn:
For interest, we added support in `.conflicts()` for a string or an array after making a similar mistake during development: https://github.com/tj/commander.js/pull/1678#issuecomment-1047210739

shadowspawn:
Shipped in 10.0.1

Thanks @canopy-js-user

canopy-js-user:
@shadowspawn thank _you_!

PR #1854, opened 2023-03-12
Assume a string parameter to implies is name of boolean option.

shadowspawn:
# Pull Request

## Problem

It is an easy mistake to pass a string to `.implies()` instead of a hash. We have other routines that do just take option names, like `.conflicts()`.

See: #1852

## Solution

Assume the users intent was to set the option to true, like a simple boolean option.

An alternative would be to throw an error, but that would probably need to wait for a semver major release since it would break programs that might otherwise be functional.

(Both of these were included as possible approaches in #1852.)

## ChangeLog

(Leaving out of changelog as this is a quiet fixup rather than a feature as such.)

### DIFF
--- a/examples/options-implies.js
+++ b/examples/options-implies.js
@@ -3,7 +3,6 @@
 const { Command, Option } = require('../'); // include commander in git clone of commander repo
 const program = new Command();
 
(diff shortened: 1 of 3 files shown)


abetomo (approved):
👍

shadowspawn:
Shipped in 10.0.1

Issue #1822, opened 2022-11-18
Wrapping the second and following line in a multiline description doesn't work if the line starts from space


bk201-:
I wanted to add an indent to the second line in the description.  

```javascript
program
  .option('--config <path>', 'Configuration file path')
  .option(
    '-M, --message-pattern <RegExp>',
    `Message pattern to configure where JIRA ticket number will be inserted
    * Symbols '$J' 
    * Symbols '$M' `,
  );
```

I've expected something like this

```cli
Usage: index [options]   
                                       
Options:                          
  --config <path>                 Configuration file path
  -M, --message-pattern <RegExp>  Message pattern to configure where JIRA ticket number will be inserted
                                     * Symbols `$J` 
                                     * Symbols `$M`
  -h, --help                      display help for command         
```

But got:

```cli
Usage: index [options]   
                                       
Options:                          
  --config <path>                 Configuration file path
  -M, --message-pattern <RegExp>  Message pattern to configure where JIRA ticket number will be inserted
      * Symbols '$J'
      * Symbols '$M'
  -h, --help                      display help for command  
```

I tried to change the template literal to a string but got the same result.
```javascript
program
  .option('--config <path>', 'Configuration file path')
  .option(
    '-M, --message-pattern <RegExp>',
    'Message pattern to configure where JIRA ticket number will be inserted\n' +
    '   * Symbols `$J`\n' +
    '   * Symbols `$M`',
 // ↑ three spaces
  );
```

But if I remove leading spaces from the second and third line, the indent will work correctly;
```cli
Usage: index [options]

Options:
  --config <path>                 Configuration file path
  -M, --message-pattern <RegExp>  Message pattern to configure where JIRA ticket number will be inserted
                                  * Symbols `$J`
                                  * Symbols `$M`
  -h, --help                      display help for command
```

shadowspawn:
Thanks for the detailed examples.

This is an intended behaviour rather than a bug. The wrapping code does not wrap text that looks pre-formatted. When you put spaces on the start of the second line, the wrapping code assumes you have pre-formatted the text.

The current wrapping behaviour was introduced in  #956 and #1051, and the issue they took into account with pre-formatted text was #396.

bk201-:
Thank you for the clarification. I read https://github.com/tj/commander.js/issues/396, but couldn't figure out what it talking about.

In any case, it would be so great if you could add more details to the documentation.

shadowspawn:
[...]

PR #1859, opened 2023-04-06
Add help description wrapping to README

shadowspawn:
## Problem

People with more complex help descriptions may want to know about the auto-wrapping, to have their description wrapped by Commander or pre-formatted by the author.

See: #1822

## Solution

Add a brief description that there is auto-wrapping, and what is considered pre-formatted text and not wrapped.

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -761,6 +761,8 @@ shell help spawn
 shell spawn --help
 ```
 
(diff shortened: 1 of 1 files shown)


abetomo (approved):
👍

shadowspawn:
Shipped in 10.0.1

Issue #1843, opened 2023-01-18
[Help] Struggling to implement configuration file options


gomain:
Toy program:
```ts
//cli.js
const optionA = new Option("--a <a>").makeOptionMandatory();
const optionB = new Option("--b <b>").makeOptionMandatory();

const useA = new Command("useA")
  .addOption(optionA)
  .action((opts) => console.log(opts.a));
const useAandB = new Command("useAandB")
  .addOption(optionA)
  .addOption(optionB)
  .action((opts) => console.log(opts.a, opts.b));

program
  .addCommand(useA)
  .addCommand(useAandB)
  .parse();
```
Now I want to include a `config.json` file. Say it is
```json
{
  "a": "config-A"
}
```
Such that when run
```bash
$ node cli.js useA
config-A

$ node cli.js useA --a cli-A
cli-A

$node cli.js useAandB
error: required option '--b <b>' not specified

$node cli.js useAandB --a cli-A
error: required option '--b <b>' not specified

$node cli.js useAandB --b cli-B
config-A cli-B

$node cli.js useAandB --a cli-A --b cli-B
cli-A cli-B
```
Use cases will extend to combinations of options having `.default` or `.env` where precedence will be cli > config > env > default.

I have looked into `.getOptionValue` and friends but struggle to find the right place to use them. Documentations lack practical examples.

Some concerns:
* Options may have `.argParser`. I wish the json to store values as strings and be parsed as if provided on cli.
* Missing option errors are triggered _after_ all option sources have been consulted.
* Not all options are consumed by every command (as in --b). I want to keep the unknown-option-error when provided at cli but unknown options in config file are ignored.
* Able to have global option `--config <path>` to pass location of config file.

shadowspawn:
A couple of historical references.

There was a discussion about improvements to support config file processing in #1584. Several of the people commenting had already written implementations. The outcome was adding `. getOptionValueSource()` to make it easier to manage precedence.

The use case in this comment is processing a config file: https://github.com/tj/commander.js/issues/1713#issuecomment-1091521617

shadowspawn:
> Options may have .argParser. I wish the json to store values as strings and be parsed as if provided on cli.

There are not methods to do this directly. I think the safest route is probably looking up the option and doing the conversion yourself.

You could try emitting an option event (e.g. `option:b`) with the string value to trigger the built-in option processing, but the processing assumes it is from the CLI so you may run into problems. I have not done this myself from client code.

[...]

Issue #1787, opened 2022-08-30
It's not possible to customize environment variables with booleans.


mshima:
It's not possible to customize environment variables parsing like:
```sh
FOO=false command
```
Because the variable value is not forwarded:
https://github.com/tj/commander.js/blob/f2aec2648d68b01d8947298b27093df3fdb551a7/lib/command.js#L1565-L1568

[GitHub Actions exposes](https://docs.github.com/en/actions/creating-actions/metadata-syntax-for-github-actions#example-specifying-inputs) inputs using the pattern:
```
INPUT_${option.name().toUpperCase()}=${value}
```
Where allowed boolean values are ['true', 'True', 'TRUE', 'false', 'False, 'FALSE'];

The proposal is to drop the condition
https://github.com/tj/commander.js/blob/f2aec2648d68b01d8947298b27093df3fdb551a7/lib/command.js#L1562

And add source argument to `argParser()` changing:
https://github.com/tj/commander.js/blob/f2aec2648d68b01d8947298b27093df3fdb551a7/lib/command.js#L539

to:
```
val = option.parseArg(val, oldValue, valueSource);
```

This would allow something like:
```
option.argParser((val, oldVal, valueSource) => valueSource === 'env' ? ['false', 'False', 'FALSE'].includes(val) : val);
```

shadowspawn:
I am interested in understanding the use case a bit better. Are you using Commander to generate a metadata file, or to consume environment variables produced from a metadata file?

In the metadata file I think the "booleans" are per YAML which has a wider range: https://yaml.org/type/bool.html

I don't know what environment variable value will be produced from a "boolean" but expect it will be just a pair of possible values (like say TRUE/FALSE).

shadowspawn:
You do get custom parsing of the environment variable by declaring the option as taking a value. Say:

```
program.addOption(new Option('--foo [bool]').env('FOO').argParser(myConvertToBool))
```

mshima:
> I am interested in understanding the use case a bit better. Are you using Commander to generate a metadata file, or to consume environment variables produced from a metadata file?

Both.
Produce metadata:
https://github.com/mshima/commander-github-actions/blob/main/src/dev.ts

Consuming:
https://github.com/mshima/commander-github-actions/blob/main/src/index.ts
 
> In the metadata file I think the "booleans" are per YAML which has a wider range: https://yaml.org/type/bool.html

That's version 1.1.
[...]

PR #1860, opened 2023-04-07
Add documentation on parsing life cycle and hooks

shadowspawn:
# Pull Request

## Problem

It can be hard to work out where to plugin extra functionality using the hooks. A specific example is reading configuration files.

See: #1843

> Perhaps the full life cycle of executing commands should be documented.

See also: #1762 https://github.com/tj/commander.js/issues/1787#issuecomment-1236027554

## Solution

Add a separate description of the processing life cycles and where the hooks fit in. This will be useful when discussions come up about adding more parsing affected hooks, and thinking about where config files can fit in.


abetomo (approved):
👍

shadowspawn:
Shipped in 10.0.1

Issue #1863, opened 2023-04-15
.helpOption(false) also disables subcommand help command


mitsukuri:
In this example, I'd like to disable impilcit help **option** and only leave impicit help **command**.
`test.mjs`:
```
#!/usr/bin/env node
import { Command } from 'commander';
const cmd = new Command ();
cmd
   .helpOption (false)
   .command ('meow')
   .option ('-l, [--loud]')

cmd.parse ();
```
If I run `test.mjs`, I get 
```
Usage: test [command]

Commands:
  meow [options]
  help [command]  display help for command
```
as expected, but if I run `test.mjs help meow`, I get an error:

`error: unknown option '--help' // what option, I thought I was calling help command?`,

while expected outcome would be
```
Usage: test meow [options]

Options:
  -l, [--loud]
```

shadowspawn:
You have found a shortcoming in the help support! 

The `help` command is implemented by passing `--help` into the subcommand. It was done this way so it could also work with stand-alone external subcommands.
- https://github.com/tj/commander.js/blob/33195f189b1ffa568c232503fb6ac0cf5548eb74/lib/command.js#L1254

I think it would be reasonable to call directly into the help for a known (non-external) command. This would make the support more robust.

shadowspawn:
This is fixed in v11

Thanks @mitsukuri

PR #1864, opened 2023-04-15
Have help command call help directly for subcommands, when possible

shadowspawn:
# Pull Request

## Problem

The implementation of the help command like `my-util help foo` is a bit fragile. The implementation invokes the subcommand passing `this._helpLongFlag`. This can fail if:
- the current command does not have a long flag
- the subcommand uses different help flags
- the subcommand has the help option disabled

See: #1863 (help option disabled)

## Solution

Call `.help()` directly for a known (non-external) subcommand. Just use the help flags as a fallback for external subcommands.

This takes a little extra code, but I have never been happy about passing through the help flags so I think it is worth it.

## ChangeLog

- Fixed: help command works when help option is disabled

### DIFF
--- a/lib/command.js
+++ b/lib/command.js
@@ -1085,6 +1085,26 @@ Expecting one of '${allowedValues.join("', '")}'`);
   }
 
   /**
+   * Invoke help directly if possible, or dispatch if necessary.
+   * e.g. help foo
+   *
+   * @api private
+   */
+
+  _dispatchHelpCommand(subcommandName) {
+    if (!subcommandName) {
+      this.help();
+    }
+    const subCommand = this._findCommand(subcommandName);
+    if (subCommand && !subCommand._executableHandler) {
+      subCommand.help();
+    }
+
+    // Fallback to parsing the help flag to invoke the help.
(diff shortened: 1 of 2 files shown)


lib/command.js
```diff
@@ -1084,6 +1084,26 @@ Expecting one of '${allowedValues.join("', '")}'`);
...
+   * e.g. help foo
+   *
+   * @api private
+   */
+
+  _dispatchHelpCommand(subcommandName) {
+    if (!subcommandName) {
+      this.help();
+    }
+    const subCommand = this._findCommand(subcommandName);
```
abetomo, line 1098: If `subcommandName` is `undefined`, `this._findCommand(subcommandName)` also returns `undefined`.

```js
if (!subcommandName) {
  this.help();
} else {
  const subCommand = this._findCommand(subcommandName);
  ...
}
```

shadowspawn: The help method calls exit, so the following code will not be reached if `subcommandName` is `undefined`.

abetomo (approved):
👍

lib/command.js
abetomo: That makes sense.

Issue #1869, opened 2023-04-25
main command action handler options empty


machinshin:
Hey all, i'm trying to separate out the action handler for a top-level command with the code below. 

the options variable in my action handler is undefined, and it's unclear why.

```typescript 
  // src/ts/program.ts
import { Command } from 'commander';

const program = new Command('cmd-name');
program
  .version('0.0.1') // real code loads this from package.json
  .description("program description string")
  .argument('<input-file>', 'file for operations')
  .requiredOption('-c --customer-name <comp>', 'description');
  //add more .option()

export default program;
```
```typescript
//src/ts/cli.ts
import program from './program';
program.action((input_file: string, options: Object): Promise<any> => {
  console.log(`input_file: ${input_file}`);
  console.log(`options: ${JSON.stringify(options, null, 2)}`);
 /* more code here */
});
(async () => {
  return await program.parseAsync(process.argv);
})();
```
so per the documentation, options should be an Object, but running this code, i get
`$> node dist/cli-bundle.js -- -c upt ./foo-bar.txt`
which displays: 
```
input_file: ./foo-bar.txt
options: undefined
```

shadowspawn:
I am unable to reproduce your problem using the sample code, and haven't worked out what might be wrong.

There is something odd in your steps, in this command:
```
$> node dist/cli-bundle.js -- -c upt ./foo-bar.txt
```
the `--` disables the option processing so `-c` is not recognised as an option and the missing required option causes an error to be displayed. Perhaps you were running it though npm?

Oh, having suggested that... npm will eat all the options if you don't use the `--`! That would fit with the problem you described.

Here is what I see running your example code:

```console
% node dist/cli.js -- -c upt ./foo-bar.txt
error: required option '-c --customer-name <comp>' not specified

% node dist/cli.js -c upt ./foo-bar.txt 
input_file: ./foo-bar.txt
options: {
  "customerName": "upt"
}
```

machinshin:
@shadowspawn  Thank you for responding so quickly! you are correct , I have a couple of npm scripts.

```json
{
  "main": "dist/bundle.js",  
  "scripts": {  
    "build": "rollup -c rollup.config.mjs"  -- generates dist/bundle.js
    "start" : "node ./dist/bundle.js"
  }
}
```
It seemed to be that the `--` are needed as testing any of the cli options, like, `npm run start -h` fails as the option is being handled by npm.

but still when I remove the `--` , now i get a different issue.

[...]

Issue #1820, opened 2022-11-14
options Undefined


arnaud422:
Hello, could someone tell me why the all option is always undefined? (even with options.all)
![option-undefinedPNG](https://user-images.githubusercontent.com/48210106/201750401-0745c24a-42e5-4a93-9c2b-8b84156ca00e.PNG)

shadowspawn:
1) What are you typing on the command-line to call the program?

2) Are you calling `program.parseAsync()`? (You need to call one of the two parse routines, and since you have an async action handler it should be `parseAsync`.)

3) If you copy and paste your program into a code block in the issue comment, it is more useful for searches and for reproducing the problem than a picture. Code blocks in markdown have three back-ticks before and after. You can optionally put the code language after the back-ticks, with `js` being a code for JavaScript. Like this:

````
```js
code goes here
```
````

arnaud422:
Thank you for your quick response! I do use "npm start" for testing. So I'll change that. Yes I made a capture without thinking about the three  back-ticks ;-)

shadowspawn:
You can use `npm run`, but need to stop npm from eating your options. See #1761 for more.

arnaud422:
great thanks

Issue #1769, opened 2022-07-17
Option always undefined


NathanC:
Options are ***always*** showing up as undefined (both in the error message and in my action handler).

I have my program set up like this:
```
program.command("deploy")
    .description("kick off a deploy")
    .addOption(
        new Option(
            "--only", "perform a partial deploy"
        ).choices(["frontend", "backend"])
    )
    .action((args, options) => console.log("only deploying: " + options.only))
    .parse();
```

When I try to run this command:
`myScript deploy --only frontend`

I get this error:
`error: option '--only' argument 'undefined' is invalid. Allowed choices are frontend, backend.`

If I try it with a disallowed choice (e.g., `--only foo`), I still get
`error: option '--only' argument 'undefined' is invalid. Allowed choices are frontend, backend.`

I logged out my argv, and the script is correctly getting all the args:
```
args:  [
  'C:\\Program Files\\nodejs\\node.exe',
  {script path},
  'deploy',
  '--only',
  'frontend'
]
```

If I don't pass an only flag at all (e.g, `myScript deploy`), I get the handler console log statement:
`only deploying: undefined`

This is happening on Windows, with Node v16.4.2,

Any idea what might be happening? Am I constructing something wrong, or is this a bug?

shadowspawn:
(Quick answer without trying code, so may be some errata.)

1) You have accidentally created a boolean option and the choices will be broken. Add an option-argument to the declaration:

```js
 new Option("--only <partial>"...
```

2) Your command does not have any command-arguments, at least in the example. So your action handler should look like:

```js
.action((options)...
```

shadowspawn:
(Good details in the description, thanks.)

NathanC:
Perfect, that solved my issue! Thanks for the prompt response. Great library by the way, I really like the help messages it generates.

Loque18:
I'm having the same issues 

```
const { program } = require('commander');

program.name('template scripts').version('1.0.0').description('Set of scripts to bootstrap development');

program
    .command('gen-template')
    .description('Generate a template page')
    .argument('<page-name>', 'Page name')
    .option('-o, --output <folder>', 'Output folder')
    .action((pageName, options) => {
        console.log('pageName', pageName);
        console.log('options', options);
    });
    
   program.parse();
```
npm run scripts gen-template --output=./pages test

I always get 
pageName test
options {}

shadowspawn:
@Loque18 
[...]

Issue #1761, opened 2022-06-25
Options are always empty


PsotnyPerun:
Hi, 
I've been trying both my own code and code from the samples, and whatever I do options are always empty. 
But let's stick with the examples - I've been trying using this one: https://github.com/tj/commander.js/blob/master/examples/pizza. I've been trying to pass a custom string to --cheese option but whatever I do, options in program.opts() are always empty. They only get populated if there is a default value specified in option() function. 

On a Win machine, I've been running the script with npm run start --cheese=my_value or npm run start -c my_value and nothing would work. Even if the options are bools, adding --peppers still wouldn't work. 

I've been trying both the latest version (9.3.0) and some old ones from a year ago. 

Am I missing something obvious here?
Thanks!

shadowspawn:
Lots of people encounter this problem! The issue is invoking your script through `npm`. It is silently eating the options and they do not reach your script. I'll run through some general tips and the way to solve it.

A useful debugging step is to check what is reaching your script: `console.log(process.argv)`

Try running your script directly with node to check it is working as expected:
```
node script.js --cheese=my_value
```

For calling through `npm run`, you can use `--` to tell npm to stop looking for options. The `--` is a commonly supported way to tell something looking for options to stop looking! (Commander supports this too.) You can put the `--` anywhere after the options intended for npm and before the options intended for your script. 

```
npm run -- start --cheese=my_value
npm run start -- --cheese=my_value
```

The npm documentation page shows the `--` in the synopsis:

```
npm run-script <command> [-- <args>]
```

https://docs.npmjs.com/cli/v8/commands/npm-run-script

PsotnyPerun:
Oh, I see what you mean - it does make sense now. 
Adding -- when calling `npm run` did solve the issue.

Thank you for your help and for getting back to me so quick.

shadowspawn:
An answer was provided, and no further activity in a month. Closing this as resolved.

Feel free to open a new issue if it comes up again, with new information and renewed interest.

kunanyi:
Oh thank God I found this lol

Maybe consider adding the npm/-- trick to the docs?  I almost gave up :)

PR #1872, opened 2023-04-30
Add npm run-script to README

shadowspawn:
We often get people reporting problems because npm eats the options before their program is called. I wondered about a FAQ or common problems section, but we already have some similar tips scattered through the "Bits and pieces". So I have added another section for `npm run-script`. 

See: #1262 #1654 #1761 https://github.com/tj/commander.js/issues/1769#issuecomment-1197187862 #1820 #1869


abetomo (approved):
👍

PR #1877, opened 2023-05-08
export suggestSimilar (incl. type definition)

jpilgrim:
# Pull Request

## Problem

Similar to internal commander functionality I needed a suggestSimilar function in my client (using commander as a dependency).

## Solution

Instead of writing the function twice or using another dependency, I simply exported the suggestSimilar function already included in commander.js. That includes its type definitions.

This is a perfect solution, since I needed exactly the format produced by suggestSimilar -- my client is a CLI after all, which is why I use commander.js in the first place.

Just one note: I also updated readme.md and copied the section into readme_zh-CN.md (in english) to simplify translation.

## ChangeLog

- export suggestSimilar to be used in client code as well

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -46,6 +46,7 @@ Read this in other languages: English | [简体中文](./Readme_zh-CN.md)
     - [Legacy options as properties](#legacy-options-as-properties)
     - [TypeScript](#typescript)
     - [createCommand()](#createcommand)
+    - [suggestSimilar()](#suggestsimilar)
     - [Node options such as `--harmony`](#node-options-such-as---harmony)
     - [Debugging stand-alone executable subcommands](#debugging-stand-alone-executable-subcommands)
     - [Display error](#display-error)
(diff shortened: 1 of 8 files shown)


shadowspawn:
I am pleased that you think it is useful, but my initial reaction is that I do not want to expose the routine as it is tailored for internal use and not core to what Commander offers. There are multiple libraries that are focused on suggestions.

I'll leave it for a while and think about it some more.

(For interest, the sort of opposite direction I had considered was whether people would want to supply their own implementation. I decided it was not particularly likely so left out, and reconsider if requested and upvoted.)

jpilgrim:
@shadowspawn I understand your concerns. In order to make the function a little bit more generic (i.e., less "tailored for internal use") I have extracted the core part into a new function computeSimilar. I also added tests. Now, only the more general function computeSimilar is exposed.

Commander does not use any dependencies, which is great. For small command line tools, I also want to keep the number of dependencies small. Suggesting a similar word (e.g., for wrongly written filenames) is a typical CLI feature (in a GUI, this usually does not happen as the file is picked from a list). This is a very typical case for CLIs, and I would hate to add another dependency (or otherwise duplicate the solution already found in commander.

If someone needs a more specific solution, one can always use another packages as well.

shadowspawn:
Thanks for the suggestion. This is not something I wish to make public at this time.

I think this is the first time the suggestion code has come up for discussion. If there are future issues and demand this can be revisited.

Issue #1880, opened 2023-05-11
Revisit TypeScript definitions for dual cjs/esm


shadowspawn:
I think we got this technically wrong in #1704. Basically, we need a separate type definition file for each of cjs and esm. This is a common mistake and in our defence the documentation may have got clearer in the last year!

From https://www.typescriptlang.org/docs/handbook/esm-node.html

> It’s important to note that the CommonJS entrypoint and the ES module entrypoint each needs its own declaration file, even if the contents are the same between them. Every declaration file is interpreted either as a CommonJS module or as an ES module, based on its file extension and the `"type"` field of the `package.json`, and this detected module kind must match the module kind that Node will detect for the corresponding JavaScript file for type checking to be correct. 

Looking back, @ChocolateLoverRaj did make a second file when investigating, and just reexporting the cjs types is hopefully sufficient like in comment: https://github.com/tj/commander.js/issues/1703#issuecomment-1068713960

The TypeScript gurus actually suggest not having a `"types"` entry and just having type definition sitting beside the entry point. However, we have historically had the definitions tucked away in a folder and I'm ok with continuing with that.

Related links:
- https://github.com/microsoft/TypeScript/issues/50058#issuecomment-1404411380
- https://arethetypeswrong.github.io/?p=commander%4010.0.1
- https://github.com/microsoft/TypeScript/pull/51973#issuecomment-1376001267

shadowspawn:
Released in v11

PR #1886, opened 2023-05-14
Add separate type file for esm per TypeScript guidelines

shadowspawn:
# Pull Request

## Problem

TypeScript wants a type definition file which matches the implementation file, like foo.js + foo.d.ts, bar.mjs + bar.d.mts. 

See #1880 for more detail.

## Solution

Adding the file was easy. Working out how to just reexport the existing types was tricky! I am using:

```
// This is a bit indirect. There is not an index.js, but TypeScript will look for index.d.ts for types.
export * from './index.js';
```

The direct `./esm.mjs` entry is stale, but tidy it up too.

## ChangeLog

- changed: refine "types" exports for ESM to follow TypeScript guidelines

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -14,7 +14,7 @@ const javascriptSettings = {
 };
 
 const typescriptSettings = {
-  files: ['*.ts'],
+  files: ['*.ts', '*.mts'],
   parserOptions: {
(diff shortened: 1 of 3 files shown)


abetomo (approved):
Awesome!

PR #1887, opened 2023-05-14
Deprecate import from commander/esm.mjs

shadowspawn:
I noticed the old `commander/esm.mjs` entry point while updating the exports. Add a note that it is deprecated.

(The plugin in the editor that is updating the table of contents has tweaked a couple entries with `*`. I'll check they still work from GitHub markdown. Update: links are working.)


abetomo (approved):
👍

PR #1888, opened 2023-05-27
Drop node14

shadowspawn:
Node 14 is out of LTS. This PR updates to minimum of Node 16.

I have updated the CHANGELOG ready for Commander v11 with this change and the pending releases. I put in date of 16 June as I'll be away this coming weekend, but could be sooner or later if suits you better @abetomo


abetomo (approved):
👍

PR #1895, opened 2023-06-20
Docs Update: Improved Function and Class Descriptions

coudron:
This pull request offers concise yet comprehensive updates to function and class documentation.

This PR contains no functional changes, only documentation improvements.

### DIFF
--- a/lib/argument.js
+++ b/lib/argument.js
@@ -2,6 +2,11 @@ const { InvalidArgumentError } = require('./error.js');
 
 // @ts-check
 
+/**
+ * Represents a command argument with a name, description, and various options.
+ * The default is that the argument is required, and you can explicitly indicate this with <> around the name.
+ * Put [] around the name for an optional argument.
+ */
 class Argument {
   /**
    * Initialize a new command argument with the given name and description.
@@ -96,6 +101,12 @@ class Argument {
    * @return {Argument}
    */
 
+  /**
+   * Sets the allowed choices for the argument and returns the current instance of the class.
+   * @param {Array} values - An array of allowed choices.
+   * @returns {Object} - The current instance of the class.
+   * @throws {InvalidArgumentError} - If the argument is not one of the allowed choices.
+   */
   choices(values) {
     this.argChoices = values.slice();
     this.parseArg = (arg, previous) => {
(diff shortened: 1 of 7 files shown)


shadowspawn:
Has this been generated by OtterDoc?

There are some potentially useful changes, but also several places where a new JSDoc is added instead of updated,  comments in the code hoisted into the JSDoc, and `throw` incorrectly described as being thrown.

coudron:
@shadowspawn Yeah, these were generated using OtterDoc. We just started developing it and looking to get feedback and determine how useful people think it is.

I'm wondering if people consider this too verbose with its additions?  

Yeah, it's hoisting some of the params up from the outer scope. That's probably not the desired way to document it.

The `throws` appeared correct to me? I appreciate it when docs tell me what is possibly thrown, but curious how others view it.

Feedback welcome. Hope this PR could be massaged into being helpful as it's not perfect on the first pass.

lib/argument.js
```diff
@@ -96,6 +101,12 @@ class Argument {
    * @return {Argument}
    */
 
+  /**
+   * Sets the allowed choices for the argument and returns the current instance of the class.
+   * @param {Array} values - An array of allowed choices.
+   * @returns {Object} - The current instance of the class.
+   * @throws {InvalidArgumentError} - If the argument is not one of the allowed choices.
```
shadowspawn, line 108: This routine contains a throw in the code, but does not actually throw an error that can be caught when it is called.
1) The throw is in a custom`.parseArg` function which is called later when the command-line arguments are being parsed.
2) The error is actually caught during processing and repackaged.

lib/help.js
```diff
@@ -26,6 +26,11 @@ class Help {
    * @returns {Command[]}
    */
 
+  /**
```
shadowspawn, line 29: There is already an existing JSDoc for this routine. This should presumably be a modification rather than an addition. (This is true of many of the additions.)

shadowspawn:
I added a couple of comments to explain two of the issues I noticed. 

I think the suggestions would be welcome while I was actually adding new code or working on a routine. But not as a change-lots PR.

In this project we have manually maintained dual documentation in the JavaScript files and the TypeScript interface definition files. I think there is too much work to take the generic suggested improvements in the JavaScript and fix up the problems, and reproduce the changes.

coudron:
Understood. Appreciate the feedback.

This is a good use-case to understand and make adjustments to OtterDoc. I think a lot of people feel the same way.

Issue #1885, opened 2023-05-13
Main help output doesn't show custom `usage` of commands


yume-chan:
Version: `"commander": "^10.0.1",`

There are several cases I'm using the `.usage()` option:

* When a command only has limited number of options, I want to display all of them in `usage`, instead of `[options]`

    ```js
    program
        .command("devices")
        .usage("[-l]")
        .option("-l", "long output", false)
    ```

* When a command requires `--` before arguments because otherwise commander parses them as options

    ```js
    program
        .command("shell [args...]")
        .description(
            "run remote shell command (interactive shell if no command given). `--` is required before command name."
        )
        .usage("[options] [-- <args...>]")
    ```

It's working in each command's individual help message, like 

```text
> node .\esm\index.js help devices
Usage: tango-cli devices [-l]

Options:
  -l          long output (default: false)
  -h, --help  display help for command
```

But not in the main help

```text
> node .\esm\index.js help
Usage: tango-cli [options] [command]

Options:
  -H <host>                  name of adb server host (default: "127.0.0.1")
  -P <port>                  port of adb server (default: 5037)
  -h, --help                 display help for command

Commands:
  devices [options]
  shell [options] [args...]  run remote shell command (interactive shell if no       
                             command given). `--` is required before command name.   
  help [command]             display help for command
```

In particular the second case might confuse the users if they don't know `help [command]` will display a different usage.

shadowspawn:
There was a PR recently to show the custom usage in the list of subcommands. I was concerned the custom help might be long if displaying all the options, and not work well in the list of subcommands.

Interestingly, you qualified your description with "When a command only has limited number of options, I want to display all of them in usage, instead of [options]" so would probably work fine for you.

You can modify the behaviour yourself to use the custom usage by customising the help:
https://github.com/tj/commander.js/pull/1853#issuecomment-1465034399

shadowspawn:
On a separate but related note, for your `shell` subcommand, you might be interested in `.passThroughOptions()` which is intended to simply this scenario and remove the need for the end user to use  `--`. You would also need to `.enablePositionalOptions()` on the program, so it would change where the `-H` and `-P` can go on the command-line.

- https://github.com/tj/commander.js#parsing-configuration
- https://github.com/tj/commander.js/blob/master/examples/pass-through-options.js

yume-chan:
Thank you! Now I'm adding `usage` to all commands and using `subcommandTerm` to customize the display.

[...]

PR #1896, opened 2023-06-25
Add example of displaying custom usage in subcommand list

shadowspawn:
# Pull Request

## Problem

People who define custom usage for their subcommands are likely to want to display them in the list of subcommands. 

See: #1853 #1885

## Solution

Add an example file which shows how to use custom usage in the list of subcommands.


abetomo (approved):
👍

shadowspawn:
Example added in Commander v11.1.0

Issue #1900, opened 2023-07-02
Consider supporting async option/argument parsers


aweebit:
I would like to parse an argument that is expected to equal an index to an array that can only be read asynchronously. Currently, I see no better way than to use an async argument parser and handle the promise it returns in the action handler. Here is a simple example with [lowdb](https://github.com/typicode/lowdb):

```javascript
import { Command, InvalidArgumentError } from 'commander';

import { Low } from 'lowdb';
import { JSONFile } from 'lowdb/node';

const adapter = new JSONFile('test.json');
const db = new Low(adapter, ['foo', 'bar']);
await db.read();
await db.write();

const indexParser = async (val) => {
  if (val) {
    val = Number(val);
    if (Number.isInteger(val) && val >= 0) {
      await db.read();
      if (val < db.data.length) return val;
    }
  }
  throw new InvalidArgumentError('Index out of bounds');
};

const program = new Command()
  .argument('index', 'Array element to access', indexParser)
  .action(function(indexPromise) {
    indexPromise.then(async function(index) {
      await db.read();
      console.log(db.data[index]);
    });
  });

// go modify test.json manually or whatever...

setTimeout(() => program.parse(), 10000);
```

Obviously, it would be better and more intuitive if the action handler execution were deferred until after the promise is resolved.

Possible implementations:
- explicit, something along these lines:

    ```javascript
    program.addArgument(new Argument('index').argParser(indexParser).async())
    ```
- defer action handler execution whenever parser returns a thenable object
  - if a certain new method was called on the command, e.g. `program.respectAsyncArgParsers(true)`
  - or if the `parseAsync()` method was called (breaking)

shadowspawn:
It is quite tricky supporting mixed sync and async, but it is something I am interested in if if can be done in a way that does not require changing too much code.

A suggestion about your example work-around. You can have an async action handler, in which case you call `program.parseAsync()` instead of `program.parse()`. (I don't think the `setTimeout` is helping much and you can use await in the action handler.)

The two callbacks that currently support async are the action handler, and the event hook.

aweebit:
I only added `setTimeout()` to better illustrate the problem, namely that `db.data` might have changed, and that the only way to find out is by calling `db.read()` at the place where we need it.

aweebit:
I have come up with a very simple solution thanks to your mention of event hooks.

```javascript
async function awaitHook(thisCommand, actionCommand) {
  actionCommand.processedArgs = await Promise.all(
[...]

PR #1902, opened 2023-07-06
Add Command.awaitHook(), Argument.chainArgParserCalls() and Option.chainArgParserCalls()

aweebit:
# Pull Request

## Problem

#1900

At the moment, [a few lines of extra code](https://github.com/tj/commander.js/issues/1900#issuecomment-1622757806) are required to enable intuitive use of asynchronous custom processing of arguments and option-arguments by means of `async` argument parsers.

## Solution

The extra code is wrapped in the `awaitHook()` method of the `Command` class. It additionally awaits values that the custom processing is not applied to, namely default argument values as well as default and implied option values.

A `chainArgParserCalls()` method causing thenables returned from `parseArg()` to be chained is added to both `Argument` and `Option` classes. This is relevant when `parseArg()` is called with second argument supplied, which is the case for variadic arguments and options and repeated non-variadic options.

### DIFF
--- a/lib/command.js
+++ b/lib/command.js
@@ -1211,16 +1211,16 @@ Expecting one of '${allowedValues.join("', '")}'`);
    * @api private
    */
 
-  _handleChainError(target, value, handleError) {
+  _catchChainError(target, value, handleError) {
     if (target.chained && typeof value?.then === 'function') {
-      value.then(x => x, handleError);
+      return value.then(x => x, handleError);
     }
+
+    return value;
   }
 
   /**
    * Internal implementation shared by ._processArguments() and option and optionEnv event listeners.
-   *
-   * @api private
    *
    * @param {Argument|Option} target
    * @param {*} value
@@ -1239,14 +1239,14 @@ Expecting one of '${allowedValues.join("', '")}'`);
         (result) => {
           result = target.parseArg(value, result);
           // call with result and not parsedValue to not catch handleError exception repeatedly
-          this._handleChainError(target, result, handleError);
+          result = this._catchChainError(target, result, handleError);
           return result;
         }
       );
     } else {
       try {
         parsedValue = target.parseArg(value, previous);
-        this._handleChainError(target, parsedValue, handleError);
+        parsedValue = this._catchChainError(target, parsedValue, handleError);
       } catch (err) {
         handleError(err);
       }
--- a/lib/command.js
+++ b/lib/command.js
@@ -58,6 +58,14 @@ class Command extends EventEmitter {
     this._showHelpAfterError = false;
     this._showSuggestionAfterError = true;
 
+    this._chainedOptionEntryFilter = ([key]) => {
+      // TODO: Maybe replace by an attribute-name-to-option map
+      const option = this.options.find(
+        option => option.attributeName() === key
+      );
+      return option && option.chained;
+    };
+
     // see .configureOutput() for docs
     this._outputConfiguration = {
       writeOut: (str) => process.stdout.write(str),
@@ -551,7 +559,7 @@ Expecting one of '${allowedValues.join("', '")}'`);
     // handler for cli and env supplied values
     const handleOptionValue = (val, invalidValueMessage, valueSource) => {
       const handleError = (err) => {
-        if (err.code === 'commander.invalidArgument') {
+        if (err?.code === 'commander.invalidArgument') {
           const message = `${invalidValueMessage} ${err.message}`;
           this.error(message, { exitCode: err.exitCode, code: err.code });
         }
@@ -969,16 +977,24 @@ Expecting one of '${allowedValues.join("', '")}'`);
 
     // Throws errors from custom processing.
     // Also prevents ERR_UNHANDLED_REJECTION by leaving no uncaught promises.
-    const toAwait = this.processedArgs.slice(0, this.args.length);
-    Object.entries(this.opts()).forEach(([key, value]) => {
-      const source = this.getOptionValueSource(key);
-      if (source === 'cli' || source === 'env') {
-        toAwait.push(value);
-      }
-    });
-    Object.values(this._overwrittenOptionValues).forEach((values) => {
-      values.forEach(value => toAwait.push(value));
-    });
+    const toAwait = this.processedArgs
+      .slice(0, this.args.length) // remove defaults
+      .filter((arg, i) => {
+        return this._args[i].chained;
+      });
+    Object.entries(this.opts())
+      .filter(this._chainedOptionEntryFilter)
+      .forEach(([key, value]) => {
+        const source = this.getOptionValueSource(key);
+        if (source === 'cli' || source === 'env') {
+          toAwait.push(value);
+        }
+      });
+    Object.entries(this._overwrittenOptionValues)
+      .filter(this._chainedOptionEntryFilter)
+      .forEach(([key, values]) => {
+        values.forEach(value => toAwait.push(value));
+      });
     await Promise.all(toAwait);
 
     return this;
@@ -1218,7 +1234,7 @@ Expecting one of '${allowedValues.join("', '")}'`);
   _processArguments() {
     const myParseArg = (argument, value, previous) => {
       const handleError = (err) => {
-        if (err.code === 'commander.invalidArgument') {
+        if (err?.code === 'commander.invalidArgument') {
           const message = `error: command-argument value '${value}' is invalid for argument '${argument.name()}'. ${err.message}`;
           this.error(message, { exitCode: err.exitCode, code: err.code });
         }
--- a/lib/command.js
+++ b/lib/command.js
@@ -58,6 +58,14 @@ class Command extends EventEmitter {
     this._showHelpAfterError = false;
     this._showSuggestionAfterError = true;
 
+    /** @type {boolean | undefined} */
+    this._asyncParsing = undefined;
+    /** @type {boolean | undefined} */
+    this._awaitHook = undefined;
+    /** @type {{ preSubcommand: number, postArguments: number } | null} */
+    this._awaitHookIndices = null;
+    this.awaitHook();
+
     // see .configureOutput() for docs
     this._outputConfiguration = {
       writeOut: (str) => process.stdout.write(str),
@@ -400,7 +408,7 @@ class Command extends EventEmitter {
    */
 
   hook(event, listener) {
-    const allowedValues = ['preSubcommand', 'preAction', 'postAction'];
+    const allowedValues = ['preSubcommand', 'postArguments', 'preAction', 'postAction'];
     if (!allowedValues.includes(event)) {
       throw new Error(`Unexpected value for event passed to hook : '${event}'.
 Expecting one of '${allowedValues.join("', '")}'`);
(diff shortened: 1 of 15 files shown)


shadowspawn:
Interesting that this implicitly supports defaults which are promises.

I am not certain this is robust enough. Some code has run and tested the promises rather than the values, like the processing of implied values. But the code I looked though only cared about the value being defined, and Commander can't make assumptions about what a custom parser returns, so may actually be ok.

package-lock.json
```diff
@@ -6,7 +6,7 @@
   "packages": {
     "": {
       "name": "commander",
-      "version": "10.0.1",
+      "version": "11.0.0",
```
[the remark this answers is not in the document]
shadowspawn: Oh, this in in `package.lock`. Should be 11. Ok to leave in.

lib/command.js
```diff
@@ -412,6 +412,27 @@ Expecting one of '${allowedValues.join("', '")}'`);
...
   }
 
+  /**
+   * Add hook to await argument and option values before calling action handlers for this command and its nested subcommands.
+   * Useful for asynchronous custom processing (parseArg) of arguments and option-arguments.
+   */
+
+  awaitHook() {
+    this.hook('preAction', async function awaitHook(thisCommand, actionCommand) {
+      actionCommand.processedArgs = await Promise.all(
```
shadowspawn, line 422: `processedArgs` may include an element which is an array for a variadic argument. I am guessing this is not supported by `Promise.all` so might need more complex processing.

lib/command.js
```diff
@@ -412,6 +412,27 @@ Expecting one of '${allowedValues.join("', '")}'`);
...
+        actionCommand.processedArgs
+      );
+
+      for (const [key, value] of Object.entries(actionCommand.opts())) {
+        actionCommand.setOptionValueWithSource(
+          key, await value, actionCommand._optionValueSources[key]
```
shadowspawn, line 428: 1) The option value may be an array of values for a variadic option.

2) I think each call to await will delay execution until the next tick. I suggest only call for `thenable` values.

3) I would slightly prefer not to call `setOptionValueWithSource` on every value, which also fits in with testing for `thenable` in above item.

shadowspawn: There is an existing test for `thenable` here: https://github.com/tj/commander.js/blob/4ef19faac1564743d8c7e3ce89ef8d190e1551b4/lib/command.js#L1192

shadowspawn:
I am intrigued whether this will be good enough to build in rather than be opt-in! See how it goes...

shadowspawn:
> At the moment, [a few lines of extra code](https://github.com/tj/commander.js/issues/1900#issuecomment-1622757806) are required 

The few lines will cover many programs, but it is harder building the code into Commander where we have to consider all the ways arguments and options may be declared and used.

shadowspawn:
A paranoid thought. Are repeated (non-variadic) options a problem because they will create dangling promises? 

Suppose lookup does an async parseArg. The promise for `bar` will overwrite the promise for `foo` in the option values.
```
my-utility --lookup foo --lookup bar
```

aweebit:
[4a6e07c](https://github.com/tj/commander.js/pull/1902/commits/4a6e07c16195b8d13e72377ecf3cabc11ffca21b) introduces a breaking change which is necessary to enable asynchronous custom processing of variadic arguments and options.

lib/command.js
aweebit: 1. Due to the "previous value" paradigm of custom variadic option processing, the only possible sources for such an array are default and implied values which are assumed to be preprocessed and should ideally be left untouched. If the array really needs to be awaited elementwise, one could simply use `Promise.all(arr)` as the default / implied value.
2. Solved by [4738e3c](https://github.com/tj/commander.js/pull/1902/commits/4738e3c3ea63ce8c02f85915d13b5de6d31c1df7).
3. Solved by [4738e3c](https://github.com/tj/commander.js/pull/1902/commits/4738e3c3ea63ce8c02f85915d13b5de6d31c1df7).

aweebit:
> A paranoid thought. Are repeated (non-variadic) options a problem because they will create dangling promises?
> 
> Suppose lookup does an async parseArg. The promise for `bar` will overwrite the promise for `foo` in the option values.
> 
> ```
> my-utility --lookup foo --lookup bar
> ```

Dangling pointers don't really pose any threat as long as their exceptions are caught, which I took care for in [4738e3c](https://github.com/tj/commander.js/pull/1902/commits/4738e3c3ea63ce8c02f85915d13b5de6d31c1df7) by running `parsedValue.catch(handleError)` on [line 1181](https://github.com/tj/commander.js/pull/1902/commits/4738e3c3ea63ce8c02f85915d13b5de6d31c1df7#diff-d8db36db6bfb32591336006f1dc1adee9e2caf7b021469995aa687c5c5723d0cR1181).

aweebit:
> [4a6e07c](https://github.com/tj/commander.js/pull/1902/commits/4a6e07c16195b8d13e72377ecf3cabc11ffca21b) introduces a breaking change which is necessary to enable asynchronous custom processing of variadic arguments and options.

On second thought, there might be use cases where custom processing is expected to produce promises without chaining them, for example
```javascript
program.argument('<file...>', 'files', (filename, promise) => {
  const newPromise = new Promise((resolve, reject) => {
    // Do something with filename...
  });
  return promise ? Promise.all([promise, newPromise]) : newPromise;
});
```
So maybe this change does not have to be breaking and should be made available via a method, e.g. `program.chainParsedVariadicPromises()`.

aweebit:
Or via an argument method for that matter: `new Argument('<file...>', 'files', parseFiles).chainThenables()`.

shadowspawn:
Good point that promises and variadic only happen with a custom parseArgs, and author is in control of what is stored. (This was a design decision when adding variadic as parseArgs can be more than just a coercion.)

aweebit:
> Dangling pointers don't really pose any threat as long as their exceptions are caught, which I took care for in [4738e3c](https://github.com/tj/commander.js/pull/1902/commits/4738e3c3ea63ce8c02f85915d13b5de6d31c1df7) by running `parsedValue.catch(handleError)` on [line 1181](https://github.com/tj/commander.js/pull/1902/commits/4738e3c3ea63ce8c02f85915d13b5de6d31c1df7#diff-d8db36db6bfb32591336006f1dc1adee9e2caf7b021469995aa687c5c5723d0cR1181).

Line 1181 turned out to be not enough since `handleError` also throws, and we need a way to let users catch those errors. I managed to solve this in [c0b7e74](https://github.com/tj/commander.js/pull/1902/commits/c0b7e74f081faf5939c72a56440b907e0aee9b01).

After spending a lot of time figuring out these unhandled rejections, I now understand your concern about repeated non-variadic options better. The code seems to pass the test I wrote for this case without errors, but I honestly have no idea why.

aweebit:
Oh, now I see it is because we treat non-variadic options the same way as variadic when they are repeated and a parser is provided, so as long as we chain calls to `parseArg()`, there will be no problem. It is something to consider though because I am planning to make the chaining optional.

aweebit:
Problem solved by creating an object with backups of overwritten values. Should it just be restricted to thenable values?

aweebit:
The changes are still breaking because `parseAsync()` can now throw, e.g.
```javascript
let promise;

program
  .argument('[arg]', 'desc', () => {
    promise = Promise.reject(new Error());
    return promise;
  })
  .action(() => {});

try {
  await program.parseAsync(['1'], { from: 'user' });
  console.error('did not catch');
} catch (err) {
  console.error('caught');
} finally {
  promise.catch(() => {});
}
```

aweebit:
Ok, now asynchronous custom processing (`awaitHook()`) and parseArg call chaining (`chainArgParserCalls()`) are two completely independent things, and neither affects `parseAsync()` when not used, so there are no more breaking changes.

lib/command.js
```diff
@@ -1126,6 +1184,59 @@ Expecting one of '${allowedValues.join("', '")}'`);
...
 
+  /**
+   * @param {Argument|Option} target
+   * @param {*} value
+   * @param {Function} handleError
+   * @api private
+   */
+
+  _handleChainError(target, value, handleError) {
+    if (target.chained && ['then', 'catch'].every(
```
shadowspawn, line 1195: Why test for `catch` too, instead of just "thenable"?

aweebit: Because `catch` is called right after, and one could theoretically supply a thenable without a `catch` function.. The call could be replaced by `value.then(null, handleError)` or `value.then(x => x, handleError)` though.

shadowspawn: Oh right, thanks.

shadowspawn:
What use case is `chainArgParserCalls` for?

I was wondering about dropped promises, but perhaps that is relatively easy for the author to handle.

One of the custom parse examples is:
```
function collect(value, previous) {
  return previous.concat([value]);
}
program.option('-c, --collect <value>', 'repeatable value', collect, []);
```

If we make that async, could it just be:
```
async function collect(value, previous) {
   const previousArray = await previous;
   const newValue = await someAsyncCall(value);
   return previousArray.concat(newValue);
}
```

If we were were doing a simple one-value coerce/parse, but similarly wanting to avoid dangling promises:
```
async function coerce(value, previous) {
   await previous; // allow previous promise to settle
   return someAsyncCall(value);
}
```

aweebit:
I guess it is just a shortcut that makes it possible to write async parsers that work with two values just like normal ones, rather than with one value and one promise.

Quite an artificial example, but imagine there is a deeply nested structure somewhere on the server, and we are writing a program that accepts keys needed to access a certain property nested in the structure as arguments. We want to join the keys into one key string the server understands, but also we want all substructurs accessed on the way to the property to satisfy certain criteria for the input to be considered valid, and the only way to check it is by sending requests to the server for all partial key strings constructed along the way. The parser could then look like this:
```javascript
async function parseKeyPart(value, previous) {
  const key = previous ? `${previous}.${value}` : value;
  if (!(await satisfiesCriteria(key))) {
    throw new InvalidArgumentError(
      `Partial key ${key} does not satisfy the criteria`
    );
  }
  return key;
}
```

shadowspawn:
Some musing, I am not sure of how best to approach these, just at the thinking stage.

I don't think the existing hooks are a good match for  settling the option promises at the higher levels of a deeply nested command, which I think are not currently covered. It might make sense to do that in `._dispatchSubcommand()` so the options are settled at each level before calling into the subcommand.

That leaves the options on the leaf command itself and the processed arguments, which are currently covered by the `preAction` implementation.

A complication is that the various places that `_processArguments()` is called, only the action handler branch of code has support for chaining, so won't get consistent coverage without more work.

shadowspawn:
For a user implementation of this, walking up the command chain from `preAction` to settle promises might be fine. The author knows whether they are relying on any option promises to be resolved before their action handler is called, or indeed whether there are any global options at all.

aweebit:
> I don't think the existing hooks are a good match for settling the option promises at the higher levels of a deeply nested command, which I think are not currently covered. It might make sense to do that in `._dispatchSubcommand()` so the options are settled at each level before calling into the subcommand.

Well, then `preSubcommand` seems to be a good match, we just need to make sure the hook is added at each level. I managed to achieve this in my last commit, so please check it out. Also, the new test I added fails when simply _walking up the command chain from `preAction` to settle promises_.

aweebit:
> A complication is that the various places that `_processArguments()` is called, only the action handler branch of code has support for chaining, so won't get consistent coverage without more work.

Could you please elaborate on this? I am not sure I understand the problem.

aweebit:
You probably mean the fact the exception is not caught in the following code, although it is when an action is added:

```js
program
  .argument('arg', 'desc', async() => {
    throw 123;
  })
  .awaitHook();

try {
  await program.parseAsync(['abc'], { from: 'user' });
} catch {
  console.log('caught');
}
```

Well, in that case, adding a new `postArguments` event for hooks that is always triggered right after `_processArguments()` is called might be a good idea.

But also, it is kind of weird that the following snippets produce different results.

```js
program
  .argument('arg', 'desc', () => {
    throw 123;
  });

try {
  program.parse(['abc'], { from: 'user' });
} catch {
  console.log('caught');
}
```

```js
program
  .argument('arg', 'desc', async() => {
    throw 123;
  });

try {
  await program.parseAsync(['abc'], { from: 'user' });
} catch {
  console.log('caught');
}
```

I am thinking whether it would be reasonable to make the behavior of `awaitHook()` the default to get rid of this discrepancy, at the expense of introducing a breaking change. What are your thought on this?

aweebit:
Not sure I will have time for this in the coming weeks, so I just went on and implemented the changes I meant. 

`awaitHook()` is now the default with `parseAsync()`, but it can be disabled by calling `awaitHook(false)`, or applied forcefully to the command and its descendants by calling `awaitHook(true)`. This is a breaking change.

Callbacks are added for both `preSubcommand` and `postArguments` events, the latter being a new feature. They are always pushed to the end of the corresponding `_lifeCycleHooks` arrays. If `awaitHook()` is called multiple times, the old callback is removed before pushing the new one. All of this ensures the flexibility I [mentioned](https://github.com/tj/commander.js/issues/1900#issuecomment-1622757806) in the discussion of the original issue.

The code is much more complex now and will require thorough testing.

shadowspawn:
> The code is much more complex now and will require thorough testing.

Yes. This is getting too complex for my liking.

High level, your idea of using the stored promises from calling `parseArgs` to settle later is an appealing way of adding async support, if there are good places to handle the pause. The hooks support async so pausing "near" hooks is likely to be easier.

I wonder how simple this could be leaving out extras like `chainArgParserCalls`. I'm interested in investigating this myself too, so no worries if this doesn't fit into your ideas or you are don't have time for a while.

Adding a few lines to the "Parsing life cycle and hooks" description, there looks like good places to place the extra calls.

Starting with top-level command (program):

- parse options: parse recognised options (for this command) and remove from args
- parse env: look for environment variables (for this command)
- process implied: set any implied option values (for this command)
- if the first arg is a subcommand
    - *settle promises from async Option.parseArgs*
    - call `preSubcommand` hooks
    - pass remaining arguments to subcommand, and process same way

Once reach final (leaf) command:

- check for missing mandatory options
- check for conflicting options
- check for unknown options
- process remaining args as command-arguments
- *settle promises from async Option.parseArgs*
- *settle promises from async Argument.parseArgs*
- call `preAction` hooks
- call action handler
- call `postAction` hooks

aweebit:
> I wonder how simple this could be leaving out extras like `chainArgParserCalls`.

`chainArgParserCalls()` does not have to do anything with `awaitHook()` and can be left out even now if you do not think it is a useful feature, but I think it is because of [this remark](https://github.com/tj/commander.js/pull/1902#issuecomment-1623993097), so in my opinion we should keep it and maybe make the default when calling `parseAsync()`, since we now have breaking changes anyway.

`awaitHook()` now has three modes, `undefined` being the default one. If it was called with `true` rather than `undefined` as the parameter value, the promises will be awaited even if `parse()` is called. Do you think we should keep it that way, or only ever await when `parseAsync()` is called? If we keep it like this, there is a threat of unhandled promise rejections when calling `parse()` (but this threat is also present when using async hooks with `parse()`, so maybe we should keep it like this for consistency and assume the user knows what they are doing when using both `awaitHook(true)` and `parse()`).

Either way, the same behavior should be implemented for `chainArgParserCalls()` if we make it the default.
- Either we only chain when `parseAsync()` is called if we are in the default `undefined` mode, but allow to force it even on `parse()` calls by calling `chainArgParserCalls(true)`;
- or we only have the `true` and `false` modes and only chain when `parseAsync()` is called if we are in the default `true` mode.

`awaitHook()` needs three modes in either case because the `undefined` mode also means the decision about whether to await the promises is inherited from the nearest ancestor command that is in either the `true` or `false` mode. This makes it possible to only call `awaitHook(false)` on the top-level command to disable the hook for all descendant commands, despite it now being enabled in all of them by default.

So as you can see, some of the new complexity is unavoidable if we want to have the hook enabled by default: `_shouldAwait()` has to be kept, for example. The ways to reduce complexity I see are the following:
- Eliminating `_awaitHookIndices` and therefore the freedom to decide how soon the underlying hooks are called (before, after or in between the other user-supplied `preSubcommand` and `postArguments` hooks). The underlying hooks can be added in the constructor instead if we are okay with them always being the first ones, or in `parseAsync()` / `_parseCommand()` if we want them to be the last
- Or not using the existing hook infrastructure altogether and instead awaiting in `_dispatchSubcommand()` and `_processArguments()` _(what you seem to offer in your "Parsing life cycle and hooks" description)_
- Additionally getting rid of the new `postArguments` event _(but to me, it looks like something that has been missing in the library as the need for it to implement `awaitHook()` using the existing hook infrastructure has shown)_

This is what the "Parsing life cycle and hooks" description should look like if we do not use the existing hook infrastructure but keep the `postArguments` event, and always await before calling the user-supplied hooks:

<details>
<summary>Parsing life cycle and hooks</summary>

> Starting with top-level command (program):
> 
> * parse options: parse recognised options (for this command) and remove from args
> * parse env: look for environment variables (for this command)
> * process implied: set any implied option values (for this command)
> * if the first arg is a subcommand
>   * _if `parseAsync()` was called or the command's (possibly inherited) `awaitHook` mode is `true`_
>     * _settle promises from async `Option.parseArgs`_
>   * call `preSubcommand` hooks
>   * pass remaining arguments to subcommand, and process same way
> 
> Once reach final (leaf) command:
> 
> * check for missing mandatory options
> * check for conflicting options
> * check for unknown options
> * process remaining args as command-arguments
> * _if `parseAsync()` was called or the command's (possibly inherited) `awaitHook` mode is `true`_
>   * _settle promises from async `Option.parseArgs`_
>   * _settle promises from async `Argument.parseArgs`_
> * _call `postArguments` hooks_
> * _if an action handler is provided_
>   * call `preAction` hooks
>   * call action handler
>   * call `postAction` hooks
</details>

shadowspawn:
> * Or not using the existing hook infrastructure altogether ... _(what you seem to suggest in your "Parsing life cycle and hooks" description)_

Yes, that was what I had in mind. I opened a PR with the experiment in #1913

aweebit:
> The ways to reduce complexity I see are the following:
> 
> * Eliminating `_awaitHookIndices` and therefore the freedom to decide how soon the underlying hooks are called (before, after or in between the other user-supplied `preSubcommand` and `postArguments` hooks). The underlying hooks can be added in the constructor instead if we are okay with them always being the first ones, or in `parseAsync()` / `_parseCommand()` if we want them to be the last
> * Or not using the existing hook infrastructure altogether and instead awaiting in `_dispatchSubcommand()` and `_processArguments()` _(what you seem to suggest in your "Parsing life cycle and hooks" description)_
> * Additionally getting rid of the new `postArguments` event _(but to me, it looks like something that has been missing in the library as the need for it to implement `awaitHook()` using the existing hook infrastructure has shown)_

I have implemented the simplifications in #1915.

aweebit:
Using the existing hook infrastructure turned out to be a bad idea giving nothing but unnecessary complexity. Closing in favor of #1915.

PR #1905, opened 2023-07-11
Update lock version

shadowspawn:
# Pull Request

Minor.

## Problem

PR are including an update to `package-lock.json`, because it gets updated to be self-consistent when developers do a clean install.

## Solution

Bump version so self consistent.


abetomo (approved):
👍

Issue #1897, opened 2023-06-26
Add support for custom help groups


smeijer:
I think it'd be helpful if we had a way to group commands and options, to create sections in `--help`.

There's a prior issue to this; https://github.com/tj/commander.js/issues/78, but that's over 9 years old. I'm also aware of `addHelpText` (https://github.com/tj/commander.js/pull/1296), but when using that to group commands, it adds boilerplate due to the need of needing to "hide" commands, and later formatting and properly indenting help ourselves.

What I wish to achieve is something like:

```
Usage: magicbell [options] [command]

Work with MagicBell from the command line

Options:
  -V, --version             output the version number
  -p, --profile <string>    Profile to use (default: "default")
  -h, --help                display help for command

Commands:
  broadcasts                Manage broadcasts
  imports                   Manage imports
  listen                    Listen to events for a users inbox
  metrics                   Manage metrics
  notification-preferences  Manage notification preferences
  notifications             Send and retrieve notifications
  push-subscriptions        Manage push subscriptions
  subscriptions             Manage subscriptions
  users                     Manage all known users

Other commands:
  config                    Display or change config values
  login                     Login to your account
  logout                    Logout of your account
  help                      Display help for a command
```

This separates 'config commands' from 'core commands'. If you'd run `stripe --help` or `gh --help` you can see similar command grouping.

At this moment, I achieve the above via:

```js
for (const command of commands) {
  program.addCommand(command);
}

for (const command of otherCommands) {
  program.addCommand(command, { hidden: true });
}

program.addHelpCommand(false);
const padding = Object.values([...commands, ...otherCommands]).reduce((acc, command) => {
  return Math.max(acc, command.name().length);
}, 0);

program.addHelpText(
  'after',
  `
Other commands:
  ${otherCommands.map((c) => `${c.name().padEnd(padding, ' ')}  ${c.description()}`).join('\n  ')}
  ${'help'.padEnd(padding, ' ')}  Display help for a command
`,
);
```

That's doable, but I think it'd be nice to have native support for it. Something like:

```js
program.addCommand(command, { group: 'Other commands' });
program.addOption(option, { group: 'Global flags' });
```

The `group` can both serve as group title, and grouping key.

shadowspawn:
Thanks for the example, and creative partial work-around.

I see option and command groups in "big" utilities and have wondered about grouping or separator support in Commander, but have not progressed to ideas.

I had one related old comment saved:

[...]

PR #1908, opened 2023-07-15
Help groups for commands and options

shadowspawn:
## Problem

Some people with larger programs would like to break up their commands and options into groups in the help. There have not been a lot of requests for this, but it is something I had noticed as a possible enhancement.

See:  https://github.com/tj/commander.js/issues/374#issuecomment-220442296 #1897

Help groups are common in "big" programs, but big programs often have custom help systems anyway! (e.g. git, npm)

## Solution

I am currently trying to implement some of the API ideas in #1897 to see how they go. If I can't decide which one or ones are best, I might open a poll for feedback on preferred approach.

Two caveats from work so far:
- supporting group for the built-in help option and help command is extra work, except for the "map" api where easy
- it takes a bit of code to sort out the groups. Want to consider whether worth the extra complexity.

## ChangeLog

-----

Still a work in progress, I am coding to try ideas so anything could change. 🛠

What is working currently. Can mix and match.

Map, which implies order of groups displayed, and order of commands and options displayed. Help option and help command are included without special code. Options allow short or long option (or attribute name).
```js
program.commandGroups = new Map([
  ['A Commands:', ['a3', 'a2', 'a1']],
  ['B Commands:', ['b3', 'b2', 'b1']],
  ['Help Commands:', ['help']]
]);

program.optionGroups = new Map([
  ['Animal Options:', ['-r', '--armadillo', '--no-sheep', 'dog', '--no-dog']],
  ['Element Options:', ['--carbon', '-o', '-n']],
  ['Help Options:', ['--help']]
]);
```

Group specified on command and option. This is nice for setting up everything about a command or option in one place. Small downsides are group specified multiple times and have to build Option.

```js
program
  .command('b99')
  .group('B Commands:')
  .action(() => console.log('Called subcommand'));

program
  .addOption(new Option('-s, --sheep').group('Animal Options:'))
```

### DIFF
--- a/lib/command.js
+++ b/lib/command.js
@@ -67,10 +67,12 @@ class Command extends EventEmitter {
     };
 
     this._group = undefined;
+    this._defaultCommandGroup = undefined;
+    this._defaultOptionGroup = undefined;
     /** @type Map<string, string[]> */
-    this.commandGroups = null;
+    this.commandGroups = new Map();
     /** @type Map<string, string[]> */
-    this.optionGroups = null;
+    this.optionGroups = new Map();
 
     this._hidden = false;
     this._hasHelpOption = true;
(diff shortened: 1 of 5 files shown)


shadowspawn:
I got everything going, but was not happy with extra properties and extra code. I did not want to offer them all anyway! Started again with support for just `helpGroup` on Option and Group with extensions for the built-ins.

See #1910

Issue #78, opened 2012-08-19
Show an arbitrary banner message (Feature request)


rstacruz:
Something like:

``` javascript
require('commander')
  .banner("Output options:")
  .option("-o, --output <path>", "Write to path")
  .option("--debug", "Write debug info to file")
  .banner("Style options:")
  .option("--red", "Choose the red style")
  .option("--blue", "Choose the blue style")
  .option("--green", "Choose the green style")
```

Expected output:

```
  Usage: foobar [options]

  Options:

    -h, --help             output usage information
    -V, --version          output the version number

  Output options:

    -o, --output <path>    Write to path
    --debug                Write debug info to file

  Style options:

    --red                  Choose the red style
    --green                Choose the green style
    --blue                 Choose the blue style
```

aka, "option groups" or "banners" or "messages" (adding this in for anyone searching issues :-) )

Issue #1870, opened 2023-04-25
Make it easy to show subcommands before options


xixixao:
We have a main command that only takes --version and --help, but has a bunch of subcommands. In the main help printout we want to show the subcommands first, because those matter. For example of others programs doing something like this see Vercel's CLI.

It doesn't seem to be possible currently to reorder these sections without large reimplementation of help?

Would you be open to a PR?

shadowspawn:
> It doesn't seem to be possible currently to reorder these sections without large reimplementation of help?

Currently, this could be achieved by overriding `formatHelp` and changing the order of the sections. Yes, this is a chunk of code to copy-paste.

shadowspawn:
> Would you be open to a PR?

Certainly willing to discuss or look! No promise to land something. Did you have some approach in particular in mind?

(The current help setup was added in #1365. Part of the intent was to make customisation easier, but of course only some things are actually easy as a result.)

xixixao:
Thanks for checking @shadowspawn ! Roughly the idea would be to pull out each of the currently `// commented` blocks from `formatHelp` into a method. The built-in `formatHelp` would then look like:

```js
[].concat(
  helpUsage(),
  helpCommandDescription(),
  helpArguments(),
  helpOptions(),
  helpGlobalOptions(),
  helpCommands(),
).join('\n')
```

The methods can return strings or lines (lines might allow more customization but might "expose" the implementation more than you'd like?).

To satisfy my original PR request, I could then implement `formatHelp` using these methods, like:

```js
[].concat(
  helpUsage(),
  helpCommands(),
  helpOptions(),
).join('\n')
```

shadowspawn:
My feeling is it is going to refactor the code into quite small pieces without adding a lot of flexibility, or solving issues that have been reported in past. 

I am happy to leave this issue open to see if it attracts support, or if other issues come up which overlap with the idea.

You could try some code for a proof of concept PR if you are keen. (But if it doesn't change my mind in short-term then I'll close the PR and keep the issue open.)

shadowspawn:
I was having a look at some other utilities with lots of commands, and none of git or hg or npm display information about their global options by default in their command-line help.

So a different approach for your case might be to just suppress the options in the top level help. Not saying it is a great solution, but it is simpler to achieve than reordering!

```
import { Command } from 'commander';

const program = new Command('my-command');

program.command('add').description('helpful informative text goes here');
program.command('subtract').description('helpful informative text goes here');
[...]

PR #1910, opened 2023-07-18
Add helpGroup for options and commands

shadowspawn:
## Problem

Some people with larger programs would like to break up their commands and options into groups in the help. There have not been a lot of requests for this, but it is something I had noticed as a possible enhancement.

See: #78 https://github.com/tj/commander.js/issues/374#issuecomment-220442296 #1897

Help groups are common in "big" programs, but big programs often have custom help systems anyway! (e.g. git, npm)

## Solution

Add `.helpGroup()` to Command and Option. Use the help groups as titles in the help.

To support built-in features and external commands, add `{helpGroup}` configuration to:
- `.version()`
- `.helpOption()`
- `.addHelpCommand()`
- external command configuration

```js
program.command('build').helpGroup('Demo Commands:').action(() => {});
program.command('serve', 'external command', { helpGroup: 'Demo Commands:' });
program.addOption(new Option('-e, --example').helpGroup('Demo Options:'));
program.version('1.2.3', '--version', 'Show version', { helpGroup: 'Demo Options:' });
program.helpOption('-s, --support', 'Show help', { helpGroup: 'Demo Options:' });
program.addHelpCommand('help [command]', 'Show help', { helpGroup: 'Demo Commands:' });
```

```console
% node pr-example.js
Usage: pr-example [options] [command]

Demo Options:
  -e, --example
  --version       Show version
  -s, --support   Show help

Demo Commands:
  build
  serve           external command
  help [command]  Show help
```

### DIFF
--- a/lib/command.js
+++ b/lib/command.js
@@ -72,11 +72,14 @@ class Command extends EventEmitter {
     this._helpDescription = 'display help for command';
     this._helpShortFlag = '-h';
     this._helpLongFlag = '--help';
+    this._helpOptionGroup = undefined;
     this._addImplicitHelpCommand = undefined; // Deliberately undefined, not decided whether true or false
     this._helpCommandName = 'help';
     this._helpCommandnameAndArgs = 'help [command]';
     this._helpCommandDescription = 'display help for command';
+    this._helpCommandGroup = undefined;
     this._helpConfiguration = {};
+    this._helpGroupTitle = undefined;
   }
 
   /**
@@ -153,6 +156,7 @@ class Command extends EventEmitter {
     cmd._hidden = !!(opts.noHelp || opts.hidden); // noHelp is deprecated old name for hidden
     cmd._executableFile = opts.executableFile || null; // Custom name for executable file, set missing to null to match constructor
     if (args) cmd.arguments(args);
+    if (opts.helpGroup) cmd.helpGroup(opts.helpGroup);
     this.commands.push(cmd);
     cmd.parent = this;
     cmd.copyInheritedSettings(this);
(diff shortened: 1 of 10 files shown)


shadowspawn:
I am still wondering about:
```
program.version('1.2.3', '--version', 'Show version', { helpGroup: 'Demo Options:' });
program.helpOption('-s, --support', 'Show help', { helpGroup: 'Demo Options:' });
program.addHelpCommand('help [command]', 'Show help', { helpGroup: 'Demo Commands:' });
```

I want to think a bit more about consistency across the API and whether an options-bag is the best approach.

(And have been busy reviewing other PRs.)

sinedied:
It's not the prettiest but I can't think of any better alternative, and it makes sense in regard to the existing api.
It's also probably not something that everyone wants to use all the time, so having it as an extra option seems the best approach to me.

shadowspawn:
(Thanks for comments @sinedied )

shadowspawn:
If I can't decide for now on a built-in way of customising the version option and help option and help command, could maybe leave it to the author! Thinking of this as an interim work-around rather than permanent solution.

```js
  createCommand(name) {
    const cmd =  new MyCommand(name);

    // Set group on the special built-in command.
    if (cmd.name() === 'help')
      cmd.helpGroup('Demo Commands:');

    return cmd;
  }

  createOption(flags, description) {
    const option = super.createOption(flags, description);

    // Set group on the special built-in options.
    if (option.long === '--version' || option.long === '--help')
      option.helpGroup('Demo Options:');

    return option;
  }
}
```

sinedied:
This would be totally fine for me given that's probably an infrequent use case.
Though the `. helpGroup()` method isn't implemented yet, right?

shadowspawn:
> Though the `.helpGroup()` method isn't implemented yet, right?

Correct. Still need that work from this PR.

shadowspawn:
There hasn't been much interest in this feature to date. There is a reasonable chunk of work to do a full solution. Putting this draft away for now.

This issue prompted a couple of refactors so may be easier to customise built-in help option and help command in future: #2006 #2087

I liked "section" approach used in #1897 example implementation which could possibly also help with #1870. (Not included in this draft.)

PR #1913, opened 2023-07-25
Feature/async parse args experiment

shadowspawn:
This is an experiment based on concepts in #1902 to support async Option.parseArgs and Arguments.parseArgs. Blindly store the return value from async calls  to parseArgs as usual (although actually a pending promise), and then settle the promises in a batch before proceeding to next stage of parsing. In particular, settle the promises before calling user hooks or action handler so client code is unlikely to see the promises.

Simplified approach compared to #1902, here:
- no configuration
- built-in support rather than using hooks
- <strike>handling previous value promises from repeated calls to parseArgs left to client</strike>

---
Reminder to add co-author credit if proceed with this PR.
```
Co-authored-by: Wee Bit <aweebit64@gmail.com>
```

### DIFF
--- a/lib/command.js
+++ b/lib/command.js
@@ -1160,6 +1160,7 @@ Expecting one of '${allowedValues.join("', '")}'`);
       processedArgs[index] = value;
     });
     this.processedArgs = processedArgs;
+    return this._settleArgumentPromises();
   }
 
   /**
@@ -1264,12 +1265,36 @@ Expecting one of '${allowedValues.join("', '")}'`);
   _settleOptionPromises() {
     const promises = [];
 
+    // Look through the options for promises from async parseArgs (or other sources).
     Object.entries(this.opts()).forEach(([key, value]) => {
       if (thenable(value)) {
         promises.push(value);
         // Resave value after promise resolves.
         value.then((settledValue) => {
           this.setOptionValueWithSource(key, settledValue, this.getOptionValueSource(key));
+        });
+      }
+    });
+
+    if (promises.length > 0) {
+      return Promise.all(promises);
+    }
+  }
+
+  /**
+   * @api private
+   */
+
+  _settleArgumentPromises() {
+    const promises = [];
+
+    // Variadic array is either ordinary, or a promise for an array.
+    this.processedArgs.forEach((arg, index) => {
+      if (thenable(arg)) {
+        promises.push(arg);
+        // Resave value after promise resolves.
+        arg.then((settledArg) => {
+          this.processedArgs[index] = settledArg;
         });
       }
     });
--- a/lib/command.js
+++ b/lib/command.js
@@ -1308,13 +1308,16 @@ Expecting one of '${allowedValues.join("', '")}'`);
         this.unknownOption(parsed.unknown[0]);
       }
     };
+    const settleAndProcessArguments = () => {
+      const chain = this._settleOptionPromises();
+      return this._chainOrCall(chain, () => this._processArguments());
+    };
 
+    let chain;
     const commandEvent = `command:${this.name()}`;
     if (this._actionHandler) {
       checkForUnknownOptions();
-      this._processArguments();
-
-      let chain = this._settleOptionPromises();
+      chain = settleAndProcessArguments();
       chain = this._chainOrCallHooks(chain, 'preAction');
       chain = this._chainOrCall(chain, () => this._actionHandler(this.processedArgs));
       if (this.parent) {
@@ -1323,12 +1326,10 @@ Expecting one of '${allowedValues.join("', '")}'`);
         });
       }
       chain = this._chainOrCallHooks(chain, 'postAction');
-      return chain;
-    }
-    if (this.parent && this.parent.listenerCount(commandEvent)) {
+    } else if (this.parent && this.parent.listenerCount(commandEvent)) {
       checkForUnknownOptions();
-      this._processArguments();
-      this.parent.emit(commandEvent, operands, unknown); // legacy
+      chain = settleAndProcessArguments();
+      chain = this._chainOrCall(chain, () => { this.parent.emit(commandEvent, operands, unknown); }); // Legacy
     } else if (operands.length) {
       if (this._findCommand('*')) { // legacy default command
         return this._dispatchSubcommand('*', operands, unknown);
@@ -1340,7 +1341,7 @@ Expecting one of '${allowedValues.join("', '")}'`);
         this.unknownCommand();
       } else {
         checkForUnknownOptions();
-        this._processArguments();
+        chain = settleAndProcessArguments();
       }
     } else if (this.commands.length) {
       checkForUnknownOptions();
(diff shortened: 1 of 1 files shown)


aweebit:
What about checking for whether `parseAsync()` has been called? Don't we want to leave the option for the user to handle promises returned from parsers manually at least when `parse()` is used like in my original example in #1900, and thus make the change non-breaking at least for `parse()` calls?

shadowspawn:
Good question. From that point of view, either the change is breaking or non-breaking. It does not matter whether `parseAsync` has been called.

One way of shipping it in a minor version would be to make it opt-in, before making it always-on in the next major version.

A possible fallback is to just add an example file for now! It all looked nice and easy for limited use case like: https://github.com/tj/commander.js/issues/1900#issuecomment-1622685764

aweebit:
Yeah, I know it is going to be breaking in the end anyway, I just meant that by checking for `parseAsync()`, we could ensure nothing changes from users' perspective at least when `parse()` is called. After all, a user might want to handle promises returned from parsers manually, but by awaiting them even if `parse()` is used, we not only deprive them of this option, but also introduce the risk of unhandled promise rejections.

aweebit:
I have submitted a new PR (#1915) which gets rid of all the complexity due to hook usage in #1902 but keeps all the parts missing in your solution that I think are necessary (the motivation is explained in detail in the new PR's description).

shadowspawn:
Add in support for multiple calls to `parseArgs` to resolve previous value (known as `chainArgParserCalls` in #1915). This is what the user would probably expect, and I mainly left it out because it is tricky to implement!

shadowspawn:
> After all, a user might want to handle promises returned from parsers manually, ...

I think we probably only need to support built-in async. If the user wants some variant of the promise resolving they can move the async processing to a hook, so not blocked. Less clever but less magic too.

lib/command.js
```diff
@@ -1249,6 +1233,88 @@ Expecting one of '${allowedValues.join("', '")}'`);
...
+      result = previous.then(resolvedPrevious => target.parseArg(value, resolvedPrevious));
+    } else {
+      try {
+        result = target.parseArg(value, previous);
+      } catch (err) {
+        refineError(err);
+      }
+    }
+
+    if (thenable(result)) result.catch(refineError);
```
aweebit, line 1268: `refineError()` throws, so there will be an unhandled promise rejection. The result of the `catch()` call needs to be assigned to `result` to prevent it.

But if `previous` is already thenable, it is better to call `catch()` in the `previous.then()` callback (and return the result). Otherwise, `catch()` might be called on an already failed promise chain in which an error has already been caught and rethrown (because that is what `refineError()` does), so `refineError()` would be called again which does not make sense.

aweebit: Also, I think calling `result.then(x => x, refineError)` instead of `result.catch(refineError)` is more robust.

aweebit:
> Added support for multiple calls to `parseArgs` to resolve previous value (known as `chainArgParserCalls` in #1915). This is what the user would probably expect, and I mainly left it out until now because it is tricky to implement!

But [there are use cases](https://github.com/tj/commander.js/pull/1902#issuecomment-1623993097) where the user might not want that, so making it configurable now suggests itself. That is why I added `chainArgParserCalls()` :)

aweebit:
> I think we probably only need to support built-in async. If the user wants some variant of the promise resolving they can move the async processing to a hook, so not blocked. Less clever but less magic too.

But all currently available hooks are only called after the available thenable option and argument values are resolved, so this does not seem like a good option to me. The user would have no access to the promises returned by async parsers at all.

In its simplest non-magical form (independent from whether `parse()` or `parseAsync()` was called and inherited via `copyInheritedSettings()`), awaiting configuration requires 10 lines of code which I don't think is such a big deal.
- 1 line for property initialization in constructor
- 1 line for property inheritance in `copyInheritedSettings()`
- 4 lines for configuration function (`await()` in [#1915](https://github.com/tj/commander.js/pull/1915))
- 2 lines for if statements in each of `_settleOptionPromises()` and `_settleArgumentPromises()`

lib/command.js
shadowspawn: > The result of the `catch()` call needs to be assigned to `result`

Oops, yes, good catch.

> Otherwise, `catch()` might be called on an already failed promise chain in which an error has already been caught and rethrown

I think this is correct pattern, but I'll try it before explaining my reasoning.

> Also, I think calling result.then(x => x, refineError) instead of result.catch(refineError) is more robust.

Because we only test for `.then()`? Yes, I have thought about that in past and then forgot. I want to just assume they are promise-like since the primary use-case is for async, but it isn't hard to only rely on `then` and cover a few more edge uses.

shadowspawn:
> But all currently available hooks are only called after the available thenable option and argument values are resolved, so this does not seem like a good option to me. The user would have no access to the promises returned by async parsers at all, even when calling `parse()`.

Yes, the work-around I was suggesting for what are hopefully rare cases was use synchronous `parseArg`.

I understand this is not the flexibility you have proposed.

lib/command.js
shadowspawn: > But if `previous` is already thenable, it is better to call `catch()` in the `previous.then()` callback (and return the result). Otherwise, `catch()` might be called on an already failed promise chain in which an error has already been caught and rethrown (because that is what `refineError()` does), so `refineError()` would be called again which does not make sense.

Took me a while to understand this. I was expecting a long catch chain, but see it can be avoided. 👍

lib/command.js
```diff
@@ -1249,6 +1233,83 @@ Expecting one of '${allowedValues.join("', '")}'`);
...
+    if (thenable(previous)) {
+      result = previous.then(resolvedPrevious => {
+        let innerResult = target.parseArg(value, resolvedPrevious);
+        if (thenable(innerResult)) innerResult = innerResult.then(null, refineError); // .catch
+        return innerResult;
+      });
+    } else {
+      try {
+        result = target.parseArg(value, previous);
+        if (thenable(result)) result = result.then(null, refineError); // .catch
```
aweebit, line 1267: `result` is going to be the next `previous` value, so for `resolvedPrevious` to be correct, `x => x` has to be used instead of `null`.

In #19115, I also went on to use `_chainOrCall()` to follow the DRY principle.

aweebit:
> Yes, the work-around I was suggesting for what are hopefully rare cases was use synchronous `parseArg`.

Did not get it back then, but I see what you mean now 👍

shadowspawn:
This is pretty cool, but I am not happy enough with it:
- uses an indirect approach to achieve async with reprocessing options (and args), not direct support for async functions
- not a nice place to slot in async for program which has neither action handler nor external command
- would add complexity to one of the longer routines and somewhat fragile routines

User can achieve most of the functionality themselves fairly easily because they know how their program is constructed. 

I am currently thinking async action handlers would be a neat example, but not a great built-in feature.

PR #1915, opened 2023-07-25
Add async custom processing support. Add `chainArgParserCalls()` for configuration. Additionally await thenable implicit and default option values and thenable default argument values

aweebit:
This is a rework of #1902 not relying on the hook infrastructure and not introducing `postArguments` hooks, i.e. all simplifications from [this comment](https://github.com/tj/commander.js/pull/1902#issuecomment-1647767306) have been implemented.

#### Added
- `Option.chainArgParserCalls()` and `Argument.chainArgParserCalls()` to control whether to chain `parseArg` calls
- warnings about thenable option and argument values in `.parse()` calls, including default and implied values and ones returned from async argParsers, with a suggestion to use `.parseAsync()` instead (shown unless `process.env.NODE_ENV === 'production'`)

#### Changed
- _Breaking:_ chain `parseArg` calls by default when `.parseAsync()` is called
- _Breaking:_ await thenable option and argument values when `.parseAsync()` is called

#### `Option.chainArgParserCalls()` / `Argument.chainArgParserCalls()`
- When set to `true` (the default), next call to the function provided via `.argParser()` will be chained to its return value if it is thenable.

### New life cycle description

The processing starts with an array of args. Each command processes and removes the options it understands, and passes the remaining args to the next subcommand. The final command calls the action handler.

Starting with top-level command (program):

* _process options for this command (includes calling or chaining provided argParsers[^1])_
  * _recognised options from args_
  * _options based on environment variables_
* set implied option values (for this command)
* _remove recognised options from args_
* _await thenable option values if parsing with `parseAsync()`_
* if the first arg is a subcommand
  * call _or chain_ `preSubcommand` hooks[^1]
  * pass remaining arguments to subcommand, and process same way

Once reach final (leaf) command:

* check for missing mandatory options
* check for conflicting options
* check for unknown options
* process remaining args as command-arguments _(includes calling or chaining provided argParsers[^1])_
* _await thenable argument values if parsing with `parseAsync()`_
* _if an action handler is provided_
  * call _or chain_ `preAction` hooks[^1]
  * call _or chain_ action handler[^1]
  * call _or chain_ `postAction` hooks[^1]
</details>

[^1]: _output a warning suggesting use of `parseAsync()` when parsing with `parse()` and a thenable is returned, unless `process.env.NODE_ENV === 'production'`_

## Some explanations

<h3 id="why-enable-for-parse-async-by-default">Why enable chaining and awaiting for <code>parseAsync()</code> by default?</h3>

A good question, especially because this introduces breaking changes. The decision is motivated by the fact it is most logical for the following code snippets to produce the same results:

```js
program
  .option('--fail', 'desc', () => {
    throw 'exception';
  })
  .argument('[args...]', 'desc', (value, previous) => (
    previous === undefined ? value : previous + value
  ))
  .action(processed => {
    console.log(processed);
  });

program.parse(['1', '2'], { from: 'user' });

try {
  program.parse(['--fail'], { from: 'user' });
} catch (err) {
  console.log('caught', err);
}
```

```js
program
  .option('--fail', 'desc', async() => {
    throw 'exception';
  })
  .argument('[args...]', 'desc', async(value, previous) => (
    previous === undefined ? value : previous + value
  ))
  .action(processed => {
    console.log(processed);
  });

await program.parseAsync(['1', '2'], { from: 'user' });

try {
  await program.parseAsync(['--fail'], { from: 'user' });
} catch (err) {
  console.log('caught', err);
}
```

Without chaining, a promise resolving to `'[object Promise]2'` is logged from inside the action instead of `12` when running the second snippet.

Without awaiting, an unhandled promise rejection is reported instead of logging `caught exception` when running the second snippet.

### Why not enable chaining and awaiting for `parse()`?

- To make the changes as "non-breaking" as possible (although they will be breaking anyway).

  There might be production code like the one in the original example in #1900 somewhere. By not enabling awaiting for `parse()` by default, we make sure this code still works after the upgrade.

- Chaining and awaiting in `parse()` introduces the risk of unhandled promise rejections, so they should not be enabled unless the user explicitly requests it.

### Why is `chainArgParserCalls()` necessary?

- The need for the default `true` mode is motivated [here](https://github.com/tj/commander.js/pull/1902#issuecomment-1626863946) and in the [Why enable chaining and awaiting for `parseAsync()` by default?](#user-content-why-enable-for-parse-async-by-default) section.
- The need for the `false` mode is motivated [here](https://github.com/tj/commander.js/pull/1902#issuecomment-1623993097).

### Why store and await overwritten option values?

An end user might repeat an option even if it is not supposed to be repeated, i.e. if the parser ignores the old value supplied to it in the second parameter. If `chainArgParserCalls(false)` had been called on the option, this can result in unhandled promise rejections if overwritten option values are not awaited.

### Why pass `handleError` to `_parseArg()` and not just the error message for invalid argument errors?

It seems to be more or less an arbitrary design decision to give invalid argument errors thrown from option-argument and command-argument parsers the same `.code` values (`'commander.invalidArgument'`). They could just as well be different, or there could theoretically be other specially handled errors specific only to option-argument parsers or only to command-argument parsers, so it just does not seem right for `_parseArg()` to take over the error handling.

### Why use the word "await" and not "settle"?

To settle a promise means to fulfill or to reject it. That happens internally in the promise callback, an observer can only _await_ the promise.

### …solving similar problems
- #1917 (introduces warnings for async hook and action calls in `parse()` in addition to the warnings for async argParser calls introduced here)

- #1917
- #1931
- #1938
- #1940

- #1955

## See also
- #1913 ([differences](https://github.com/tj/commander.js/pull/1915#issuecomment-1655478510))

### DIFF
--- a/lib/command.js
+++ b/lib/command.js
@@ -924,7 +924,6 @@ Expecting one of '${allowedValues.join("', '")}'`);
 
   parse(argv, parseOptions) {
     this._asyncParsing = false;
-    this._setShouldAwait();
 
     const userArgs = this._prepareUserArgs(argv, parseOptions);
     this._parseCommand([], userArgs);
@@ -953,7 +952,6 @@ Expecting one of '${allowedValues.join("', '")}'`);
 
   async parseAsync(argv, parseOptions) {
     this._asyncParsing = true;
-    this._setShouldAwait();
 
     const userArgs = this._prepareUserArgs(argv, parseOptions);
     await this._parseCommand([], userArgs);
@@ -972,22 +970,6 @@ Expecting one of '${allowedValues.join("', '")}'`);
   await(enabled) {
     this._await = enabled === undefined ? undefined : !!enabled;
     return this;
-  }
-
-  /**
-   * If `_await` is `undefined`, inherit await behaviour or base it on whether `.parseAsync()` has been called on the top-level command.
-   *
-   * @api private
-   */
-
-  _setShouldAwait() {
-    this._shouldAwait = undefined;
-    let cmd = this;
-    do {
-      this._shouldAwait = cmd._await;
-      cmd = cmd.parent;
-    } while (this._shouldAwait === undefined && cmd);
-    this._shouldAwait ??= this._asyncParsing;
   }
 
   /**
@@ -1415,6 +1397,11 @@ Expecting one of '${allowedValues.join("', '")}'`);
    */
 
   _parseCommand(operands, unknown) {
+    // Use explicitly set await mode,
+    // or inherit await behaviour from parent,
+    // or base it on whether .parseAsync() has been called.
+    this._shouldAwait = this._await ?? this.parent?._shouldAwait ?? this._asyncParsing;
+
     let chain;
 
     const parsed = this.parseOptions(unknown);
--- a/lib/command.js
+++ b/lib/command.js
@@ -1488,7 +1488,6 @@ Expecting one of '${allowedValues.join("', '")}'`);
         }
       };
       const processAndAwaitArguments = () => {
-        checkForUnknownOptions();
         this._processArguments();
         chain = this._chainOrCall(chain, () => this._awaitArgumentResavePromises());
       };
--- a/lib/command.js
+++ b/lib/command.js
@@ -1271,8 +1271,16 @@ Expecting one of '${allowedValues.join("', '")}'`);
       // already have a promise, chain callback
       return chain.then(fn);
     }
+
     // callback might return a promise
-    return fn(chain);
+    chain = fn(chain);
+    if (isThenable(chain) && !this._asyncParsing) {
+      chain.then(x => x, () => {}); // prevent unhandled rejection
+      process.exitCode = 1; // additionally indicates to the user that user code (parsers, hooks, action) is not the error source
+      throw new Error(`.parse() is incompatible with async argParsers, hooks and actions.
+Use .parseAsync() instead.`);
+    }
+    return chain;
   }
 
   /**
--- a/lib/command.js
+++ b/lib/command.js
@@ -517,14 +517,21 @@ Expecting one of '${allowedValues.join("', '")}'`);
     const name = option.attributeName();
 
     // store default value
-    if (option.negate) {
+    const setDefaultOptionValue = (value) => {
+      this.setOptionValueWithSource(name, value, 'default');
+      if (!PRODUCTION && !this._asyncParsing && isThenable(value)) {
+        console.warn(`.parse() is incompatible with thenable default option values.
+Use .parseAsync() instead`);
+      }
+    };
+    if (option.defaultValue !== undefined || !option.negate) {
+      setDefaultOptionValue(option.defaultValue);
+    } else {
       // --no-foo is special and defaults foo to true, unless a --foo option is already defined
       const positiveLongFlag = option.long.replace(/^--no-/, '--');
       if (!this._findOption(positiveLongFlag)) {
-        this.setOptionValueWithSource(name, option.defaultValue === undefined ? true : option.defaultValue, 'default');
+        setDefaultOptionValue(true);
       }
-    } else if (option.defaultValue !== undefined) {
-      this.setOptionValueWithSource(name, option.defaultValue, 'default');
     }
 
     // register the option
--- a/lib/command.js
+++ b/lib/command.js
@@ -807,11 +807,11 @@ Expecting one of '${allowedValues.join("', '")}'`);
 
   setOptionValueWithSource(key, value, source) {
     if (!PRODUCTION && !this._asyncParsing && isThenable(value)) {
-        console.warn(`.parse() is incompatible with ${
-          ['cli', 'env'].includes(source)
-            ? 'async argParsers'
-            : 'thenable option values'
-        }.
+      console.warn(`.parse() is incompatible with ${
+        ['cli', 'env'].includes(source)
+          ? 'async argParsers'
+          : 'thenable option values'
+      }.
 Use .parseAsync() instead`);
     }
     if (this._storeOptionsAsProperties) {
--- a/lib/command.js
+++ b/lib/command.js
@@ -63,8 +63,6 @@ class Command extends EventEmitter {
     this._asyncParsing = undefined;
     /** @type {boolean | undefined} */
     this._await = undefined;
-    /** @type {boolean | undefined} */
-    this._shouldAwait = undefined;
 
     // see .configureOutput() for docs
     this._outputConfiguration = {
@@ -114,6 +112,7 @@ class Command extends EventEmitter {
     this._enablePositionalOptions = sourceCommand._enablePositionalOptions;
     this._showHelpAfterError = sourceCommand._showHelpAfterError;
     this._showSuggestionAfterError = sourceCommand._showSuggestionAfterError;
+    this._await = sourceCommand._await;
 
     return this;
   }
@@ -973,6 +972,14 @@ Expecting one of '${allowedValues.join("', '")}'`);
   }
 
   /**
+   * @api private
+   */
+
+  _shouldAwait() {
+    return this._await || (this._await === undefined && this._asyncParsing);
+  }
+
+  /**
    * @return {Promise[]}
    * @api private
    */
@@ -1019,7 +1026,7 @@ Expecting one of '${allowedValues.join("', '")}'`);
    */
 
   _optionalAwait(getPromises) {
-    if (this._shouldAwait) {
+    if (this._shouldAwait()) {
       const toAwait = getPromises();
       if (toAwait.length) return Promise.all(toAwait);
     }
@@ -1374,7 +1381,7 @@ Expecting one of '${allowedValues.join("', '")}'`);
    */
 
   _parseArg(target, value, previous, handleError) {
-    const handleThenableError = this._shouldAwait ? handleError : undefined;
+    const handleThenableError = this._shouldAwait() ? handleError : undefined;
 
     try {
       if (this._shouldChainArgParserCalls(target)) {
@@ -1397,11 +1404,6 @@ Expecting one of '${allowedValues.join("', '")}'`);
    */
 
   _parseCommand(operands, unknown) {
-    // Use explicitly set await mode,
-    // or inherit await behaviour from parent,
-    // or base it on whether .parseAsync() has been called.
-    this._shouldAwait = this._await ?? this.parent?._shouldAwait ?? this._asyncParsing;
-
     let chain;
 
     const parsed = this.parseOptions(unknown);
--- a/lib/command.js
+++ b/lib/command.js
@@ -953,59 +953,6 @@ Expecting one of '${allowedValues.join("', '")}'`);
   }
 
   /**
-   * @api private
-   */
-
-  _optionalAwait(getPromises) {
-    if (this._asyncParsing) {
-      const toAwait = getPromises();
-      if (toAwait.length) return Promise.all(toAwait);
-    }
-  }
-
-  /**
-   * @api private
-   */
-
-  _awaitThenableOptions() {
-    return this._optionalAwait(() => {
-      const chainedPromises = Object.entries(this.opts())
-        .filter(([_, value]) => isThenable(value))
-        .map(([key, thenable]) => (async() => {
-          this.setOptionValueWithSource(
-            key, await thenable, this.getOptionValueSource(key)
-          );
-        })());
-
-      const notChainedNonVariadicPromises = Object.entries(
-        this._handledNotChainedNonVariadicOptionValues
-      )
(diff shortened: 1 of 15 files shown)


lib/command.js
```diff
@@ -1257,81 +1449,93 @@ Expecting one of '${allowedValues.join("', '")}'`);
...
-      if (this.parent) {
-        actionResult = this._chainOrCall(actionResult, () => {
+      // We do not always call this check to avoid masking a "better" error, like unknown command.
+      const checkForUnknownOptions = () => {
+        if (parsed.unknown.length > 0) {
+          this.unknownOption(parsed.unknown[0]);
+        }
+      };
+      const processAndAwaitArguments = () => {
+        checkForUnknownOptions();
```
shadowspawn, line 1491: I didn't call `checkForUnknownOptions` here in my experiment in case not used in some branches, but in fact everywhere that `processAndAwaitArguments` is called, `checkForUnknownOptions` is the line before. So it is ok.

aweebit: Actually, `processAndAwaitArguments()` was originally simply called `preprocess()`, but then I decided the unknown option check and argument processing were too unrelated to keep them in one function, and also I noticed that `checkForUnknownOptions()` was called without `_processArguments()` in one place, so I wanted to remove the `checkForUnknownOptions()` call from the function, but forgot to do it and ended up only renaming the function.

So we should either remove the `checkForUnknownOptions()` call from `processAndAwaitArguments()` as planned, or rename the function back to `preprocess()` or something along these lines and remove the redundant `checkForUnknownOptions()` calls. What do you prefer?

lib/command.js
```diff
@@ -1257,81 +1449,93 @@ Expecting one of '${allowedValues.join("', '")}'`);
    */
 
   _parseCommand(operands, unknown) {
+    let chain;
+
     const parsed = this.parseOptions(unknown);
     this._parseOptionsEnv(); // after cli, so parseArg not called on both cli and env
     this._parseOptionsImplied();
+    chain = this._chainOrCall(chain, () => this._awaitOptionResavePromises());
```
shadowspawn, line 1457: Just a comment. 

This is a nice logical position for processing the options just after they are defined. It does unfortunately spread the chaining logic through the top of this routine. I went for positions close to the hooks since I knew there was already chaining support for hooks.

(And processing async parseArgs for arguments when is no action handler spreads the chaining logic down through routine.)

lib/command.js
shadowspawn: I prefer removing `checkForUnknownOptions` from `processAndAwaitArguments` too, not related enough so harder to understand what is being checked where.

(As understanding of the pattern evolves we might come up with a routine name where it makes sense included again!)

lib/command.js
aweebit: > It does unfortunately spread the chaining logic through the top of this routine.

But on the other hand, it unifies the option value awaiting for parent and leaf commands, and `_dispatchSubcommand()` can remain unchanged. Also, it is actually quite nice to have the `chain` declaration at the very top and `return chain` at the bottom, that kind of sets the logical frame for the function.

lib/command.js
aweebit: Resolved in [905dac0](https://github.com/tj/commander.js/pull/1915/commits/905dac07a17d017029afc7d1a7de1540ed93a228).

shadowspawn:
> Chaining and awaiting in `parse()` introduces the risk of unhandled promise rejections, so they should not be enabled unless the user explicitly requests it.

I am not sure what scenario you have in mind here. Is this when the user has written their own async handling which the built-in chaining would effectively bypass?

shadowspawn:
> Why use the word "await" and not "settle"?

I laughed when I saw this, I have used "settle" in a couple of dubious places. Still getting terminology straight.

shadowspawn:
> The need for the default undefined mode is motivated by the fact we want the mode to be inherited by subcommands by default. 

The way Commander usually does this is copy "common" settings to subcommands as they are added. The routine that does the work is `copyInheritedSettings`. This is a bit clunky but in practice works quite well.

(This might simplify some of your tri-state behaviours down to two.)

lib/command.js
```diff
@@ -931,12 +1012,124 @@ Expecting one of '${allowedValues.join("', '")}'`);
...
+   * If `_await` is `undefined`, inherit await behaviour or base it on whether `.parseAsync()` has been called on the top-level command.
+   *
+   * @api private
+   */
+
+  _shouldAwait() {
+    let shouldAwait;
+    let cmd = this;
+    do {
+      shouldAwait = cmd._await ?? cmd._asyncParsing;
```
aweebit, line 1050: Here, `TypeError: Cannot read properties of null (reading '_await')` is thrown when `parse()` / `parseAsync()` is called on a command that does not "own" the subcommand anymore since it has been added as a subcommand to another parent command like in #1916.

The fact the error is thrown indicates
- either the wrong use of the library, in which case the while condition should be changed to `shouldAwait === undefined && cmd` and an error with a clear explanation of what has been done wrong should be thrown if `shouldAwait` is still `undefined` after the loop;
- or problems in the library's logic, in which case the suggestions from #1916 should be implemented.

aweebit:
> > Chaining and awaiting in `parse()` introduces the risk of unhandled promise rejections, so they should not be enabled unless the user explicitly requests it.
> 
> I am not sure what scenario you have in mind here. Is this when the user has written their own async handling which the built-in chaining would effectively bypass?

I mean the user should not rely on the library for async handling and instead always have their own when calling `parse()` since there is no way to catch errors thrown from async parsers. That's why not chaining and awaiting in `parse()` calls is the reasonable default, but the user still has the freedom to rely on the library by calling `await(true)` which might be useful if they know the parsers will never throw.

aweebit:
> > The need for the default undefined mode is motivated by the fact we want the mode to be inherited by subcommands by default.
> 
> The way Commander usually does this is copy "common" settings to subcommands as they are added. The routine that does the work is `copyInheritedSettings`. This is a bit clunky but in practice works quite well.
> 
> (This might simplify some of your tri-state behaviours down to two.)

If we only have the `true` and `false` modes for `chainArgParserCalls()` and `await()` (with `true` modes being the default ones), no chaining and awaiting should be done in `parse()` calls in the respective `true` modes for reasons explained in [Why not enable chaining and awaiting for `parse()` by default?](#user-content-why-not-enable-for-parse-by-default)

There are following problems with only having those two modes:
- Chaining and awaiting cannot be enforced in `parse()` calls.

  That is not too bad, just weird that even when `await(true)` / `chainArgParserCalls(true)` is called explicitly, it is silently ignored when `parse()` is called.
- The mode is not inherited automatically by explicitly constructed commands added with `addCommand()`.

  Now that is pretty bad since disabling awaiting for all descendant commands is not as simple as running `await(false)` on the top-level command anymore.

aweebit:
> And then there would also be no way for a subcommand to overwrite the inherited mode. That is just awful.

It might even be a good idea to make all inherited settings `undefined` initially, and only inherit or set them to default values when they are `undefined` (`copyInheritedSettings()` would take care of that). `copyInheritedSettings()` calls could then be moved to `_parseCommand()` since it would not be a problem to call the method on commands added with `addCommand()` anymore, and that would be an elegant solution to #1916 ([section](https://github.com/tj/commander.js/issues/1916#user-content-inherit)).

Why do all of that? Because the issue I described in the previous comment is actually relevant for all inherited settings.

shadowspawn:
> Is adding subsubcommands only after the subcommand was added to its parent also the intended pattern

Yes. In particular if you have changed any inherited default configuration on the root command, then need to do in this order to get settings propagated.

(With default settings, I expect can construct in any order.)

aweebit:
The most important of the latest commits is [ea25170](https://github.com/tj/commander.js/pull/1915/commits/ea251708588074b074917a62fb784e1f8fe8cc0a). When an argParser, a hook or an action returns a thenable, it is assigned an empty error handler and an error is thrown from within `_chainOrCall()`. This
- prevents unhandled promise rejections,
- eliminates the need to think about `parse()` calls in this PR, and thus leaves no reasons to have three modes for `chainArgParserCalls()` and `await()`. As a result, awaiting is not configurable now, and chaining no only has the `true` and `false` modes.

lib/command.js
```diff
@@ -1181,20 +1242,27 @@ Expecting one of '${allowedValues.join("', '")}'`);
...
+    if (isThenable(chain)) {
       // already have a promise, chain callback
-      return promise.then(() => fn());
+      return chain.then(fn);
     }
+
     // callback might return a promise
-    return fn();
+    chain = fn(chain);
+    if (isThenable(chain) && !this._asyncParsing) {
```
shadowspawn, line 1259: I had been wondering about whether and how to encourage or enforce parseAsync for async code. Are there benefits to doing the check here rather than in `Command.parse()`?

lib/command.js
```diff
@@ -1181,20 +1242,27 @@ Expecting one of '${allowedValues.join("', '")}'`);
...
+      chain.then(x => x, () => {}); // prevent unhandled rejection
+      process.exitCode = 1; // additionally indicates to the user that user code (parsers, hooks, action) is not the error source
+      throw new Error(`.parse() is incompatible with async argParsers, hooks and actions.
```
shadowspawn, line 1262: An alternative to throwing would be to log a warning and carry on, so preserving existing behaviour but encouraging more careful coding by author. This does increase risk authors miss the change in behaviour as unit tests still likely to pass, but does not hard-break end users. Thoughts?

(On that note, some authors have asked why they need to call `.parseAsync()` when `.parse()` seems to work fine for them.)

lib/command.js
aweebit: How would you check it in `parse()`? I think you would have to call all user functions with God knows what parameter values, and even that would not be enough because a function can return a thenable in one cases and something else in other ones.

lib/command.js
aweebit: I like the idea, see [385fe48](https://github.com/tj/commander.js/pull/1915/commits/385fe48b00ce0bfdab8702bf83e899893ef65b5f).

lib/command.js
aweebit: > since the promise gets returned for `.parseAsync()` to wrap

That does not apply to promises returned from arg parsers.

aweebit: Well, actually it does indirectly via the calls to `_awaitThenableOptions()` and `_awaitThenableArguments()`, but only when `parseAsync()` has been called. We do not want unnecessary work in `parse()` calls, that is why the methods are no-op when calling `parse()`.

aweebit:
Very similar to #1913 now. The difference are:
- chaining is configurable via `chainArgParserCalls()` (motivation in PR description)
- a warning is shown when making async calls in `parse()`
- no awaiting is done in `parse()` calls to avoid extra work (`_asyncParsing` is added for checks)
- overwritten non-variadic option thenable values are stored in `_handledNotChainedNonVariadicOptionValues` and awaited
- thenable options are awaited right after they have been parsed and processed
- error handling is not taken over by `_parseArg()` (motivated by separation of concerns, more details in PR description)
- `_chainOrCall()` is made use of in `_parseArg()`, which is especially important since we want to show warnings about async argParser calls in `parse()`, too
- `_parseArgSubroutine()` is added to follow the DRY principle
- the word "await" is used instead of "settle"
- the thenable check function is named `isThenable()` to prevent confusion when a variable named `thenable` is used to store a thenable object

aweebit:
TODO: Warn about thenable implicit and default option values in `parse()` calls.

aweebit:
Now has to be merged together with #1917 for consistent async warnings in `parse()` calls.

shadowspawn:
> It seems to be more or less an arbitrary design decision to give invalid argument errors thrown from option-argument and command-argument parsers the same `.code` values (`'commander.invalidArgument'`). 

For interest, the rationale for that design decision was to allow an author's custom parser to be used with both options and command-arguments. Like say `myParseScalar()`.

https://github.com/tj/commander.js/pull/1508

shadowspawn:
This was an epic journey from [a few extra lines of code](https://github.com/tj/commander.js/pull/1902#issue-1790910698) to a deep exploration of async implementation! Special mention of lots of explanation and lots of unit tests. I learnt more about async while reviewing and trying out these ideas, especially the error handling.

I do not want to proceed with this at this time. My reservations are:
- not much requested
- complicates `_parseCommand` which is already a long complicated routine
- subtle and complicated async error handling to maintain
- requires much more use of async chaining in the implementation (e.g. `_chainOrCall`)
- does not provide a clean implementation of async: delayed resolution after other code, out of order resolution, promises temporarily stored instead of values. This will lead to some occasional subtle problems for authors. To be clear, this is pretty much impossible to avoid without a drastic refractor! But means there are downsides.
- simple use cases can be handled by author with "a few lines of extra code", but really hard to handle in Commander for general use. In particular, in existing hooks (such as in early implementations).

I think simple author implemented async support for a limited scenario could make a nice example file. Like in https://github.com/tj/commander.js/issues/1900#issuecomment-1622685764

PR #1917, opened 2023-07-28
React to wrong library usage

### LINKED ISSUE #1916: On command and subcommand reusability and setting inheritance

Here is an example:

```js
program
  .option('-a')
  .option('-b');

```

Here is an example:

```js
  .action(() => {});

program
  })
  .addCommand(sub);

  .addCommand(sub);

```

Here is an example:

```js

```

aweebit:
___Update:___ Ideally, this PR should be superseded by #1938 and #1940. See [comment](https://github.com/tj/commander.js/pull/1917#issuecomment-1666467716) for details.

Partially fixes #1916. I think it is a good idea to react to wrong library uses even if they have not been featured in many issues.

#### Why `console.warn()` and not throw from within `parse()` / `parseAsync()`?

Errors thrown from `parse()` / `parseAsync()` are expected to have originated in user-supplied code (argParsers, hooks, actions).

An alternative could be to `console.error()` and `process.exit()` with a non-zero exit code, effectively forbidding all wrong usage, but what I don't like is the discrepancy between this approach and how all other library errors are simply thrown and can be handled by the user.

### Added
- _Breaking:_ throw an error when trying to add a subcommand that already has a parent command
- warnings about parse calls on subcommands
- warnings about async hook and action calls made in `.parse()`

- #1915
- #1931
- #1938
- #1940

### Parse call subroutine (`_parseSubroutine()`) needs to be consistent with…
- #1919

### Incompatible with…
- #1955: parse methods shall never be called on subcommands, so if warnings about such calls are added, the JSDoc and the ChangeLog entry for `.suppressWarnings()` introduced there become incorrect. The fact there is currently no well-defined meaning for such calls is one of the reasons why [#1938](https://github.com/tj/commander.js/pull/1938) and [#1940](https://github.com/tj/commander.js/pull/1940) should be preferred over this PR ([#1917](https://github.com/tj/commander.js/pull/1917))!

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -546,8 +546,6 @@ subcommand is specified ([example](./examples/defaultCommand.js)).
 
 You can add alternative names for a command with `.alias()`. ([example](./examples/alias.js))
 
-`.command()` automatically copies the inherited settings from the parent command to the newly created subcommand. This is only done during creation, any later setting changes to the parent are not inherited.
-
 For safety, `.addCommand()` does not automatically copy the inherited settings from the parent command. There is a helper routine `.copyInheritedSettings()` for copying the settings when they are wanted.
 
 ### Command-arguments
--- a/Readme.md
+++ b/Readme.md
@@ -546,6 +546,8 @@ subcommand is specified ([example](./examples/defaultCommand.js)).
 
 You can add alternative names for a command with `.alias()`. ([example](./examples/alias.js))
 
+Commands added with `.command()` automatically inherit settings for which inheritance is meaningful from the parent command, but only upon the subcommand creation. The setting changes made after calling `.command()` are not inherited.
+
 For safety, `.addCommand()` does not automatically copy the inherited settings from the parent command. There is a helper routine `.copyInheritedSettings()` for copying the settings when they are wanted.
 
 ### Command-arguments
--- a/Readme.md
+++ b/Readme.md
@@ -546,7 +546,7 @@ subcommand is specified ([example](./examples/defaultCommand.js)).
 
 You can add alternative names for a command with `.alias()`. ([example](./examples/alias.js))
 
-Commands added with `.command()` automatically inherit settings for which inheritance is meaningful from the parent command, but only upon the subcommand creation. The setting changes made after calling `.command()` are not inherited.
+`.command()` automatically copies the inherited settings from the parent command to the newly created subcommand. This is only done during creation, any later setting changes to the parent are not inherited.
 
 For safety, `.addCommand()` does not automatically copy the inherited settings from the parent command. There is a helper routine `.copyInheritedSettings()` for copying the settings when they are wanted.
 
--- a/lib/command.js
+++ b/lib/command.js
@@ -1229,7 +1229,7 @@ Call on top-level command instead`);
     // callback might return a promise
     const result = fn();
     if (!PRODUCTION && !this._asyncParsing && isThenable(result)) {
-      console.warn(`.parse() is incompatible with async argParsers, hooks and actions.
+      console.warn(`.parse() is incompatible with async hooks and actions.
 Use .parseAsync() instead.`);
     }
     return result;
--- a/lib/command.js
+++ b/lib/command.js
@@ -12,6 +12,8 @@ const { suggestSimilar } = require('./suggestSimilar');
 
 // @ts-check
 
+const PRODUCTION = process.env.NODE_ENV === 'production';
+
 class Command extends EventEmitter {
   /**
    * Initialize a new `Command`.
@@ -77,6 +79,9 @@ class Command extends EventEmitter {
     this._helpCommandnameAndArgs = 'help [command]';
     this._helpCommandDescription = 'display help for command';
     this._helpConfiguration = {};
+
+    /** @type {boolean | undefined} */
+    this._asyncParsing = undefined;
   }
 
   /**
@@ -264,6 +269,10 @@ class Command extends EventEmitter {
     if (!cmd._name) {
       throw new Error(`Command passed to .addCommand() must have a name
 - specify the name in Command constructor or using .name()`);
+    }
+    if (cmd.parent) {
+      throw new Error(`'${cmd._name}' cannot be added using .addCommand()
+- it already has a parent command`);
     }
 
     opts = opts || {};
@@ -888,6 +897,28 @@ Expecting one of '${allowedValues.join("', '")}'`);
   }
 
   /**
+   * @param {boolean} async
+   * @param {Function} userArgsCallback
+   * @param {string[]} [argv]
+   * @param {Object} [parseOptions]
+   * @param {string} [parseOptions.from]
+   * @return {Command|Promise}
+   * @api private
+   */
+
+  _parseSubroutine(async, userArgsCallback, argv, parseOptions) {
+    this._asyncParsing = async;
+    const methodName = async ? 'parseAsync' : 'parse';
+    if (!PRODUCTION && this.parent) {
+      console.warn(`Called .${methodName}() on subcommand '${this._name}'.
+Call on top-level command instead`);
+    }
+
+    const userArgs = this._prepareUserArgs(argv, parseOptions);
+    return userArgsCallback(userArgs);
+  }
+
+  /**
    * Parse `argv`, setting options and invoking commands when defined.
    *
    * The default expectation is that the arguments are from node and have the application as argv[0]
@@ -905,10 +936,10 @@ Expecting one of '${allowedValues.join("', '")}'`);
    */
 
   parse(argv, parseOptions) {
-    const userArgs = this._prepareUserArgs(argv, parseOptions);
-    this._parseCommand([], userArgs);
-
-    return this;
+    return this._parseSubroutine(false, (userArgs) => {
+      this._parseCommand([], userArgs);
+      return this;
+    }, argv, parseOptions);
   }
 
   /**
@@ -931,10 +962,10 @@ Expecting one of '${allowedValues.join("', '")}'`);
    */
 
   async parseAsync(argv, parseOptions) {
-    const userArgs = this._prepareUserArgs(argv, parseOptions);
-    await this._parseCommand([], userArgs);
-
-    return this;
+    return this._parseSubroutine(true, async(userArgs) => {
+      await this._parseCommand([], userArgs);
+      return this;
+    }, argv, parseOptions);
   }
 
   /**
@@ -1071,6 +1102,7 @@ Expecting one of '${allowedValues.join("', '")}'`);
   _dispatchSubcommand(commandName, operands, unknown) {
     const subCommand = this._findCommand(commandName);
     if (!subCommand) this.help({ error: true });
+    subCommand._asyncParsing = this._asyncParsing;
 
     let hookResult;
     hookResult = this._chainOrCallSubCommandHook(hookResult, subCommand, 'preSubcommand');
@@ -1189,12 +1221,18 @@ Expecting one of '${allowedValues.join("', '")}'`);
 
   _chainOrCall(promise, fn) {
     // thenable
-    if (promise && promise.then && typeof promise.then === 'function') {
+    if (isThenable(promise)) {
       // already have a promise, chain callback
       return promise.then(() => fn());
     }
+
     // callback might return a promise
-    return fn();
+    const result = fn();
+    if (!PRODUCTION && !this._asyncParsing && isThenable(result)) {
+      console.warn(`.parse() is incompatible with async argParsers, hooks and actions.
+Use .parseAsync() instead.`);
+    }
+    return result;
   }
 
   /**
@@ -2191,6 +2229,16 @@ function getCommandAndParents(startCommand) {
     result.push(command);
   }
   return result;
+}
+
+/**
+ * @param {*} value
+ * @returns {boolean}
+ * @api private
+ */
+
+function isThenable(value) {
+  return typeof value?.then === 'function';
 }
 
 exports.Command = Command;
--- a/lib/command.js
+++ b/lib/command.js
@@ -937,7 +937,11 @@ Call on top-level command instead`);
 
   parse(argv, parseOptions) {
     return this._parseSubroutine(false, (userArgs) => {
-      this._parseCommand([], userArgs);
+      const result = this._parseCommand([], userArgs);
+      if (!PRODUCTION && isThenable(result)) {
+      console.warn(`.parse() is incompatible with async hooks and actions.
+Use .parseAsync() instead.`);
+      }
       return this;
     }, argv, parseOptions);
   }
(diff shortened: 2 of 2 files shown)


aweebit:
Idea: allow adding a subcommand that already has a parent to a different parent and `console.warn()` about incorrect usage if `parse()` / `parseAsync()` is called on the new command. That might be better because
- no breaking change is introduced
- there is nothing bad about reusing a subcommand as long as the old parent is no longer used.
But since we would be doing parent checks in `parse()` / `parseAsync()` anyway, we could just as well set the `parent` property for all subcommands in `this.commands` not upon adding them with `command()` / `addCommand()`, but in `parse()` / `parseAsync()` instead. That would change the expected behavior mentioned in #1916 ([comment](https://github.com/tj/commander.js/issues/1916#issuecomment-1652726166)) and enable reuse of subcommands.

shadowspawn:
A comment. There are about four different independent things here. I struggle to avoid adding extra things myself while on a roll making code changes. But as a reviewer, I dislike it!

Readme.md
```diff
@@ -546,6 +546,8 @@ subcommand is specified ([example](./examples/defaultCommand.js)).
 
 You can add alternative names for a command with `.alias()`. ([example](./examples/alias.js))
 
+Commands added with `.command()` automatically inherit settings for which inheritance is meaningful from the parent command, but only upon the subcommand creation. The setting changes made after calling `.command()` are not inherited.
```
shadowspawn, line 549: I have struggled with whether and how to describe this in the past so thanks for having a go. I like the direction here, fairly high level but clarifies the implicit ordering. Maybe:

> Commands created with `.command()` automatically inherit relevant settings from the parent command. This is done during creation, and any later setting changes to the parent are not inherited.

lib/command.js
```diff
@@ -12,6 +12,8 @@ const { suggestSimilar } = require('./suggestSimilar');
 
 // @ts-check
 
+const PRODUCTION = process.env.NODE_ENV === 'production';
```
shadowspawn, line 15: Hmm. I don't think `PRODUCTION` will often be relevant for a CLI? I get the desire to avoid spamming an end-user who can't fix the problem directly. But I don't think Commander has a strong enough production/non-production case to make the distinction?

lib/command.js
```diff
@@ -265,6 +270,10 @@ class Command extends EventEmitter {
       throw new Error(`Command passed to .addCommand() must have a name
 - specify the name in Command constructor or using .name()`);
     }
+    if (cmd.parent) {
+      throw new Error(`'${cmd._name}' cannot be added using .addCommand()
```
shadowspawn, line 274: I quite like the format of the error just above. Maybe:
```
throw new Error(`Command '${cmd.name()}' passed to .addCommand() already has a parent
 - commands can not be reused or shared`);
```

lib/command.js
```diff
@@ -887,6 +896,28 @@ Expecting one of '${allowedValues.join("', '")}'`);
...
+   * @param {string} [parseOptions.from]
+   * @return {Command|Promise}
+   * @api private
+   */
+
+  _parseSubroutine(async, userArgsCallback, argv, parseOptions) {
+    this._asyncParsing = async;
+    const methodName = async ? 'parseAsync' : 'parse';
+    if (!PRODUCTION && this.parent) {
+      console.warn(`Called .${methodName}() on subcommand '${this._name}'.
```
shadowspawn, line 913: I don't think I care enough about this error to add. For interest, did you hit this in real world usage?

If I did add it, I think I would rather have some small code repetition than an extra level of code.

lib/command.js
```diff
@@ -1189,12 +1221,18 @@ Expecting one of '${allowedValues.join("', '")}'`);
...
-    if (promise && promise.then && typeof promise.then === 'function') {
+    if (isThenable(promise)) {
       // already have a promise, chain callback
       return promise.then(() => fn());
     }
+
     // callback might return a promise
-    return fn();
+    const result = fn();
+    if (!PRODUCTION && !this._asyncParsing && isThenable(result)) {
```
shadowspawn, line 1231: Why not have this test in `.parse()` itself? Save having a global as well as have the code right where it is relevant.

lib/command.js
```diff
@@ -2193,4 +2231,14 @@ function getCommandAndParents(startCommand) {
   return result;
 }
 
+/**
+ * @param {*} value
+ * @returns {boolean}
+ * @api private
+ */
+
+function isThenable(value) {
```
shadowspawn, line 2240: I did this in a my recent PR too, so apologies I am having second thoughts about where to put the code. Buckle up for some subtle considerations.

An unexported global function is not available to authors who subclass Command, and in general I don't want to explicitly export internal utility functions. I have wondered about static methods, but they can't be overridden (not that I want to explicitly support overriding it as such!).

So what I have done in a couple of places is make some common code a private method as the best compromise (at least when used by only a single class).

With the optional chaining it is pretty compact so could just be inlined, but I do like naming the test.

aweebit:
> A comment. There are about four different independent things here. I struggle to avoid adding extra things myself while on a roll making code changes. But as a reviewer, I dislike it!

Okay, I will take that into account next time.

Readme.md
aweebit: Not sure I like "relevant", seems a little vague. Maybe simplify what I suggested to "for which that is meaningful”?

I like the rest though 👍

lib/command.js
aweebit: I haven't used `NODE_ENV` in CLI applications myself, but thought that having this distinction wouldn't harm, so why not add it. It gives the developer the freedom to ignore the warnings in production, which might be meaningful after #1915 is merged so that the developer can insist on manually handling thenable argument and option values by using `parse()`.

Readme.md
shadowspawn: Ha, thanks, I was a little dubious about the `relevant` too. Given your reaction, I think leave it out, just "inherited settings". It isn't the whole story, but any attempt to hint at more just leads to questions about what is being hinted at without actually helping. (e.g. "some", "relevant", "for which inheritance is meaningful")

aweebit: "inherit inherited settings" just doesn't sound good because of the repetition.

shadowspawn: > Commands created with .command() automatically inherit settings from the parent command. This is done during creation, and any later setting changes to the parent are not inherited.

lib/command.js
aweebit: I like it, but I would leave out "reused" because "shared" means the same thing here but is more specific. I would maybe change "commands" to "subcommands" on the second line, too.

shadowspawn: Yup, I like both.

lib/command.js
aweebit: > I don't think I care enough about this error to add.

It seems a little weird to add the error in `addCommand()` but not this one. But everything could be fixed by implementing the suggestion from [the first comment](https://github.com/tj/commander.js/pull/1917#issuecomment-1656365692). Why do you dislike it? I think it is very easy to implement. If I'm not mistaken, it would require even fewer lines of code than the combination of this warning and the error in `addCommand()` (cannot check it right now unfortunately since I'm on a small trip and haven't taken my laptop with me). Only the error from `passThroughOptions()` would need to be turned into a warning in `_parseSubroutibe()`, and something done about the `parent` use in `outputHelp()`.

> For interest, did you hit this in real world usage?

No, just trying on the library developer mindset and thinking of different ways it could be used 🙂

> If I did add it, I think I would rather have some small code repetition than an extra level of code.

More shared code can be added to `_parseSubroutibe()` later, like after the suggestion from the first comment is implemented or #1919 is merged.

lib/command.js
aweebit: `_chainOrCall()` could later be used not only for the chain returned from `_parseCommand()`, see #1915 ([comment](https://github.com/tj/commander.js/pull/1915#discussion_r1277289430)).

Readme.md
aweebit: Now it makes it seem like all settings are inherited 🤷‍♂️

lib/command.js
aweebit: I don't think we want to let users overwrite the thenable check, do we? If not, I think the function is placed right where it needs to be.

lib/command.js
shadowspawn: The high level author error that does sometimes happen is chucking `parse()` on the end of the subcommand definition. This would catch that particular case. It does not come up often and I wonder if there are valid use cases for calling `.parse()` down tree in fancy custom programs.

```
const { program } = require('commander')
   .exitOverride()
   .option('-g, --global')
   .command('sub')
   .option('--sub-option')
   .parse(); // oops
```

To encourage a usage pattern that avoids several potential problems, I use this style as much as possible in the hopes people will just copy it:
- define `program`
- configure program
- new statement per subcommand
- call `.parse()`

```

program
   .exitOverride()
   .option('-g, --global');

program.command('sub')
   .option('--sub-option');

program.parse();
```

lib/command.js
aweebit: Resolved in https://github.com/tj/commander.js/pull/1917/commits/55b9146cbf23fc20e9b32bedadb27d2169ee41a1.

lib/command.js
aweebit: How about we propagate `parse()` / `parseAsync()` calls to the root command? Then chucking a call at the end of a subcommand definition would not be a problem, and we would not need the warning since subcommands would not be usable as stand-alones.

aweebit: The suggestion from the first comment would also become irrelevant.

Readme.md
shadowspawn: Mind dump, not condensed!

I wondered about "all" and initially thought it would be obvious that not all settings are copied. But I guess some users might carefully set settings after adding the commands to avoid them being inappropriately copied. And I would hate to cause that.

Words: "relevant", "some", "selected", "appropriate"

In the working model in my mind there are "inherited settings" which are copied, with a somewhat circular definition of what that means. I'll try writing something out...

"automatically copy inherited settings from the parent command"

Hmm, maybe! That fits in with my longterm working model, matches the subsequent mention of `copyInheritedSettings` in the next paragraph, which provides a clue for people to find out the actual code. (And on that note, I like what I put in the JSDoc for that routine.)

aweebit: Going for "`.command()` automatically copies the inherited settings" just like in the next paragraph, how do you like it?

```suggestion
`.command()` automatically copies the inherited settings from the parent command to the newly created subcommand. This is only done during creation, and any later setting changes to the parent are not inherited.
```

shadowspawn: LGTM

lib/command.js
shadowspawn: Hmmm... No, I think not. Either the user is intentionally doing something interesting outside normal usage, or they are confused about how Commander works. Doing something different that _might_ be what they intended is masking the situation.

Readme.md
aweebit: Resolved in https://github.com/tj/commander.js/pull/1917/commits/dfe2fc74a53db43153f4305ec5f9f1cc6a13c64f.

lib/command.js
aweebit: Actually, it might be a good idea to show a different warning for argParsers and wrap the relevant code in a new method called `_chainOrCallArgParser()`.

aweebit: Oh, actually, I see now the right place for the argParser warning in #1915 is `_parseArg()`.

lib/command.js
aweebit: As of now, there are no meaningful semantics / intended usage patterns for `parse()` / `parseAsync()` calls on subcommands anyway, and any such call should be considered wrong. The only meaningful semantics for such a call I can think of are exactly that, propagating it to the root command. So why not assign a meaning to calls that currently have no meaning at all, giving the developer the freedom to chuck a parse call at the end of a subcommand chain? That is not _doing something different that might be what they intended_, that is doing the only meaningful thing when receiving such a call instead of doing complete nonsense (like what the library currently does) or having to come up with ways to tell the user they are using the library in a wrong way (like what this PR currently does).

lib/command.js
aweebit: Resolved in https://github.com/tj/commander.js/pull/1917/commits/3572657066e281b98eb5a653d00b26de7ceee958 (`PRODUCTION` still has to be global, though).

aweebit:
Please don't pull now gonna reset a commit I accidentally committed to a wrong branch lol

aweebit:
(Force push is due to a hard reset needed because I committed on a wrong branch.)

lib/command.js
shadowspawn: The "global" I was referring to was `_asyncParsing`. I should have said an instance property. It is still in the code, but I think no longer being used after latest changes.

aweebit: Oh yes, that's right. Fixed in https://github.com/tj/commander.js/pull/1917/commits/adde7ffcf0a4b02bbcef50dc7b8dc5b4261b490d.

lib/command.js
shadowspawn: There are a few other global functions which have not caused problems to date, and easy workaround (copy). I'll put the paranoia aside!

lib/command.js
aweebit: Two days later, I am having second thoughts about my comment on the suggestion from the first comment becoming irrelevant. I think I was wrong saying that, it will still be relevant because it deals with parse calls on a parent command that has lost ownership of a subcommand. So I would still like to hear why you disliked that suggestion. Implementing it would make the warning this thread is about unnecessary since parsing subcommands as stand-alones would become possible. Could still leave the warning for the case when parse call is chucked a the end of an implicit subcommand chain, though, for example by adding a `_parseCallsUndesired`  field with the default value of `false` to the `Command` class.

lib/command.js
aweebit: I don't even know why you think global functions in a module should cause any problem, they are not available in user code even if the entire module is `import`ed (and not just some variables) because [strict mode is always enabled in modules](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Strict_mode?retiredLocale=de#strict_mode_for_modules).

aweebit:
Now, I would like this PR to be superseded by #1940 (the warning part taken from this PR) and #1938 (implementing a better version of the changes suggested in [comment](https://github.com/tj/commander.js/pull/1917#issuecomment-1656365692) and [comment](https://github.com/tj/commander.js/pull/1917#discussion_r1279217920)).

Due to the support for sharing and stand-alone parsing introduced in #1938, fewer use cases would be considered wrong and there would be need for only one new warning, instead of two new warnings and one new error like in this PR.

aweebit:
Since #1940 is closed now, I could also remove the warning about async calls from this issue, but I don't really want to change anything here because i strongly advocate for this issue being superseded by #1938.

shadowspawn:
Related: https://stackoverflow.com/questions/76896860/test-suite-fails-when-commander-error-is-thrown-in-jest-even-with-catch

aweebit:
> Related: https://stackoverflow.com/questions/76896860/test-suite-fails-when-commander-error-is-thrown-in-jest-even-with-catch

Related to the now-closed #1940 component of this PR, to be more more specific. I couldn't ask for a better demonstration of why closing it was a mistake.

Also, the comment you left there is related to #1922: accepting the PR would make it easier to copy settings to an entire command hierarchy when it can only be done after subcommands were added.

shadowspawn:
There are multiple alternative preferred PRs by same author, and this PR was somewhat of a learning experience. In particular #1940 has been rejected, for now.

> I think it is a good idea to react to wrong library uses even if they have not been featured in many issues.

I think this is an easy approach in a new library, as users can be directed to the intended use patterns. I move more slowly and conservatively in a mature library, where every change may break an upgrade for some existing users who may or may not appreciate being forced to change their "working" program.

Issue #1829, opened 2022-12-05
Subcommand Holds onto `_optionValues` Between Executions


manunamz:
I am setting up some unit tests around a commander.js cli tool. I verify that the args and opts the program is holding onto internally matches expectations for each test. It seems that subcommands, and possibly the `Command` object in general, is holding onto options between executions.

This means, if I test a command with options and then test the same command without any options, the second test will still have the options from the first execution.

```js
// test with options pass
program.parse(['program', 'command', '-o', 'stuff']);
// test without options still have previous options
program.parse(['program', 'command']); // '-o' and 'stuff' will still be parsed
```

I am currently working around this issue by manually clearing options between each test -- see the following:

```js
  afterEach(() => {
    const cmd: any = program.commands.find((cmd) => cmd.name() === 'command');
    cmd._optionValues = {};
  });
```

This is just a short description to notify the commander.js maintainer. I think the issue is probably a simple reset of `_optionValues` or class properties more generally...If more detail is required, I can fill out a more detailed ticket upon request.

shadowspawn:
Thanks for the clear description.

Yes, there isn't any code to reset state after calling `.parse()`. The intended pattern is make a new command for each test. 

Related: #438 #1565 #1819

manunamz:
Thanks for the issue numbers! The relevant conversation is useful, particularly your other comment [here](https://github.com/tj/commander.js/issues/1819#issuecomment-1309276026).

This feels like a duplicate of https://github.com/tj/commander.js/issues/1819 , so I'll close this out. Thank you again!

shadowspawn:
Support was added in Commander v13 for calling parse more than once: https://github.com/tj/commander.js/pull/2299

Issue #1819, opened 2022-11-09
Command instance re-using old arguments on subsequent calls in test env


yamalight:
I've encountered an odd behavior where subcommands in test env will re-use optional arguments from previous calls. 
At the top level - you have a program and a subcommand with optional flags. Following calls behave oddly:

1. Call `myapp subcommand --init` (correctly passes `init` as argument to handler)
2. Call `myapp subcommand --other 1234` (correctly passes `other` as argument to handler, but also includes `init` from first call)
3. Call `myapp subscommand` (passes previous values for `init` and `other` instead of empty args)

Is this behavior intended? If yes, how would one reset args between tests? The only way I've found so far was to monkey-patch subcommand `_optionValues` which feels wrong 😅

I've created a small repro case in stackblitz: https://stackblitz.com/edit/node-nsyb1h?file=app.test.js
Calling `npm test` executes it. 
Uncommenting monkey-patching in the last test case makes it behave as expected (i.e. not pass any args when none given to `parse` function).

shadowspawn:
Thanks for the detail and example.

The short answer is make new Commands for each test. i.e. both program and subcommand.

The code isn't written to allow multiple calls to parse, and there isn't a routine to reset state.

shadowspawn:
See #1565 for some additional tips and link to an example.

shadowspawn:
(Feel free to ask more questions, just pointing you in the right direction!)

yamalight:
@shadowspawn gotcha! thanks for the clarification 👍 
I'll close the issue since it's expected behavior.

shadowspawn:
Support was added in Commander v13 for calling parse more than once: https://github.com/tj/commander.js/pull/2299

Issue #1565, opened 2021-07-09
What is the best way to unit test a commander cli?


tiagoporto:
I would like to cover my cli (commander.js) with unit test.

What is the best way to do that?

shadowspawn:
A couple of tips:

Have a function that creates a new `Command` object for your cli, so each test can use a fresh program.

If your tests include error cases detected by Commander, pass in some configuration when creating the program to make it easier to catch errors (`.exitOverride()`) or suppress output (`configureOutput()`).

tiagoporto:
@shadowspawn do you have some examples to share, will be really helpful?
Thank you

shadowspawn:
Example of a function to create the "program", with some extra options for using in unit tests:
- https://github.com/shadowspawn/forest-arborist/blob/fca5ffcc5b300660ae9e1f6c4a8667d72feb0822/src/command.ts#L48

(I am  not claiming this is the best way, just what I used!)

tiagoporto:
Thank you.

shadowspawn:
For interest, support was added in Commander v13 for calling parse more than once: https://github.com/tj/commander.js/pull/2299

PR #1919, opened 2023-07-28
Clean up state from previous parse call when calling `parse()` / `parseAsync()`

aweebit:
The better solution for #438 #1565 #1819 #1829 #1916 suggested in #1916.

<details>
<summary>Outdated</summary>

While implementing it, I noticed that the type restriction for the `source` parameter of the public `setOptionValueWithSource()` method are too loose. The only really supported source value seems to be `'config'` as of now, so I changed the type to this single value. This has the benefit of preventing messing up the internal logic by supplying option values with the source being one of `'default'`, `'cli'`, `'env'` and `'implied'` (at least when using TypeScript, might also want to add a real check in code).
</details>

### Changed
- _Breaking:_ previous `parse()` / `parseAsync()` call state is cleaned up in the beginning of a new call (enables command reuse for testing, REPL etc.)

### Fixed
- option value source type (set to `string | undefined` to enable custom source use)

### Merge this PR after…
- #1915 (`_asyncParsing` is used in this PR instead of simply adding a boolean variable indicating whether the command is currently being parsed because it is assumed the PRs are merged together. If they are not, use a boolean variable instead)

### Parse call subroutine (`_parseSubroutine()`) needs to be consistent with…
- #1917

### DIFF
--- a/lib/command.js
+++ b/lib/command.js
@@ -10,12 +10,6 @@ const { Help } = require('./help.js');
 const { Option, splitOptionFlags, DualOptions } = require('./option.js');
 const { suggestSimilar } = require('./suggestSimilar');
 
-/**
- * TypeScript import types for JSDoc, used by Visual Studio Code IntelliSense and `npm run typescript-checkJS`
- * https://www.typescriptlang.org/docs/handbook/jsdoc-supported-types.html#import-types
- * @typedef { import('../typings').OptionValueSource } OptionValueSource
- */
-
 // @ts-check
 
 class Command extends EventEmitter {
@@ -797,8 +791,7 @@ Expecting one of '${allowedValues.join("', '")}'`);
    */
 
   setOptionValue(key, value) {
-    this._setPersistentOptionValueWithSource(key, value);
-    return this;
+    return this.setOptionValueWithSource(key, value);
   }
 
   /**
@@ -806,7 +799,7 @@ Expecting one of '${allowedValues.join("', '")}'`);
     *
     * @param {string} key
     * @param {Object} value
-    * @param {'config'} source
+    * @param {string} [source]
     * @return {Command} `this` command for chaining
     */
 
@@ -818,7 +811,7 @@ Expecting one of '${allowedValues.join("', '")}'`);
   /**
     * @param {string} key
     * @param {Object} value
-    * @param {'default' | 'config'} [source]
+    * @param {string} [source]
     * @api private
     */
 
@@ -831,7 +824,7 @@ Expecting one of '${allowedValues.join("', '")}'`);
   /**
     * @param {string} key
     * @param {Object} value
-    * @param {OptionValueSource} [source]
+    * @param {string} [source]
     * @api private
     */
 
@@ -849,7 +842,7 @@ Expecting one of '${allowedValues.join("', '")}'`);
     * Expected values are default | config | env | cli | implied
     *
     * @param {string} key
-    * @return {OptionValueSource | undefined}
+    * @return {string | undefined}
     */
 
   getOptionValueSource(key) {
@@ -861,7 +854,7 @@ Expecting one of '${allowedValues.join("', '")}'`);
     * Expected values are default | config | env | cli | implied
     *
     * @param {string} key
-    * @return {OptionValueSource | undefined}
+    * @return {string | undefined}
     */
 
   getOptionValueSourceWithGlobals(key) {
--- a/lib/command.js
+++ b/lib/command.js
@@ -12,7 +12,71 @@ const { suggestSimilar } = require('./suggestSimilar');
 
 // @ts-check
 
-class Command extends EventEmitter {
+class CommandBase extends EventEmitter {
+  constructor() {
+    super();
+
+    // The proxy only treats keys not present in the instance and its prototype chain as keys for _optionValues when _storeOptionsAsProperties is set to true.
+    // Setting option values for keys present in the instance and its prototype chain is still possible by calling .setOptionValue() or .setOptionValueWithSource(),
+    // but such values will not be accessible as instance properties because the instance and its prototype chain have precedence.
+    // However, they will be accessible via .getOptionValue(), .opts() and .optsWithGlobals().
+    return new Proxy(this, {
+      get(target, key, receiver) {
+        if (target._storeOptionsAsProperties && !(key in target)) {
+          target = receiver = receiver._optionValues;
+        }
+        return Reflect.get(target, key, receiver);
+      },
+      set(target, key, value, receiver) {
+        if (target._storeOptionsAsProperties && !(key in target)) {
+          target = receiver = receiver._optionValues;
+        }
+        return Reflect.set(target, key, value, receiver);
+      },
+      has(target, key) {
+        if (target._storeOptionsAsProperties && !(key in target)) {
+          target = target._optionValues;
+        }
+        return Reflect.has(target, key);
+      },
+      deleteProperty(target, key) {
+        if (target._storeOptionsAsProperties && !(key in target)) {
+          target = target._optionValues;
+        }
+        return Reflect.deleteProperty(target, key);
+      },
+      defineProperty(target, key, descriptor) {
+        if (target._storeOptionsAsProperties && !(key in target)) {
+          target = target._optionValues;
+        }
+        return Reflect.defineProperty(target, key, descriptor);
+      },
+      getOwnPropertyDescriptor(target, key) {
+        if (target._storeOptionsAsProperties && !(key in target)) {
+          target = target._optionValues;
+        }
+        return Reflect.getOwnPropertyDescriptor(target, key);
+      },
+      ownKeys(target) {
+        const result = Reflect.ownKeys(target);
+        if (target._storeOptionsAsProperties) {
+          result.push(...Reflect.ownKeys(target._optionValues).filter(
+            key => !(result.includes(key)) // remove duplicates
+          ));
+        }
+        return result;
+      },
+      preventExtensions(target) {
+        if (target._storeOptionsAsProperties) {
+          Reflect.preventExtensions(target._optionValues);
+        }
+        return Reflect.preventExtensions(target);
+      }
+    });
+  }
+}
+
+class Command extends CommandBase {
   /**
    * Initialize a new `Command`.
    *
@@ -70,6 +134,12 @@ class Command extends EventEmitter {
     this._helpCommandDescription = 'display help for command';
     this._helpConfiguration = {};
 
+    // Because of how the proxy returned from the CommandBase constructor works in order to support options-as-properties,
+    // all instance properties have to be defined when _storeOptionsAsProperties is set to false.
+    // Ideally, that should happen as soon as in the constructor, even if it seems unnecessary because the initial values are undefined like here.
+    this._version = undefined;
+    this._versionOptionName = undefined;
+
     this._unprocessedName = name || '';
     this._persistentOptionValues = {};
     this._persistentOptionValueSources = {};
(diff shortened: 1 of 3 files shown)


shadowspawn:
> While implementing it, I noticed that the type restriction for the `source` parameter of the public `setOptionValueWithSource()` method are too loose. The only really supported source value seems to be `'config'` as of now, so I changed the type to this single value.

It is intended that the author can use any value, whether one of the ones Commander uses internally, or a custom value which means something to the author but not to Commander. The option-source was added in part to make configuration files easier to implement by user, and in part to support adding environment variables.

It is a bit difficult for Commander to decided how to treat unknown values, and in general I dislike adding support for speculative uses. In this case I was encouraged by having multiple use cases I wanted to support in mind from the start, and it then proved useful for implementing `.conflicts()` too.

lib/command.js
```diff
@@ -10,6 +10,12 @@ const { Help } = require('./help.js');
 const { Option, splitOptionFlags, DualOptions } = require('./option.js');
 const { suggestSimilar } = require('./suggestSimilar');
 
+/**
```
shadowspawn, line 13: My high level approach is to keep the JSDoc fairly pure for the environments that do not support TypeScript. I am unclear how many people this benefits! I use Visual Studio Code myself.

A little ironically, the only check for JSDoc consistency is the TypeScript JavaScript test (`typescript-checkJS`) which did and does pick up more JSDoc issues than coding issues.

My favourite JSDoc reference is: https://jsdoc.app

shadowspawn:
To be clear, I don't think I'll want this! So choose how much further work to do accordingly.

But I'll work through and make comments, and interested in whether it could work. I find there is only so much I can work out in theory, and usually learn more once I try writing or running the code.

lib/command.js
```diff
@@ -780,34 +797,64 @@ Expecting one of '${allowedValues.join("', '")}'`);
...
+  _setNonPersistentOptionValueWithSource(key, value, source) {
     if (this._storeOptionsAsProperties) {
       this[key] = value;
     } else {
       this._optionValues[key] = value;
     }
     this._optionValueSources[key] = source;
-    return this;
+
+    delete this._persistentOptionValues[key];
```
shadowspawn, line 848: Setting a non-persistent value should not clear the underlying persistent value?

shadowspawn:
On to the reset itself.

There are two specific things which are not working currently in this PR:
- resetting default values
- resetting if legacy `.storeOptionsAsProperties()` is enabled. This is definitely legacy but has not been deprecated yet. It could be forbidden rather than supported. Other than setting option values to `undefined`, there isn't a high-level way of "deleting" a value so hard to fully restore property state.

High level things I am dubious about include
- author set option values: before parse, from hooks, after parse
- implicitly set option values (might be ok)

If aiming for a high level model that a second `parse()` starts from the same state as the first parse, a more direct model might be saving the state on first entry and resetting state on subsequent entries. This bypasses trying to decide whether to persist changes as they occur. 

An entirely different design is that calling `.parse()` does not change any state and returns the parse result. This is very much _not_ the way Commander is written.

I looked back at some of the issue and PR history. There was an early PR for a reset in #499. It is way simple, but does have one interesting insight that the "known" options are available from the options array.

> To be clear, I don't think I'll want this! So choose how much further work to do accordingly.

Warning still applies. 😄

lib/command.js
aweebit: Hmm, `@typedef`s are already used in `Help`  though.

lib/command.js
aweebit: Oh yes, you are right.

lib/command.js
shadowspawn: Yup, yup. I wanted `Command` and `Option` and `Argument` available in Help when the code got split up into multiple files, and the first attempt created an infinite loop in the TypeScript parsing, and the second attempt hoisted some code into `index.js` just for typing. Importing the definitions the TypeScript way kept the "real" code clean while not making Help just use Object everywhere for zero type checking.

lib/command.js
aweebit: Resolved in https://github.com/tj/commander.js/pull/1919/commits/ebf672aca9fc3187e9aa8f8161d854397f44b2ae.

lib/command.js
aweebit: I think using it TypeScript-flavored JSDoc in one place makes it possible to use it everywhere. But I could just as well replace the import by the actual type, not a problem.

aweebit:
> - resetting default values

Default values were actually reset. I made it more clear in https://github.com/tj/commander.js/pull/1919/commits/7fd0b22f1adc75a97bdb4af262a71126cb9f3b8f.

> - resetting if legacy `.storeOptionsAsProperties()` is enabled. This is definitely legacy but has not been deprecated yet. It could be forbidden rather than supported. Other than setting option values to `undefined`, there isn't a high-level way of "deleting" a value so hard to fully restore property state.

How about storing all properties of `this` as persistent option values in this mode?

aweebit:
> It is intended that the author can use any value, whether one of the ones Commander uses internally, or a custom value which means something to the author but not to Commander.

I did not think this was the case because of the `OptionValueSource` use in typings.

aweebit:
> - author set option values: before parse, from hooks, after parse

Could make `setOptionValueWithSource()` call `_setNonPersistentOptionValueWithSource()` when run from hooks.

aweebit:
> - implicitly set option values (might be ok)

They depend on input so should definitely be left non-persistent.

shadowspawn:
> > or a custom value which means something to the author but not to Commander.
> 
> I did not think those were supported because of the `OptionValueSource` use in typings.

Sorry about that. I was looking into that before I saw your message as I was expecting the types would make it clear and surprised when I found explicit list. Took me a while to find the clarification which I have trialed first in companion repo: https://github.com/commander-js/extra-typings/pull/3

lib/command.js
aweebit: It could be a problem if commander-js/extra-typings#3 is merged into the main repo, though.

aweebit: Resolved in https://github.com/tj/commander.js/pull/1919/commits/940b111eb40e865f10a96b6fdbbb866c990b7fec.

aweebit:
> Could make `setOptionValueWithSource()` call `_setNonPersistentOptionValueWithSource()` when run from hooks.

Implemented in https://github.com/tj/commander.js/pull/1919/commits/94e439d75d4b7510dc06cbed6e741f235034f236.

aweebit:
Sorry have been editing code from mobile, gonna check what's wrong when I come home.

shadowspawn:
No rush. (And somewhat related, I have had an unusual amount of time available in the last week, I will be less responsive this week.)

lib/command.js
```diff
@@ -780,34 +794,61 @@ Expecting one of '${allowedValues.join("', '")}'`);
...
-    * @param {string} source - expected values are default/config/env/cli/implied
+    * @param {string} [source]
     * @return {Command} `this` command for chaining
     */
 
   setOptionValueWithSource(key, value, source) {
+    const set = this._asyncParsing === undefined
+      ? this._setPersistentOptionValueWithSource
+      : this._setNonPersistentOptionValueWithSource;
+    set(key, value, source);
```
shadowspawn, line 813: I suspect you may need to use "bind" or "apply" to the method to have `this` defined correctly inside the called routine?

(I hit an `undefined` error while manually testing and had this on call stack.)

shadowspawn: Steps to reproduce:

```
import { Command, Option } from 'commander';
program.setOptionValue('beforeFirstParse', "1");
```

shadowspawn:
Reset is not being called in subcommands, but still ok as a proof of concept with root command.

lib/command.js
aweebit: Resolved in https://github.com/tj/commander.js/pull/1919/commits/56eed63dee40a7000aac08fd08197c641c3f6e70 (I think?)

aweebit:
> Reset is not being called in subcommands, but still ok as a proof of concept with root command.

Fixed in https://github.com/tj/commander.js/pull/1919/commits/4a7bc412e10e017499595b43ba523a13b8969a71.

aweebit:
All checks should pass now.

I think the only problem left is the one with options-as-properties. Do you think I should give [the idea with proxies]( https://github.com/tj/commander.js/pull/1919#issuecomment-1656577426) a try? Or is there maybe a reason why you would not want to use proxies in the library?

aweebit:
The command constructor now returns a proxy as suggested.

I personally think the proxy use is meaningful even outside the context of this PR

lib/command.js
```diff
@@ -1256,81 +1368,87 @@ Expecting one of '${allowedValues.join("', '")}'`);
...
 
-  _parseCommand(operands, unknown) {
-    const parsed = this.parseOptions(unknown);
-    this._parseOptionsEnv(); // after cli, so parseArg not called on both cli and env
-    this._parseOptionsImplied();
-    operands = operands.concat(parsed.operands);
-    unknown = parsed.unknown;
-    this.args = operands.concat(unknown);
+  _parseCommand(operands, unknown, async) {
+    this._asyncParsing = async;
```
shadowspawn, line 1372: Is the `_asyncParsing` property bring used to track whether in the middle of a phase, rather than "async"?

aweebit: Yes, for motivation see commit message of [94e439d](https://github.com/tj/commander.js/pull/1919/commits/94e439d75d4b7510dc06cbed6e741f235034f236). #1917 does not use `_asyncParsing` anymore, but there would still be a nasty redundancy that I think would not even result in a merge conflict when #1915 is merged, which I do hope will happen eventually. Maybe it is not such a good idea to keep other PRs in mind, though. I am not sure since I have almost zero experience with contributing to open source.

shadowspawn: > Maybe it is not such a good idea to keep other PRs in mind, though.

In my opinion it is ok to keep them in mind, and helpful to warn about potential merge issues. 

(Even ok in my opinion to suggest it should land after another PR, but then be prepared for a rewrite if that other one gets rejected!)

aweebit:
### TODO
- [ ] Take closing comments from #1921 into consideration and rethink returning a proxy from the constructor

aweebit:
> * resetting if legacy `.storeOptionsAsProperties()` is enabled. This is definitely legacy but has not been deprecated yet. It could be forbidden rather than supported. Other than setting option values to `undefined`, there isn't a high-level way of "deleting" a value so hard to fully restore property state.

 Your [comment](https://github.com/tj/commander.js/pull/1921#issuecomment-1661437046) from #1921 reads:

> > If you say this discrepancy is desired because keeping the old approach as backwards compatible as possible is
> 
> Effectively yes. The old approach is still available for backwards compatibility, not as an alternative mechanism for accessing options.

The way I interpret this is that supporting modern features in the legacy mode is of low importance. So how about we just resort to warnings in repeated parse calls (similar to those in #1917) when options-as-properties are enabled so that proxy use can be avoided? (I still think the solution with proxies is not bad, though.)

shadowspawn:
> avalanche of new PRs

Yes. Yes indeed. 😢 

Which PR would you like me to look at first? I am going to mark the others as on hold.

aweebit:
> Which PR would you like me to look at first? I am going to mark the others as on hold.

It does not really matter that much to me, but if I were to write a PR list sorted by the importance of the changes, it would look something like this:

### Patches
- #1941
- #1935 (might want to decouple from [#1929](https://github.com/tj/commander.js/issues/1929))
- #1922
- #1930 (the `.version()` fix needs to agree with [#1954](https://github.com/tj/commander.js/pull/1954))

### Enhancements fixing open issues
- #1924 (fixes [#1903](https://github.com/tj/commander.js/issues/1903))
  - The closely related #1923
- #1951 (fixes [#1901](https://github.com/tj/commander.js/issues/1901), quite robust and thus ready for review in my opinion, only tests missing)

### Other enhancements
- #1929
- #1934

### Warnings about presumably wrong library usage
- #1940 (currently closed, but additional motivation has been provided in [#1940 (comment)](https://github.com/tj/commander.js/pull/1940#issuecomment-1675873642))
- #1938 (includes an arguably questionable enhancement to parent handling, so might require more time to think about than the previous one, but the warning is very, very, [VERY](https://github.com/tj/commander.js/pull/1938#issuecomment-1667037227) useful)
- #1931 (not as useful as the previous one because it is likely the develop is okay with the help option being obscured, but a good complement to [#1923](https://github.com/tj/commander.js/issues/1923))
- #1955 (a reasonable addition since all warnings introduced in the PRs above are for usage patterns that are not always wrong, just often connected with a developer mistake)

### Using proxies
- #1919 (awaiting feedback)
- #1921 (awaiting feedback – currently closed, but has been reworked to minimize proxy use and so should be reopened)

### Optional changes
- #1950

### Should be superseded
- #1917 (by [#1938](https://github.com/tj/commander.js/issues/1938) and [#1940](https://github.com/tj/commander.js/issues/1940))

### Closed but might need reconsideration
- #1933 (the version number part, see [#1954 (comment)](https://github.com/tj/commander.js/pull/1954#issuecomment-1675952316))
- #1946 (discussion finished in [#1945](https://github.com/tj/commander.js/issues/1945), but reiterated in [#1947 (comment)](https://github.com/tj/commander.js/issues/1947#issuecomment-1675914455))
- #1927

### Issues that might need a PR
- #1936 (ongoing discussion)
- #1947
- #1949 (on hold)

----

Could make an issue out of the list so that it is not placed so randomly.

shadowspawn:
Support was added in Commander v13 for calling parse more than once: https://github.com/tj/commander.js/pull/2299

PR #1922, opened 2023-07-31
Make `copyInheritedSettings()` recursive

aweebit:
# Pull Request

## Problem

Pretty much all the inherited settings are such that it is only really meaningful for them to have the same values across the entire command hierarchy. However, `copyInheritedSettings()` only copies them at one level instead of doing it recursively. Because of that, the intended usage pattern is currently that the user adds subsubcommands only after the subcommand was added to its parent if the parent's inherited settings are meddled with (as clarified [here](https://github.com/tj/commander.js/pull/1915#issuecomment-1652787454)).

### Demo

```js
const sub = new Command('sub');
const subsub = sub.command('subsub');

program
  .showHelpAfterError()
  .addCommand(sub);
sub.copyInheritedSettings(program);

console.log(sub._showHelpAfterError); // true
console.log(subsub._showHelpAfterError); // false
```

## Solution

Make `copyInheritedSettings()` recursive.

### Changed
- _Breaking:_ make `.copyInheritedSettings()` recursive (settings are now inherited by the entire command hierarchy)

### DIFF
--- a/lib/command.js
+++ b/lib/command.js
@@ -106,6 +106,10 @@ class Command extends EventEmitter {
     this._showHelpAfterError = sourceCommand._showHelpAfterError;
     this._showSuggestionAfterError = sourceCommand._showSuggestionAfterError;
 
+    this.commands.forEach(command => {
+      command.copyInheritedSettings(this);
+    });
+
     return this;
   }
 
--- a/Readme.md
+++ b/Readme.md
@@ -546,7 +546,7 @@ subcommand is specified ([example](./examples/defaultCommand.js)).
 
 You can add alternative names for a command with `.alias()`. ([example](./examples/alias.js))
 
-For safety, `.addCommand()` does not automatically copy the inherited settings from the parent command. There is a helper routine `.copyInheritedSettings()` for copying the settings when they are wanted.
+For safety, `.addCommand()` does not automatically copy the inherited settings from the parent command. There is a helper routine `.copyInheritedSettings()` for copying the settings when they are wanted. The copying is done recursively, so the settings are inherited by the entire subcommand hierarchy.
 
 ### Command-arguments
 
--- a/tests/command.copySettings.test.js
+++ b/tests/command.copySettings.test.js
@@ -18,6 +18,28 @@ test('when add subcommand with .command() then calls copyInheritedSettings from 
   expect(copySettingMock).toHaveBeenCalledWith(program);
 });
 
+test('when copyInheritedSettings on command with subcommand hierarchy then copies recursively', () => {
+  const source = new commander.Command();
+  const cmd = new commander.Command();
+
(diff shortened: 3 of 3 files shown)


lib/command.js
```diff
@@ -106,6 +106,10 @@ class Command extends EventEmitter {
     this._showHelpAfterError = sourceCommand._showHelpAfterError;
     this._showSuggestionAfterError = sourceCommand._showSuggestionAfterError;
 
+    this.commands.forEach(command => {
+      command.copyInheritedSettings(this);
```
shadowspawn, line 110: Somewhat cosmetic after the copy, but I suggest: 
```suggestion
      command.copyInheritedSettings(sourceCommand);
```

aweebit:
At the risk of making the `.copyInheritedSettings()` behavior more confusing, I suggest making calls to it with no arguments result in inherited settings being copied from `this` to all its descendants.

Use case: want to change behavior of an entire command hierarchy when the root command is returned from a function with all subcommands already registered (like in https://stackoverflow.com/questions/76896860 first referenced in [#1917 (comment)](https://github.com/tj/commander.js/pull/1917#issuecomment-1678335650)), for example for testing purposes (setting `exitOverride` like in the example etc.)

Even with changes introduced in this PR, the `.commands` array would still have to be iterated to achieve the desired result. If the new suggestion is accepted, one call will be enough.

Could also dump this PR if the suggestion is accepted. The change would be non-breaking then.

shadowspawn:
Currently `copyInheritedSettings` does one simple thing, which I like. It exposes what used to be an internal step so authors can use it for assembling commands. I don't like making routines more flexible (and complicated) without a clear vision. In your case, you are already wondering about two different implementations. My instinct is not clear enough yet (i.e. I want to see another/other issues).

Given it is easy enough for author to implement exactly what they want, and I think a different approach would be better for the `.exitOverride()` example, I am going to close this for now.

The apply-settings-after-creating-program situation might be handled by something like:
```
program.recursiveSet((cmd) -> {
   cmd.exitOverride();
   cmd.helpOption('-a, --assist', 'assistance');
});
```

aweebit:
> The apply-settings-after-creating-program situation might be handled by something like:
> 
> ```
> program.recursiveSet((cmd) -> {
>    cmd.exitOverride();
>    cmd.helpOption('-a, --assist', 'assistance');
> });
> ```

`.recursiveSet()` would just have to do exactly what a `.copyInheritedSettings()` call with no arguments would do in my suggestion, just after calling the callback function. There is no need to make it so complicated.

> In your case, you are already wondering about two different implementations.

It's a good thing this PR was kept open long enough for me to come up with a better idea :)) I think just adding support for the call with no arguments is great because the suggested behavior is likely what most users really want when they use `.copyInheritedSettings()`, but at the same time, the change wouldn't be breaking and finer control would still be possible for cases where it's needed.

Issue #1903, opened 2023-07-06
Commander unsafely allows commands with duplicate `name`s to be registered


huntie:
We ran into an issue recently in a project that depends on `commander`, which dynamically defines commands based on plugins. Commander's internal array data structure for commands, and the `Array.find` based lookup on `name`, meant that repeat calls to `command()` produced unexpected behaviour (the most recent call does **_not_** redeclare the command).

At minimum, I believe this should **error**.

- https://github.com/react-native-community/cli/pull/1999

**Example**

```js
const { Command } = require('commander');
const program = new Command();
program.command('split').action((str, options) => console.log('1st'));
program.command('split').action((str, options) => console.log('2nd'));
program.parse();
```
Output will always be the first.
```sh
$ node example
Usage: example [options] [command]
Options:
  -h, --help      display help for command
Commands:
  split
  split
  help [command]  display help for command
$ node example split
1st
```

**Expectation**

Shouldn't allow duplicates and should throw an exception if a `name` already exists.

shadowspawn:
Commander does not have tests for all possible mistakes/accidental-misuse by the "author", but willing to consider more as they come up. Mixing in plugins does make it possible for the issue to come up for the end-user rather than for the author as such.

The duplicate primary name does have a dubious outcome. I am less worried about aliases clashing, but could perhaps actually handle this and drop clashing alias.

aweebit:
Opened #1924 with a fix.

shadowspawn:
Reworked PR in #2509.

shadowspawn:
Released in Commander v12.0.0.

PR #1923, opened 2023-07-31
Prevent options with flags already in use from being added

aweebit:
# Pull Request

## Problem

Adding options with flags that are already in use is allowed.

### Demo
```js

var program = new Command();
program
  .option('-a, --first')
  .option('-a, --second');
console.log(program.opts()); // { first: true }

var program = new Command();
program
  .option('-V, --version [arg]', 'added via .option()')
  .version('1.0.0')
  .action(() => {
    console.log('called action handler');
    console.log('version value is', program.opts().version);
  });
program.outputHelp(); // both version option and option added via .option() are shown
program.parse(['-V', '2.0.0'], { from: 'user' }); // action handler not called
```

### Changed
- _Breaking:_ throw an error when `.addOption()` is called with an option whose flags are already in use

### …solving similar problems
- #1924
- #1931 (introduces `_registerOption()`, too – watch out for merge conflicts!)

### DIFF
--- a/lib/command.js
+++ b/lib/command.js
@@ -507,6 +507,31 @@ Expecting one of '${allowedValues.join("', '")}'`);
    * @return {Command} `this` command for chaining
    */
   addOption(option) {
+    const flagString = [option.short, option.long]
+      .filter(flag => flag)
+      .join(', ');
+    const flagConflictCommonMessage = `Cannot add option ${flagString}`;
+
+    const shortConflictOption = this.options.find((other) => (
+      option.short && option.short === other.short
+    ));
+    const longConflictOption = this.options.find((other) => (
+      option.long && option.long === other.long
+    ));
+
+    if (shortConflictOption) {
+      if (shortConflictOption === longConflictOption) {
+        throw new Error(`${flagConflictCommonMessage}.
+An option with the same flags has already been added`);
+      }
+      throw new Error(`${flagConflictCommonMessage}.
+An option with flag ${shortConflictOption.short} has already been added`);
+    }
+    if (longConflictOption) {
+      throw new Error(`${flagConflictCommonMessage}.
+An option with flag ${longConflictOption.long} has already been added`);
+    }
+
     const oname = option.name();
     const name = option.attributeName();
 
--- a/lib/command.js
+++ b/lib/command.js
@@ -515,17 +515,15 @@ Expecting one of '${allowedValues.join("', '")}'`);
     const shortConflictOption = option.short && this._findOption(option.short);
     const longConflictOption = option.long && this._findOption(option.long);
 
-    if (shortConflictOption) {
+    if (shortConflictOption || longConflictOption) {
       if (shortConflictOption === longConflictOption) {
         throw new Error(`${flagConflictCommonMessage}.
 An option with the same flags has already been added`);
       }
       throw new Error(`${flagConflictCommonMessage}.
-An option with flag ${shortConflictOption.short} has already been added`);
-    }
-    if (longConflictOption) {
-      throw new Error(`${flagConflictCommonMessage}.
-An option with flag ${longConflictOption.long} has already been added`);
+An option with flag ${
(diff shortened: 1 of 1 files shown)


shadowspawn:
Commander only checks some of the possible mistakes/accidental-misuse by the "author", but willing to consider more as they come up. 

Related: #1903

shadowspawn:
I am hopeful this can be done with less code by using the internal `Command._findOption`.

shadowspawn:
> Adding options with flags that are already in use is allowed.

Optional: explain briefly why this is a problem.

(As an exercise for you, not because I need an explanation.)

aweebit:
> I am hopeful this can be done with less code by using the internal `Command._findOption`.

Done in [5ec2249](https://github.com/tj/commander.js/pull/1923/commits/5ec2249201c37fad75c001115e8d922bd146e836). Further simplifications would make the error message less detailed.

aweebit:
> > Adding options with flags that are already in use is allowed.
> 
> Optional: explain briefly why this is a problem.
> 
> (As an exercise for you, not because I need an explanation.)

I did not want to spend time on it because I thought the demo already made it obvious.

shadowspawn:
> I did not want to spend time on it because I thought the demo already made it obvious.

Ok.

shadowspawn:
The way an option is usually referred to in the end-user messages is `'${option.flags}'` and we can use same format in author facing error. My reasoning for this format is the combo is potentially more recognisable than a single flag.

I suggest listing the found item in same format too. I am deliberately simplifying here and not doing the error breakdown you did (although I admire the effort). I think it will be clear enough to the author? 

```
    const matchingOption = (option.short && this._findOption(option.short)) || (option.long && this._findOption(option.long));
    if (matchingOption) {
      throw new Error(`Cannot add option '${option.flags}' as an option using same flag has already been been added
- conflicts with option '${matchingOption.flags}'`);
    }
```

aweebit:
> The way an option is usually referred to in the end-user messages is `'${option.flags}'` and we can use same format in author facing error. My reasoning for this format is the combo is potentially more recognisable than a single flag.

A combo is already used, but with option-argument and extra spaces removed: see `flagString` on line 510. I think it is better than what you suggest because the actual flags are the only thing that matters here.

shadowspawn:
I did forget you were using one combo! But no need to calculate it, just use `'${option.flags}'` which matches what the author specified and other messages.

I think identifying the conflicting option is important and the whole `matchingOption.flags` is easier to find than a flag alone.

aweebit:
Fair enough. I used your suggestion in [9721ae8](https://github.com/tj/commander.js/pull/1923/commits/9721ae86702fa9f446d88b444a734d7536fdd7ae), just added the command name and changed "as" to "since" because it reads better (with "as", I first read it as `(Cannot add option '${option.flags}' as an option using same flag) has already been been added` and was confused).

lib/command.js
```diff
@@ -507,6 +507,13 @@ Expecting one of '${allowedValues.join("', '")}'`);
    * @return {Command} `this` command for chaining
    */
   addOption(option) {
+    const matchingOption = (option.short && this._findOption(option.short))
+      || (option.long && this._findOption(option.long));
+    if (matchingOption) {
+      throw new Error(`Cannot add option '${option.flags}'${this._name && ` to '${this._name}'`} since an option using same flag has already been added
```
shadowspawn, line 513: command name is good addition 👍

shadowspawn (changes requested):
Code looks good to me, simple and informative and effective.

Just needs a couple of tests, for short and long conflicts.

shadowspawn:
(I am concerned this may be a blocking change for people upgrading big programs from older versions of Commander, because their code will be broken until they resolve the clashes. I might want to add configuration to make this optional specifically to ease migration, but won't do that now. We can do a prerelease and see if people hit issues that they considered an obstacle.)

shadowspawn:
Carried on in #2055

PR #1924, opened 2023-08-01
Prevent aliases already in use and subcommands with such an alias or the name already in use from being added

aweebit:
# Pull Request

## Problem

Adding aliases that are already in use and subcommands with such an alias or the name that is already in use is allowed.

### Demo
```js
program.command('sub').action(() => console.log('first'));
program.command('sub').action(() => console.log('second'));
program.parse(['sub'], { from: 'user' }); // output will always be first
```

```js
program.command('sub').action(() => console.log('first'));
const explicit = new Command('explicit')
  .alias('sub')
  .action(() => console.log('second'));
program.addCommand(explicit);
program.parse(['sub'], { from: 'user' }); // output will always be first
```

```js
program.command('sub');
const explicit = new Command('explicit');
program.addCommand(explicit);
explicit.alias('sub'); // does not throw
```

### Original issue
- #1903

### Changed
- _Breaking:_ throw an error when trying to add a subcommand with a name or an alias that is already in use
- _Breaking:_ throw an error when trying to add an alias that is already in use to a subcommand

### …solving similar problems
- #1923

### DIFF
--- a/lib/command.js
+++ b/lib/command.js
@@ -110,14 +110,40 @@ class Command extends EventEmitter {
   }
 
   /**
-   * @param {string} name
+   * @param {string} names
+   * @param {string[]} aliases
+   * @param {Command} matchingCommand
+   * @return {string}
    * @api private
    */
 
-  _checkForConflictingCommandNames(name) {
-    if (this._findCommand(name)) {
-      throw new Error(`Cannot add command '${name}'${this._name && ` to '${this._name}'`}.
-A subcommand with the same name has already been added`);
+  _getConflictingCommandNamesMessageDetail(name, aliases, matchingCommand) {
+    const matchingCommandNames = matchingCommand._aliases.concat(
+      matchingCommand._name
+    );
+    if (matchingCommandNames.includes(name)) return 'same name';
+    const matchingAlias = aliases.find(
+      alias => matchingCommandNames.includes(alias)
+    );
+    return `alias '${matchingAlias}'`;
+  }
+
+  /**
+   * @param {string} names
+   * @param {string[]} aliases
+   * @api private
+   */
+
+  _checkForConflictingCommandNames(name, aliases = []) {
+    const names = aliases.concat(name);
+    const matchingCommand = names
+      .map(name => this._findCommand(name))
+      .filter(found => found)[0];
+    if (matchingCommand) {
+      const messageDetail = this._getConflictingCommandNamesMessageDetail(
+        name, aliases, matchingCommand
+      );
+      throw new Error(`Cannot add command '${name}'${this._name && ` to '${this._name}'`} since a command using the ${messageDetail} has already been added`);
     }
   }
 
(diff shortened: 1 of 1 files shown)


shadowspawn:
I had done some thinking when issue first added, and aliases are a complication. With current PR code aliases on the first command get checked when the second command is added. But I think not vice versa.

Some possibilities:
- don't worry, covered much more than before!
- do the whole conflict command check from start of parse rather than as command added (this is reasonable, just slightly annoying to mix logic). Still highly likely to be detected by author rather than found by end-user, plugins excepted.
- check at runtime and throw if ambiguous name used (bad, because depends on parameters used so may be missed by author. But mentioned it because I considered it!)

aweebit:
> I had done some thinking when issue first added, and aliases are a complication. With current PR code aliases on the first command get checked when the second command is added. But I think not vice versa.
> 
> Some possibilities:
> 
> * don't worry, covered much more than before!
> * do the whole conflict command check from start of parse rather than as command added (this is reasonable, just slightly annoying to mix logic). Still highly likely to be detected by author rather than found by end-user, plugins excepted.
> * check at runtime and throw if ambiguous name used (bad, because depends on parameters used so may be missed by author. But mentioned it because I considered it!)

I have come up with a better solution, check [bd12c90](https://github.com/tj/commander.js/pull/1924/commits/bd12c90881c296c7c27b3af34a8aa8b398081cd1) and the updated PR description.

aweebit:
> I have come up with a better solution, check [bd12c90](https://github.com/tj/commander.js/pull/1924/commits/bd12c90881c296c7c27b3af34a8aa8b398081cd1) and the updated PR description.

My suggestion from [the first comment](https://github.com/tj/commander.js/pull/1917#issuecomment-1656365692) in #1917 enabling sharing and stand-alone use of subcommands won't work if this change is accepted, though. The checks would have to be done at the beginning of parse calls as you suggested.

shadowspawn:
The way commands with alias get displayed in the help is like `one|1`

I tried a rework to use that form in error messages:

```
  _checkForConflictingCommandNames(name, aliases = []) {
    let newKey;
    let oldKey;
    const makeKey = (nameOrAlias, matchingCommand) => {
      if (nameOrAlias === matchingCommand.name()) return nameOrAlias;
      return `${matchingCommand.name()}|${nameOrAlias}`;
    };
    let matchingCommand = this._findCommand(name);
    if (matchingCommand) {
      newKey = name;
      oldKey = makeKey(name, matchingCommand);
    } else {
      const alias = aliases.find((alias) => {
        matchingCommand = this._findCommand(alias);
        return matchingCommand;
      });
      if (matchingCommand) {
        newKey = `${name}|${alias}`;
        oldKey = makeKey(alias, matchingCommand);
      }
    }
    if (matchingCommand) {
      throw new Error(`Cannot add command '${newKey}'${this._name && ` to '${this._name}'`} since command '${oldKey}' has already been added`);
    }
  }
```

Example error when use `addCommand()` and alias clashes with alias :
```
Error: Cannot add command 'three|1' since command 'one|1' has already been added
```

shadowspawn:
Closing in favour of #2059

PR #1925, opened 2023-08-01
Add documentation for implicitly constructed commands' setting inheritance

### LINKED ISSUE #1916: On command and subcommand reusability and setting inheritance

aweebit:
Partially fixes #1916.

Cherry-picked from #1917 because the docs extension discussion is [already finalized](https://github.com/tj/commander.js/pull/1917#discussion_r1278270441) and the change is independent from the other changes introduced there, so it can be merged even now.

### Added
- documentation for implicitly constructed commands' setting inheritance

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -546,6 +546,8 @@ subcommand is specified ([example](./examples/defaultCommand.js)).
 
 You can add alternative names for a command with `.alias()`. ([example](./examples/alias.js))
 
+Commands added with `.command()` automatically inherit settings for which inheritance is meaningful from the parent command, but only upon the subcommand creation. The setting changes made after calling `.command()` are not inherited.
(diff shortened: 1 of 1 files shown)


aweebit:
(Force push is due to a commit amendment changing the commit message so that the `Co-authored-by` trailer is placed correctly.)

shadowspawn (approved):
I have wondered about adding something informative but concise about inherited settings for a long time. 👍

abetomo (approved):
👍

PR #1926, opened 2023-08-02
Enforce correct option usage and support custom option flag and name extraction in `helpOption()`

aweebit:
# Pull Request

## Problem

```js
var program = new Command();
var helpOption = program.createOption(program._helpFlags, program._helpDescription);
var helpOptionName = helpOption.attributeName();
console.log(helpOptionName); // 'help'
program.setOptionValue(helpOptionName, 123); // allowed!

var program = new Command();
program.option('--my-help');
program.helpOption('--my-help'); // allowed!

var program = new Command();
program.setOptionValue('myHelp', 123);
program.helpOption('--my-help'); // allowed!

var program = new Command();
program.version('1.0.0');
console.log(program.version()); // '1.0.0'
var versionOption = program.createOption(program._versionOptionName);
var versionOptionName = versionOption.attributeName();
console.log(versionOptionName); // 'version'
program.setOptionValue(versionOptionName, '1.0.1'); // allowed!
console.log(program.version()); // still '1.0.0'

var program = new Command();
program.option('--my-version');
program.version('1.0.0', '--my-version'); // allowed!

var program = new Command();
program.setOptionValue('myVersion', 123);
program.version('1.0.0', '--my-version'); // allowed!

var program = new Command();
program.setOptionValue('foo', 'bar');
console.log(program.getOptionValue('foo')); // 'bar'
console.log(program.opts()); // { foo: 'bar' }
program.storeOptionsAsProperties(); // allowed!
console.log(program.getOptionValue('foo')); // undefined
console.log(program.opts()); // {}
```

## Solution

Throw errors in all demonstrated cases (partially borrowed from [a5afe93](https://github.com/tj/commander.js/pull/1921/commits/a5afe935defd7f928a236a4bf237d73d79011a50) in the now-closed #1921).

Additionally, the code for the help option has been refactored to support custom option flag and name extraction mechanisms defined in subclasses. To achieve that, an `Option` instance returned from a `createOption()` call is used for extraction in `helpOption()` (the same is currently done in `version()`), which is now called in the constructor and `copyInheritedSettings()`.

The first two commits are tiny improvements to the code that have been made while working on the solution and don't really belong anywhere else.

### Changed
- _Breaking:_ throw errors when trying to set values for the help and version options
- _Breaking:_ throw error when trying to add help or version option with `.opts()` name already in use
- _Breaking:_ throw error when calling `.storeOptionsAsProperties()` after setting an option value
[...]

PR #1927, opened 2023-08-02
Handle falsy values in `addHelpCommand()` properly

aweebit:
Cherry-picked from #1926 for better separation of concerns.

## Problem

Only `false` is treated as falsy in `addHelpCommand()`.

### Fixed
- falsy parameter value handling in `.addHelpCommand()`

### DIFF
--- a/lib/command.js
+++ b/lib/command.js
@@ -365,7 +365,7 @@ class Command extends EventEmitter {
    */
 
   addHelpCommand(enableOrNameAndArgs, description) {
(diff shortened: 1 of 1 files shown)


shadowspawn:
The `false` check is as intended.

(I would normally leave this open for discussion, but you are have opened ~15 Pull Requests in the last week with ~160 comments between us. I don't have the time to continue that level of feedback.)

PR #1928, opened 2023-08-02
Throw error on `storeOptionsAsProperties()` after setting option values

aweebit:
Cherry-picked from https://github.com/tj/commander.js/pull/1926 for better separation of concerns.

Additionally, a unit test has been added to cover the error.

## Problem

```js
program.setOptionValue('foo', 'bar');
console.log(program.getOptionValue('foo')); // 'bar'
console.log(program.opts()); // { foo: 'bar' }
program.storeOptionsAsProperties(); // allowed!
console.log(program.getOptionValue('foo')); // undefined
console.log(program.opts()); // {}
```

### Changed
* _Breaking:_ throw error when calling `.storeOptionsAsProperties()` after setting an option value

### DIFF
--- a/tests/commander.configureCommand.test.js
+++ b/tests/commander.configureCommand.test.js
@@ -84,3 +84,11 @@ test('when storeOptionsAsProperties() after adding option then throw', () => {
     program.storeOptionsAsProperties(false);
   }).toThrow();
 });
+
+test('when storeOptionsAsProperties() after setting option value then throw', () => {
+  const program = new commander.Command();
+  program.setOptionValue('foo', 'bar');
+  expect(() => {
+    program.storeOptionsAsProperties(false);
+  }).toThrow();
+});
--- a/tests/commander.configureCommand.test.js
+++ b/tests/commander.configureCommand.test.js
@@ -81,7 +81,7 @@ test('when storeOptionsAsProperties() after adding option then throw', () => {
   const program = new commander.Command();
   program.option('--port <number>', 'port number', '80');
   expect(() => {
-    program.storeOptionsAsProperties(false);
+    program.storeOptionsAsProperties();
   }).toThrow();
 });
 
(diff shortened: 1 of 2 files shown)


tests/commander.configureCommand.test.js
```diff
@@ -84,3 +84,11 @@ test('when storeOptionsAsProperties() after adding option then throw', () => {
     program.storeOptionsAsProperties(false);
   }).toThrow();
 });
+
+test('when storeOptionsAsProperties() after setting option value then throw', () => {
+  const program = new commander.Command();
+  program.setOptionValue('foo', 'bar');
+  expect(() => {
+    program.storeOptionsAsProperties(false);
```
shadowspawn, line 92: Passing `false` is same as the default state, so it isn't the interesting case and pretty unlikely to be encountered:
```suggestion
    program.storeOptionsAsProperties();
```

(I know the throw does not currently care, but I would still like to test the interesting case.)

shadowspawn (changes requested):
This might help someone making a gradual (and big) transition away from storing options as properties and as first step switching from `program.foo = true` to `program.setOptionValue('foo', true)`.

One minor change.

tests/commander.configureCommand.test.js
aweebit: I actually just copied the call from above, and before doing that, I took a minute to think about why you'd decided to pass `false` in that previous test. The way I ended up explaining it to myself is that you'd probably wanted to test the unproblematic call that would not even change anything if it were allowed because that demonstrates how intransigent the check is. It is obvious an error will be thrown when passing `true` or nothing if even passing `false` results in an error (despite the fact that the call would simply be no-op otherwise). That last bit is not so obvious, so that is why I think the `false` case is actually the more interesting one here.

shadowspawn: > I actually just copied the call from above

Oh, digging... The previous test logic dates back to when the default was to store options as properties, and was not correctly updated when the default behaviour changed. Please change the previous test too! Either pass true or no param, either is fine.

shadowspawn (approved):
LGTM

abetomo (approved):
+1

PR #1930, opened 2023-08-03
[PR with tiny fixes] Fix help command for subcommands with an executable handler and only a short help flag. Do not use undefined long help option flag in legacy code. Change initial variable values in test for better error messages. 

aweebit:
Tiny fixes that don't really belong anywhere else.

### Changed
- ~do not call `preSubcommand` hooks when displaying help for a subcommand with an executable handler~
  - ~_(technically breaking, but hey, are we really going to assume someone relied on this?)_~
- ~deviating help flags specified for a subcommand with an executable handler via `.helpOption()` are now respected by the parent's help command (passed to the subcommand instead of the parent's help flag)~

### Fixed
- ~`.version()` optional parameter type~
- subcommands with an executable handler and only a short help flag are now handled correctly by the parent's help command
- stopped using undefined long help option flag in legacy code

### DIFF
--- a/lib/command.js
+++ b/lib/command.js
@@ -1809,7 +1809,7 @@ Expecting one of '${allowedValues.join("', '")}'`);
    *
    * You can optionally supply the  flags and description to override the defaults.
    *
-   * @param {string} [str]
+   * @param {string} str
    * @param {string} [flags]
    * @param {string} [description]
    * @return {this | string} `this` command for chaining, or version string if no arguments
--- a/lib/command.js
+++ b/lib/command.js
@@ -1101,9 +1101,9 @@ Expecting one of '${allowedValues.join("', '")}'`);
     }
 
     // Fallback to parsing the help flag to invoke the help.
-    if (this._helpLongFlag) {
-      return this._dispatchSubcommand(subcommandName, [], [this._helpLongFlag]);
-    }
+    return this._dispatchSubcommand(subcommandName, [], [
+      this._helpLongFlag || this._helpShortFlag
+    ]);
   }
 
   /**
--- a/lib/command.js
+++ b/lib/command.js
@@ -1807,7 +1807,7 @@ Expecting one of '${allowedValues.join("', '")}'`);
    *
    * You can optionally supply the  flags and description to override the defaults.
    *
-   * @param {string} str
+   * @param {string} [str]
    * @param {string} [flags]
    * @param {string} [description]
    * @return {this | string} `this` command for chaining, or version string if no arguments
--- a/lib/command.js
+++ b/lib/command.js
@@ -1101,7 +1101,9 @@ Expecting one of '${allowedValues.join("', '")}'`);
     }
 
     // Fallback to parsing the help flag to invoke the help.
-    return this._dispatchSubcommand(subcommandName, [], [this._helpLongFlag]);
+    if (this._helpLongFlag) {
+      return this._dispatchSubcommand(subcommandName, [], [this._helpLongFlag]);
+    }
   }
 
   /**
@@ -2034,7 +2036,9 @@ Expecting one of '${allowedValues.join("', '")}'`);
     }
     context.write(helpInformation);
 
-    this.emit(this._helpLongFlag); // deprecated
+    if (this._helpLongFlag) {
+      this.emit(this._helpLongFlag); // deprecated
+    }
     this.emit('afterHelp', context);
     getCommandAndParents(this).forEach(command => command.emit('afterAllHelp', context));
   }
--- a/typings/index.d.ts
+++ b/typings/index.d.ts
@@ -293,7 +293,7 @@ export class Command {
    *
    * You can optionally supply the  flags and description to override the defaults.
    */
-  version(str?: string, flags?: string, description?: string): this;
+  version(str: string, flags?: string, description?: string): this;
 
   /**
    * Define a command, implemented using an action handler.
--- a/typings/index.d.ts
+++ b/typings/index.d.ts
@@ -293,7 +293,7 @@ export class Command {
    *
    * You can optionally supply the  flags and description to override the defaults.
    */
-  version(str: string, flags?: string, description?: string): this;
+  version(str?: string, flags?: string, description?: string): this;
 
   /**
    * Define a command, implemented using an action handler.
--- a/tests/command.addHelpText.test.js
+++ b/tests/command.addHelpText.test.js
@@ -181,7 +181,7 @@ describe('context checks with full parse', () => {
   });
 
   test('when help requested then context.error is false', () => {
-    let context = {};
+    let context;
     const program = new commander.Command();
     program
       .exitOverride()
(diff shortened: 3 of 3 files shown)


aweebit:
The `.version()` type fix has to agree with whatever decision about the future of the call with no parameters is made in #1954.

### Note for myself
One. Change. Per. PR. **Always.** (no matter how tiny they are)

lib/command.js
```diff
@@ -1101,7 +1101,9 @@ Expecting one of '${allowedValues.join("', '")}'`);
     }
 
     // Fallback to parsing the help flag to invoke the help.
-    return this._dispatchSubcommand(subcommandName, [], [this._helpLongFlag]);
+    if (this._helpLongFlag) {
+      return this._dispatchSubcommand(subcommandName, [], [this._helpLongFlag]);
```
shadowspawn, line 1105: I think this needs to try something and not just ignore.

I suggest use `--help` when `this._helpLongFlag` undefined?

lib/command.js
```diff
@@ -1807,7 +1809,7 @@ Expecting one of '${allowedValues.join("', '")}'`);
    *
    * You can optionally supply the  flags and description to override the defaults.
    *
-   * @param {string} str
+   * @param {string} [str]
```
shadowspawn, line 1812: No, I don't want churn on something I want to deprecate.

lib/command.js
```diff
@@ -2034,7 +2036,9 @@ Expecting one of '${allowedValues.join("', '")}'`);
     }
     context.write(helpInformation);
 
-    this.emit(this._helpLongFlag); // deprecated
+    if (this._helpLongFlag) {
```
shadowspawn, line 2039: Skipping the emit is ok, but for consistent behaviour I think do same as other case and use `'--help'` if undefined.

typings/index.d.ts
```diff
@@ -294,6 +294,7 @@ export class Command {
    * You can optionally supply the  flags and description to override the defaults.
    */
   version(str: string, flags?: string, description?: string): this;
+  version(): string;
```
shadowspawn, line 297: No churn.

typings/index.test-d.ts
```diff
@@ -36,6 +36,7 @@ expectType<commander.Command | null>(program.parent);
 expectType<commander.Command>(program.version('1.2.3'));
 expectType<commander.Command>(program.version('1.2.3', '-r,--revision'));
 expectType<commander.Command>(program.version('1.2.3', '-r,--revision', 'show revision information'));
+expectType<string>(program.version());
```
shadowspawn, line 39: No churn.

shadowspawn (changes requested):
Couple of good fixes for the reasonably recent possibility the help option might only have short flag.

Changes noted in separate comments.

lib/command.js
shadowspawn: Thinking again, should use short flag since user specifically set it up that way.

lib/command.js
shadowspawn: Thinking again, I think keep what you have. Skipping the emit when no long flag is fine. There is new better way of appending help so anyone using help option with only a short flag can do it the new way.

lib/command.js
aweebit: Oh, I didn't even look at it carefully, just thought it was some deprecated stuff like in [the other place](https://github.com/tj/commander.js/pull/1930/files/f3a45fb082dc1d7a5c85c7f13d20c615d9b2b613#diff-d8db36db6bfb32591336006f1dc1adee9e2caf7b021469995aa687c5c5723d0cR2039) where only `_helpLongFlag` is checked below. My bad.

lib/command.js
```diff
@@ -1101,7 +1101,9 @@ Expecting one of '${allowedValues.join("', '")}'`);
     }
 
     // Fallback to parsing the help flag to invoke the help.
-    return this._dispatchSubcommand(subcommandName, [], [this._helpLongFlag]);
+    if (this._helpLongFlag) {
+      return this._dispatchSubcommand(subcommandName, [], [this._helpLongFlag]);
+    }
```
aweebit, line 1106: Could use the flags of `subCommand` that are always inherited from the parent command but can be changed by the library user via `.helpOption()` later if they know different help flags are used in the executable.

```suggestion
    return this._dispatchSubcommand(subcommandName, [], [
      subCommand._helpLongFlag || subCommand._helpShortFlag
    ]);
```

What do you think?

aweebit: I actually don't like the `._dispatchSubcommand()` call here at all.
- It introduces an inconsistency between how commands with and without an executable handler are handled since the `preSubcommand` hooks are only called for the former.
- It makes a new `._findCommand()` call internally, which is redundant.

So I suggest
```suggestion
    this._executeSubCommand(subCommand, [
      subCommand._helpLongFlag || subCommand._helpShortFlag
    ]);
```

lib/command.js
```diff
@@ -1096,12 +1096,16 @@ Expecting one of '${allowedValues.join("', '")}'`);
       this.help();
     }
     const subCommand = this._findCommand(subcommandName);
-    if (subCommand && !subCommand._executableHandler) {
+    if (!subCommand) {
+      this.help({ error: true });
```
shadowspawn, line 1100: No, this is wrong. This routine is only called to implement the help command. It might be invoked by end-user with a subcommand name:
```
program help
program help subcommand
```

It might help to look at the PR, which was about invoking help on a subcommand more robust: https://github.com/tj/commander.js/pull/1864

shadowspawn: Oh, but might not be valid subcommand name.
Oh, but that is handled by fall through case.

In summary, this piece of code is about directly calling subcommand help if we can reach it, so don't need to worry about help flags or help command.

lib/command.js
shadowspawn: No. Just fix the issue with the undefined help flag. Stay focused.

lib/command.js
aweebit: What exactly is wrong?

If you mean the `this.help({ error: true })` call, that is exactly what currently happens in `._dispatchSubcommand()` if no subcommand with the supplied name is found.

shadowspawn: Right. Which I am happy happening there. This routine is short-circuiting one case without refactoring the existing somewhat complex code.

lib/command.js
aweebit: Okay, gonna open a new PR for this.

shadowspawn (changes requested):
New changes have gone from fixing a bug with undefined (easy to approve) to refactoring code (opinionated and time consuming to review).

I want the simple fix please.

aweebit:
Sorry about the force pushes, there is just already enough clutter in other PRs due to this one, and I don't want to introduce even more.

shadowspawn (approved):

aweebit:
Thank you for the patience, I know this PR is a mess.

abetomo (approved):
+1

PR #1931, opened 2023-08-03
Warn about obscured help option

aweebit:
An extension of #1929 adding warnings about obscured help option as suggested in #1926 ([comment](https://github.com/tj/commander.js/pull/1926#issuecomment-1663029635)).

Additionally proves the usefulness of storing the `Option` instance returned from `createOption()` in the `_helpOption` property (a change introduced by [9857b1e](https://github.com/tj/commander.js/pull/1929/commits/9857b1eea2e64e42ab8710ab80cef98e0a71b354) in #1929), since in the solution, not only the individual flags but also the originally provided `flags` value needs to be accessed, and in the future it could be the `.name()`, the `.attributeName()` or something else pertaining to the help option. It makes no sense to add a property to the `Command` instance each time such a property is needed, it is way better to simply store the `Option` instance.

Has been merged with the already approved tiny fixes PR https://github.com/tj/commander.js/pull/1930 because the parent #1929 has been merged with it.

## Problem

States in which the help options is obscured are supported by the library (see `Help.visibleOptions()` and the corresponding unit tests), but definitely not desirable.

```js
const desc = 'added via .option()';

var program = new Command();
program
  .option('-h, --help <arg>', desc) // help option is obscured
  .outputHelp();

var program = new Command();
program
  .option('-h, --host <arg>', desc)
  .option('--help <arg>', desc) // help option is obscured
  .outputHelp();

var program = new Command();
program
  .option('-c, --custom-help <arg>', desc)
  .helpOption('-c, --custom-help') // help option is obscured
  .outputHelp();

// Support for this option combination is currently missing because of a wrong Help.visibleOptions() implementation.
// I will open another PR to address that soon. (Update: Opened #1935)
var program = new Command();
program
  .option('-c, --custom-help <arg>', desc)
  .helpOption('-c') // help option is obscured
  .outputHelp();

var program = new Command();
program
  .option('-c, --custom-help <arg>', desc)
  .helpOption('--custom-help') // help option is obscured
  .outputHelp();

var program = new Command();
program
  .option('-c <arg>', desc)
  .option('--custom-help <arg>', desc)
  .helpOption('-c, --custom-help') // help option is obscured
  .outputHelp();

var program = new Command();
program
  .option('-c <arg>', desc)
  .helpOption('-c') // help option is obscured
  .outputHelp();

var program = new Command();
program
  .option('--custom-help <arg>', desc)
  .helpOption('--custom-help') // help option is obscured
  .outputHelp();
```

## Solution

Add warnings when such a state is reached similar to those in #1915 and #1917.

### Added
- warnings about obscured help option

### Changed
- (from [#1929](https://github.com/tj/commander.js/pull/1929)) _Breaking:_ use `.createOption()` in `.helpOption()` to support custom option flag extraction

### …solving similar problems
- #1923 (introduces `_registerOption()`, too – watch out for merge conflicts!)

- #1915
- #1917
- #1938
- #1940

- #1955

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -14,7 +14,7 @@ const javascriptSettings = {
 };
 
 const typescriptSettings = {
-  files: ['*.ts'],
+  files: ['*.ts', '*.mts'],
   parserOptions: {
     project: './tsconfig.json'
   },
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -12,7 +12,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        node-version: [14.x, 16.x, 18.x, 20.x]
+        node-version: [16.x, 18.x, 20.x]
         os: [ubuntu-latest, windows-latest, macos-latest]
 
     steps:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,18 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 <!-- markdownlint-disable MD024 -->
 <!-- markdownlint-disable MD004 -->
 
+## [11.0.0] (2023-06-16)
+
+### Fixed
+
+- help command works when help option is disabled ([#1864])
+
+### Changed
+
+- leading and trailing spaces are now ignored by the .arguments() method ([#1874])
+- refine "types" exports for ESM to follow TypeScript guidelines ([#1886])
+- *Breaking:* Commander 11 requires Node.js v16 or higher
+
 ## [10.0.1] (2023-04-15)
 
 ### Added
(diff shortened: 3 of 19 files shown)


shadowspawn:
A quick reaction. I may not accept this PR, or at least not to "Warn about obscured help option". The zero-config obscuring was added in #1247 and raised in #645. The ability to explicitly suppress the built-in help option was added later in #1325 and perhaps the obscuring is no longer as appropriate, so I'll think about it a bit.

Separately, storing the help option configuration in an actual Option is something I considered again while working on #1910. I have not had a deep look yet myself. (I resisted attempting it in the same PR.)

aweebit:
> (I resisted attempting it in the same PR.)

I realized only later that I had had to do the change in the base #1929, that is why there is the empty [371751e](https://github.com/tj/commander.js/pull/1931/commits/371751e05a9d6b8fa593cf7223b7d4bc71aa68fd).

shadowspawn:
I think the commit specific to the title for this PR is https://github.com/tj/commander.js/pull/1931/commits/d03566b2eade32479bd41f8927fdd6420f22fa48

The current behaviour evolved out of a sort of auto-configuration for the implicit help option, which slowly became more robust as doing other work. I suspect the behaviour is more appropriate for simpler programs where the author uses `-h` say and we just get out of the way with the default help option.

That leads to an interesting thought that could treat explicit and implicit help options differently. But not solving a problem I am invested in for now, and behaviour is as intended.

(I have lost track of how many of your PRs have the help configuration refactored into a help option. That is something I am interested in as a concept, but not urgently, and may refer back to this in the future.)

PR #1933, opened 2023-08-03
Allow dynamically updating version number and version option flags and description

aweebit:
## Problem

```js
// demo.js

const flags = '-v, --my-version';
const description = 'custom version option description';

var program = new Command();
program
  .version('1.0.0')
  .version('2.0.0', flags, description); // expectation: the version number and option's flags and description are updated
program.outputHelp(); // reality: a new option is added

var program = new Command();
program
  .version('1.0.0')
  .version('2.0.0'); // expectation: the version number is updated
program.outputHelp(); // reality: a new option is added

var program = new Command();
program
  .version('1.0.0')
  .version(program._version, flags); // expectation: the version option's flags are updated
program.outputHelp(); // reality: a new option is added

var program = new Command();
program
  .version('1.0.0')
  .version(program._version, undefined, description); // expectation: the version option's description is updated
program.outputHelp(); // reality: a new option is added

var program = new Command();
const result = program
  .version('1.0.0')
  .version(undefined, flags, description); // expectation: the version option's flags and description are updated
program.outputHelp(); // reality: the flags are not updated
console.log(result); // the version number is returned instead

// A real-world example
var program = new Command();
program.version('1.0.0');
if ('EARLY_ADOPTER' in process.env) {
  program.version('2.0.0'); // expectation: the version number is simply updated
}
program.outputHelp(); // reality: contains version option twice if early adopter
program.parse(['--version'], { from: 'user' }); // output is always '1.0.0'
```

```bash
EARLY_ADOPTER= node demo.js
```

## Solution

Rework `version()` to make all the above examples behave as expected.

Includes various improvements to the code for the version option, including the type fix from #1930. For reasoning behind those improvements, see the commit descriptions.

### Added
- support for dynamically updating version number and version option flags and description via `.version()`

### Fixed
-  (from [#1930](https://github.com/tj/commander.js/pull/1930)) `.version()` optional parameter type

### …solving similar problems
- #1929 and it's extensions (storing the help option instance in `_helpOption` instead of only the individual flags in `_helpShortFlag` and `_helpLongFlag`)

### DIFF
--- a/lib/option.js
+++ b/lib/option.js
@@ -11,21 +11,8 @@ class Option {
    */
 
   constructor(flags, description) {
-    this.flags = flags;
-    this.description = description || '';
-
-    this.required = flags.includes('<'); // A value must be supplied when the option is specified.
-    this.optional = flags.includes('['); // A value is optional when the option is specified.
-    // variadic test ignores <value,...> et al which might be used to describe custom splitting of single argument
-    this.variadic = /\w\.\.\.[>\]]$/.test(flags); // The option can take multiple values.
+    this.setFlagsAndDescription(flags, description);
     this.mandatory = false; // The option must have a value after parsing, which usually means it must be specified on command line.
-    const optionFlags = splitOptionFlags(flags);
-    this.short = optionFlags.shortFlag;
-    this.long = optionFlags.longFlag;
-    this.negate = false;
-    if (this.long) {
-      this.negate = this.long.startsWith('--no-');
-    }
     this.defaultValue = undefined;
     this.defaultValueDescription = undefined;
     this.presetArg = undefined;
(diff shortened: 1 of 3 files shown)


shadowspawn:
A quick reaction. I don't think being able to call `.version()` twice is important. On a related note in terms of early detection by author that it does not work that way, there is the PR open to add an error for overlapping option flags.

aweebit:
> I don't think being able to call `.version()` twice is important.

The last example in the demo code shows it could be important.

Being able to call `.version()` twice with new flags and description is not that important, but added for completeness and to provide meaning to calls that currently don't have any well-defined meaning at all.

> On a related note in terms of early detection by author that it does not work that way, there is the PR open to add an error for overlapping option flags.

If you mean #1923, I updated it a few hours ago so that version option flag conflicts are handled, too. But it solves a different problem.

lib/option.js
```diff
@@ -11,21 +11,8 @@ class Option {
...
 
   constructor(flags, description) {
-    this.flags = flags;
-    this.description = description || '';
-
-    this.required = flags.includes('<'); // A value must be supplied when the option is specified.
-    this.optional = flags.includes('['); // A value is optional when the option is specified.
-    // variadic test ignores <value,...> et al which might be used to describe custom splitting of single argument
-    this.variadic = /\w\.\.\.[>\]]$/.test(flags); // The option can take multiple values.
+    this.setFlagsAndDescription(flags, description);
```
shadowspawn, line 14: I was quite surprised this refactor did not break the `typescript-checkJS` run-script. At one time I am pretty sure we were getting errors for JavaScript properties which were not declared/set in the constructor.

TypeScript may have removed that check as too strict for real-world JavaScript.

https://github.com/microsoft/TypeScript/issues/22896

shadowspawn:
I do not currently want this functionality or refactor of Option, thanks. The `.version()` support is a simple convenience to add a version option with an event handler to display the version.

I see one commit was initialising `this._version` and `this._versionOption` in the constructor. That is consistent with our normal pattern if you want to suggest that separately. 

(`.version()` started out looking like a getter/setter for a "version" property, including returning it in `.opts()`. Now I don't think it should be acting as a getter and I don't want to store `_version` or `_versionOption`, so I should declare that behaviour deprecated!)

shadowspawn:
> I see one commit was initialising this._version and this._versionOption in the constructor. That is consistent with our normal pattern if you want to suggest that separately.

Actually, no. I'll leave it undeclared and deprecate it as moving in the preferred direction.

PR #1937, opened 2023-08-04
Add missing check for broken passThrough

aweebit:
## Problem

```js

try {
  new Command('parent')
    .command('sub')
    .passThroughOptions(); // not allowed without enablePositionalOptions for parent
} catch (err) {
  console.error(err);
}

try {
  const parent = new Command('parent');
    .passThroughOptions();
  parent.addCommand(sub); // allowed without enablePositionalOptions for parent
} catch (err) {
  console.error(err);
}
```

<details>
<summary>A check for this case is considered unnecessary, see <a href="https://github.com/tj/commander.js/pull/1937#discussion_r1284917245">comment</a></summary>

```js
try {
  const parent = new Command('parent')
    .enablePositionalOptions();
  parent.command('sub')
    .passThroughOptions();
  parent.enablePositionalOptions(false); // allowed despite passThroughOptions for sub
} catch (err) {
  console.error(err);
}
```
</details>

## Solution

Throw errors in all demonstrated cases. Additionally make the error message more meaningful by including the command name (similar to #1923 and #1924).

### Changed
- _Breaking:_ throw when a subcommand with  `passThroughOptions` is added to a command without `enablePositionalOptions`

### DIFF
--- a/lib/command.js
+++ b/lib/command.js
@@ -272,6 +272,8 @@ class Command extends EventEmitter {
 
     this.commands.push(cmd);
     cmd.parent = this;
+    this._checkForIllegalPassThroughOptions(cmd, this);
+
     return this;
   }
 
@@ -721,6 +723,9 @@ Expecting one of '${allowedValues.join("', '")}'`);
    */
   enablePositionalOptions(positional = true) {
     this._enablePositionalOptions = !!positional;
+    this.commands.forEach((command) => {
+      this._checkForIllegalPassThroughOptions(command, this);
+    });
     return this;
   }
 
@@ -735,10 +740,20 @@ Expecting one of '${allowedValues.join("', '")}'`);
    */
   passThroughOptions(passThrough = true) {
     this._passThroughOptions = !!passThrough;
-    if (!!this.parent && passThrough && !this.parent._enablePositionalOptions) {
-      throw new Error('passThroughOptions can not be used without turning on enablePositionalOptions for parent command(s)');
-    }
+    this._checkForIllegalPassThroughOptions(this, this.parent);
     return this;
+  }
+
+  /**
+   * @param {Command} command
+   * @param {Command | null} parent
+   * @api private
+   */
+
+  _checkForIllegalPassThroughOptions(command, parent) {
+    if (parent && command._passThroughOptions && !parent._enablePositionalOptions) {
+      throw new Error(`passThroughOptions cannot be used for '${command._name}' without turning on enablePositionalOptions for parent command${parent._name ? ` '${parent._name}'` : ''}`);
+    }
   }
 
   /**
--- a/lib/command.js
+++ b/lib/command.js
@@ -272,7 +272,7 @@ class Command extends EventEmitter {
 
     this.commands.push(cmd);
     cmd.parent = this;
-    this._checkForBrokenPassThrough(cmd, this);
+    cmd._checkForBrokenPassThrough();
 
     return this;
   }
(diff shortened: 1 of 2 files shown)


tests/command.positionalOptions.test.js
```diff
@@ -362,13 +362,36 @@ describe('program with action handler and positionalOptions and subcommand', ()
...
 
-  expect(() => {
-    sub.passThroughOptions();
-  }).toThrow();
+    expect(() => {
+      program.addCommand(sub);
+    }).toThrow();
+  });
+
+  test('when program has subcommand with passThroughOptions and reset to non-positional then error', () => {
```
shadowspawn, line 385: I do not want to bother testing for this. Resetting `enablePositionalOptions` is a very unlikely use case.

aweebit: Hmm, okay, so should I even remove the check from the code?

lib/command.js
```diff
@@ -735,12 +740,22 @@ Expecting one of '${allowedValues.join("', '")}'`);
...
     return this;
   }
 
+  /**
+   * @param {Command} command
+   * @param {Command | null} parent
+   * @api private
+   */
+
+  _checkForIllegalPassThroughOptions(command, parent) {
```
shadowspawn, line 753: The high level check is that the settings are likely to work together. `Illegal` is too strong for my liking, and it is the combination rather than just `PassThroughOptions`. Suggest: `_checkForBrokenPassThrough`.

tests/command.positionalOptions.test.js
aweebit: Would need to update the ChangeLog, the change would need to be more specific. It is a bit strange to exclude one particular case in which the inconsistent state is reached…

shadowspawn: Yes, thanks.

shadowspawn:
I like catching the misconfiguration when using `addCommand()`.

lib/command.js
```diff
@@ -735,12 +740,22 @@ Expecting one of '${allowedValues.join("', '")}'`);
...
+    if (parent && command._passThroughOptions && !parent._enablePositionalOptions) {
+      throw new Error(`passThroughOptions cannot be used for '${command._name}' without turning on enablePositionalOptions for parent command${parent._name ? ` '${parent._name}'` : ''}`);
```
shadowspawn, line 755: I was assuming that the throw would include the line number so easy enough to track down, but I do like adding the subcommand name. Easier to jump into fixing.

Adding the parent command name is lower value and makes the code harder to read.
```suggestion
      throw new Error(`passThroughOptions cannot be used for '${command._name}' without turning on enablePositionalOptions for parent command(s)`);
```

shadowspawn:
This is the example case that convinced me this is worth doing, bypassed the check:
```js
const parent = new Command('parent');
    .passThroughOptions();
parent.addCommand(sub); // allowed without enablePositionalOptions for parent
```

lib/command.js
aweebit: I agree the parent command name is not important here.

I have actually been working on a proof of concept PR introducing support for multiple parent commands. But before it is added, I wonder what exactly is the purpose of the `(s)` part at the end of the error message?

shadowspawn: For safety, positional options should be turned on all the way up the tree. In general I expect it will be turned on at the root, and that intermediate levels won't have options, so unlikely to have a practical effect, and not attempting to robustly enforce. But I thought it was worth adding the `(s)` to hint at the multiple levels.

shadowspawn:
(Introduces a `throw` so better go in a major version.)

shadowspawn (changes requested):
I am adding this review as a confirmation that requested changes will make it likely that merge accepted.

tests/command.positionalOptions.test.js
aweebit: Resolved in [9d4c96a](https://github.com/tj/commander.js/pull/1937/commits/9d4c96a4fc40590dced784d24f26661d183fb512).

lib/command.js
aweebit: Resolved in [43d9faa](https://github.com/tj/commander.js/pull/1937/commits/43d9faa402345f0424ee642203535b2c1aa25bb2).

aweebit:
> I am adding this review as a confirmation that requested changes will make it likely that merge accepted.

All the requested changes have been made now.

lib/command.js
```diff
@@ -735,12 +737,22 @@ Expecting one of '${allowedValues.join("', '")}'`);
...
     return this;
   }
 
+  /**
+   * @param {Command} command
+   * @param {Command | null} parent
+   * @api private
+   */
+
+  _checkForBrokenPassThrough(command, parent) {
```
shadowspawn, line 750: I think remove the parameters and call the check on the subcommand?

aweebit: Sure. Resolved in [7689506](https://github.com/tj/commander.js/pull/1937/commits/7689506699b9db61a121baea40ba455e8d89c78d).

shadowspawn (approved):

shadowspawn:
(Thanks for all the tweaks.)

abetomo (approved):
+1

aweebit:
Thanks for the merge!

> (Introduces a `throw` so better go in a major version.)

Should I set base to the next release branch when there is a breaking change? Or always to `develop`?

shadowspawn:
Using next release branch for base is appropriate when know PR contains breaking changes.

PR #1938, opened 2023-08-05
Add support for sharing and stand-alone parsing of subcommands. Warn about parse calls on commands added with `.command()`

### LINKED ISSUE #1916: On command and subcommand reusability and setting inheritance

aweebit:
An extension of #1937 showcasing a better version of the changes I suggested in #1917 ([comment](https://github.com/tj/commander.js/pull/1917#discussion_r1279217920)).

This was originally a proof of concept PR that expected to give new life to #1917 and #1924.

Now, I would like this PR and #1940 to supersede #1917. See [#1917 (comment)](https://github.com/tj/commander.js/pull/1917#issuecomment-1666467716) for details.

Partially fixes #1916.

### Added
- support for sharing and stand-alone parsing of subcommands
- warnings about parse calls on commands added with `.command()`

## Motivation
- gives meaning to currently meaningless parse calls on subcommands
- the meaning given is what most users probably expect
-  makes the error introduced in [#1917](https://github.com/tj/commander.js/pull/1917) unnecessary, and the warning introduced there not absolutely necessary but just useful for chains with `.command()`

## Implementation details
- backwards compatibility is preserved: the behavior has not changed in code using the current and previous versions of the library as intended (only adding commands to one parent and parsing from the parent)
- the implementation is quite low-effort
  - very few code changes
  - unlikely something could go wrong
  - here is a summary:
    - additionally set `parent` to the invoking parent command at the beginning of `_parseCommand()`
    - add `parents` to store all parent commands so that no parent checks have to be moved to parse calls, which I wrongly assumed would be necessary in my [comment](https://github.com/tj/commander.js/pull/1924#issuecomment-1660039439) to [#1924](https://github.com/tj/commander.js/pull/1924) (see how `_checkForBrokenPassThrough()` is implemented in this PR, for example)
    - add `_implicitlyCreated` indicating whether the command was created with `.command()`
    - check its value to warn the developer about parse calls at the end of a `.command()` chain

- #1915
- #1917
- #1931
- #1940

- #1955

### DIFF
--- a/examples/help-subcommands-usage.js
+++ b/examples/help-subcommands-usage.js
@@ -1,0 +1,29 @@
+// const commander = require('commander'); // (normal include)
+const commander = require('../'); // include commander in git clone of commander repo
+
+// By default the subcommand list includes a fairly simple usage. If you have added a custom usage
+// to the subcommands you may wish to configure the help to show these instead.
+//
+// See also ./configure-help.js which shows configuring the subcommand list to have no usage at all
+// and just the subcommand name.
+
+const program = new commander.Command()
+  .configureHelp({ subcommandTerm: (cmd) => cmd.name() + ' ' + cmd.usage() });
+
+program.command('make <FILENAME>')
+  .usage('-root ROOTDIR [-abc] <FILENAME>')
+  .requiredOption('--root <ROOTDIR>')
+  .option('-a')
+  .option('-b')
+  .option('-c')
+  .summary('example command with custom usage')
+  .description('this full description is  displayed in the full help and not in the list of subcommands');
+
+program.command('serve <SERVICE>')
+  .option('--port <PORTNUMBER>')
+  .description('example command with default simple usage');
+
+program.parse();
+
+// Try the following:
+//    node help-subcommands-usage help
--- a/lib/command.js
+++ b/lib/command.js
@@ -939,8 +939,8 @@ Expecting one of '${allowedValues.join("', '")}'`);
   _parseSubroutine(async, userArgsCallback, argv, parseOptions) {
     const methodName = async ? 'parseAsync' : 'parse';
     if (this._implicitlyCreated) {
-      console.warn(`Called .${methodName}() on subcommand '${this._name}' added with .command().
-Call on top-level command instead`);
+      console.warn(`Called .${methodName}() on subcommand '${this._name}' added with .command()
+- meant to call on the top-level command?`);
     }
     const userArgs = this._prepareUserArgs(argv, parseOptions);
     return userArgsCallback(userArgs);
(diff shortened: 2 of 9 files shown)


shadowspawn:
Nice clear labelling as "Proof of concept" and "Draft", and good description to explain. 

The approach of having a stable forward/subcommand link while wiring up the reverse/parent links while parsing is an interesting idea. The wiring itself is indeed "low effort".

The downside is a more complex model for maintainers and to a lesser extent authors, to support some exotic reuse configurations.

I am comfortable with the current simple model of one stable tree with matching forwards and backwards links. This POC is not something I am interested in proceeding with at this time.

aweebit:
Hmm, expected the already merged commits to not be shown anymore after changing base and merging with the new one. Well, at least the file changes are not cluttered anymore.

aweebit:
> The downside is a more complex model for maintainers and to a lesser extent authors, to support some exotic reuse configurations.

What kind of authors would the more complex model affect? Authors of libraries extending the functionality of Commander somehow? I don't think any maintenance complications are implied for authors of CLI applications using Commander as currently intended (only adding commands to one parent and parsing from that parent) since the change is fully backwards compatible and invisible in such cases (I have even got rid of `currentParent` in the most recent commit because it was redundant). And if they do parse subcommands from different parents or as stand-alones, then they know what they are doing.

> I am comfortable with the current simple model of one stable tree with matching forwards and backwards links. This POC is not something I am interested in proceeding with at this time.

One reason I want to support the exotic reuse configurations is because right now, parsing subcommands as stand-alones is allowed despite being a meaningless operation (since the `.parent` is not reset to `null`), so it would be reasonable to forbid it, but there is no good way to do it because parse calls are only supposed to throw errors originating in user code (argParsers, hooks and actions), so warnings have to be used instead, and I would like to avoid that as much as possible.

aweebit:
> The downside is a more complex model for maintainers

I can agree some overhead is introduced for contributors, but really insignificant. Would just have to think whether iterating over `parents` is more appropriate when using `parent` in code, that's it.

aweebit:
I would now like this PR and #1940 to supersede #1917. See [#1917 (comment)](https://github.com/tj/commander.js/pull/1917#issuecomment-1666467716) for details.

aweebit:
Just a quick note. I keep finding myself putting `.parse()` at the end of chains that have `.command()` in them by mistake when writing tests and demos for my pull requests. Introducing a warning for such parse calls would indeed be very, very useful.

shadowspawn:
This is a useful note, thanks: https://github.com/tj/commander.js/pull/1938#issuecomment-1667037227

I added one previous warning because of a mistake I kept making: #1655

shadowspawn:
These changes are not something I am interested in proceeding with.

PR #1939, opened 2023-08-05
Refactor `getCommandAndParents()` into `Command._getCommandAndAncestors()` and use consistently

aweebit:
Cherry-picked from #1938.

A purely cosmetic PR Introducing no functional changes or changes visible to the user.

Replaces `getCommandAndParents()` with `Command._getCommandAndAncestors()`
- …because an instance method is where the function logically belongs.
- Renamed because "ancestors" is more precise than "parents".

Also changes the code to make use of the function whenever appropriate. (I thought it was a good idea after introducing `currentParent` replacing `parent` in #1938. I ended up getting rid of it and simply upgrading the `parent` semantics, but who knows what changes to the parent resolution could be made in the future. It is better to keep the code for getting the ancestors in one place.)

### DIFF
--- a/lib/command.js
+++ b/lib/command.js
@@ -1756,13 +1756,14 @@ Expecting one of '${allowedValues.join("', '")}'`);
     if (flag.startsWith('--') && this._showSuggestionAfterError) {
       // Looping to pick up the global options too
       let candidateFlags = [];
-      for (const command of this._getCommandAndAncestors()) {
+      let command = this;
+      do {
         const moreFlags = command.createHelp().visibleOptions(command)
           .filter(option => option.long)
           .map(option => option.long);
         candidateFlags = candidateFlags.concat(moreFlags);
-        if (command._enablePositionalOptions) break;
-      }
+        command = command.parent;
+      } while (command && !command._enablePositionalOptions);
       suggestion = suggestSimilar(flag, candidateFlags);
     }
 
--- a/lib/command.js
+++ b/lib/command.js
@@ -1378,13 +1378,13 @@ Expecting one of '${allowedValues.join("', '")}'`);
 
   _checkForMissingMandatoryOptions() {
     // Walk up hierarchy so can call in subcommand after checking for displaying help.
-    for (let cmd = this; cmd; cmd = cmd.parent) {
+    this._getCommandAndAncestors().forEach((cmd) => {
       cmd.options.forEach((anOption) => {
         if (anOption.mandatory && (cmd.getOptionValue(anOption.attributeName()) === undefined)) {
           cmd.missingMandatoryOptionValue(anOption);
         }
       });
-    }
+    });
   }
 
   /**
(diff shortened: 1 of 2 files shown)


aweebit:
(Force push is due to a reset in order to change the commit message borrowed from [aa280af](https://github.com/tj/commander.js/pull/1938/commits/aa280afbf8b0fde08566ebd9c20cc85df2a62c6a) so that it does not say anything about the `parents` array only introduced in #1938.)

shadowspawn:
This comment is just musing. No action items.

I had considered these when I wrote the code, but was 50/50 on a couple, and willing to change.

> …because an instance method is where the function logically belongs.

Ok.

> Renamed because "ancestors" is more precise than "parents".

My concern was that ancestors might be more correct but less understandable to some readers. However, ancestor is certainly tree terminology. Ok.

https://en.wikipedia.org/wiki/Tree_(data_structure)
**Ancestor**
A node reachable by repeated proceeding from child to parent.

lib/command.js
```diff
@@ -1743,14 +1756,13 @@ Expecting one of '${allowedValues.join("', '")}'`);
     if (flag.startsWith('--') && this._showSuggestionAfterError) {
       // Looping to pick up the global options too
       let candidateFlags = [];
-      let command = this;
-      do {
+      for (const command of this._getCommandAndAncestors()) {
```
shadowspawn, line 1759: Can this be a `forEach`? (I don't like the `for ( of )`.

lib/help.js
```diff
@@ -101,10 +101,10 @@ class Help {
     if (!this.showGlobalOptions) return [];
 
     const globalOptions = [];
-    for (let parentCmd = cmd.parent; parentCmd; parentCmd = parentCmd.parent) {
-      const visibleOptions = parentCmd.options.filter((option) => !option.hidden);
+    cmd._getCommandAndAncestors().slice(1).forEach((ancestorCmd) => {
```
shadowspawn, line 104: No. Having to use the slice makes this too subtle for my liking.

shadowspawn: (Ok with the active variable changing to ancestor though.)

lib/help.js
```diff
@@ -240,11 +240,11 @@ class Help {
     if (cmd._aliases[0]) {
       cmdName = cmdName + '|' + cmd._aliases[0];
     }
-    let parentCmdNames = '';
-    for (let parentCmd = cmd.parent; parentCmd; parentCmd = parentCmd.parent) {
-      parentCmdNames = parentCmd.name() + ' ' + parentCmdNames;
-    }
-    return parentCmdNames + cmdName + ' ' + cmd.usage();
+    let ancestorCmdNames = '';
+    cmd._getCommandAndAncestors().slice(1).forEach((ancestorCmd) => {
```
shadowspawn, line 244: No. Having to use the slice makes this too subtle for my liking.
Ok with the active variable changing to ancestor though.

shadowspawn:
> So using the helper routine and array methods to replace manual loops is ok when it is an improvement or at least directly equivalent. But it is not appropriate when it requires extra steps or is awkward.

This only ended up affecting a couple of the changes! I had seen them on first scan and feared there might be more, so wanted to explain my reasoning. Added comments on the lines of concern.

shadowspawn (changes requested):
Was dubious about this but after working through my concerns, yup, ok!

Changes requested on three of the lines.

lib/command.js
aweebit: Could be if not for the early exit that is impossible with `forEach`. I guess this is one of the cases where iterating manually via `.parent` is more appropriate.

shadowspawn: 👍

shadowspawn (approved):
LGTM

abetomo (approved):
+1

PR #1941, opened 2023-08-05
Add missing support for command-arguments to commands added with `.addCommand()`

aweebit:
## Problem

```js

{
  program.command('sub <arg>').action(arg => console.log(arg));
  program.parse(['sub', 'arg'], { from: 'user' }); // arg
}

{
  const sub = new Command('sub <arg>').action(arg => console.log(arg));
  program.addCommand(sub);
  program.parse(['sub', 'arg'], { from: 'user' }); // error: unknown command 'sub'
}
```

## Solution

Move declared command-argument parsing from `.command()` to constructor.

### Fixed
- added missing support for command-arguments to commands added with `.addCommand()`

### DIFF
--- a/lib/command.js
+++ b/lib/command.js
@@ -21,6 +21,15 @@ class Command extends EventEmitter {
 
   constructor(name) {
     super();
+    const nameMatchResult = name?.match(/([^ ]+) *(.*)/);
+    /** @type {string} */
+    this._name = nameMatchResult?.[1] ?? '';
+    /** @type {Argument[]} */
+    this._args = [];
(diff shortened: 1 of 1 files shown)


shadowspawn:
I was considering this PR and the related changes, and realised I do not want to change the parameter to `createCommand()` to include args. Working backwards, I do not want to change the Command constructor.

This is a long-standing behaviour, and is now somewhat less of an issue. The more recent and now somewhat preferred way of adding arguments is using `.argument()`, which is also the way to add an argument description if desired.

Issue #1936, opened 2023-08-04
`passThroughOptions` and unknown option handling


aweebit:
Is there a particular reason why `.parseOptions()` stops processing args when an unknown option is encountered in the `passThroughOptions` mode? That complicates the help option refactor I am currently working on in #1934 that aims to achieve maximum consistency between how the help option and all other options are handled, because help flags end up ignored if they come after an unknown option and there is no subcommand.

Checkout [02fde0e](https://github.com/tj/commander.js/pull/1934/commits/02fde0e5d5824dd21546358d313f231b2861f7be) and run the following code for a demo:
```js
const program = require('commander');
program
  .passThroughOptions()
  .allowUnknownOption();
program.parse(['--unknown', '--help'], { from: 'user' }); // no help displayed
console.log(program.args); // [ '--unknown', '--help' ]
program
  .allowUnknownOption(false);
program.parse(['--unknown', '--help'], { from: 'user' }); // expected help, got error: unknown option '--unknown'
```

I tried to change the behavior to the one implied by the docs and comments in code so that the processing is only stopped when a command-argument is encountered (see [fcfbec1](https://github.com/tj/commander.js/pull/1934/commits/fcfbec1ff6bfbc10d5e6cc908ab5e8d3335f872e)), but that resulted in a test failing.

#### Relevant test from `tests/command.positionalOptions.test.js` (line 423)
```js
  test('when passThroughOptions and unknown option then arguments from unknown passed through', () => {
    const program = new commander.Command();
    program
      .passThroughOptions()
      .allowUnknownOption()
      .option('--debug');

    program.parse(['--unknown', '--debug'], { from: 'user' });
    expect(program.args).toEqual(['--unknown', '--debug']);
  });
```

The behavior I would expect: `--debug` is consumed and so `program.opts()` equals to `{ debug: true }` while `program.args` equals to `['--unknown']`.

#### Relevant lines in `.parseOptions()` code (on current `develop` branch)
```js
      // If using passThroughOptions, stop processing options at first command-argument.
      if (this._passThroughOptions) {
        dest.push(arg);
        if (args.length > 0) dest.push(...args);
        break;
      }
```

The condition on line 2 should be `this._passThroughOptions && !maybeOption(arg)` so that the comment on line 1 describes what happens well and does not miss the fact the processing is also stopped at unknown options (currently missed in all explanations of `passThroughOptions`).

shadowspawn:
[...]

PR #1942, opened 2023-08-05
Add clarifications about `passThroughOptions` handling of unknown options to docs and comments

aweebit:
## Problem
#1936

## Solution
What caused my confusion turned out to be desired behavior.

This PR includes clarifications that could help avoid confusion in the future.

### Fixed
- extended documentation about `.passThroughOptions()` by a missing mention of unknown option handling

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -981,8 +981,8 @@ program -b subcommand
 program subcommand -b
 ```
 
-By default options are recognised before and after command-arguments. To only process options that come
-before the command-arguments, use `.passThroughOptions()`. This lets you pass the  arguments and following options through to another program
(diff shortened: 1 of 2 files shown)


shadowspawn:
I suspect this was driven in part by confusion with interaction with `.allowUnknownOption()` which is a second-class setting.  

On balance, I don't think the comment in the README is helpful for the majority of readers, and the comment in the code was unclear to me and I know what the code does. Given how many other issues and PRs you have open at the moment, I am just closing this one.

Related comments about `.allowUnknownOption()` :
- https://github.com/tj/commander.js/issues/1945#issuecomment-1667103439
- https://github.com/tj/commander.js/issues/1936#issuecomment-1666694788

Issue #1945, opened 2023-08-07
`allowUnknownOption` handling in `.parseOptions()`


aweebit:
The docs says:

> By default the option processing shows an error for an unknown option. To have an unknown option treated as an ordinary command-argument and continue looking for options, use `.allowUnknownOption()`.

However, unknown options are not treated as command-arguments by `.parseOptions()` when `allowUnknownOption` is used. That can be shown by a simple example:

```js
const program = new Command().allowUnknownOption();
console.log(program.parseOptions(['arg'])); // { operands: [ 'arg' ], unknown: [] }
console.log(program.parseOptions(['--option'])); // { operands: [], unknown: [ '--option' ] }
```

This might seem harmless, but actually has consequences. The following test from [`tests/command.helpCommand.test.js`](https://github.com/tj/commander.js/blob/master/tests/command.helpCommand.test.js), [line 24](https://github.com/tj/commander.js/blob/4ef19faac1564743d8c7e3ce89ef8d190e1551b4/tests/command.helpCommand.test.js#L24) would have to fail if unknown options were treated as described in the docs since `caughtErr.code` would equal to `'commander.unknownCommand'` instead of `'commander.help'`.

```js
test('when program has subcommands and specify only unknown option then display help', () => {
  const program = new commander.Command();
  program
    .configureHelp({ formatHelp: () => '' })
    .exitOverride()
    .allowUnknownOption()
    .command('foo');
  let caughtErr;
  try {
    program.parse(['--unknown'], { from: 'user' });
  } catch (err) {
    caughtErr = err;
  }
  expect(caughtErr.code).toEqual('commander.help');
});
```

Some possible solutions I see:
- treat unknown options as command-arguments in `.parseOptions()` as suggested by the docs and
  - either remove / change the test accordingly
  - or make `.unknownCommand()` treat options specially
- remove the part about treatment as a command-argument from the docs so that it is not misleading anymore

aweebit:
> treat unknown options as command-arguments in `.parseOptions()` as suggested by the docs and
> * either remove / change the test accordingly

I opened #1946 with an implementation of this suggestion.

shadowspawn:
`.parseOptions()` is a low level routine that only processes the arguments using the context of the current command.

`.allowUnknownOption` is not relevant to `.parseOptions()`. The comment quoted from the README is about the `.parse()` outcomes. 

In `.parseOptions()` the input arguments are either:
- a recognised option and consumed, including option-argument if appropriate
[...]

PR #1946, opened 2023-08-07
Correctly handle `allowUnknownOption` in `.parseOptions()`

aweebit:
## Problem

#1945

## Solution

Implement the first solution suggested in the issue.

Unknown options are now treated as command-arguments in `.parseOptions()` as suggested by the docs.

The incorrect test has been removed.

### Fixed
- _Breaking:_ treat unknown options as command-arguments in `.parseOptions()` as suggested by the docs

### DIFF
--- a/lib/command.js
+++ b/lib/command.js
@@ -1445,6 +1445,8 @@ Expecting one of '${allowedValues.join("', '")}'`);
 
     // parse options
     let activeVariadicOption = null;
+    let onlyKnownOptionsSoFar = true;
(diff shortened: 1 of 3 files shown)


shadowspawn:
The current behaviour is as intended. `.parseOptions()` is an internal routine not affected by `.allowUnknownOption()`.

See longer explanation of `.parseOptions()` in issue: https://github.com/tj/commander.js/issues/1945#issuecomment-1667103439

PR #1948, opened 2023-08-07
Watch for npm package updates with dependabot

abetomo:
# Pull Request

## Solution

How about checking npm package update in dependabot?

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,7 +1,7 @@
 version: 2
 updates:
-- package-ecosystem: "github-actions"
-  target-branch: "develop"
-  directory: "/"
-  schedule:
+  - package-ecosystem: "github-actions"
+    target-branch: "develop"
+    directory: "/"
+    schedule:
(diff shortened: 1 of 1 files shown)


shadowspawn:
I find it noisy in repos that aren't actively maintained, but have never tried it. Happy to give it a go. Dependency maintenance is considered a best practice these days.

shadowspawn (approved):

abetomo:
Thanks for the review!

shadowspawn:
I think the recent merges have included all the PR commits. I suggest doing a "Squash and merge" normally. I think the single commit for the PR is more useful when researching code history than hitting wee commits from during the PR development. I don't feel too strongly about it and if there is a reason to normally pick a different merge type then let me know!

I think Github just leaves the last selected merge type as the default for the next one, so easy to miss.

(I just picked one of the recent PRs to make a comment on. This PR only had two commits anyway.)

abetomo:
I got it.
I'll do that next time.

PR #1950, opened 2023-08-08
Fix tests (Node.js v18.17.0)

aweebit:
Tests of `node` child processes stopped working on Windows after upgrading to Node.js v18.17.0. The reason turned out to be the fact Jest allows an internally set value of the [`FORCE_COLOR`](https://nodejs.org/api/cli.html#force_color1-2-3) environment variable to leak through to tests, see jestjs/jest#14391. This results in colored output of the child processes, and so comparisons of the output with expected values fail.

 This PR is a quick fix unsetting the variable for the child processes.

### DIFF
--- a/tests/command.default.test.js
+++ b/tests/command.default.test.js
@@ -3,7 +3,12 @@ const commander = require('../');
 const path = require('path');
 const util = require('util');
 
-const execFileAsync = util.promisify(childProcess.execFile);
+const execFileAsync = function(...args) {
+  if (args.length < 3) args.push({});
+  args[2].env ??= { ...process.env };
+  delete args[2].env.FORCE_COLOR;
+  return util.promisify(childProcess.execFile).apply(this, args);
+};
 
 describe('default executable command', () => {
   // Calling node explicitly so pm works without file suffix cross-platform.
--- a/tests/command.executableSubcommand.inspect.test.js
+++ b/tests/command.executableSubcommand.inspect.test.js
@@ -2,7 +2,12 @@ const childProcess = require('child_process');
 const path = require('path');
 const util = require('util');
 
-const execFileAsync = util.promisify(childProcess.execFile);
+const execFileAsync = function(...args) {
+  if (args.length < 3) args.push({});
+  args[2].env ??= { ...process.env };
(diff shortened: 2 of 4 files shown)


shadowspawn:
I do not see `FORCE_COLOR` being used in the Commander code base.

I do see some odd test failures using Node v18.17.0 on Mac, and in the Jest tests being run automatically in the VSCode editor. I'll add a comment in the linked issue.

I don't think is a problem that needs to be solved in Commander though.

shadowspawn:
> I don't think is a problem that needs to be solved in Commander though.

However, I am glad to know about the issue before stumbling across it myself! 😅

aweebit:
Force pushed to make merges less cluttered in case the PR does end up considered useful.

shadowspawn:
If the PR GitHub Checks start failing or other developers trip over this then will do something, but currently passing and hopefully gets fixed in Jest.

For readers, temporary developer work-around to run tests is use `npx jest --worker-threads` or `npx jest --runInBand`
- https://github.com/tj/commander.js/pull/1950#issuecomment-1669523410
- https://github.com/jestjs/jest/issues/14391#issuecomment-1668840119

shadowspawn:
I don't like patching the code with detailed work-around for `FORCE_COLOR`.

But perhaps modifying the Jest call in the run-script to use `--worker-threads` for now? Has it been well-behaved for you? 

(I am currently running at the previous node to avoid the problem!)

aweebit:
> But perhaps modifying the Jest call in the run-script to use `--worker-threads` for now? Has it been well-behaved for you?

For me, it has, but the feature is still experimental, so still not sure it should be relied on in general.

shadowspawn:
Would be temporary, but good point, no need to use experimental. Tests with `--runInBand` take 10 s on my computer, which is fine.

aweebit:
> Would be temporary, but good point, no need to use experimental. Tests with `--runInBand` take 10 s on my computer, which is fine.

It's 50 vs 20 seconds on mine 🤨

shadowspawn:
Closing this in favour of #2011 which has a simpler work-around (with @aweebit added as co-author)

PR #1951, opened 2023-08-08
Add `.strictOptionalOptionArguments()` for POSIX-style handling of options with optional option-arguments

### LINKED ISSUE #1901: Add mode for optional option-argument from embedded value only (POSIX style)
Short version: support POSIX style parsing mode which only reads optional value with embedded forms `--optional=VALUE` and `-oVALUE`.

Say: `program.optionOptionalValuesEmbedded()`

(Shorter meaningful name suggestions welcome!)

----- 
Long version

Commander supports an option which may be used as a boolean option but may optionally take an option-argument (declared with square brackets like `-o, --opt [value]`). This can lead to [parsing ambiguities](https://github.com/tj/commander.js/blob/release/10.x/docs/options-taking-varying-arguments.md) due to the space-separated `VALUE` being confused with non-option arguments.

These are all the optional variations:

```
foo --optional
foo -o
foo --optional VALUE
foo -o VALUE
foo --optional=VALUE
foo -oVALUE
```

POSIX takes a stricter approach, which avoids the parsing ambiguity. The option-argument has to be embedded with the option. 

> If the SYNOPSIS shows an optional option-argument (as with `[ -f[option_argument]]` in the example), a conforming application shall place any option-argument for that option directly adjacent to the option in the same argument string, without intervening <blank> characters. If the utility receives an argument containing only the option, it shall behave as specified in its description for an omitted option-argument; it shall not treat the next argument (if any) as the option-argument for that option.

So the POSIX approach only allows:
```
foo --optional
foo -o
foo --optional=VALUE
foo -oVALUE
```

Reference:
  - https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap12.html#tag_12_01
  - https://pubs.opengroup.org/onlinepubs/9699919799/xrat/V4_xbd_chap12.html

aweebit:
Fixes #1901.

### Added
- `.strictOptionalOptionArguments()` for POSIX-style handling of options with optional option-arguments

### DIFF
--- a/lib/command.js
+++ b/lib/command.js
@@ -1575,6 +1575,7 @@ Expecting one of '${allowedValues.join("', '")}'`);
 
       // Not a recognised option by this command. Might be
       // - a subcommand,
+      // - the help option (currently not handled by this routine),
       // - an option unknown to this command,
       // - or a command-argument.
 
--- a/lib/command.js
+++ b/lib/command.js
@@ -46,6 +46,7 @@ class Command extends EventEmitter {
     this._defaultCommandName = null;
     this._exitCallback = null;
     this._aliases = [];
+    this._strictOptionalOptionArguments = false;
     this._combineFlagAndOptionalValue = true;
     this._description = '';
     this._summary = '';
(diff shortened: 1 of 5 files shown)


shadowspawn:
This is something I have been thinking about for years. I will be interested to see how you have approached it, but may choose to do it myself.

aweebit:
(Force pushed to undo the merge with #1950.)

shadowspawn:
Leaving  #1901 open to see if there is any interest or compelling use cases. 

There are some extra commits in this PR, the relevant commits to the strict optionals are:
- https://github.com/tj/commander.js/pull/1951/commits/1517423ca372a1b1cf2a7a713f2f8da5a049b496
- https://github.com/tj/commander.js/pull/1951/commits/50d88dece3216c3039908e5962af9cedfec72183

See also a different approach in #1984 where the help shows the flags as supplied by author and up to author to format appropriately.

PR #1955, opened 2023-08-11
Add `.suppressWarnings()` for warnings in #1915 #1931 #1938 #1940

aweebit:
- #1915
- #1931
- #1938
- #1940

### Incompatible with…
- #1917: parse methods shall never be called on subcommands, so if warnings about such calls are added, the JSDoc and the ChangeLog entry for `.suppressWarnings()` introduced here become incorrect. The fact there is currently no well-defined meaning for such calls is one of the reasons why [#1938](https://github.com/tj/commander.js/pull/1938) and [#1940](https://github.com/tj/commander.js/pull/1940) should be preferred over [#1917](https://github.com/tj/commander.js/pull/1917)!

### Proposed mistake handling model

Based on https://github.com/tj/commander.js/pull/1940#issuecomment-1675075727. Could be added to the docs.

* suspicious library usage patterns (although not always entirely wrong): assume author mistake and show a warning unless `suppressWarnings` is on – a setting added specifically so that warnings can be hidden from end users
* end user mistakes, including arguments deemed invalid by argParsers: display an error message and exit the program (by default)
* other errors originating in callbacks (argParsers, hooks, actions): `throw` the error for the author to handle – should not be seen by end user

#### Added
- `.suppressWarnings()` to suppress warnings about presumably wrong library usage

### Why `.suppressWarning()` instead of an environment variable such as `NODE_ENV` (as originally proposed)?

Environment variables are used for stuff the end user should be able to control, but in our case, the tool author is the one to decide whether the warnings are to be shown.

See also: https://github.com/tj/commander.js/pull/1940#issuecomment-1675099093.

### DIFF
--- a/lib/command.js
+++ b/lib/command.js
@@ -57,6 +57,7 @@ class Command extends EventEmitter {
     this._showHelpAfterError = false;
     this._showSuggestionAfterError = true;
 
+    this._suppressWarnings = false;
     // see .configureOutput() for docs
     this._outputConfiguration = {
       writeOut: (str) => process.stdout.write(str),
@@ -88,6 +89,7 @@ class Command extends EventEmitter {
    * @return {Command} `this` command for chaining
    */
   copyInheritedSettings(sourceCommand) {
+    this._suppressWarnings = sourceCommand._suppressWarnings;
     this._outputConfiguration = sourceCommand._outputConfiguration;
     this._hasHelpOption = sourceCommand._hasHelpOption;
     this._helpFlags = sourceCommand._helpFlags;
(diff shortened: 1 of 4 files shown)


shadowspawn:
Lots of mind-dump comments.

Warnings are something I think about in general, and applicable to multiple of your suggested PRs. Not ready for a big control switch yet.

A somewhat related concept is "convention over configuration": https://en.wikipedia.org/wiki/Convention_over_configuration

I think suppressing the warnings would need a narrower more explicit focus than just "warnings".

In general I prefer messages to be for high-value issues that don't need to be made conditional.

On a vaguely related note, I have thought about added in combined make-old-programs-work configuration, but got talked out of it the one time I wrote code. And I think it doesn't work as a concept in the longterm. Instead, I have been explicit about documenting deprecated patterns, but not removing them until there is a gain for the client pain.

There is potentially a tension between the mistakes someone makes when very first attempting to use the library, versus the things authors try adding isolated custom behaviour or making major additions.

PR #1957, opened 2023-08-13
Add support for executable subcommands with help flags differing from those of the parent command

aweebit:
**Disclaimer:** The change introduced here is of low importance.

This PR builds upon the already approved #1930.

Diff only for the change introduced by this PR:
https://github.com/aweebit/commander.js/compare/feature/fixes...feature/help-command-executable-help-flags

## Problem

Currently, help commands don't have support for executable subcommands whose help flags differ from those of the parent command.

Such support could be useful especially when third-party executables are used.

## Solution

Get help flags to pass to an executable subcommand from the subcommand instance rather than the parent instance in `._dispatchHelpCommand()`.

Make it possible to set custom flags for such a subcommand instance by including `helpFlags` in the options passed to `.command()`.

```js
// program.js

program.command('executable', 'description', {
  helpFlags: '-c, --custom-help'
});
program.parse();
```

```js
// program-executable.js

const program = new Command()
  .helpOption('-c, --custom-help');
program.parse();
```

```bash
node program help executable
```

### Added
- support for executable subcommands with help flags differing from those of the parent command (must be specified in the `helpFlags` option when calling `.command()`)

### DIFF
--- a/lib/command.js
+++ b/lib/command.js
@@ -1809,7 +1809,7 @@ Expecting one of '${allowedValues.join("', '")}'`);
    *
    * You can optionally supply the  flags and description to override the defaults.
    *
-   * @param {string} [str]
+   * @param {string} str
    * @param {string} [flags]
    * @param {string} [description]
    * @return {this | string} `this` command for chaining, or version string if no arguments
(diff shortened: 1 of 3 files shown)


shadowspawn:
I like the concept, but so low value and you have so many other PR that closing.

May _never_ come up in the real world!

Issue #1907, opened 2023-07-14
Allow method _chainOrCall to return the promise result.


Rmannn:
Hi, I am the developper of a nestjs module using commander [nestjs-console](https://www.npmjs.com/package/nestjs-console)
Our module is built to partially wrap your module logic into a nestjs module.

We were extracting response from the command until the new version.

https://github.com/tj/commander.js/blob/4ef19faac1564743d8c7e3ce89ef8d190e1551b4/lib/command.js#L1301-L1305
As you can see here, actionResult will always returns undefined.

Can we add a line to return the result  ?
```typescript
if (this.parent) { 
   actionResult = this._chainOrCall(actionResult, () => { 
     this.parent.emit(commandEvent, operands, unknown); // legacy 
     // new line to allow us extracting the response
     return actionResult
   }); 
 } 
```

I could help opening a PR if you want.

Thx

shadowspawn:
I might be misunderstanding your question, so ask again if this answer does not make sense.

Commander does not expect the action handler to return a response, other than a promise if it is async. You might well think it does based on the `actionResult` variable name, but that it purely for supporting promises. The `actionResult` is used to work out whether to chain the next piece of code or to call it directly. A non-promise return value from the action handler is promptly forgotten in the next call to `_chainOrCall`.

For interest, you can see the action handler response was ignored before support for async was added. This is the PR adding the first version of async support: https://github.com/tj/commander.js/pull/1118/files

shadowspawn:
The link to #1118 was more relevant than I expected, it looks like you were relying on original implementation to indirectly get the action response, although `_actionResults` was only being retained to settle promises.

https://github.com/Pop-Code/nestjs-console/pull/345/files#diff-84b68bce2abfcb25c3328df0286963d7dacc924940adc41923bf4f502a6bc95aR105

Again, the internal variable name could lead to incorrect assumptions. Sorry about that. But the public facing API has the action handler returning void:
https://github.com/tj/commander.js/blob/4ef19faac1564743d8c7e3ce89ef8d190e1551b4/typings/index.d.ts#L497

Rmannn:
Thanks for your response.

PR #1958, opened 2023-08-13
Improve chain variable name

shadowspawn:
# Pull Request

Just a rename, no functional changes.

## Problem

The variable used for chaining is named for a particular result, but the real point is the promise chain and it has no other purpose. In particular, the settled promise value is not relevant.

This may have contributed to some confusion in #1907

## Solution

Rename variable in a couple of places to be all about the chaining.


abetomo (approved):
+1

PR #1960, opened 2023-08-13
Fix broken `typescript-checkJS` script

aweebit:
## Problem
Currently, `npm run typescript-checkJS` results in

```
error TS6053: File 'lib/*.js' not found.
  The file is in the program because:
    Root file specified for compilation
```

on Windows (even when run from Git Bash), and in normal termination with no output even when there are type errors in the files on [Linux](https://github.com/tj/commander.js/pull/1960#issuecomment-1676463561) and [macOS](https://github.com/tj/commander.js/pull/1960#issuecomment-1676457677).

## Solution
Fix the script command.

The configuration for the JS check has been moved to a dedicated config file.

The script has been renamed to `typescript-checkJs` for better consistency with [the TSConfig option name](https://www.typescriptlang.org/tsconfig#checkJs). (The script has been broken for a long time without someone noticing, so it is safe to assume it is not used in any contributors' custom scripts, and so the rename won't break anything.)

### …building upon this one
- #1962

### DIFF
--- a/lib/suggestSimilar.js
+++ b/lib/suggestSimilar.js
@@ -1,3 +1,5 @@
+// @ts-check
+
 const maxDistance = 3;
 
 function editDistance(a, b) {
--- a/package.json
+++ b/package.json
@@ -25,7 +25,7 @@
     "test": "jest && npm run test-typings",
     "test-esm": "node ./tests/esm-imports-test.mjs",
     "test-typings": "tsd",
-    "typescript-checkJS": "tsc -p tsconfig.checkJs.json",
+    "typescript-checkJs": "tsc -p tsconfig.checkJs.json",
     "test-all": "npm run test && npm run lint && npm run typescript-checkJS && npm run test-esm"
   },
   "files": [
--- a/package.json
+++ b/package.json
@@ -25,7 +25,7 @@
     "test": "jest && npm run test-typings",
     "test-esm": "node ./tests/esm-imports-test.mjs",
     "test-typings": "tsd",
-    "typescript-checkJS": "tsc --allowJS --checkJS index.js lib/*.js --noEmit",
+    "typescript-checkJS": "tsc -p tsconfig.checkJs.json",
     "test-all": "npm run test && npm run lint && npm run typescript-checkJS && npm run test-esm"
   },
   "files": [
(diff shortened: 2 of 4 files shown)


aweebit:
Enough rebases, should be good to go now.

shadowspawn:
The script runs fine on Mac, I use it as part of `npm run test-all`. I expect the underlying problem is it is relying on wildcard expansion by shell.

```
% npm run typescript-checkJS

% echo $?
0

```

aweebit:
> The script runs fine on Mac, I use it as part of `npm run test-all`. I expect the underlying problem is it is relying on wildcard expansion by shell.
> 
> ```
> % npm run typescript-checkJS
> 
> > commander@11.0.0 typescript-checkJS
> > tsc --allowJS --checkJS index.js lib/*.js --noEmit
> 
> % echo $?
> 0
> ```

Checked it on Debian and the result was the same as what you get.

However, it isn't correct. There are currently a lot of type errors in the code and we are supposed to see them, instead we see nothing.

aweebit:
Even when running the command for one individual file that has type errors (a lot of them) and with the exact option names from TypeScript's documentation, it still doesn't output anything:

```
$ npx tsc --allowJs --checkJs lib/command.js --noEmit
$ echo $?
0
```

So now it simply looks like the command has been broken the entire time  even though it has been in use, so renaming it might not be a good idea.

aweebit:
Good read: [Why you should always quote your globs in NPM scripts](https://medium.com/@jakubsynowiec/you-should-always-quote-your-globs-in-npm-scripts-621887a2a784)

shadowspawn:
There are (at least!) 3 different things going on:
- `// @ts-check` only works as the first line of the file
- `lib/*.js` in run-script `typescript-checkJS` does not work on Windows
- run-script `typescript-checkJS` generates far less errors than when enable checking in VSCode editor 

The command-line help gives one clue, specifying the files apparently uses the default compiler options:
```
tsc app.ts util.ts
  Ignoring tsconfig.json, compiles the specified files with default compiler options.
```

I haven't yet managed to get VSCode to do the gentler type checking that `typescript-checkJS` implicitly uses, but it may just be difference between `tsconfig.json` and default compiler settings.

shadowspawn:
Managed to get looser checks matching run-script in VSCode with both of these in `tsconfig`, so at least reproducible behaviour now:

```
   "checkJs": true,
   "strict": false, 
```

aweebit:
Closing in favor of #1969.

PR #1963, opened 2023-08-13
Fix option value source type

aweebit:
The change has been manually cherry-picked from https://github.com/tj/commander.js/pull/1919 so that it can be discussed independently.

## Problem

[#1919 (comment)](https://github.com/tj/commander.js/pull/1919/#issuecomment-1656490158)
[#1919 (comment)](https://github.com/tj/commander.js/pull/1919/#issuecomment-1656578103)
[#1919 (comment)](https://github.com/tj/commander.js/pull/1919/#issuecomment-1656581492)

## Solution

Compared with the original solution from #1919 (see https://github.com/tj/commander.js/compare/7fd0b22f1adc75a97bdb4af262a71126cb9f3b8f...52d5885140017009c930e77044b99f4baaf8c9c7), the one suggested here keeps `OptionValueSource` as an exported type (could be used in subclasses to narrow down the allowed custom sources) and uses the `OptionValueSource | string` union instead of just `string` for additional clarity.

### Fixed
- option value source type (`undefined` and arbitrary custom source strings are allowed now)

- #1919
- #1967

### DIFF
--- a/lib/command.js
+++ b/lib/command.js
@@ -780,7 +780,7 @@ Expecting one of '${allowedValues.join("', '")}'`);
    */
 
   setOptionValue(key, value) {
-    return this.setOptionValueWithSource(key, value, undefined);
+    return this.setOptionValueWithSource(key, value);
   }
 
   /**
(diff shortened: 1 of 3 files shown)


shadowspawn:
This isn't quite right, and ultimately I want to bring back the changes first tried in https://github.com/commander-js/extra-typings/pull/3

aweebit:
> This isn't quite right, and ultimately I want to bring in the changes first tried in [commander-js/extra-typings#3](https://github.com/commander-js/extra-typings/pull/3)

Okay, but #1967 won't work without the problem for this type being solved.

PR #1966, opened 2023-08-13
Fix `.description()`: do not set command description if `undefined` is provided in the first parameter

aweebit:
This is a ridiculously unimportant change, but it fixes one of the errors revealed by a run of the TypeScript check fixed in #1960.

I've learnt my [lesson](https://github.com/tj/commander.js/pull/1930#issuecomment-1676039813), and since the change is technically visible to the outside world, I am making a separate PR for it.

- #1967

### DIFF
--- a/lib/command.js
+++ b/lib/command.js
@@ -1837,7 +1837,9 @@ Expecting one of '${allowedValues.join("', '")}'`);
    */
   description(str, argsDescription) {
     if (str === undefined && argsDescription === undefined) return this._description;
-    this._description = str;
+    if (str) {
+      this._description = str;
+    }
     if (argsDescription) {
       this._argsDescription = argsDescription;
     }


lib/command.js
```diff
@@ -1837,7 +1837,9 @@ Expecting one of '${allowedValues.join("', '")}'`);
    */
   description(str, argsDescription) {
     if (str === undefined && argsDescription === undefined) return this._description;
```
aweebit, line 1839: Alternative solution that would make the change invisible to the outside world:

```suggestion
    if (str === undefined) return this._description;
```

shadowspawn: I actually like this alternative solution better, and `argsDescription` is deprecated. But it would technically be a breaking change. I think simpler just to leave it alone (and one day get rid of `argsDescription`).

shadowspawn:
Can revisit this if/when it gets in the way of adding more strict checks to the JavaScript.

Issue #1823, opened 2022-11-23
List all command's arguments using Helper


grallm:
I want to list all the arguments (or sub-commands, not sure of the difference) of my command.

If I look the property `_args`, I can confirm my command has an argument:
```json
[
  {
    "description": "",
    "variadic": false,
    "required": false,
    "_name": "commands"
  }
]
```
<details>
<summary>Screenshot of the Argument object</summary>
<img width="284" alt="image" src="https://user-images.githubusercontent.com/11904039/203580547-4471b0c2-c63c-4f1f-8092-d99135de3b92.png">
</details>

To not use a private property, I try to access this argument with the `Help` class and the `visibleArguments` method ([see the code](https://github.com/tj/commander.js/blob/master/lib/help.js#L90))

The problem, I do not have any description for my argument and the implementation of `visibleArguments` requires one.

Is this a problem? If not, can someone explain it to me? Is there any other clean way?

Thank you! 😁

shadowspawn:
> I want to list all the arguments (or sub-commands, not sure of the difference) of my command.

When you type something on the command line like
```
git clone my-repo
```

`clone` is a subcommand, and `my-repo` is a command-argument.

> To not use a private property, I try to access this argument with the `Help` class and the `visibleArguments` method [...]
> The problem, I do not have any description for my argument and the implementation of `visibleArguments` requires one.

Using `visibleArguments` to get the argument objects is an intended as a way to access the arguments, but as you discovered the "visible" part is that they are only returned if at least one of them has a description so that they appear in the help.

There are not any other methods to return the arguments from the private property `_args`.

As a work-around, could you add a description to your arguments so `visibleArguments` returns them? (What are you listing them for?)

grallm:
Thank you for your answer @shadowspawn 😄 

> clone is a subcommand, and my-repo is a command-argument.

The example is clear, but in the documentation of Commander I am just lost when speaking of sub-commands in the part of arguments while you can declare them with `.command` 😅 

Is there an explanation of why a Visible Argument is one only with a description? I may have missed something in your explanation, but still, the argument is displayed in the usage part of the help 🤔 

To describe my situation, I am using [NestJS Console](https://github.com/Pop-Code/nestjs-console/), based on top of Commander, and it seems I can't add argument description 😊
For the listing part, I am trying to generate dynamic configuration files.

Thanks again for the answer !

shadowspawn:
[...]

PR #1970, opened 2023-08-14
Make command-argument array public (in `readonly` mode)

aweebit:
## Problem

The array of registered command-arguments (currently `._args`) is not made publicly available unlike the arrays of registered options (`.options`) and subcommands (`.commands`). That is not only an obvious inconsistency, but also hinders third-party library development.

## Solution

Rename `._args` to `.registeredArguments` (because `.arguments` is already used for a method) and add the field to the type declaration for `Command` in `readonly` mode.

Other names considered:
- `.commandArguments`: confusing since the word "command" is not included in the names of `.argument()`, `.arguments()` and `.addArgument()`
- `.addedArguments`: makes it seem like only arguments added vis `.addArguments()` are included

The use of the word "registered" is also consistent with the name of `._registerOption()`, an internal routine suggested in #1923 and #1931.

### Added
- `.registeredArguments` (array with `Argument` instances for all registered command-arguments)

### DIFF
--- a/docs/deprecated.md
+++ b/docs/deprecated.md
@@ -15,6 +15,7 @@ They are currently still available for backwards compatibility, but should not b
   - [InvalidOptionArgumentError](#invalidoptionargumenterror)
   - [Short option flag longer than a single character](#short-option-flag-longer-than-a-single-character)
   - [Import from `commander/esm.mjs`](#import-from-commanderesmmjs)
+  - [.\_args](#_args)
 
 ## RegExp .option() parameter
 
@@ -207,3 +208,17 @@ import { Command } from 'commander';
 ```
 
 README updated in Commander v9. Deprecated from Commander v9.
+
+## ._args
+
+The registered command argument instances have been made accessible via `.registeredArguments`.
+
+```js
+const program = new Command().arguments('arg1 <arg2> [arg3]');
+const argumentNamesForUsage = program.registeredArguments
+  .map(arg => arg.required ? `< ${arg.name()} >` : `[ ${arg.name()} ]`)
+  .join(' ');
+program.usage('program ' + argumentNamesForUsage);
+```
+
+The older name of the property was `_args`. It was never documented but is still available so that old code relying on it keeps working. Do not use it.
--- a/typings/index.d.ts
+++ b/typings/index.d.ts
@@ -282,7 +282,7 @@ export class Command {
   readonly commands: readonly Command[];
   readonly options: readonly Option[];
   readonly registeredArguments: readonly Argument[];
-  /** @deprecated */
+  /** @deprecated Use `.registeredArguments` instead. */
   readonly _args: readonly Argument[]; // added here for strikethrough highlighting in editors
   parent: Command | null;
 
--- a/typings/index.d.ts
+++ b/typings/index.d.ts
@@ -282,6 +282,8 @@ export class Command {
   readonly commands: readonly Command[];
   readonly options: readonly Option[];
   readonly registeredArguments: readonly Argument[];
+  /** @deprecated */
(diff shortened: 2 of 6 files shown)


shadowspawn:
This fixes #1823

`registeredArguments` is potentially better than `declaredArguments` which was the best I came up with in:
https://github.com/tj/commander.js/issues/1823#issuecomment-1499716170

shadowspawn (changes requested):
To avoid immediately breaking existing client code which may have directly accessed `_args`, I suggest `this._args = this.registeredArguments` in constructor (and deprecating `_args` to remove later).

typings/index.d.ts
```diff
@@ -282,6 +282,8 @@ export class Command {
   readonly commands: readonly Command[];
   readonly options: readonly Option[];
   readonly registeredArguments: readonly Argument[];
+  /** @deprecated */
+  readonly _args: readonly Argument[]; // added here for strikethrough highlighting in editors
```
shadowspawn, line 286: When is this better than leaving it undeclared, and hence an error to use?

docs/deprecated.md
```diff
@@ -207,3 +208,17 @@ import { Command } from 'commander';
 ```
 
 README updated in Commander v9. Deprecated from Commander v9.
+
+## ._args
```
shadowspawn, line 212: Right idea, and I do want an entry in this doc, thanks. I will change this text. Shall I just do that after (if) gets merged, or would you like to incorporate rewrite?

typings/index.d.ts
aweebit: When coding in JavaScript and `checkJs` is not turned on in the TSConfig.

docs/deprecated.md
aweebit: Feel free to do it after it is merged, I don't mind.

typings/index.d.ts
shadowspawn: Feels slightly backwards, but it is typical to add warnings about deprecated things before removing them. Ok.

typings/index.d.ts
```diff
@@ -281,6 +281,9 @@ export class Command {
   processedArgs: any[];
   readonly commands: readonly Command[];
   readonly options: readonly Option[];
+  readonly registeredArguments: readonly Argument[];
+  /** @deprecated Use `.registeredArguments` instead. */
+  readonly _args: readonly Argument[]; // added here for strikethrough highlighting in editors
```
shadowspawn, line 286:
```suggestion
  readonly _args: readonly Argument[]; 
```

typings/index.d.ts
```diff
@@ -281,6 +281,9 @@ export class Command {
   processedArgs: any[];
   readonly commands: readonly Command[];
   readonly options: readonly Option[];
+  readonly registeredArguments: readonly Argument[];
+  /** @deprecated Use `.registeredArguments` instead. */
```
shadowspawn, line 285: Use same style as other deprecated messages
```suggestion
  /** @deprecated since v11, instead use `.registeredArguments` */
```

shadowspawn (changes requested):
Looks noisy because the name is so much longer! But not many authors will be typing it, and best name so far (and get away from the existing args/arguments/arguments). Should be accessible for symmetry with `commands` and `options` for people adding functionality.

Couple of minor changes for the comments adding *and* deprecating `_args` to typings. (Which surprised me at first, but happy now.)

(I'll update deprecation documentation later to more closely follow match the style I am using.)

typings/index.d.ts
aweebit: I actually thought about specifying the version, but decided against it because the property had never been supported officially.

shadowspawn:
Closing in favour of #2010, where will list @aweebit as co-author.

PR #1971, opened 2023-08-15
Make `Option.env()` behave as a getter when called with no arguments

aweebit:
Another tiny enhancement third-party libraries could benefit from (like #1970).

## Problem
I had to use the internal `Option.envVar` to build informative error messages in the [async-commander](https://github.com/aweebit/async-commander) library I am currently working on.

## Solution
Make `Option.env()` return the value of the variable when it is called with no parameters.

### Added
- `Option.env()` now returns the previously provided environment variable name when called with no argument

### DIFF
--- a/lib/option.js
+++ b/lib/option.js
@@ -109,16 +109,17 @@ class Option {
   }
 
   /**
-   * Set environment variable to check for option value.
+   * Get or set environment variable to check for option value.
    *
    * An environment variable is only used if when processed the current option value is
    * undefined, or the source of the current value is 'default' or 'config' or 'env'.
    *
-   * @param {string} name
-   * @return {Option}
+   * @param {string} [name]
+   * @return {Option | string | undefined}
    */
 
   env(name) {
+    if (!name) return this.envVar;
     this.envVar = name;
     return this;
   }
--- a/typings/index.d.ts
+++ b/typings/index.d.ts
@@ -142,12 +142,13 @@ export class Option {
   implies(optionValues: OptionValues): this;
 
   /**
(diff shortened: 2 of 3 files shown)


shadowspawn:
Unless the property name starts with an underscore, it is public. e.g. `option.public`, `option._private` So it is fine to access `option.envVar`.

Option was a simple prototype class with all public properties. As new features and Option became more full-featured the existing properties were preserved and the all-public pattern was continued. (Which does lead to some naming gymnastics to have both a property and a chaining setter.)

aweebit:
> Unless the property name starts with an underscore, it is public. e.g. `option.public`, `option._private` So it is fine to access `option.envVar`.

It is not included in the typings, though, that's why I though it was private. Had to include it in the manually augmented module declaration along with `Command._args` because of that, see [typings/commander.d.ts](https://github.com/aweebit/async-commander/blob/master/typings/commander.d.ts).

So do you think it should be added to the library's type declarations instead? Or can we also keep the change I suggest here? It's not always easy to remember which methods are setter-only and which can also be used as getters. Since there is no other meaningful interpretation for an `Option.env()` call with no arguments, I think it's okay if we keep it.

shadowspawn:
> It is not included in the typings, though, that's why I though it was private.

Oops, that explains it, I had wondered why you hadn't just used the property!

> So do you think it should be added to the library's type declarations instead?

Yes.

> Or can we also keep the change I suggest here? 

No. `Option` does not have any properties using the overloaded getter/setter that is used quite a bit in `Command`.

PR #1967, opened 2023-08-13
Fix all type errors

aweebit:
Fixes all type errors revealed by the TypeScript check fixed in #1960.

Builds on the fix from #1960, excluding the [questionable](https://github.com/tj/commander.js/pull/1960#issuecomment-1676467230) script rename [273a35d](https://github.com/tj/commander.js/pull/1960/commits/273a35de31bbc8c1dab14ef284cb6f6561f6d362).

Diff only for the changes introduced by this PR:
https://github.com/aweebit/commander.js/compare/5c5c9726c6c71cf2a53f67183ea5e16864a77c26...fix/type-errors

Includes fixes merged from #1963 #1965 #1966.

### Fixed
- (from [#1963](https://github.com/tj/commander.js/pull/1963)) option value source type (`undefined` and arbitrary custom source strings are allowed now)
- (from [#1965](https://github.com/tj/commander.js/pull/1965)) return type of `.executableDir()`

### Changes are to be merged into…
- #1975

Issue #1949, opened 2023-08-07
On `@api` JSDoc tag


aweebit:
The `@api` JSDoc tag has been used in the library code since [the initial commit](https://github.com/tj/commander.js/commit/672c7d01d8382257226d67c39c6e1002c881d95f). However, I have not been able to find any documentation on it. The fact its values are properly highlighted in VS Code is due to microsoft/vscode@9ad4bcd66f8c7d9bd71a400efd07f1fdaa454ab2, which includes no explanation on why the highlighting was added. My best guess is that it was added just to cover de facto usage like that in Commander (see microsoft/vscode#189812).

The well-documented alternatives to `@api private` (which is currently used throughout the library) are [`@private`](https://jsdoc.app/tags-private.html) and [`@access private`](https://jsdoc.app/tags-access.html). Should we use one of them instead?

aweebit:
Found the progenitor commit adding the highlighting later adopted by VS Code: [31d2a5f](https://github.com/atom/language-javascript/pull/512/commits/31d2a5f3d321761a1b2d1ec85ed3d507f76b5812)

My assumption was right. The commit description reads:

> Not part of the official JSDoc spec, yet occasionally used in-the-wild.

aweebit:
I found other usage examples and opened jsdoc/jsdoc#2066. Excited to see where it leads, but not expecting quick reaction since there has not been much activity in the repo recently.

aweebit:
I've just realized `@package` should be used instead of `@private` or `@api private` because classes use each others' "private" methods all over the library.

shadowspawn:
`@package` is technically accurate, but I had never seen it and had to look it up to confirm the meaning. I think @private is much clearer for Commander users seeing the tag.

aweebit:
> `@package` is technically accurate, but I had never seen it and had to look it up to confirm the meaning.

Unfortunately, it is indeed quite uncommon ([only 10k results](https://github.com/search?q=language%3AJavaScript+OR+language%3ATypeScript+%22*+%40package%22+NOT+%22*+%40packageDocumentation%22+NOT+%22*+%40package+%22&type=code) in GitHub Search).

> I think `@private` is much clearer for Commander users seeing the tag.

The problem is that if you replace all `@api private` occurrences by `@private`, you will see a bunch of TypeScript errors that were not show before because `@api private` is not valid JSDoc, and in particular not currently supported by the TypeScript flavor of JSDoc, but that could change in the future.

[...]

PR #1975, opened 2023-08-17
[Proof of concept] The ultimate TypeScript setup rework

aweebit:
### Core components of this PR
- [ccc0245](https://github.com/tj/commander.js/pull/1975/commits/ccc02454af440182b6bba4eefda3a8cdbe21edde):
  - Generate `.d.ts` typings automatically from JSDoc (see [TypeScript docs](https://www.typescriptlang.org/docs/handbook/declaration-files/dts-from-js.html)) with `npm run build`
  - Modularize TSConfig by using [project references](https://www.typescriptlang.org/docs/handbook/project-references.html) so that only relevant setting are applied for each kind of JS/TS files (separate configs for `lib`, Jest tests and tsd tests)
  - Remove the `typescript-checkJS` script. Add `typecheck`, `typecheck-jest` and `typecheck-all` instead (Jest tests are currently not included in `typecheck`)
  - The `test-typings` script now also includes a typecheck
- #1974 (augmented by [71d0966](https://github.com/tj/commander.js/pull/1975/commits/71d0966e6e700d5c50c5ed5486ceb50f79b5a685), but there might be better solutions)
- #1967 (includes the now-closed #1963 that might need to be reconsidered) 
- [77d256c](https://github.com/tj/commander.js/pull/1975/commits/77d256c8c123fa8f94178a98d7df196a51066cee): move `index.js` and `esm.mjs` to `lib` where they belong (to `lib/exports`, to be precise)
- [cd4338b](https://github.com/tj/commander.js/pull/1975/commits/cd4338b15234d0a95ec7ed1193c979f20ffacd97): replace `@api private` with `@package` although unfortunately both mean nothing to TypeScript, more information at #1949

### Why generate `.d.ts` from JSDoc?
- one source of truth
- therefore no inconsistencies between types in `.d.ts` and JSDoc like currently
- therefore no chance contributors forget to add a new field / method to `.d.ts`
- less work since no manual management of `.d.ts` is required
- source maps are generated (source code shown when using Go to Definition, much more useful than just the declaration and developer doesn't have to click around searching for the source file)
- all fields and methods are exposed as [recommended](https://github.com/microsoft/TypeScript/issues/1867#issuecomment-209018980) by the TypeScript team
  - Unfortunately, use of methods marked with `@package` outside Commander is not prevented by TypeScript, but that has little to do with this PR. See [#1949 (comment)](https://github.com/tj/commander.js/issues/1949#issuecomment-1682354349) for details

### TODO
- Manually eliminate inconsistencies between JSDoc and the old `.d.ts`
- Add `.d.ts` files to `.gitignore? (Svelte [does that](https://github.com/sveltejs/svelte/blob/master/packages/svelte/.gitignore) for example)
- Draw more inspiration from Svelte and SvelteKit

### See also
- [TypeScript is ‘not worth it’ for developing libraries, says Svelte author, as team switches to JavaScript and JSDoc](https://devclass.com/2023/05/11/typescript-is-not-worth-it-for-developing-libraries-says-svelte-author-as-team-switches-to-javascript-and-jsdoc/)


shadowspawn:
This is an interesting refactor, but _not_ something I think we would adopt. If we went all-in on typing then I think it is more likely we would switch to TypeScript implementation. We do certainly have parallel JSDoc and TSDoc, but a benefit is tailoring the two interfaces separately. (Although I wouldn't recommend this in a new program!)

PR #1977, opened 2023-08-23
Refactor to wrap invalid argument

shadowspawn:
# Pull Request

This is an optional refactor.

## Problem

The extra try/catch code to handle `command.invalidArgument` means calling `option.parseArgs` and `argument.parseArgs` is noisy.

## Solution

I liked the refactoring in #1913 which wrapped exception and promise handling. Here it is just the exception handling, but still cleans up the calling code.


abetomo (approved):
+1

PR #1982, opened 2023-08-25
Types for version getter

shadowspawn:
# Pull Request

## Problem

The implementation for `.version()` has always supported returning the (previously defined) version string, but the
types have never included this. I did briefly consider deprecating the whole thing (#1954), but it is being used in the wild
and not hard to support!

## Solution

Small change to types and JSDoc and TSDoc to include the getter behaviour.

### DIFF
--- a/lib/command.js
+++ b/lib/command.js
@@ -1818,17 +1818,16 @@ Expecting one of '${allowedValues.join("', '")}'`);
   }
 
   /**
-   * Set the program version to `str`.
+   * Get or set the program version.
    *
-   * This method auto-registers the "-V, --version" flag
-   * which will print the version number when passed.
+   * This method auto-registers the "-V, --version" option which will print the version number.
    *
-   * You can optionally supply the  flags and description to override the defaults.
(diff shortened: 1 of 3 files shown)


lib/command.js
```diff
@@ -1818,17 +1818,16 @@ Expecting one of '${allowedValues.join("', '")}'`);
...
    *
-   * You can optionally supply the  flags and description to override the defaults.
+   * You can optionally supply the flags and description to override the defaults.
    *
-   * @param {string} str
+   * @param {string} [str]
    * @param {string} [flags]
    * @param {string} [description]
-   * @return {this | string} `this` command for chaining, or version string if no arguments
+   * @return {this | string | undefined} `this` command for chaining, or version string  if no arguments
```
abetomo, line 1830: Do you need this space?

```
version string[ ][ ]if no arguments
```

shadowspawn: No! Fixed.

abetomo (approved):
👍

PR #1984, opened 2023-08-26
Strict optionals

shadowspawn:
# Pull Request

## Problem

Commander does not support strict optional option-arguments per POSIX. This is an interesting mode as it resolves the ambiguity between an optional option-argument and a (positional) command-argument.

```
program -p
program -pPORT
program --port
program --port=PORT
program ARG
program -p ARG
program --port ARG
```

See #1901 (also #1951)

## Solution

Add `.strictOptionalOptionArguments()` to enable the POSIX treatment.

Example:

```js

program
  .strictOptionalOptionArguments()
  .option('-C[num], --context[=num]', 'grep style flag help')
  .option('-t, --track[=(direct|inherit)]', 'git style flag help');

program.parse();
console.log({ opts: program.opts(), args: program.args });
```

Note: the flags displayed in the help are as specified in the `.option()` call, so author can describe them in their preferred usage style .

```console
% node index.mjs --help
Usage: index [options]

Options:
  -C[num], --context[=num]        grep style flag help
  -t, --track[=(direct|inherit)]  git style flag help
  -h, --help                      display help for command
% node index.mjs -C -t 
{ opts: { context: true, track: true }, args: [] }
% node index.mjs -C 1 -t 2
{ opts: { context: true, track: true }, args: [ '1', '2' ] }
% node index.mjs -C1 -t2  
{ opts: { context: '1', track: '2' }, args: [] }
% node index.mjs --context 1 --track 2
{ opts: { context: true, track: true }, args: [ '1', '2' ] }
% node index.mjs --context=1 --track=2
{ opts: { context: '1', track: '2' }, args: [] }
```

### DIFF
--- a/lib/command.js
+++ b/lib/command.js
@@ -47,6 +47,7 @@ class Command extends EventEmitter {
     this._exitCallback = null;
     this._aliases = [];
     this._combineFlagAndOptionalValue = true;
+    this._strictOptionalOptionArguments = false;
     this._description = '';
     this._summary = '';
     this._argsDescription = undefined; // legacy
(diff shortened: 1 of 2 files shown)


shadowspawn:
Not sure if there is enough interest in the POSIX compatible processing to make it worth adding. Leaving  #1901 open to gather up-votes and/or compelling use cases. Can reopen this if there is support.

PR #2005, opened 2023-09-02
Move Jest configuration to config file

shadowspawn:
This is an optional refactor to move the Jest configuration from `package.json` to a configuration file. I used to quite like using `package.json` for extra configuration and reducing number of files, but I do find configuration for packages a little easier to find and work with when the settings when in an external file, and it is annoying being unable to add comments in `package.json`.

Reference: https://jestjs.io/docs/configuration


abetomo (approved):
+1

Issue #689, opened 2017-09-05
Don't Show options when there are no options


mateodelnorte:
The simplest possible executable with a single command and no options still forces showing of options: 
```
const program = require('commander');

program
  .command('cmd', 'a single cmd with no options')

program
  .parse(process.argv);
```
yields: 
```
./bin/my-cmd

  Usage: my-cmd [options] [command]


  Options:

    -h, --help  output usage information


  Commands:

    cmd         a single cmd with no options
    help [cmd]  display help for [cmd]
```

Why show options here? Help is already redundant, and the resulting UI is less usable because the intended call to action obscured by a big in-your-face prompt about `-h`. 

Can we allow setting a property to now show options?

grant:
The line to change would be here:
https://github.com/tj/commander.js/blob/master/index.js#L956

Probably a change like this:

## Current
```js
Command.prototype.optionHelp = function() {
  var width = this.largestOptionLength();

  // Append the help information
  return this.options.map(function(option) {
      return pad(option.flags, width) + '  ' + option.description
        + (option.defaultValue !== undefined ? ' (default: ' + option.defaultValue + ')' : '');
  }).concat([pad('-h, --help', width) + '  ' + 'output usage information'])
    .join('\n');
};
```

## Proposed
```js
Command.prototype.optionHelp = function() {
  if (!this.options.length) return [];
  var width = this.largestOptionLength();

  // Append the help information
  return this.options.map(function(option) {
      return pad(option.flags, width) + '  ' + option.description
        + (option.defaultValue !== undefined ? ' (default: ' + option.defaultValue + ')' : '');
  }).concat([pad('-h, --help', width) + '  ' + 'output usage information'])
    .join('\n');
};
```

Only join if the array is not empty.

shadowspawn:
This has got a few likes and I have thought the same myself at times, so although no recent activity on this issue I am adding a comment rather than closing as stale.

A reason for not returning empty `.optionHelp` when no custom options (as suggested) is that the help may be displayed programatically and in that case the help entry itself is not redundant. Otherwise this would be a nice simple approach.

If we end up with a parameter for other help behaviours that have been requested, like changing exit behaviour, then there would at least be a place to add a property for suppressing option help as suggested here.

oa495:
Any update on this?

shadowspawn:
Any new information @oa495 ?

Not included in https://github.com/tj/commander.js/milestone/3

shadowspawn:
[...]

PR #2006, opened 2023-09-02
Refactor help option implementation to hold actual Option

shadowspawn:
# Pull Request

The main motivation for this refactor is to potentially simplify future work that uses more Option properties. For example, adding help group to Option like in #1910.

Instead of extending `.helpOption()` to add parallel features like:
```
program.helpOption('-s, --support', 'Show help', { helpGroup: 'Demo Options:' });
```

could instead use same approach as normal options:
```
program.addHelpOption(new Option('-s, --support', 'Show help').group('Demo Options:'));
```

(This overlaps with motivation and implementation in #1929.)

## Problem

Additions to `Option` that can be used with `.addOption()` can't be used with the help option. For example, hiding the help option (#689).

There are an annoying number of help option related properties on `Command`.

Determining the help option flags does not go through `cmd.createOption()` so misses some potential custom class behaviour, such as custom flags parsing. (Minor!)

## Solution

Store a help Option rather than the necessary deconstructed properties.

Add `.addHelpOption()` to add author created option, exposing more configuration just like `.addOption()` does.

## ChangeLog

- Added: `.addHelpOption()` as another way of configuring built-in help option
- Changed: refactor internal implementation of built-in help option

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -904,6 +904,8 @@ program
   .helpOption('-e, --HELP', 'read more information');
 ```
 
+You can also customise the help option using `.addHelpOption()` with an `Option` you create yourself.
+
 ### .addHelpCommand()
 
 A help command is added by default if your command has subcommands. You can explicitly turn on or off the implicit help command with `.addHelpCommand()` and `.addHelpCommand(false)`.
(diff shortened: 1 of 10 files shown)


danielsann038:
Hi everyone

shadowspawn:
Still wondering about possible problems with this, but pretty happy with how it has worked out so opening it for review.

shadowspawn:
Fixing merge conflicts...

shadowspawn:
Updated with release/12.x changes.

shadowspawn:
Thanks @abetomo  🚀

shadowspawn:
I will do another prerelease this weekend with the help command and help option refactor, and then we could perhaps release in a couple of weeks.

Issue #1953, opened 2023-08-09
Add regex parameter to `.argument`


MaxumCSmith:
I find myself having to validate arguments more thoroughly using Regex. I end up doing this from within the action method, but it would be nice if we could pass another parameter into `.argument` as a regex that would be used to validate.

Maybe something like this would be nice?

`argument(name: string, description?: string, defaultValue?: unknown): this;` -> `argument(name: string, description?: string, defaultValue?: unknown, validationRegex: Regex): this;`

shadowspawn:
You can write a custom processing function that validates the argument with a regular expression. 

See https://github.com/tj/commander.js#custom-argument-processing and have a look at `myParseInt` in the [Custom option processing](https://github.com/tj/commander.js#custom-option-processing) section

To avoid writing multiple parsers if you want to use different regex you can write a function to generate the parser from the regex, see later comments in #1130

MaxumCSmith:
I see. I didn't notice the ability to pass in a processing function. `fn` should enable what I need.
* `argument<T>(flags: string, description: string, fn: (value: string, previous: T) => T, defaultValue?: T): this;`

Thank you.

shadowspawn:
I'll close this issue, but #1909 is still open and flagged for reviewing the documentation.

Issue #1909, opened 2023-07-15
[docs] Add clarifying example of `fn` param to option()


gregfenton:
The description for what the `fn` parameter of `option()` is sparse and there are no examples of its use.  Looking at the code for it, the use is non-obvious.

gregfenton:
Might also consider adding an example of a parameter being passed that is quoted:
```
$ node options-common.js --pizza-type='veggie lovers' -s

pizza details:
- small pizza size
- veggie lovers
```

See the [Options docs ](https://github.com/tj/commander.js/tree/master#options) and the [options-common.js](https://github.com/tj/commander.js/blob/master/examples/options-common.js) examples.

shadowspawn:
See the README section on custom option processing:

https://github.com/tj/commander.js#custom-option-processing

gregfenton:
Okay, cool that it is in there.  But it was not obvious to see/find.  It comes after a long scroll and after topics labelled "Other" and "Misc"....at which point I assumed I had hit the end of the Options section.

Feel free to close as the docs are there.  But I recommend some sort of tweaking to make this more obvious.  I did not link  the `fn` parameter to be "custom processing".

shadowspawn:
Why were you looking for more about the `fn` parameter? From in-editor assistance while typing, or looking at the Javascript code, or looking at the TypeScript definition?

gregfenton:
In-editor (VSCode).  I was looking to do something new but realized that I didn't have sufficient information to know what `fn` did, or whether it could be a valid approach.  Ultimately chose to do something different, but that was after some frustration trying to understand `fn`.

Again, delighted that the docs are there.  But didn't find those docs on my own.

Thanks!

shadowspawn:
Another person who saw `fn` but not the custom parsing: #1953

MaxumCSmith:
Thank you.

shadowspawn:
Opened #2009 with a rework of the inline documentation. Single most relevant change was renaming `fn` to `parseArg`!

shadowspawn:
Shipped in Commander v11.1.0

PR #2009, opened 2023-09-10
Rework option inline doc

shadowspawn:
# Pull Request

## Problem

The inline documentation for `.option()` dates back to early days when it was used to generate the online documentation. It was quite lengthy covering the flags but not much else. In particular the custom argument parsing was named simply `fn` and easily missed.

Related: #1909 #1953

## Solution

Rename `fn` to `parseArg` for better clue! Shorten the examples, and call-out to the README and other related functions.

### DIFF
--- a/lib/command.js
+++ b/lib/command.js
@@ -632,57 +632,32 @@ Expecting one of '${allowedValues.join("', '")}'`);
   }
 
   /**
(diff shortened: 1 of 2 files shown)


abetomo (approved):
👍

shadowspawn:
Shipped in Commander v11.1.0

PR #2010, opened 2023-09-10
Add public Arguments property

shadowspawn:
# Pull Request

## Problem

There is no public property for the command `Argument[]`, like there is with `program.options` and `program.commands`.

The `Help.visibleArguments()` is not a working alternative since unless the argument have descriptions, there are no visible arguments.

See: #1823

## Solution

Add `.registeredArguments`, as suggested in #1970

Reminder to self: preserve co-author commit comment.

## ChangeLog

- add `cmd.registeredArguments` to access the command `Argument` array


abetomo (approved):
+1

shadowspawn:
Shipped in Commander v11.1.0

PR #2011, opened 2023-09-10
Work around Jest bug

shadowspawn:
# Pull Request

## Problem

Work-around a Jest issue that shows up in recent Node.js 18 versions. Ascii colours are being included in spawned command output in certain tests. As first raised in #1950.

Reference: https://github.com/jestjs/jest/issues/14391

## Solution

Suppress colours in the failing tests with environment variable `FORCE_COLOR`. 

This is a simpler version of #1950. (@aweebit added as co-author in commit comment)

### DIFF
--- a/tests/fixtures/inspect.js
+++ b/tests/fixtures/inspect.js
@@ -2,6 +2,8 @@
 
 const program = require('../../');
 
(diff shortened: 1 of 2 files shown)


abetomo (approved):
+1

PR #2013, opened 2023-09-11
Fix factory functions with typescript whole module import

shadowspawn:
# Pull Request

A subtle problem with Commander exports got reported in #1974

## Problem

The factory functions like `createOption` were not available using a TypeScript whole module import (`import * as commander`) when `esModuleInterop: true`.

## Solution

Explicitly export the factory functions, rather than just rely on implicitly being available via global program. This is also consistent with eventual goal of dropping deprecated default export of global program.

Inspired by #1974. Lists @aweebit as co-author in commit comment.

## ChangeLog

- fix: global factory functions now available when using TypeScript whole module import (`import * as commander`)

### DIFF
--- a/index.js
+++ b/index.js
@@ -10,18 +10,30 @@ const { Option } = require('./lib/option.js');
  * Expose the root command.
  */
 
-exports = module.exports = new Command();
-exports.program = exports; // More explicit access to global command.
-// Implicit export of createArgument, createCommand, and createOption.
+const program = new Command();
+exports = module.exports = program; // default export (deprecated)
+exports.program = program; // more explicit access to global command
 
 /**
  * Expose classes
  */
 
-exports.Argument = Argument;
 exports.Command = Command;
-exports.CommanderError = CommanderError;
+exports.Option = Option;
+exports.Argument = Argument;
 exports.Help = Help;
+
+exports.CommanderError = CommanderError;
 exports.InvalidArgumentError = InvalidArgumentError;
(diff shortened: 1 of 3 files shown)


abetomo (approved):
LGTM

shadowspawn:
I did think it looked like a circular reference, but if it made TypeScript happy...

However, a little amusingly and annoyingly, TypeScript thinks it is circular too! (Apparently I didn't try the full tests before now.)

It was interesting learning about the issue and I am still happy the tests are improved, but planning to comment out the factory functions again. There is an easy work-around if anyone hits the issue, which is to import the factory function using a named import.

```
% npm run typescript-checkJS

index.js:38:1 - error TS2300: Duplicate identifier 'createCommand'.

38 exports.createCommand = (name) => new Command(name);
   ~~~~~~~~~~~~~~~~~~~~~

  lib/command.js:188:3
    188   createCommand(name) {
          ~~~~~~~~~~~~~
    'createCommand' was also declared here.

index.js:39:1 - error TS2303: Circular definition of import alias 'createArgument'.

39 exports.createArgument = program.createArgument;
   ~~~~~~~~~~~~~~~~~~~~~~

index.js:40:1 - error TS2303: Circular definition of import alias 'createOption'.

40 exports.createOption = program.createOption;
   ~~~~~~~~~~~~~~~~~~~~


Found 3 errors in the same file, starting at: index.js:38
```

PR #2014, opened 2023-09-12
Revert factory reexport

shadowspawn:
# Pull Request

## Problem

Adding reexports of the factory function in #2013 causes a TypeScript error from one of the optional checks:
```
% npm run typescript-checkJS

index.js:37:1 - error TS2303: Circular definition of import alias 'createCommand'.

37 exports.createCommand = program.createCommand;
   ~~~~~~~~~~~~~~~~~~~~~

index.js:38:1 - error TS2303: Circular definition of import alias 'createArgument'.

38 exports.createArgument = program.createArgument;
   ~~~~~~~~~~~~~~~~~~~~~~

index.js:39:1 - error TS2303: Circular definition of import alias 'createOption'.

39 exports.createOption = program.createOption;
   ~~~~~~~~~~~~~~~~~~~~


Found 3 errors in the same file, starting at: index.js:37
```

## Solution

The failure the re-export was fixing is that the factory functions are not available with a whole-module import (`import * as commander`) when `esModuleInterop: true`. 

A simple work-around if someone does encounter the problem is to use a named import for the factory function(s) they want.

So reverting the ex-export! (And the related tests can come back one day.)

### DIFF
--- a/index.js
+++ b/index.js
@@ -14,6 +14,8 @@ const program = new Command();
 exports = module.exports = program; // default export (deprecated)
 exports.program = program; // more explicit access to global command
 
+// createArgument, createCommand, and createOption are implicitly available as they are methods on program.
+
 /**
  * Expose classes
  */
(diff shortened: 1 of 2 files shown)


abetomo (approved):
😢

Issue #2016, opened 2023-09-15
Drop default import of global Command (for CommonJS)


shadowspawn:
The default import of a global Command object (aka `program`) has been out of favour since Commander v5, [deprecated](https://github.com/tj/commander.js/blob/master/docs/deprecated.md#default-import-of-global-command-object) since v7, removed from the TypeScript in v8, and never implemented for the explicit esm support (#1440).

In general with the deprecated features and behaviours, I am willing to leave them alone until they cause problems rather than aggressively remove them, to minimise breaking changes when people upgrade from old versions of Commander.

The global Command object was involved in recent problems with TypeScript (#2013 #2014), so I am suggesting removing it while I am still annoyed! 😠 

To be clear, this will still work:
```js
const { program } = require('commander');
program.parse();
```

And this:
```js
const commander = require('commander');
const program = new commander.Command();
program.parse();
```

But this won't work anymore:
```js
const commander = require('commander');
commander.parse();
```

shadowspawn:
Released in Commander v12.0.0.

PR #2017, opened 2023-09-15
Remove default export of global program

shadowspawn:
# Pull Request

## Problem

We have been slowly moving towards removing the default export of global Command instance. Already removed from the README, removed from the TypeScript, and omitted from the esm interface.

Recently observed some problems due to the legacy export: #2013 #2014

Deprecation: https://github.com/tj/commander.js/blob/4ef19faac1564743d8c7e3ce89ef8d190e1551b4/docs/deprecated.md?plain=1#L41-L60

Issue: #2016

## Solution

Remove from the CommonJS exports too!

## ChangeLog

- Removed: _breaking_: removed default export of a global Command instance from CommonJS (use the named `program` export instead)

### DIFF
--- a/docs/deprecated.md
+++ b/docs/deprecated.md
@@ -58,6 +58,7 @@ const program = new Command()
 - Removed from README in Commander v5.
 - Deprecated from Commander v7.
 - Removed from TypeScript declarations in Commander v8.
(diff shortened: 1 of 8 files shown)


abetomo (approved):
+1

PR #2018, opened 2023-09-15
Switch @api private to official JSDoc

shadowspawn:
# Pull Request

## Problem

Commander uses `@api private` which didn't make it into official JSDoc, although supported in some implementations.

See: #1949

## Solution

Switch to `@private`, or `@package` for routines that are intended for internal use (or happen to be used internally!).

Made `.attributeName()` public in the JSDoc to match the TypeScript, which is the intended visibility as needed to look up option values.

(I targeted v12 in case breaks some client use in an environment which actually starts warning about use of private routines.)

### DIFF
--- a/lib/argument.js
+++ b/lib/argument.js
@@ -52,7 +52,7 @@ class Argument {
   }
 
   /**
-   * @api private
+   * @package internal use only
    */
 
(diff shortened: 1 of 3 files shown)


abetomo (approved):
+1

Issue #1972, opened 2023-08-15
TypeScript definitions missing Option.envVar


shadowspawn:
`Option.envVar` is a public property of the JavaScript class.

The `index.d.ts` TypeScript definition is missing the `envVar` property for the `Option` class.

(Raised in #1971)

aweebit:
`presetArg`, `conflictsWith` and `implied` are also missing.

Same with `parseArg`, `defaultValue`, `defaultValueDescription` and `argChoices` in the `Argument` class. Also might want to consider making `_name` public (for comparison, option-argument names can be accessed by extracting them from the public `flags` property).

shadowspawn:
> are all explicitly marked as private in JSDoc despite having no underscores in their names. Historical reasons, I guess?

Yes indeed. I did add an issue about renaming those, but decided to minimise breakages and leave them alone:  #1104

shadowspawn:
The TypeScript has never had 1:1 match with properties visible from JavaScript, but adding them as needed, and intend to usually add them for new properties.

(The dual purpose getter/setter does at least reduce the number of properties exposed, much as I dislike the overload on return type!)

shadowspawn:
Shipped in Commander v11.1.0

PR #2019, opened 2023-09-17
Add more Option and Argument properties to typings

shadowspawn:
# Pull Request

## Problem

Not all of the Option and Argument properties are covered in the typings. I have mixed feeling about covering everything since most people don't need them and they appear in the editor completion suggestions, but appropriate to add things that people ask for or might need, and make Option and Argument symmetrical.

See: #1972

## Solution

Added a few more properties, and typing tests.

### DIFF
--- a/typings/index.d.ts
+++ b/typings/index.d.ts
@@ -50,6 +50,9 @@ export class Argument {
   description: string;
   required: boolean;
   variadic: boolean;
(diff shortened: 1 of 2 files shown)


abetomo (approved):
+1

shadowspawn:
Shipped in Commander v11.1.0

Issue #2021, opened 2023-09-21
Exit Code is Zero if Subcommand Killed by Signal


dsanders11:
If a subcommand child process is killed by a signal (e.g. `SIGKILL`), then commander will exit with exit code zero. I'm seeing this where OOM was killing the subcommand but commander was exiting with zero.

Relevant code looks to be here. The arguments to the `close` event are `(code, signal)` and in the case of killed by signal, `code` will be `null`, so `process.exit` gets called with `null` which leads to exiting with zero. Additionally, should the `exitCallback` case (on line 1036) be using `proc.exitCode` instead of `process.exitCode`? The latter doesn't seem correct.

https://github.com/tj/commander.js/blob/4ef19faac1564743d8c7e3ce89ef8d190e1551b4/lib/command.js#L1032-L1038

Possible fix for the issue could be to bubble up signals, although I'm not sure how to handle the `exitCallback` case then, since AFAIK there's no way to get an exit code for a signal from Node.js (other than mapping it yourself).

```js
proc.on('close', (code, signal) => {
  if (signal) {
    process.kill(process.pid, signal);
  } else {
    process.exit.bind(process)(code);
  }
});
```

Alternatively a fixed exit code of 1 could be used in the case of a signal, which would work for the `exitCallback` case. However, that's not desirable since it will hide the underlying exit code from the subcommand.

Manually mapping signals to exit codes might be the simplest way to handle this for both cases.

shadowspawn:
Commander does listen for some signals here: https://github.com/tj/commander.js/blob/4ef19faac1564743d8c7e3ce89ef8d190e1551b4/lib/command.js#L1021

But when I attempted to reproduce, those never fire! I wonder if the signal listener only works within the current process, and instead Commander should be listening for `close` which gets passed the signal and exit code: https://nodejs.org/dist/latest-v18.x/docs/api/child_process.html#event-close

We do listen for `close` already but pay not attention to the parameters.

shadowspawn:
Apologies @dsanders11, you are quite correct. I didn't understand your initial description on first read and made a mess of experimenting with signals. Now your description makes sense, and good detail, thanks. 

I'll take another look! 😊

shadowspawn:
> Additionally, should the `exitCallback` case (on line 1036) be using `proc.exitCode` instead of `process.exitCode`?

Yes, I agree.

shadowspawn:
[...]

PR #2023, opened 2023-09-23
Exit with non-zero code when subprocess terminated by signal

shadowspawn:
# Pull Request

## Problem

There are a couple of problems with the `'close'` event handler from spawned executable subcommand:
- exit code is effectively zero if subprocess terminated due to a signal
- exit code passed to exit override is (probably) always zero

See #2021

## Solution

Use the event `code`argument, or `1` if code is `null` because of a signal. 

Why `1` for signal? The exit code seen in the shell when a process terminates due to a signal is often the signal number plus an offset, like 128+n in bash. We don't have a good way of mapping a signal name to a signal number, and don't have a way of returning the right offset for the current shell through `process.exit()`. Using `kill` to terminate main process with the same signal as subprocess feels a bit heavy handed and dubious. So going for exit code of `1` as simple and an incremental improvement from existing code.

Targeting a major version as changing years old behaviour, although I think it is effectively a bug fix.

## ChangeLog

- fix: use non-zero exit code when spawned executable subcommand terminates due to a signal

### DIFF
--- a/lib/command.js
+++ b/lib/command.js
@@ -1067,15 +1067,15 @@ Expecting one of '${allowedValues.join("', '")}'`);
     }
 
     // By default terminate process when spawned process terminates.
-    // Suppressing the exit if exitCallback defined is a bit messy and of limited use, but does allow process to stay running!
     const exitCallback = this._exitCallback;
-    if (!exitCallback) {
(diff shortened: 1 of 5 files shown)


abetomo (approved):
LGTM

PR #2025, opened 2023-10-07
Update CHANGELOG for 11.1.0

shadowspawn:
Preparing for a minor release. I have gone through the Pending Release label, and taken a look at branch difference too.

I usually update the version number too in PR, but would it be more convenient if I leave that for you at release time @abetomo so you can use (say) `npm version minor` and get the tagging done at the same time?

I noticed #1928 introduces a throw, so I think it would be better in a major version. I'll remove it from `develop` branch and add it to `release/12.x`. (#2026)


abetomo:
Thanks.
I think it would be good to update the version number in the PR as it would be easier to understand the changes.

abetomo (approved):
👍

shadowspawn:
Released, woop! Thanks @abetomo

PR #2027, opened 2023-10-07
Drop Node.js 16

shadowspawn:
# Pull Request

## Problem

Long Term Support for Node.js 16 has finished.

## Solution

Bump engine and drop support for Node.js 16 in the next major release.

## ChangeLog

- *Breaking:* Commander 12 requires Node.js v18 or higher


abetomo (approved):
👍

PR #2028, opened 2023-10-07
Refactor headings

shadowspawn:
# Pull Request

## Problem

There are now multiple sections in the options documentation. This is not reflected in the headings.

## Solution

Refactor the headings so the Table of Contents make sense.


abetomo (approved):
+1

PR #2029, opened 2023-10-07
Restore extra sanity check when enabling storeOptionsAsProperties

shadowspawn:
This is just enabling #1928 again, to land in a major version release.


abetomo (approved):
👍

PR #2035, opened 2023-10-15
Add migration tips for default import

shadowspawn:
# Pull Request

## Problem

Removing the default export is going to break old programs.

## Solution

Include explicit Migration Tips in the CHANGELOG to highlight the necessary changes.


abetomo (approved):
+1

PR #2036, opened 2023-10-15
Add explicit note to only call parse once

shadowspawn:
# Pull Request

## Problem

Commander does not reset state when `.parse()` is called. I don't consider this as a bug as starting afresh with a new program/Command is simple and robust. But we haven't explicitly stated this is the expected model.

See: #438 #1565 #1819 #1829 #1916

## Solution

Add a note to README. Will still catch some people out, but at least it is documented!

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -967,6 +967,8 @@ program.parse(); // Implicit, and auto-detect electron
 program.parse(['-f', 'filename'], { from: 'user' });
 ```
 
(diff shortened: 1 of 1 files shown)


abetomo (approved):
+1

shadowspawn:
Opened a PR to allow multiple calls to `.parse()`: #2299

Issue #1978, opened 2023-08-23
Poll: check for more usage errors


shadowspawn:
Commander could check for more usage errors. These are probably a win for new programs and people new to Commander, but may mean more work when upgrading Commander in an old program (to fix problems that had not been noticed previously).

Have a look and upvote 👍 errors you wish Commander would check for, and downvote 👎  errors that would get in the way of how you are using Commander or you wouldn't want to fix when upgrading.

shadowspawn:
1. Calling `program.parse()` when the action handler or hook is async . Should be calling `program.parseAsync()`.

Note: this error can not be detected except during parsing, so may not show up on call calls.

shadowspawn:
2. Adding options with overlapping flags.

```
program
   .option('-p, --print')
   .option('p, --port');
```

shadowspawn:
3. Adding commands with overlapping names.

```
program.command('runnable')
   .alias('run')
   // more configuration

program.command('run') // oops, already using 'run'
   // more configuration
```

shadowspawn:
4. Calling `.parse()` more than once. Parse leaves behind state, so the intention is make a new program for every parse.

```
program.parse(['--port', '80'], { from: 'user' });
assert_deep_equal(program.opts(), { port: 80 });
program.parse(['--name', 'foo'], { from: 'user' });
assert_deep_equal(program.opts(), { name: 'foo' }); // oops, has port too
```

shadowspawn:
5. Calling `.parse()` on a subcommand.

This isn't technically wrong, but probably a mistake.

```
program.command('sub')
   .action(() => { console.log('Thanks for calling subcommand'); })
   .parse();
```

shadowspawn:
6. Enforce format of option flags.

```
program
   .option('-env') // oops, single dash means a short option which is a single character
   .option('-h, -?, -help'); // oops, short or long or both, but not more than one of each
```

shadowspawn:
Not much feedback on this. Numbers at 21 Jan 2024: 2 3 3 1 1 1

Adding check for conflicting commands (#2059) and options (#2055) in Command v12.

PR #2055, opened 2023-10-30
Throw error when add option with clashing flags

shadowspawn:
# Pull Request

## Problem

There are no warnings/errors for adding options with overlapping flags.

One of the "more usage errors" suggestion in a poll: https://github.com/tj/commander.js/issues/1978#issuecomment-1689391455

## Solution

Throw an error from `.option()` or `.addOption()` if there are clashing flags.

This is building on #1923, and adds tests.

Example:
```console
% node index.mjs
/Users/john/Documents/Sandpits/commander/my-fork/lib/command.js:552
      throw new Error(`Cannot add option '${option.flags}'${this._name && ` to command '${this._name}'`} due to conflicting flag '${matchingFlag}'
            ^

Error: Cannot add option '-e, --erase' to command 'main' due to conflicting flag '-e'
-  already used by option '-e, --example'
...
```

## ChangeLog

- _Breaking_: throw an error if add an option with flags which are already in use

### DIFF
--- a/lib/command.js
+++ b/lib/command.js
@@ -548,8 +548,9 @@ Expecting one of '${allowedValues.join("', '")}'`);
     const matchingOption = (option.short && this._findOption(option.short)) ||
       (option.long && this._findOption(option.long));
     if (matchingOption) {
-      throw new Error(`Cannot add option '${option.flags}'${this._name && ` to command '${this._name}'`} since an option using same flag has already been added
-- conflicts with option '${matchingOption.flags}'`);
+      const matchingFlag = (option.long && this._findOption(option.long)) ? option.long : option.short;
+      throw new Error(`Cannot add option '${option.flags}'${this._name && ` to command '${this._name}'`} due to conflicting flag '${matchingFlag}'
+-  already used by option '${matchingOption.flags}'`);
     }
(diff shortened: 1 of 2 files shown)


yoyo837:
Is it possible to provide an option to allow adding flags with already in use options.

shadowspawn:
There is not support for turning this off currently.

What is the use case @yoyo837 ? Why do you want to be able to add flags which are already in use?

yoyo837:
When I combine different npm scripts,When I combine different npm scripts to call multiple nodejs files, they can be run separately or in combination:

it looks something like this:
```json
{
  "scripts": {
    "s1": "npm run xxx --javaenv $javaenv",
    "s2": "npm run yyy --javaenv $javaenv",
    "s3": "npm run s1 && npm run s2 --javaenv $javaenv"
  }
}

```

shadowspawn:
Sorry, I don't understand how the npm script example relates to your question.

This PR makes it an error if the author has overlapping option flags, like this where two options both use `-a`:

```js
var program = new Command();
program
  .option('-a, --first')
  .option('-a, --second');
```

Is that what you want to turn off @yoyo837 ?

PR #2057, opened 2023-11-01
Change npm package checks in dependabot to monthly

abetomo:
# Pull Request

## Problem

Weekly updates were too frequent.

https://github.com/tj/commander.js/pull/2050

## Solution

Change to monthly.

(`github-actions` will remain weekly as it is updated infrequently.)

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -4,7 +4,7 @@ updates:
     target-branch: "develop"
     directory: "/"
     schedule:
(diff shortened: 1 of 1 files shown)


shadowspawn (approved):
We don't have people asking for more frequent updates, and the dependencies are all development ones so developers can update themselves. Weekly is proving a bit noisy, so I agree lets try monthly!

PR #2058, opened 2023-11-01
Remove Readme.md from bundle

### LINKED ISSUE #2056: Remove README.md from bundle
By NPM itself is impossible to remove, `pack` command will always include the `Readme.md` in the bundle, but maybe we can use another tool to pack or just try to reduce the size of the `Readme` by splitting the documentation in another file.

This change can save 25% of the bundle size, from 177kb to 134,7kb, saving up to 4,3TB/week of bandwidth.

H4ad:
Closes #2056 

This change can save 25% of the bundle size, from 177kb to 134,8kb (45.8kb to 33.1 gzip), saving up to ~1,3TB/week (108,256,856 * 12.7 kB / 1024/ 1024) of bandwidth.

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -1,7 +1,7 @@
 # Commander.js
 
 [![Build Status](https://github.com/tj/commander.js/workflows/build/badge.svg)](https://github.com/tj/commander.js/actions?query=workflow%3A%22build%22)
-[![NPM Version](http://img.shields.io/npm/v/commander.svg?style=flat)](https://www.npmjs.org/package/commander)
+[![NPM Version](https://img.shields.io/npm/v/commander.svg?style=flat)](https://www.npmjs.org/package/commander)
 [![NPM Downloads](https://img.shields.io/npm/dm/commander.svg?style=flat)](https://npmcharts.com/compare/commander?minimal=true)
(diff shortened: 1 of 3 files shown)


H4ad:
I think we should not remove it, but instead replace it with a cheaper version, removing it will probably make the page on NPM be empty.

shadowspawn:
For interest, a different approach is to use `.github/README.md` which is preferentially used by GitHub. Nice example project showing it in action: https://www.npmjs.com/package/@usulpro/readme-pro

H4ad:
But this only works if you want to publish using Github, I think.

At least, doing the test locally using `npm pack`, didn't work.

shadowspawn:
I don't want to make changes without a clearer idea of the benefits and downsides. "Saving bandwidth" is not enough of a reason on its own.

Is the npm distribution network asking for bundles to be made smaller? Are Commander users asking for the bundle to be made smaller? Are Commander users asking for the README to be split up or moved to its own website?

I will leave the associated issue open to see if it gets feedback from users (#2056), and the related issue on npm about a higher level solution has got some interesting responses (https://github.com/npm/rfcs/issues/744).

PR #2059, opened 2023-11-04
Prevent commands being created with overlapping names or aliases

shadowspawn:
# Pull Request

## Problem

Silently allow overlapping commands, such as two commands with same name, or two commands with same alias.

## Solution

Check names and aliases when adding new command or alias. Throw if overlaps with existing command.

See: #1903 #1924

Related PR for clashing options: #2055.

## ChangeLog

- _Breaking_: throw an error if add a command with name or aliases which are already in use


abetomo (approved):
+1

PR #2077, opened 2023-11-13
Update url per npm advice

shadowspawn:
When I did an `npm publish` with npm v10 it said I should run `npm pkg fix` to fix up the URL. This is the change that makes!


abetomo (approved):
👍

PR #2078, opened 2023-11-13
Add Removed section to Deprecated

shadowspawn:
# Pull Request

## Problem

The first deprecated feature to actually be removed was just mixed in with the rest.

## Solution

Add a `Removed` section, and rework title levels so is a section in ToC.


abetomo (approved):
👍

PR #2087, opened 2023-11-26
Refactor help command implementation to hold actual Command

shadowspawn:
# Pull Request

## Problem

In order of importance:

1) the other `addFoo` routines like `addCommand()` take objects
2) we have to do custom work and add more properties for new Command features, like say helpGroup (#1910)
3) bit messy storing multiple properties to construct help command when needed (minor)

## Solution

- make `.addHelpCommand()` accept a Command
   - so similar to `.addCommand()`
- add `.helpCommand()` which behaves similarly to the old API
   - so similar to `.command()`
- `.addHelpCommand()` still accepts string/boolean/undefined for backwards compatibility, but that usage is deprecated in favour of `.helpCommand()`

## ChangeLog

- added: `.helpCommand()` for configuring built-in help command
- changed: `.addHelpCommand()` now takes a `Command` for configuring built-in help command
- deprecated: `.addHelpCommand()` being passed string or boolean for configuring built-in help command

### DIFF
--- a/lib/command.js
+++ b/lib/command.js
@@ -72,9 +72,7 @@ class Command extends EventEmitter {
     this._helpShortFlag = '-h';
     this._helpLongFlag = '--help';
     this._addImplicitHelpCommand = undefined; // Deliberately undefined, not decided whether true or false
-    this._helpCommandName = 'help';
-    this._helpCommandnameAndArgs = 'help [command]';
-    this._helpCommandDescription = 'display help for command';
+    this._helpCommand = null;
     this._helpConfiguration = {};
   }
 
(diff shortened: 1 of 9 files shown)


lib/command.js
```diff
@@ -367,39 +364,76 @@ class Command extends EventEmitter {
...
-    } else {
-      this._addImplicitHelpCommand = true;
-      if (typeof enableOrNameAndArgs === 'string') {
-        this._helpCommandName = enableOrNameAndArgs.split(' ')[0];
-        this._helpCommandnameAndArgs = enableOrNameAndArgs;
-      }
-      this._helpCommandDescription = description || this._helpCommandDescription;
+  helpCommand(enableOrNameAndArgs, description) {
+    if (typeof enableOrNameAndArgs === 'boolean') {
+      this._addImplicitHelpCommand = !!enableOrNameAndArgs;
```
abetomo, line 381: Since enableOrNameAndArgs is a boolean, isn't `!!` unnecessary?

abetomo (approved):
LGTM

lib/command.js
shadowspawn: Good point! Done.

PR #2103, opened 2023-12-16
Improve jsdoc

shadowspawn:
# Pull Request

## Problem

I noticed JSDoc type unions are meant to have brackets around them:

> This means a value can have one of several types, with the entire list of types enclosed in parentheses and separated by `|`.

https://jsdoc.app/tags-type

## Solution

Add brackets. And fixed a few `String`/`string` too.

-----

Related: I noticed Visual Studio Code was not getting the union types consistently right, but I think that is related to TSDoc ignoring the JSDoc union types so not fixed by this change! 

> API Extractor does not support this `{(number|Array)}` notation, since its grammar is complex, and the type system it implements is fairly rudimentary compared to TypeScript's.

https://api-extractor.com/pages/tsdoc/tag_returns/

### DIFF
--- a/lib/command.js
+++ b/lib/command.js
@@ -52,7 +52,7 @@ class Command extends EventEmitter {
     this._enablePositionalOptions = false;
     this._passThroughOptions = false;
     this._lifeCycleHooks = {}; // a hash of arrays
-    /** @type {boolean | string} */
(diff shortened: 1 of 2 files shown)


abetomo (approved):
👍

Issue #2126, opened 2024-01-16
Rewrite error message log function


kolserdav:
Hi, Thanks for your great work!

I have a stylized logger but commander's error messages have an another view,  how can I rewrite error messages?

```sh
error: missing required argument 'name'
```
Maybe you'll could provide a way like this for example:
```javascript
const program = new Command();
program.errorLogHandler((data) => customLogFunction(data));
```

kolserdav:
https://github.com/tj/commander.js/pull/2127

shadowspawn:
There is support for customising the error display using `.configureOutput()` and supplying `writeErr`.

https://github.com/tj/commander.js#override-exit-and-output-handling

-----

You are actually using this as the implementation in the PR. What is the problem with using it directly to achieve your goals?

Hugos68:
@kolserdav I'm trying to overwrite `error: missing required argument <arg>` message but it does not seem to work?

```js
#!/usr/bin/env node
import { program } from 'commander';
import { migrate } from './commands/migrate/index.js';
import { intro, outro, log } from "@clack/prompts";
import {bgBlue, bgGreen, bgRed, black } from 'colorette';

program.name('Skeleton');
program.description('A CLI tool for Skeleton related functionality');
program.addCommand(migrate);


program.configureOutput({
	writeErr(str: string) {
		log.error(str);
	},
	writeOut(str: string) {
		log.info(str);
	}
});

program.exitOverride();

try {
	intro(bgBlue(black(` Skeleton CLI `)));
	program.parse(process.argv);
	outro(bgGreen(black(` Success `)));
} catch (e) {
	console.log(e);
	outro(bgRed(black(` Failed `)));
}
```

Output:

```bash
┌   Skeleton CLI 
error: missing required argument 'migration'
```

The `log.info` should output:
```bash
┌   Skeleton CLI 
│
■  This is an error message
``

**Comment by shadowspawn** [comment]
@Hugos68 

Firstly, If you want to make changes to the program that affect subcommands (Which is normal) then you configure the program before adding the subcommands. 

Secondly, there is an extra step when using `.addCommand()` to copy the settings from the program to the subcommands. 

So your code would look like:
```js
program.configureOutput({ ... });
program.addCommand(migrate);
migrate.copyInheritedSettings(program);
```

Hugos68:
@shadowspawn Thank you! The `migrate.copyInheritedSettings(progam)` fixed what I was struggling with.

Hugos68:
[...]

PR #2127, opened 2024-01-16
Add customErrorHandler with example

kolserdav:
## Problem

When error messages are not stylized as the other errors in application it worsens an user experience.

## Solution
It make possible to rewrite the standard error log method
Provided the method:
```typescript
customErrorHandler(callback: (data: string) => void): this;
```


kolserdav:
https://github.com/tj/commander.js/issues/2126

Issue #2163, opened 2024-03-01
Handle the case when node called with the "-e" argument


atleta:
When you call node with the -e argument and supply the script (commands) to execute on the command line then the script parameter will be missing from process.argv (at index 1) so automatic parsing (`{from: 'node'}`) will fail.

As far as I can see this case can be handled by looking into `process.execArgv` and checking for the presence of the `-e` or `--eval` option.

(One might use this option to run non-trivial scripts e.g. when calling a node process in a docker container.)

shadowspawn:
For interest, did you encountered this lack in use or noticed in code?

Yes, this is a case we could handle.

I wrote code for this in `parseargs` where the node REPL was a use case being considered:
- https://github.com/pkgjs/parseargs/pull/20

atleta:
I ran into this when I changed the way I called a script (running in a docker instance) from another app.

It took a bit of investigation to figure out what caused the problem as node is not my main development environment and I'd say, this definitely is a surprising behavior. Knowing what the problem is, the workaround is 2 lines of code but, I think, it's better to take care of it in a higher-level library, like this.

shadowspawn:
I have opened a PR which will autodetect `node --eval` and `node --print` when _no_ arguments are specified to `.parse()`. To avoid breaking existing code or getting weird test results with custom `argv` depending on how node is launched, `{ from: 'node' }` still assumes "normal" node conventions.

shadowspawn:
Released in v12.1.0

PR #2164, opened 2024-03-03
Add auto-detection of args when node evaluating script code on command-line

shadowspawn:
# Pull Request

## Problem

`program.parse()` will not detect the user arguments correctly when script supplied with `node --print "script..."` or `node --eval "script..."`.

See: #2163

```console
% node --eval 'const { program } = require("./index.js"); console.log(program.parse().args);' 1 2 3                                            develop
[ '2', '3' ]
```

## Solution

Look for the relevant options in `process.execArgv` to detect these special cases.

References:
- https://nodejs.org/docs/latest/api/process.html#processexecargv
- https://nodejs.org/docs/latest/api/cli.html#options

```console
% node --eval 'const { program } = require("./index.js"); console.log(program.parse().args);' 1 2 3                                  feature/node-repl
[ '1', '2', '3' ]
```

## ChangeLog

- add: auto-detect special node flags `node --eval` and `node --print` when call `.parse()` with no arguments


shadowspawn:
Thinking the "node" mode could be smart about eval/print too, since it is a node behaviour. And being explicit in the README that no arguments is the auto mode,

shadowspawn:
I decided I am happy with the full-auto just when Commander is deciding everything!

Supplied arguments means we don't know where they came from, so play safe and assume normal node.

Did big rework of docs. Might be overkill for the JSDoc/TSDoc, but parse is a fundamental method to get right.

abetomo (approved):

PR #2153, opened 2024-02-16
ESLint changes preparing for Prettier

shadowspawn:
# Pull Request

## Problem

ESLint have deprecating the formatting rules, and suggest using a separate formatter.

Using StandardJS and TypeScript equivalent has been painful to maintain. (To be fair, partly because of how we ended up using it.)

See: #2104

## Solution

Use ESLint for linting. Start with ESLint recommended rules.

Use Prettier for formatting. In a way this is quite like using Standard, opinionated so less setup and debate over settings! Installed and configured in this PR, but not used to format files yet to keep the PR focused on the setup changes.

I did a rework of the naming of the npm scripts so now `check` lint/format/type, and `fix` lint/format.

Note: I saw there was a new [flat config](https://eslint.org/blog/2022/08/new-config-system-part-2/) for ESLint and liked the look of it. Flat config is going to be default in ESLint 9. I used the new config file name `eslint.config.js` so detected as flat from command-line with ESLint 8.

Note: `eslint-plugin-jsdoc` installed but not activated yet, as will complain about multiple issues. Have also installed `prettier-plugin-jsdoc` which will hopefully fix some of the lint issues! Will evaluate those in separate PR.

For interest, I found a good blog after doing most of the work: https://www.joshuakgoldberg.com/blog/configuring-eslint-prettier-and-typescript-together/

PR #2167, opened 2024-03-08
Reformat files with Prettier

shadowspawn:
For convenient viewing, this is what the file changes are from running Prettier using the setup in #2153

```
npm run fix:format
```


shadowspawn:
Will open another PR using the configuration which has landed.

PR #2168, opened 2024-03-12
Use standard import in examples

shadowspawn:
It turns out we don't need to use `'../'` for the imports in the example files to be able to run the examples from the repo. Can just import from `'commander'` like in client code and finds the parent package. Small change, but nice tidy!


abetomo (approved):
+1

Fariborzfofo (approved):

PR #2170, opened 2024-03-31
Fix: Use `node`-prefixed requires for builtins

### LINKED ISSUE #2169: Neutral JavaScript runtime support (Deno, Bun, etc)
Hey there @tj,

First of all, big fan of `commander`; I've used it for a long time, so, longtime user first time contributor 😁. I have a patch for `commander` which makes it work neutrally on all JS runtimes; it was already working fine on Bun and Node, but with the following two small changes:

- Prefixing Node API imports with `node:*`
- Importing `process` into an explicit symbol

... it will now run smoothly on Deno as well.

For context: I'm trying to ship a tool downstream, [`hashlock`](https://github.com/sgammon/hashlock), which uses commander, but I want it to be able to run on any JS runtime.

**Distribution diff:**
```diff
diff --git a/node_modules/commander/lib/command.js b/node_modules/commander/lib/command.js
index 5b16e60..e1ddedb 100644
--- a/node_modules/commander/lib/command.js
+++ b/node_modules/commander/lib/command.js
@@ -1,8 +1,8 @@
-const EventEmitter = require('events').EventEmitter;
-const childProcess = require('child_process');
-const path = require('path');
-const fs = require('fs');
-const process = require('process');
+const EventEmitter = require('node:events').EventEmitter;
+const childProcess = require('node:child_process');
+const path = require('node:path');
+const fs = require('node:fs');
+const process = require('node:process');
 
 const { Argument, humanReadableArgName } = require('./argument.js');
 const { CommanderError } = require('./error.js');
```

PR incoming shortly.

<em>This issue body was [partially generated by patch-package](https://github.com/ds300/patch-package/issues/296).</em>

sgammon:
# Pull Request

## Problem

I want to use a Commander-based CLI tool from Deno, as explained in #2169 

## Solution

Fixes and closes #2169 by using Node-prefixed imports.

**Note:** Node-prefixed imports are supported on Node 16+, so this may break older apps on _very_ old versions of Node. As a result I would suggest a major version bump to signal a compatibility change. Node 16 has reached EOL, so people really should be upgrading if they can.

No code was changed in Commander itself; just imports.

## ChangeLog

- Use `node:`-prefixed requires; this may break compatibility with older versions of Node (up to 15). Please upgrade Node, bundle your app with rewriting to remove the `node:` prefix on requires, or avoid upgrading Commander if Node cannot be upgraded.

### DIFF
--- a/lib/command.js
+++ b/lib/command.js
@@ -1,8 +1,8 @@
-const EventEmitter = require('events').EventEmitter;
-const childProcess = require('child_process');
-const path = require('path');
-const fs = require('fs');
-const process = require('process');
+const EventEmitter = require('node:events').EventEmitter;
+const childProcess = require('node:child_process');
+const path = require('node:path');
+const fs = require('node:fs');
+const process = require('node:process');
 
 const { Argument, humanReadableArgName } = require('./argument.js');
 const { CommanderError } = require('./error.js');


lib/command.js
```diff
@@ -1,8 +1,8 @@
-const EventEmitter = require('events').EventEmitter;
-const childProcess = require('child_process');
-const path = require('path');
-const fs = require('fs');
-const process = require('process');
+const EventEmitter = require('node:events').EventEmitter;
+const childProcess = require('node:child_process');
+const path = require('node:path');
+const fs = require('node:fs');
+const process = require('node:process');
```
sgammon, line 5: the whole pr

shadowspawn:
Commander v12 requires node v18 or higher, so node compatibility for using the `node:` protocol is not an issue.

https://github.com/tj/commander.js/blob/83c3f4e391754d2f80b179acc4bccc2d4d0c863d/package.json#L77

shadowspawn:
This issue only shows up because of bundling then running with Deno, and the change to fix it is only likely to break similar workflows involving transpiling and bundling tools not supporting the "node:" syntax.

The `node:` prefix for core modules has been available in Node.js since (at least) v16.0.0
- https://nodejs.org/docs/latest/api/modules.html#core-modules
- https://nodejs.org/docs/latest/api/esm.html#node-imports

So I think normal uses of Commander won't be affected by this change, and if it breaks some more exotic uses then we will roll it back.

shadowspawn (approved):
LGTM

abetomo (approved):
👍

shadowspawn:
Released in v12.1.0

sgammon:
@shadowspawn Thank you for the ping! I'll go ahead and adopt. 🥳

PR #2176, opened 2024-04-02
Add support for help groups for options

shadowspawn:
# Pull Request

## Problem

See: #1897
Related PR: #1908 #1910

## Solution

This is a proof of concept adding a map of group titles to the Help configuration. This sort of mapping has some advantages compared with adding a group property to each Option:
- specify group title once instead of multiple times
- support for grouping `help` and `version` without needing to create custom Options
- support for grouping user-created options without needing to create custom Options
- (for Command groups, would support putting external commands into groups without extra work)

However, directly supplying the map does not feel very Commander-like. Adding a method on Command to add a single option group at a time might be more in the chaining style while retaining the practical advantages. Adding a group property to the Option would be more object-oriented, and other libraries allow this (https://github.com/tj/commander.js/issues/1897#issuecomment-1610897543).

Playing with the idea partly to think about how Help could be refactored to make this more feasible to implement with an author supplied override.

Example use:
```js

program
  .option('-c, --carbon')
  .option('-r, --rabbit', 'cudly little friends')
  .option('-o, --oxygen')
  .addOption(new Option('-s, --sheep'))
  .option('--no-sheep') // negated
  .option('-n', 'neon') // short
  .option('--armadillo') // long
  .option('--zzz')
  .option('--dog', 'faithful furry companions')
  .option('--no-dog')

program.configureHelp({
  optionGroups: {
    'Animal Options:': ['rabbit', 'armadillo', 'sheep', 'dog'],
    'Element Options:': ['carbon', 'oxygen', 'n' /* neon */],
    'Help Options:': ['help']
    }
  }),

program.parse();
```

```console
% node map2-options.js --help
Usage: map2-options [options]

Options:
  --zzz

Animal Options:
  -r, --rabbit  cudly little friends
  -s, --sheep
  --no-sheep
  --armadillo
  --dog         faithful furry companions
  --no-dog

Element Options:
  -c, --carbon
  -o, --oxygen
  -n            neon

Help Options:
  -h, --help    display help for command
```

### DIFF
--- a/lib/help.js
+++ b/lib/help.js
@@ -15,6 +15,7 @@ class Help {
     this.sortSubcommands = false;
     this.sortOptions = false;
     this.showGlobalOptions = false;
+    this.optionGroups = {};
   }
 
   /**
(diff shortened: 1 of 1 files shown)


shadowspawn:
Decent proof of concept. Still wishing was easier for authors to customise the behaviour in large `formatHelp` without copying.

A defect in this PR is using `.configureHelp()` in a subcommand to get grouping will replace (lose) other changes made at root level.

PR #2182, opened 2024-04-06
Refactor eslint configs inside tseslint.config

shadowspawn:
This is a minor refactor. I didn't fully appreciate `tseslint.config()` first time around, and just used it for the tseslint related configs. In fact `tseslint.config()` adds some type safety to all the configs and typescript-tslint recommends using it.


abetomo (approved):
👍

PR #2183, opened 2024-04-07
Show `defaultValueDescription` even if `defaultValue` is not set

davidyuk:
# Pull Request

## Problem

I have an option that gets passed to the underlying library. I want to describe its default behaviour, but I can't pass any default value there except `undefined`. In case `undefined` the default value description is not printed in help.

```js

const ttl = new Option('--ttl [ttl]', 'Transaction can be mined before provided height')
  .default(undefined, 'current height increased by 3');

program.addOption(ttl);
program.parse(process.argv);
```

Output:
```
$ node ./test.js --help
Usage: test [options]

Options:
  --ttl [ttl]  Transaction can be mined before provided height
```

## Solution

I'm proposing to print `defaultValueDescription` even if it is provided without `defaultValue`. Though I'm not sure that it is not a misuse of "default value" concept.

Expected output:
```
$ ./test.js --help
Usage: test [options]

Options:
  --ttl [ttl]  Transaction can be mined before provided height (default: current height increased by 3)
```

## ChangeLog

Added: show `defaultValueDescription` even if `defaultValue` is not set

This PR is supported by the Æternity Foundation

### DIFF
--- a/lib/help.js
+++ b/lib/help.js
@@ -302,7 +302,10 @@ class Help {
         `choices: ${option.argChoices.map((choice) => JSON.stringify(choice)).join(', ')}`,
       );
     }
-    if (option.defaultValue !== undefined) {
+    if (
+      option.defaultValue !== undefined ||
+      option.defaultValueDescription !== undefined
+    ) {
       // default for boolean and negated more for programmer than end user,
       // but show true/false for boolean option as may be for hand-rolled env or config processing.
       const showDefault =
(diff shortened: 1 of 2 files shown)


shadowspawn:
Nice tidy PR.

It does feel enough like a misuse of "default value" concept that I am reluctant to add this.

> ... I can't pass any default value there except undefined.

Are you able to put in a placeholder default value (like `false` or `null` or even a symbol) and fix it up before passing to the underlying library? I appreciate this means you need to add an extra line of code in your code to do the work, so not free.

---

For example:

```js

const ttl = new Option('--ttl [ttl]', 'Transaction can be mined before provided height')
  .default(false, 'current height increased by 3');

program.addOption(ttl);
program.parse(process.argv);
if (program.opts().ttl === false) program.setOptionValue('ttl', undefined); // replace placeholder value
console.log(program.opts());
```

```console
% node index.js --help
Usage: index [options]

Options:
  --ttl [ttl]  Transaction can be mined before provided height (default: current height increased by 3)
% node index.js       
{ ttl: undefined }
```

davidyuk:
This workaround worked in my case, thank you! You may close this PR

PR #2196, opened 2024-05-18
Add sections to allow some customisation of help layout

shadowspawn:
# Pull Request

I wanted to give this a try, but not sure it adds enough value or is easy enough to use. Opening a draft with proof of concept.

## Problem

It isn't convenient to customise the order of sections in the Help.

See #1870

## Solution

Create "sections" in `formatHelp` and call `visibleSections` which can be overridden to customise the list. Some possible use cases:

- reorder sections
- hide sections
- alter section titles, like make bold or all capitals
- add own (pre-formatted) sections, like "Environment Variables" or "Examples"
- (maybe) replace single-line Usage with multi-line usages

Example client code to put options after commands:

```js
program.configureHelp({
  visibleSections: (sections) => {
    // Hide empty sections.
    const visible = sections.filter((section) => !!section.body);
    // Reorder sections.
    const order = ['usage', 'description', 'arguments', 'commands', 'options', 'globalOptions'];
    return visible.sort((a, b) => {
      return order.indexOf(a.name) - order.indexOf(b.name);
    });
  }
});
```

### DIFF
--- a/lib/help.js
+++ b/lib/help.js
@@ -360,6 +360,21 @@ class Help {
   }
 
   /**
+   * visibleSections
+   *
+   * @typedef {object} Section
+   * @property {string} name
+   * @property {string} title - may be empty
+   * @property {string} body
+   *
+   * @param {Section[]} sections
+   * @returns
+   */
+  visibleSections(sections) {
+    return sections.filter((section) => !!section.body);
+  }
+
+  /**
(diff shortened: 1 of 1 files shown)


shadowspawn:
Not enough demand or clarity that solving a popular problem. Closing for now.

PR #2197, opened 2024-05-19
set `this`context in action function

### LINKED ISSUE #2145: Missing `this` type for `action`
https://github.com/tj/commander.js/blob/83c3f4e391754d2f80b179acc4bccc2d4d0c863d/lib/command.js#L511-L527
The passed `action` function `fn` receives the current command as `this` context in line 523. However, this context is missing in the current typings.
https://github.com/tj/commander.js/blob/83c3f4e391754d2f80b179acc4bccc2d4d0c863d/typings/index.d.ts#L525

Happy to provide a PR for this!

matthyk:
Closes #2145 by adapting from commander-js/extra-typings#59

### DIFF
--- a/typings/index.d.ts
+++ b/typings/index.d.ts
@@ -544,7 +544,7 @@ export class Command {
    *
    * @returns `this` command for chaining
    */
-  action(fn: (...args: any[]) => void | Promise<void>): this;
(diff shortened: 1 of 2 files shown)


shadowspawn (approved):
LGTM

PR #2210, opened 2024-06-09
Feature/run all checks

shadowspawn:
# Pull Request

## Problem

Was not running all tests after finishing the eslint and Prettier work.

## Solution

Update GitHub action to run all `npm run check`.

This picks up a file in a recent PR which did not get Prettier applied.

NB: tests will currently fail due to issue fixed in #2209

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,2 @@
+# disable line-ending changes (otherwise Windows developers with default settings from git install see Prettier wanting to fix every file)
+* -text
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -27,6 +27,5 @@ jobs:
       run: npm ci
     - name: npm test
       run: npm test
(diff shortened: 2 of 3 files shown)


shadowspawn:
Format tests failing in Windows, which implies running Prettier on Windows produces different results than on Mac/Linux. That isn't good, and seems a little surprising.

I'll have a look on Windows and see what I can see!

abetomo (approved):

XhmikosR:
@shadowspawn just a hint in case you weren't aware of, it's better to use:

```
# Enforce Unix newlines
* text=auto eol=lf
```

This way, the line endings are normalized to LF for everyone regardless of their OS or git settings.

shadowspawn:
The rationale for using `-text` is for git to leave alone the line endings, which are specified in `.editorconfig` and enforced by prettier. Specified in one place instead of two.

PR #2214, opened 2024-06-27
Variadic arguments: fix "--" processing

Totto16:
# Pull Request

This PR fixes the behavior of variadic arguments in the context of the usage of "--"

## Problem

When passing arguments `["a","b","c","--","ERROR"]` to a variadic argument, it should only include `["a","b","c"]` and not the string `"ERROR"` at the end.

## Solution

Since this package is really complex, I just added a property to the `Command` class.
```js
   /** @type {(undefined | number)} */
    this.lastArgIndex = undefined

```
This keeps track of the "--" location, aka. the length of valid arguments. I adjusted the JSDoc types everywhere and also wrote two tests, that failed before this fix, expecting the correct behavior. They pass after this fix. 

I adjusted 3 tests, that tested the raw parser, where `lastArgIndex` was a number, so that it matches the current behavior.

The way I solved this, seems the easiest, and correctest way of doing this IMO. It also assures, that `parsed.processedArgs` and `parsed.args` is correct in all cases

## ChangeLog

Variadic Argument: don't add arguments after "--" to the list of `processedArgs`

### DIFF
--- a/lib/command.js
+++ b/lib/command.js
@@ -55,6 +55,8 @@ class Command extends EventEmitter {
     /** @type {(boolean | string)} */
     this._showHelpAfterError = false;
     this._showSuggestionAfterError = true;
+    /** @type {(undefined | number)} */
+    this.lastArgIndex = undefined
 
     // see .configureOutput() for docs
     this._outputConfiguration = {
@@ -1347,7 +1349,12 @@ Expecting one of '${allowedValues.join("', '")}'`);
       if (declaredArg.variadic) {
         // Collect together remaining arguments for passing together as an array.
         if (index < this.args.length) {
-          value = this.args.slice(index);
+          if(this.lastArgIndex === undefined){
+            value = this.args.slice(index);
+          }else{
+            value = this.args.slice(index, this.lastArgIndex);
+          }
+
           if (declaredArg.parseArg) {
             value = value.reduce((processed, v) => {
               return myParseArg(declaredArg, v, processed);
@@ -1452,6 +1459,7 @@ Expecting one of '${allowedValues.join("', '")}'`);
     operands = operands.concat(parsed.operands);
     unknown = parsed.unknown;
     this.args = operands.concat(unknown);
+    this.lastArgIndex = parsed.lastArgIndex
 
     if (operands && this._findCommand(operands[0])) {
       return this._dispatchSubcommand(operands[0], operands.slice(1), unknown);
@@ -1638,13 +1646,14 @@ Expecting one of '${allowedValues.join("', '")}'`);
    *     sub -- --unknown uuu op => [sub --unknown uuu op], []
    *
    * @param {string[]} argv
-   * @return {{operands: string[], unknown: string[]}}
+   * @return {{operands: string[], unknown: string[],lastArgIndex:undefined | number}}
    */
 
   parseOptions(argv) {
     const operands = []; // operands, not options or values
     const unknown = []; // first unknown option and remaining unknown args
     let dest = operands;
+    let lastArgIndex = undefined;
     const args = argv.slice();
 
     function maybeOption(arg) {
@@ -1659,6 +1668,7 @@ Expecting one of '${allowedValues.join("', '")}'`);
       // literal
       if (arg === '--') {
         if (dest === unknown) dest.push(arg);
+        lastArgIndex = dest.length;
         dest.push(...args);
         break;
       }
@@ -1764,8 +1774,7 @@ Expecting one of '${allowedValues.join("', '")}'`);
       // add arg
       dest.push(arg);
     }
-
-    return { operands, unknown };
+    return { operands, unknown, lastArgIndex };
   }
 
   /**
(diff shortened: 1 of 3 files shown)


shadowspawn:
> When passing arguments `["a","b","c","--","ERROR"]` to a variadic argument, it should only include `["a","b","c"]` and not the string `"ERROR"` at the end.

Why do you think it should not include the following arguments? Do you have examples of utilities or other libraries that do this? Do you have a particular use-case in mind that you wanted this behaviour?

Commander is behaving as intended. The `--` does gets used for other things by some utilities and libraries, but the primary use is to stop following arguments from being parsed as options.

The POSIX guidelines describe much of the behaviour you see in utilities on command line.

- https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap12.html

> Guideline 10:
> The first -- argument that is not an option-argument should be accepted as a delimiter indicating the end of options. Any following arguments should be treated as operands, even if they begin with the '-' character.

A very common use of `--` is when adding option-like arguments that are meant to be passed through to another command:

```
npm run --npm-option -- example --example-option
```

shadowspawn:
> Why do you think it should not include the following arguments? Do you have examples of utilities or other libraries that do this? Do you have a particular use-case in mind that you wanted this behaviour?

Rereading this, my questions sound a bit harsh, sorry if you thought that. Genuine questions to understand what you expected and why.

Totto16:
I didn't find it harsh, it's a justified question, since I didn't explain that. 😓

So let me explain, what the expectations are, there are certainly multiple solutions to this, but this solution seems the best imo.

`"--"` stops options processing. Which is fine. But if I want to have a variadic argument, but limit it somehow, to add some other things at the end, I expect the `"--"` to be either: 
a) limit the argument parsing as well
b) also be in the final list of arguments, so that it can be handled

In practice, this means this:
a)
Let's say the js is like this:
```js
const commander = require('.');

program
  .addArgument(new commander.Argument('<id>'))
  .option('-p', 'add pepper')
  .argument('[variadicArg...]');
const extraArguments = ['a', 'b', 'c'];

program.parse(['node', 'test', 'id', '-p', ...extraArguments, '--', 'ERROR']);

console.log(program.processedArgs, program.opts());

```

What I wanted to have is the option, to get these three things:

argument `id`
variadic argument `variadicArg`
options

Like e.g. `[ 'id', [ 'a', 'b', 'c' ] ] { p: true }`

but also `[ 'id', [ 'a', 'b', 'c' ´,'--', 'ERROR'] ] { p: true }`

would work, since Then I can manually remove everything after the "--" of the `variadicArg` list.

The main problem is, that with the current behaviour I get this:

`[ 'id', [ 'a', 'b', 'c', 'ERROR' ] ] { p: true }`

And there  the `"--"` is removed from the variadic Argument, so I have no idea, that `"--"` was passed by the user, and that it just got excluded, There is no access to the "raw" arguments (without options) , other than the real raw process.argv, and parsing that manually is what this lib does.

So the `"--"` only has a special meaning in the case of options, but by excluding it from the `program.processedArgs` you give it a special meaning in the arguments as well, which is, ignoring it and doing nothing special with it. A better behavior would be, to include it in variadic arguments, which I find an inferior solution, or just do the same as with options, everything after "--" is excluded.


The link from you above describes options, but argument processing isn't described there, at least not the form in which this library implements it. 

There might be uses cases, where variadic arguments also have to be limited by `"--"` So i find this feature valid.

I hope My problem and explanation is clear, otherwise don't hesitate to question it.

shadowspawn:
Option d looks like it adds an extra argument to the action handler before the options if there are post-variadic args. That is a no-no as it would break anyone not detecting the variant calling signature. I did have an extra argument tagged on the end for "extra" arguments at one point, but that has been removed. 

The existing behaviour which treats `--` as the option terminator needs to stay the same for existing code. That looks like:

```js
program
  .option('-p', 'add pepper')
  .argument('<id>')
  .argument('[variadicArgs...]')
  .action((id, args, options) => {
    console.log({ id, args, options})
  });
```

```console
% node index.mjs -p ID a b c                
{ id: 'ID', args: [ 'a', 'b', 'c' ], options: { p: true } }
% node index.mjs -p -- ID a b c --NOT-OPTION
{
  id: 'ID',
  args: [ 'a', 'b', 'c', '--NOT-OPTION' ],
  options: { p: true }
}
% node index.mjs -p ID a b c -- --NOT-OPTION
{
  id: 'ID',
  args: [ 'a', 'b', 'c', '--NOT-OPTION' ],
  options: { p: true }
}
```

shadowspawn:
I think a simple work-around might be to use a different sequence for your special behaviour, like say `---` or `//`. Then the arguments to parse could be pre-processed to strip out the "extra" arguments, or the variadic array from the Commander parsing could be post-processed.

shadowspawn:
For interest, [yargs-parser](https://github.com/yargs/yargs-parser) has configuration for `populate--` to store the arguments after `--` in a separate array:
- https://github.com/yargs/yargs-parser#populate---

[minimist](https://github.com/minimistjs/minimist) shares history with yargs and has `opts['--']` configuration to do the same thing.

Totto16:
Ok thank you for your tips ❤️ 
, so this PR is bogus, so I'm closing it

Issue #2149, opened 2024-02-08
[Bug] Unexpected arguments ignored


KaKi87:
Hi,

When passing more arguments to a command than it expects, the command is executed without error, receiving only the expected arguments.

It would be nice to throw an error instead, preventing potantially wrong actions/results unforeseen by the user.

One particular case is when an argument is a name/title/label, in which the user may include spaces while forgetting quotes.

On the other hand, I wonder if it would be possible, when said argument is the last (or only) one, to *optionally* put everything automatically in it, kinda like the spread operator in JS (but still outputting a string) ?

Thanks

shadowspawn:
> It would be nice to throw an error instead, preventing potantially wrong actions/results unforeseen by the user.

You can make this an error by calling `.allowExcessArguments(false)`.

> On the other hand, I wonder if it would be possible, when said argument is the last (or only) one, to optionally put everything automatically in it, kinda like the spread operator in JS (but still outputting a string) ?

You can explicitly collect the remaining arguments by declaring the last one as variadic like `.argument('[more-args...]').

Also, all of the arguments are available using `cmd.args`. This includes both the declared and the excess arguments.

KaKi87:
Oh, I'm sorry I didn't find these beforehand. 😅

> `allowExcessArguments(false)`

I'd like to request this to be default though, for security reasons.

Thanks

shadowspawn:
I originally intended excess arguments to be an error by default, but made it opt-in for the first release to reduce breakage at the time:

- https://github.com/tj/commander.js/pull/1399#issuecomment-752749726
- #1429

On a related note, there was an in-depth discussion during the development of `parseArgs` about unexpected positionals and we eventually decided to make positionals opt-in for strict mode (a shout-out to the thoughtful and patient championing by @bakkot). The extra step in `parseArgs` to allow positionals has not caused any user feedback that I am aware of.

- https://github.com/pkgjs/parseargs/issues/85

I am leaning towards making `allowExcessArguments` false by default in next major version of Commander.

shadowspawn:
A prerelease is available for v13. The release is tagged as next and can be installed with:

    npm install commander@next

shadowspawn:
Version 13.0.0 has been released

PR #2223, opened 2024-07-07
Disallow excess arguments

shadowspawn:
## Problem

By default, if is not an error to pass excess command-arguments. This may sometimes be as intended but
will often mean a command-line error by the user will be silently ignored. A nice example from #2149 

> One particular case is when an argument is a name/title/label, in which the user may include spaces while forgetting quotes.

I originally intended excess arguments to be an error by default, but made it opt-in for the first release to reduce breakage at the time:

- https://github.com/tj/commander.js/pull/1399#issuecomment-752749726
- #1429

See: #2149

## Solution

- Make `allowExcessArguments` false by default. 
- Fix lots of broken tests where passing undeclared arguments!
- Add migration tips to CHANGELOG for the breaking change.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,54 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 <!-- markdownlint-disable MD024 -->
 <!-- markdownlint-disable MD004 -->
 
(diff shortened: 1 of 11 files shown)


abetomo (approved):

shadowspawn:
Released in Commander v13.0.0

Issue #2236, opened 2024-08-21
[Bug] Wrong types for defaultValue


Sadzurami:
Example:
```ts
import { program } from 'commander';
program.option('-n, --num', 'test num value', 10)
```

Result:
```sh
No overload matches this call.
  Overload 1 of 3, '(flags: string, description?: string, defaultValue?: string | boolean | string[]): Command', gave the following error.
    Argument of type '10' is not assignable to parameter of type 'string | boolean | string[]'.
  Overload 2 of 3, '(flags: string, description: string, parseArg: (value: string, previous: unknown) => unknown, defaultValue?: unknown): Command', gave the following error.
    Argument of type 'number' is not assignable to parameter of type '(value: string, previous: unknown) => unknown'.
  Overload 3 of 3, '(flags: string, description: string, regexp: RegExp, defaultValue?: string | boolean | string[]): Command', gave the following error.
    Argument of type 'number' is not assignable to parameter of type 'RegExp'.
```

shadowspawn:
The TypeScript types are deliberately stricter than the JavaScript implementation to help guide users to safer patterns by default. In this case if this was silently allowed, you would have your option being a `number` if it came from the default and a `string` if from the parsed arguments.

See #1792

shadowspawn:
The Pull Request that removed multiple uses of `any` including for the `.option()` default value: #1119

shadowspawn:
In many cases, you can get the type you want _and_ strict types by also specifying a custom [option processing function](https://github.com/tj/commander.js?tab=readme-ov-file#custom-option-processing).

If that isn't a match for what you want you can also use the `Option` class directly and it allows any type in the `.default()` call:

```
.addOption(new Option('-n, --num <value>', 'test num value').default(10));
```

PR #2237, opened 2024-08-21
fix: update types for defaultValue

Sadzurami:
## Problem

Wrong typings for `defaultValue`
ref: https://github.com/tj/commander.js/issues/2236

## Solution

Force `defaultValue` to `any` type. 
Commander actually don't check types at code (except `undefined`) when use value.

We see that the tests pass successfully with types other then defined now.
https://github.com/tj/commander.js/blob/4e87829ba3d41dcb6109dea9899bf9a7e8f44cda/tests/argument.custom-processing.test.js#L52-L62

### DIFF
--- a/typings/index.d.ts
+++ b/typings/index.d.ts
@@ -411,7 +411,7 @@ export class Command {
     fn: (value: string, previous: T) => T,
     defaultValue?: T,
   ): this;
-  argument(name: string, description?: string, defaultValue?: unknown): this;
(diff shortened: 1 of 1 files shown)


shadowspawn:
I am aware that the JavaScript allows any type, but we are stricter in the TypeScript. 

See similar closed PR https://github.com/tj/commander.js/pull/1792

And the Pull Request that removed multiple uses of `any` including for the `.option()` default value: https://github.com/tj/commander.js/pull/1119

Issue #2068, opened 2023-11-08
--no is bugged with default arguments


Zamiell:
Hello, and thanks for the useful library.

Consider the following program:

```ts
const { program } = require('commander');

program.option('--no-foo', true)
program.option('--no-bar', false)
program.parse();

const options = program.opts();
console.log(options)
```

Running this program will result in the following output:

```
{ foo: true, bar: true }
```

This is a bug, because commander.js is not using the default value that was specified. Specifically, I would expect specifying `true` for the default value of `--no-foo` would mean that `false` should be the default value of `--foo`, which means that the correct output should be:

```
{ foo: false, bar: true }
```

Additionally, I will say that I do not like the special/magic behavior of commander.js to invert my `--no` arguments. It makes for a more confusing code base, *especially* when bugs like the above exist! Is there some way to disable this special/magic behavior?

shadowspawn:
Thanks for the example code. 

You are missing the description argument in your example, so the `true`/`false` are being read as a description and not a default value.

Try:
```
program.option('--no-foo', 'foo description', true)
program.option('--no-bar', 'bar description', false)
```

>  Is there some way to disable this special/magic behavior?

No, it is a built-in feature which can not be disabled.

(I think you can flip the results by specifying both a default value and a preset value, but I suggest it is not worth the effort!)

Zamiell:
Sorry about that. With the following program:

```js
const { program } = require('commander');

program.option('--no-foo', 'foo description', true)
program.option('--no-bar', 'bar description', false)
program.parse();

const options = program.opts();
console.log(options)
```

I get the following output:

```
{ foo: true, bar: false }
```

However, as I said above, this output is a bug, because the following line:

```js
program.option('--no-foo', 'foo description', true)
```

When translated to English, this line means: "By default, we want there to be no foo."
Thus, the output should be:

```
{ foo: false, bar: true }
```

I understand that Commander.js has magic behavior to transform `no-foo` --> `foo`, but this special behavior should also properly flip the default argument as well.

shadowspawn:
> I understand that Commander.js has magic behavior to transform no-foo --> foo, but this special behavior should also properly flip the default argument as well.

Here is another way of describing it. Adding an option with flags `--no-foo` creates an option named `foo`. Specifying a default value sets the starting value of option `foo`, just like with any other option.

I see your interpretation, but it isn't the only interpretation and I don't think is worth considering changing the behaviour. Not least because specifying a default value for a negated option is probably rare.

[...]

Issue #1100, opened 2019-11-16
How to disable '--no-foo' (negatable boolean and flag|value) features


snowman:
I write an program which wrap around git command, in which runs `git diff --no-index` command.

```javascript
const program = require('commander');

program
  .version('0.0.7')
  .usage('[options] [<commit>] [--] [<path>...]')
  .option('--cached', 'show diff of staging files')
  .option('--no-index', 'compare two paths on the filesystem')
  .parse(process.argv);

const options = ['--patience'];
if (program.cached) {
  options.push('--cached');
}
if (!program.index) {        // instead of using `program.noIndex"
  options.push('--no-index')
}
```

shadowspawn:
There is not a way to disable the negatable boolean feature. Your program looks as expected.

The negatable feature looks for an option name with a leading `no-` and does not trigger if the option is named differently. So you can get the code to have `noIndex` but only by using a matching command line option. (I am not suggesting you would want to do this, just mentioning the possibility.) For example:

```
  .option('--noIndex', 'compare two paths on the filesystem')
```

```
if (program.noIndex) {
  options.push('--no-index')
}
```

PR #2238, opened 2024-08-27
Allow bypassing negation

bernard-wagner:
# Pull Request

## Problem

Sometimes it is desired to have an option starting with `--no` to be treated as a regular boolean argument. I encountered this when writing a wrapper for another command and needing to pass the options as is to the sub-process.

## Solution

Since `negate` will be `true` for any option starting with `--no`, the default behavior can be overridden by explicitly changing `negate` to `false`.

### DIFF
--- a/tests/options.bool.test.js
+++ b/tests/options.bool.test.js
@@ -32,13 +32,11 @@ describe('boolean flag on program', () => {
     expect(program.opts().cheese).toBe(false);
   });
 
-  test('when bypassing negatable then value is true', () => {
+  test('when negatable boolean flag specified then negative value is undefined', () => {
     const program = new commander.Command();
-    var option = program.createOption('--no-cheese', 'remove cheese');
-    option.negate = false;
-    program.addOption(option);
+    program.option('--no-cheese', 'remove cheese');
     program.parse(['node', 'test', '--no-cheese']);
-    expect(program.opts().noCheese).toBe(true);
+    expect(program.opts().noCheese).toBeUndefined();
   });
 });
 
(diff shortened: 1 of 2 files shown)


shadowspawn:
There have been some requests whether it is possible to disable the negate support: #1100 #2068

tests/options.bool.test.js
```diff
@@ -31,6 +31,15 @@ describe('boolean flag on program', () => {
     program.parse(['node', 'test', '--no-cheese']);
     expect(program.opts().cheese).toBe(false);
   });
+
+  test('when bypassing negatable then value is true', () => {
```
shadowspawn, line 35:
```suggestion
  test('when override negatable and flag specified then value is true', () => {
```

shadowspawn (changes requested):
I would like a couple more tests, but I can add these myself later if you prefer.

I would like a test that when the overridden flag is not specified that the option value is undefined rather than true.

And an explicit test, although redundant, that the name of the stored option is just camel case. So see `noCheese` in options and `cheese' is undefined. This one is implicitly tested already, but since it is a key change I would like an explicit test covering this particular behaviour.

bernard-wagner:
> I would like a couple more tests, but I can add these myself later if you prefer.
> 
> I would like a test that when the overridden flag is not specified that the option value is undefined rather than true.
> 
> And an explicit test, although redundant, that the name of the stored option is just camel case. So see `noCheese` in options and `cheese' is undefined. This one is implicitly tested already, but since it is a key change I would like an explicit test covering this particular behaviour.

I think the tests cover all the permutations now.

shadowspawn:
Thanks for all the tests for a small code change! I did need to search the code a couple of times to check the one-liner was really going to work, so the tests are good for confirming and also catching future regressions.

One more variation. You added this:

```
  test('when negatable boolean flag is specified and negate is overridden then positive value is undefined', () => {
```

And I would also like this variation, to check the default is not getting set. This is implicitly tested by the above, but the default value being true for a negated boolean is a feature of the standard behaviour and I want to check it does not occur.)

```
  test('when negatable boolean flag is unspecified and negate is overridden then positive value is undefined', () => {
```

shadowspawn (approved):
One  more test suggested, but not needed for approval.

shadowspawn:
(Thanks for the work on the extra tests @bernard-wagner )

shadowspawn:
I am having second thoughts about this @abetomo

Basically, I don't want to commit to supporting bypassing negation as a feature, like by adding a chaining method on `Option` to clear `negate`. So I don't really want to add all the tests showing we support it (although they were good for showing it does work currently).

Two possible approaches for your consideration:
1) we could just change `attributeName` implementation as in this PR and not add any tests. So it would make it easier to bypass negation without us officially supporting this as a use case. And I think the change is reasonable on its own, makes the code a little clearer.
2) we could add an example file showing subclassing `Option` to allow bypassing negation, including a chaining method. (With a note that it isn't officially supported.)

abetomo:
I too think the 1 approach is better.

shadowspawn:
Closing in favour of #2250 which has just the code change and not the tests.

Thanks for the suggestion @bernard-wagner and work on the tests.

PR #2248, opened 2024-09-08
Add methods for formatOptions and formatCommands in Help, with example of option help groups

shadowspawn:
I am still thinking about help groups for options and commands per #1897. Still not happy I have a solid enough API to make it part of Commander.

This is a proof of concept of a refactor to make it easier for author to add their own support for help groups. Split out format of options and commands so author does not need to override whole of `formatHelp`. Show example of adding help groups for options in subclass.

Other help group experiments: #1908 #1910 #2176

### DIFF
--- a/examples/help-option-groups.js
+++ b/examples/help-option-groups.js
@@ -1,0 +1,90 @@
+const { Command, Help, Option } = require('commander');
+
+class MyHelp extends Help {
+  constructor() {
+    super();
+    this.optionGroups = {};
+  }
+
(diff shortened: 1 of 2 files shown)


shadowspawn:
The Help class is changing a bit in next major version of Commander so will close this as going to be out of date.

I think the grouping should probably be done by a different method than `.configureHelp()`, since that is usually for changes affecting all help including subcommands.

(Still thinking about grouping!)

PR #2266, opened 2024-10-12
Update README and quick-start for allowExcessArguments change

shadowspawn:
The README had not been updated to describe excess arguments as now an error by default. While doing so, I noticed some examples which need updating due to the stricter checks.

### DIFF
--- a/examples/split.js
+++ b/examples/split.js
@@ -2,7 +2,7 @@ const { program } = require('commander');
 
 // This is used as an example in the README for the Quick Start.
 
-program.option('--first').option('-s, --separator <char>');
+program.option('--first').option('-s, --separator <char>').argument('<string>');
 
 program.parse();
 
--- a/Readme.md
+++ b/Readme.md
@@ -79,7 +79,8 @@ const { program } = require('commander');
 
 program
   .option('--first')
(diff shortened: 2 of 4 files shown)


examples/split.js
```diff
@@ -2,7 +2,7 @@ const { program } = require('commander');
 
 // This is used as an example in the README for the Quick Start.
 
-program.option('--first').option('-s, --separator <char>');
+program.option('--first').option('-s, --separator <char>').argument('<string>');
```
abetomo, line 5:
```suggestion
program
  .option('--first')
  .option('-s, --separator <char>')
  .argument('<string>');
```

How about a line break to match the README?

shadowspawn: We run Prettier over the example files (but not over the source code in the README) and Prettier uses the compact layout for the short chaining calls in the simple example.

shadowspawn: I have wondered about whether should use the Prettier layout in the README, but I prefer being able to manually layout code and take extra space where I think it helps. Might just be habit though!

abetomo: I got it.

abetomo (approved):

Issue #2148, opened 2024-02-08
The root `--help` doesn't respect `.usage()` in a command


samcx:
## Example

```
.usage('[directory] [options]')
```

![CleanShot 2024-02-07 at 17 39 20@2x](https://github.com/tj/commander.js/assets/28912696/b2fe28bd-b3de-4c19-b789-3459208a6144)

Let me know if I'm missing something here!

On the latest version v12.

shadowspawn:
This is historical behaviour (which I am still wondering whether to change).

You can configure the help so the subcommand list displays your custom usage.

Example file: [help-subcommands-usage.js](https://github.com/tj/commander.js/blob/master/examples/help-subcommands-usage.js)

```
program.configureHelp({ subcommandTerm: (cmd) => cmd.name() + ' ' + cmd.usage() });
```

Related: #1853 #1885

samcx:
@shadowspawn Thank you for the confirmation!

Given how things are nicely updated in the root help output (e.g., `description()`) when things are updated in a subcommand), I was expecting it to also change for `usage()`, so maybe it would be a nice change!

shadowspawn:
This issue has not had any activity in over six months. It isn't likely to get acted on due to this report.

See also poll in #2268

Feel free to open a new issue if it comes up again, with new information and renewed interest.
 
Thank you for your contributions.

samcx:
@shadowspawn Will take a look at the poll! Thanks for responding.

samcx:
@shadowspawn Seems like I was late to voting, but casted my vote anyways!

PR #2267, opened 2024-10-12
Simplified subcommand term

shadowspawn:
# Pull Request

## Problem

The subcommand usage being slightly different than the command usage has always annoyed me a little. It may be a historical accident rather than by design.

In particular people who write a custom usage are surprised it does not show (#1853 #1885 #2148). But I am concerned about potential length of showing full usage by default: https://github.com/tj/commander.js/pull/1853#issuecomment-1465031565

Subcommand list was not showing `[command]` for subcommands with nested subcommands. (Missed when added nested subcommands.)

## Solution

Show just the arguments (or `[command]` for nested subcommands). No alias. No `'[options]`.

This makes the subcommand term clearly different that the usage, and the terms a bit shorter.

Tests skipped while thinking about approach! See first comment, as taking it further could show just the name!

### DIFF
--- a/lib/command.js
+++ b/lib/command.js
@@ -2186,7 +2186,7 @@ Expecting one of '${allowedValues.join("', '")}'`);
         .concat(
           this.options.length || this._helpOption !== null ? '[options]' : [],
           this.commands.length ? '[command]' : [],
-          this.registeredArguments.length ? args : [],
+          args,
         )
         .join(' ');
     }
--- a/lib/help.js
+++ b/lib/help.js
@@ -147,16 +147,13 @@ class Help {
    */
 
   subcommandTerm(cmd) {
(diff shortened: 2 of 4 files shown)


shadowspawn:
An alternative would be to show _only_ the subcommand name for the term in the list of subcommands. This is what larger programs do, such as `git`, `npm`, `docker`, and `hg`.

To make sure people find there is more help available, the `help` command would be listed separately. For example `git` has:
```
See 'git help <command>' or 'git help <concept>' to read about a specific subcommand or concept.
```

shadowspawn:
I opened a PR as I wanted to try changing the subcommand term and had an idea. Then I wondered about other approaches... I should make this an issue with alternatives instead!

Issue #2235, opened 2024-08-12
Show error for author when create option with invalid flags


shadowspawn:
The option parsing is quite loose. This means authors do not get feedback they are using unsupported flags when creating options.

In particular could throw an error for:
- more than two flags
    - #1862
    - #2222)
- short flag with more than 1 character
    - #1718
    - #2227
    - #2211

shadowspawn:
A prerelease is available for v13. The release is tagged as next and can be installed with:

    npm install commander@next

shadowspawn:
Version 13.0.0 has been released

Issue #2227, opened 2024-07-23
Bug in regex for lone short flag


sorawee:
https://github.com/tj/commander.js/pull/1256 added a support for lone short flag.

Unfortunately, this only works correctly on short flags with one character. Short flags with multiple characters (Java-style short flags) will incorrectly create incorrect field names. For example, `.option('-name')` will result in a field `Name`, which is not in camelCase.

The reason is that [this regex](https://github.com/tj/commander.js/blob/970ecae402b253de691e6a9066fea22f38fe7431/lib/option.js#L322) only matches a short flag with one character. This means a short flag with multiple characters will be categorized as a long flag. Consequently, Commander.js attempts to strip `--` from `-name`, which is a no-op, resulting in `-name`. Then, the camel case conversion of `-name` results in `Name`, which is not a camel case as intended.

shadowspawn:
Short flags with more than one character are not supported. Short flags are a dash followed by a single character. See #2211 for more details.

(Feel free to ask if you have more questions.)

sorawee:
Thanks @shadowspawn! That makes sense. But if Java-style short flags are intentionally not supported, shouldn't `.option('-name')` result in an error to let users know that it is not supported? I think we agree than categorizing `-name` as a long flag should not happen, right?

shadowspawn:
Closing in favour of #2235, adding an error for short flag longer than one character

shadowspawn:
FYI, there is a recent issue asking to allow multiple character short options: #2307

Issue #2222, opened 2024-07-04
[Feature request] More than 2 flags per option / case-insensitive flags


KaKi87:
Hi,

I'm trying to add a `-f, -F, --foo` option, but when doing so, the first two flags are passed as `options.F` and the last one is ignored.

So, it seems that `option()` only supports up to 2 flags.

Therefore, I would like to request being able to either specify any number of flags, or optionally mark flags as case-insensitive.

Thanks

shadowspawn:
Having more than 1 short and 1 long flag does come up very occasionally: #1862

This is not something I want to add to Commander. 

For an extra flag, you could perhaps add a hidden option which sets the "real" option value when used.

For case insensitive, you could perhaps subclass the Option class to change how options are matched

```js
import { Command, Option } from 'commander';

class MyOption extends Option {
  is(arg) {
    return this.short?.toLowerCase() === arg.toLowerCase() || this.long === arg;
  }
}

class MyCommand extends Command {
  createOption(flags, description) {
    return new MyOption(flags, description);
  }
  createCommand(name) {
    return new MyCommand(name);
  }
}


const program = new MyCommand();
program.option('-f, --foo', 'enable some foo');
program.parse();

console.log(program.opts());
```

```console
% node index.mjs -f
{ foo: true }
% node index.mjs -F
{ foo: true }
```

KaKi87:
> This is not something I want to add to Commander.

Why not ?

> For an extra flag, you could perhaps add a hidden option

Okay, for anyone else interested, here's a workaround :

```js
.option('-f, -F, --foo')
.addOption(new Option('--foo').hideHelp())
.action(options => {
    if(options.F || options.foo){}
});
```

> For an extra flag, you could perhaps add a hidden option

So you wouldn't want to implement case insensitivity either ? If no, why ?

Thanks.

cc. @mitsukuri

shadowspawn:
> Why not?

Basically, I don't think it is useful enough or commonly used enough to add to Commander.

I leave enhancement requests open for six months to see if they get upvotes or further comments. So there is a chance for other people to show interest with upvotes or comments, and make a case for the feature.

Why do you think it should be added?

- What is your use case? 
- Do you see more than two flags or case-insensitive flags in common tools you use? 
- Do you see support in other parsing libraries?

shadowspawn:
> For an extra flag, you could perhaps add a hidden option

 I had this approach in mind.

```js
[...]

Issue #2211, opened 2024-06-12
Short option name with single character or multiple characters have different result 


Brucedayton:
Steps to reproduce:

1. Have command with a required option argument with a short name (e.g -a).
2. Have command with a required option argument with a short name (e.g -bc).
Run the command with the option argument but with no space after the argument name (e.g commandname -asdf).
Run the command with the option argument but with no space after the argument name (e.g commandname -bcd).

Actual result:
1. Command runs and option argument -a has value "sdf".
2. error: unknown option '-bcd'

Can short option with single character report the same error like issue 2?

shadowspawn:
The simple answer is short options should be a single character. This is not currently treated as an immediate error, but may be in the future.

From the README:

> Each option can have a short flag (single character) and a long name

And listed as explicitly deprecated:
- https://github.com/tj/commander.js/blob/master/docs/deprecated.md#short-option-flag-longer-than-a-single-character

> Short option flags like `-ws` were never supported, but the old README did not make this clear. The README now states that short options are a single character.

shadowspawn:
(Thanks for the clear description.)

shadowspawn:
> Can short option with single character report the same error like issue 2?

No, an option with a required option argument looks for a value with or without a space. So these are equivalent:

```console
util -a AAA
util -aAAA
```

This is a widely supported syntax, although I do consider it an advanced syntax. When you don't know about it, the lack of an error for an unlucky typo can be surprising.

Brucedayton:
Thank you for the detailed explanation

shadowspawn:
Closing in favour of #2235, adding an error for short flag longer than one character

shadowspawn:
FYI, there is a recent issue asking to allow multiple character short options: #2307

PR #2270, opened 2024-10-29
Throw for unsupported option setup

shadowspawn:
# Pull Request

## Problem

People who try and use unsupported Option flags do not get an error, and have to discover themselves that there are problems at runtime.

Placeholder issue: #2235
See: #430 #479 #908 #1718 #1862 #2211 #2222 #2227

## Solution

Throw for unsupported Option flags, especially for known issues — too many flags, and short flag with more than one character.

This _will_ break some existing programs that had undetected problems. (It broke one of our tests!)

## ChangeLog

- _breaking_: throw during Option construction for unsupported option flags

### DIFF
--- a/lib/option.js
+++ b/lib/option.js
@@ -312,17 +312,32 @@ function camelcase(str) {
 function splitOptionFlags(flags) {
   let shortFlag;
   let longFlag;
-  // Use original very loose parsing to maintain backwards compatibility for now,
-  // which allowed for example unintended `-sw, --short-word` [sic].
-  const flagParts = flags.split(/[ |,]+/);
-  if (flagParts.length > 1 && !/^[[<]/.test(flagParts[1]))
-    shortFlag = flagParts.shift();
-  longFlag = flagParts.shift();
-  // Add support for lone short flag without significantly changing parsing!
-  if (!shortFlag && /^-[^-]$/.test(longFlag)) {
-    shortFlag = longFlag;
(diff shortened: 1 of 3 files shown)


abetomo (approved):

shadowspawn:
Released in Commander v13.0.0

p3k:
Hi @shadowspawn! We noticed some of our commands break with v13 because they are configured with more than one long flag. The idea is to provide a shorter version without using a one-letter option, e.g. `--es` and  `--ecmascript` or `--ddsg` and `--donaudampfschifffahrtsgesellschaft` (just kidding with the latter one).

Now that multi-letter options are also throwing an error, this feels quite limiting. I get the point of considering POSIX-compatibility with multi-letter options (I am always confused by those e.g. in Gnu’s `find` command) – does the same apply to multiple long flags or did you have other considerations?

shadowspawn:
No, I had not considered multiple long options. Like multi-character short options, they just happened to work. There is an issue open about multiple-character short options, and I'll add a reference to multiple longs: #2307

Hyrum's Law: https://www.hyrumslaw.com

> With a sufficient number of users of an API,
> it does not matter what you promise in the contract:
> all observable behaviors of your system
> will be depended on by somebody.

shadowspawn:
This PR was released in Commander 13.0

Support for dual long options to allow a short-ish flag was added in Commander 13.1: #2312

Issue #2290, opened 2024-12-05
Strange API behaviour for command description


PFight:
When adding description to command declaration, command does not work.

Working code:

```
program.command('analyze')
  .description('Do analyze.')
  .action(() => console.info("Hello explain!");
```
Execute:
```
> node index.js analyze
> Hello explain!
```

Not working code:
```
program.command('analyze', 'Do analyze.')
  .action(() => console.info("Hello explain!");
```
Execute:
```
> node index.js analyze
internal/modules/cjs/loader.js:892
  throw err;
  ^

Error: Cannot find module 'C:\Dev\stuff\npm-dependency-inspector\index-analyze'
?[90m    at Function.Module._resolveFilename (internal/modules/cjs/loader.js:889:15)?[39m
?[90m    at Function.Module._load (internal/modules/cjs/loader.js:745:27)?[39m
?[90m    at Function.executeUserEntryPoint [as runMain] (internal/modules/run_main.js:76:12)?[39m
?[90m    at internal/main/run_main_module.js:17:47?[39m {
  code: ?[32m'MODULE_NOT_FOUND'?[39m,
  requireStack: []
}
```

I spent HOURS to find a way to make commander work for simpliest command. Read all docummentation, tried several versions, and only happy accident helped me - I added description for one command, and do not added to another. So I found, that another command working and this give me a tip.

I think, when description specified, or command options specified, commander begin interpretate it as executable command. If it is true, then action() should throw error - it is usless for executable command. WTF? Why simple adding description change behaviour totally? And where in documentation this behaviour is specified?

shadowspawn:
Sorry you hit this problem. This used to be a common source of confusion: #938

The README was updated to mention the behaviour. In particular in the example code in the Commands section: https://github.com/tj/commander.js?tab=readme-ov-file#commands

https://github.com/tj/commander.js/blob/970ecae402b253de691e6a9066fea22f38fe7431/Readme.md?plain=1#L519-L541

And in the section: https://github.com/tj/commander.js?tab=readme-ov-file#stand-alone-executable-subcommands

> When `.command()` is invoked with a description argument, this tells Commander that you're going to use stand-alone executables for subcommands. Commander will search the files in the directory of the entry script for a file with the name combination `command-subcommand`, like `pm-install` or `pm-search` in the example below. The search includes trying common file extensions, like `.js`.

[...]

PR #2291, opened 2024-12-05
Add informative message for missing executable on Windows

shadowspawn:
## Problem

There is an informative runtime message to help the author when fail to find a stand-alone subcommand, but the message does not appear on Windows.

See: #2290

## Solution

Windows always tries opening the executable as an argument to node. Test for the existence of the file first.

## ChangeLog

- add: show informative error on Windows for missing stand-alone executable subcommand

### DIFF
--- a/lib/command.js
+++ b/lib/command.js
@@ -1165,6 +1165,17 @@ Expecting one of '${allowedValues.join("', '")}'`);
 
     launchWithNode = sourceExt.includes(path.extname(executableFile));
 
+    function throwMissingExecutable() {
+      const executableDirMessage = executableDir
+        ? `searched for local subcommand relative to directory '${executableDir}'`
+        : 'no directory for search for local subcommand, use .executableDir() to supply a custom directory';
+      const executableMissing = `'${executableFile}' does not exist
+  - if '${subcommand._name}' is not meant to be an executable command, remove description parameter from '.command()' and use '.description()' instead
+  - if the default executable name is not suitable, use the executableFile option to supply a custom name or path
+  - ${executableDirMessage}`;
+      throw new Error(executableMissing);
+    }
+
     let proc;
     if (process.platform !== 'win32') {
       if (launchWithNode) {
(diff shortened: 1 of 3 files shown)


shadowspawn:
Ouch. The new failure mode was hard to get working with the existing tests, but refactoring into a separate routine allowed suppressing the behaviour so existing tests could stay much the same.

lib/command.js
```diff
@@ -1094,6 +1094,24 @@ Expecting one of '${allowedValues.join("', '")}'`);
     return this;
   }
 
+  /**
+   * Note: masked in some unit tests.
+   * @param {string} executableFile
+   * @param {string} executableDir
+   * @param {string} subcommandName
+   */
+  _throwForMissingExecutable(executableFile, executableDir, subcommandName) {
```
abetomo, line 1103: The method name is `_throwForMissingExecutable`, but it may not throw, so why not name it something like `_validateExecutable`?

Or how about doing the checks outside of the method?
```js
if (!fs.existsSync(executableFile)) {
 _throwForMissingExecutable(...);
}
```

shadowspawn: > method name

Oh, yes. I originally had the check outside the routine. Will change name.

> Or how about doing the checks outside of the method?

I agree that would be tidier for the code and usage. However, I started that way, and multiple unit tests check whether `fs.existsSync` has been called and broke. I didn't think of an easy way of fixing the tests, other than moving the `fs. fs.existsSync` inside the routine. Then easy to completely bypass the new code and run the old tests.

shadowspawn: We have several `_checkForX` routines already so following that style, `_checkForMissingExecutable`.

shadowspawn: Lots!
- `_checkForMissingMandatoryOptions`
- `_checkForConflictingLocalOptions`
- `_checkForConflictingOptions`
- `_checkForBrokenPassThrough`

abetomo (approved):

shadowspawn:
Released in Commander v13.0.0

PR #2292, opened 2024-12-06
Prepare for 13.0.0-0

shadowspawn:
Updated CHANGELOG and version for 13.0.0-0. I will do a prerelease after this is approved.

(I will then do same for `commander-js/extra-typings`.)


abetomo (approved):
:package:

Issue #2263, opened 2024-10-09
`parse/parseAsync` cannot be used multiple times on the same `Command` instance


goodov:
Not sure if it's a known gotcha or an actual bug, but I've noticed that I can't use `parseAsync` multiple times with different args on the same `Command` instance (for ex. in tests), because internally all parsed options are cached.

goodov:
Hm. Found multiple closed duplicates which are all solved by referring to the added text in `README.md`. How about having this info in the parse/parseAsync method comment? This is pretty important bit that I wasn't able to find by looking into the source closely.

```
  /**
   * Parse `argv`, setting options and invoking commands when defined.
   *
   * Use parseAsync instead of parse if any of your action handlers are async. Returns a Promise.
   *
   * The default expectation is that the arguments are from node and have the application as argv[0]
   * and the script being run in argv[1], with user parameters after that.
   *
   * @example
   * ```
   * program.parseAsync(process.argv);
   * program.parseAsync(); // implicitly use process.argv and auto-detect node vs electron conventions
   * program.parseAsync(my-args, { from: 'user' }); // just user supplied arguments, nothing special about argv[0]
   * ```
   *
   * @returns Promise
   */
  parseAsync(argv?: readonly string[], options?: ParseOptions): Promise<this>;
```

goodov:
Throwing might be the best thing to do as another user mentioned here: https://github.com/tj/commander.js/issues/2246#issuecomment-2400779362

This should be improved.

shadowspawn:
There were no related issues opened for a while after the note got added to the README in #2036, but this issue and two others just recently:
- #2246
- #2247 

I expect prohibiting repeated calls will break some programs, but I also expect most of the cases will be technically unsafe and the author had not noticed there is state left behind from the first parse.

goodov:
> I expect prohibiting repeated calls will break some programs, but I also expect most of the cases will be technically unsafe and the author had not noticed there is state left behind from the first parse.

yep, I think this is what everyone are doing right now until they figure that something is wrong.

Throwing is a fast and simple take on the problem. Supporting multiple calls to `parse` would be a much better solution for the library in terms of the usability. It looks like it supposed to work like that, but was never carefully supported in this area which led to some invariants being broken with time.

shadowspawn:
Opened a PR to allow multiple calls to `.parse()`: #2299

shadowspawn:
Support for multiple parse added in version 13.0.0

Issue #2247, opened 2024-09-03
Boolean option does not reset between parse() and parseAsync() calls


edmsamuel:
### Español:
**Descripción del Problema:**
Al usar `commander.js`, la opción booleana `--json` no se restablece correctamente cuando se analiza varias veces. En el siguiente ejemplo, la opción `--json` se mantiene en `true` después de que se establece, incluso cuando no se pasa en la siguiente llamada a `parse`. Esto causa un comportamiento inesperado, ya que la opción debería restablecerse a `false` si no se incluye en la entrada.

**Código de ejemplo:**
```javascript
const { Command, createCommand } = require('commander');
const program = createCommand();

const tmp = program
  .option('--json', 'output extra debugging', false)
  .action(options => console.log(options));

const r1 = program.parse(["node", "app.js"]);
const r2 = program.parse(["node", "app.js", "--json"]);
const r3 = program.parse(["node", "app.js"]);
```

**Salida esperada:**
```javascript
{ json: false }
{ json: true }
{ json: false }
```

**Salida actual:**
```javascript
{ json: false }
{ json: true }
{ json: true }
```

**Comportamiento esperado:**
La opción `--json` debería ser `false` en la última llamada a `parse`, pero se mantiene en `true`, lo que indica que no se restablece correctamente entre llamadas.

### Inglés:
**Problem Description:**
When using `commander.js`, the boolean option `--json` does not reset correctly when parsing multiple times. In the following example, the `--json` option remains `true` after it is set, even when it is not passed in the subsequent `parse` call. This leads to unexpected behavior, as the option should reset to `false` if not included in the input.

**Example Code:**
```javascript
const { Command, createCommand } = require('commander');
const program = createCommand();

const tmp = program
  .option('--json', 'output extra debugging', false)
  .action(options => console.log(options));

const r1 = program.parse(["node", "app.js"]);
const r2 = program.parse(["node", "app.js", "--json"]);
const r3 = program.parse(["node", "app.js"]);
```

**Expected Output:**
```javascript
{ json: false }
{ json: true }
{ json: false }
```

**Actual Output:**
```javascript
{ json: false }
{ json: true }
{ json: true }
```

**Expected Behavior:**
The `--json` option should be `false` in the final `parse` call, but it remains `true`, indicating that it is not resetting properly between calls.

---

### Evidencia / Evidence

![image](https://github.com/user-attachments/assets/d8fe70c8-0a04-4d44-bd2a-00ece42195b8)

[...]

Issue #2246, opened 2024-09-03
Commands cannot be reused


gaohoward:
Hi I'm trying to create a cli app (interactive) that uses commands to perform tasks. For each task I create a command instance.
For example I provide a `get` command for user to use.
```
const getCmd = program.command('get')
.description('get information from a endpoint')
.argument('<path>', 'path of the component, [endpointName/componentType]')
.argument('[compName]', 'name of the component', '')
.option('-a, --attributes [attributeNames...]', 'get attributes from component', '')
.action(async (path, compName, options, cmd) => {
...
});
```
 In the interactive shell (I use readline lib) if user input `get args opts`  the getCmd will called
` getCmd.parseAsync(allargs [])`

then after the command is executed, the shell waiting for next command, if user inputs the same get command but with some options removed, when it calls the `getCmd.parseAsync(allargs)` again I found some of the options from the first execution remains.

So basically I assume each time I have to create a new instance of the `get` command. I wonder if there is a way to `reset` the command for reuse?

gaohoward:
looks like I can use my own options parser function... but still this is something can be improved.

shadowspawn:
> So basically I assume each time I have to create a new instance of the get command. 

Yes, that is the intended approach. 

See #2036 for some reference to past issues, and adding an explicit mention to the README.

shadowspawn:
An answer was provided, and no further activity in a month. Closing this as resolved.

Feel free to open a new issue if it comes up again, with new information and renewed interest.

(I am wondering whether Commander should display an error on a second call to parse, but that would be a very breaking change. I will continue to monitor issues where people trip over this.)

pillowfication:
I just ran into this issue, and it was very annoying to debug. I realize now that it's mentioned in the README, but I glossed over it and expected `parse()`/`parseAsync()` to simply be able to be called multiple times.

Throwing an error on a second call to `parse()` would be great. It seems that the README states that `parse()` shouldn't be called multiple times, and I'm not sure there is a great use case for its current behavior. So it seems that multiple `parse()` calls isn't supported at all currently and throwing an error wouldn't be a breaking change.

shadowspawn:
For interest @pillowfication , why were you calling parse multiple times? Was it unit tests, or a long running process using same program multiple times, or something else?

pillowfication:
> For interest @pillowfication , why were you calling parse multiple times? Was it unit tests, or a long running process using same program multiple times, or something else?

[...]

PR #2299, opened 2024-12-21
Add save/restore state to allow multiple calls to parse

shadowspawn:
# Pull Request

## Problem

People may not realise parse should only be called once, and discover by trial and error.

See: #438 #1565 #1819 #1829 #1916 #2036 #2246 #2247 #2263

## Solution

I thought it was going to be hard and messy to save/restore state and not worth it, but turned out fairly straightforward! Fingers crossed.

When `storeOptionsAsProperties: false` allow multiple calls to parse by storing and resetting the state. This is done in a lazy way, so a wide or deep hierarchy of subcommands does not need to save state on whole tree.

When `storeOptionsAsProperties: true` then throw on the second call to parse.

Not planning to support multiple calls to `.parseOptions()` as little used for direct calls, but could do that by moving the implementation into a private routine and call `._prepareForParse()` in public routine.

## ChangeLog

- support multiple calls to `.parse()` with default settings
- throw on multiple calls to `.parse()` if `storeOptionsAsProperties: true`
- add `.saveStateBeforeParse()` and `.restoreStateBeforeParse()` for use by subclasses

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -959,8 +959,6 @@ program.parse(['--port', '80'], { from: 'user' }); // just user supplied argumen
 
 Use parseAsync instead of parse if any of your action handlers are async.
 
-If you want to parse multiple times, create a new program each time. Calling parse does not clear out any previous state.
-
 ### Parsing Configuration
 
 If the default parsing does not suit your needs, there are some behaviours to support other usage patterns.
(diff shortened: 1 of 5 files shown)


shadowspawn:
Comments welcome on routine naming or more interesting cases to add to units tests.

abetomo (approved):

shadowspawn:
Released in Commander v13.0.0

Issue #2309, opened 2025-01-09
GNU-style triple-dash option (`---gnu-style-hidden-option`) no longer work in 13.0.0


wataash:
In 12.1.0, `---gnu-style-hidden-option` was parsed as `GnuStyleHiddenOption`, but due to the changes in https://github.com/tj/commander.js/pull/2270, 13.0.0 now throws an error: `Error: invalid Option flags: '---gnu-style-hidden-option'`.

```js
import { Command, Option } from "@commander-js/extra-typings";
const program = new Command();

program
  .addOption(new Option("---gnu-style-hidden-option").hideHelp().default(false))
  .action((opts) => {
    console.log(opts.GnuStyleHiddenOption);
    // 12.1.0: true or false
    // 13.0.0: Error: invalid Option flags: '---gnu-style-hidden-option'

    // N.B. @commander-js/extra-typings correctly infers the type of opts as { GnuStyleHiddenOption: boolean; }
    // (even with @commander-js/extra-typings 13.0.0)
  })
  .parse();

/*
$ node /src/program_a/a.js ---gnu-style-hidden-option
/src/program_a/node_modules/commander/lib/option.js:342
    throw new Error(`invalid Option flags: '${flags}'`);
          ^

Error: invalid Option flags: '---gnu-style-hidden-option'
    at splitOptionFlags (/src/program_a/node_modules/commander/lib/option.js:342:11)
    at new Option (/src/program_a/node_modules/commander/lib/option.js:20:25)
    at file:///src/program_a/a.js:10:16
    at ModuleJob.run (node:internal/modules/esm/module_job:271:25)
    at async onImport.tracePromise.__proto__ (node:internal/modules/esm/loader:547:26)
    at async asyncRunEntryPointWithESMLoader (node:internal/modules/run_main:116:5)

Node.js v22.12.0
*/
```

## Additional context

In GNU programs, triple-hyphen options are used as hidden debugging options that do not appear in the `program --help` output. (As far as I know, this is not documented anywhere and seems to be a convention within the GNU project. I think it existed in projects other than coreutils, but I'm not sure.)

- `stty ---debug` https://github.com/coreutils/coreutils/blob/v9.2/src/stty.c#L495 (since coreutils 9.2)
- `tail ---disable-inotify -F a.txt` https://github.com/coreutils/coreutils/blob/v9.2/src/tail.c#L240 (since coreutils 7.6)

[...]

Issue #2307, opened 2025-01-03
Return support for multiple character short options


twalshOG:
I know this change was intentional (#2270), but am hoping the team would reconsider.
A single character can be limiting and result in counterintuitive/weird short parameters. 

Real-ish example of where multi-character shorts would be nice:
``` ts
program
  .option('-k, --kill', 'kill orphaned reports')
  .option(
    '-sk, --skipExtract',
    "don't load results into database; intended for cases where you already have a fully populated database",
  )
  .option(
    '-sd, --skipPublish',
    "don't publish results the overall results; intended for cases where we are to archive the results but not push them to the entire team",
  );
```

shadowspawn:
Short options with multiple-characters worked more by accident than being intentionally supported. The README was updated in Commander v3 to make the intended single-character explicit and multiple-characters was added to the [deprecated](https://github.com/tj/commander.js/blob/master/docs/deprecated.md#short-option-flag-longer-than-a-single-character) documentation in Commander v9.

PR #599 added support for embedded option values and moved Commander closer to POSIX conventions which only have a single character for a short option.

There have been issues reported where people noted shortcomings in the (unsupported) behaviour: #2211 #2227. Likewise, code added to detect conflicting options did not consider multiple-character short options in #2055.

However, multiple character short options have been used in the wild! 2k hits with this [search](https://github.com/search?q=%22commander%22+%2F%5B%27%22%5D-%5Ba-z%5D%5Ba-z%5D%2B%5B+%2C%7C%5D%2F+language%3Ajavascript+NOT+is%3Afork&type=code&p=1)

If we add support for multiple character short options then I want it to be officially supported and not just work by happy coincidence.

Do you see multiple-character short options being used in popular command-line utilities? (Other than npm, which I know uses its own permissive command-line parsing utility with [shorthands](https://github.com/npm/nopt?tab=readme-ov-file#shorthands).)

shadowspawn:
A historical note after some investigation. Before Commander v5.0.0, short options with more than 1 character could not actually be used on the command line. In previous versions an option like `-sk` was unconditionally expanded to `-s -k`. So at the time I first noticed it in code in the wild and updated the README, it was not actually functional.

shadowspawn:
I am continuing to think about this. Using a multiple character short is not standard (POSIX, GNU) and not something I want to promote as a feature. However, I am wondering about allowing it to work in unambiguous cases since being used in the wild and will be annoying for existing users to upgrade.

[...]

PR #2312, opened 2025-01-11
Add support for dual long options when no short option

shadowspawn:
## Problem

#2270 throws an exception for unsupported option setups. This is picking up more variations than expected, and use of a short-ish form is more widespread than assumed (despite being at odds with the documented usage).

- multiple character short option: #2307
- dual long options: https://github.com/tj/commander.js/pull/2270#issuecomment-2577122902
- triple dash: #2309

In particular, people want to have a multi-character shortish option.

## Solution

The high level goals are still to restrict usage to supported formats, and follow standard conventions.

As per https://github.com/tj/commander.js/issues/2307#issuecomment-2581611847

> Short version: my latest thinking is to add explicit support for a short-ish option, but with double-dash rather than single dash.
> 
> Long version
> 
> There are strong conventions for short options (POSIX) and long options (GNU). I am uncomfortable about supporting a third and non-standard variation with multi-character short option.
> 
> But I appreciate the author desire for a more memorable short-ish option than a single character.
> 
> Adding explicit support for a double-dash short-ish option feels like a more pure approach still offering the key benefit. e.g. `--es, --ecmascript <year>`
> 
> There are no new behaviours that need to be decided or described for the short-ish option, and it is full featured. Can you embed option value with a short-ish option? Of course, `--es=2018`. Can you combined short and short-ish options in a single argument after a single dash? Of course not.

So:

- Multiple character short option: no, conflicts with standard posix and incomplete (no embedded value support). Use double dash instead. See next item.
- dual long options: yes, add support for this as the intended format for "shortish" option like `--ws, --workspace`
- triple dash: no, rare

## ChangeLog

- add: support for a pair of long option flags to allow a short-ish flag, like `.option('--ws, --workspace')`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 - *Breaking*: excess command-arguments cause an error by default, see migration tips ([#2223])
 - *Breaking*: throw during Option construction for unsupported option flags, like multiple characters after single `-` ([#2270])
+  - note: support for dual long option flags added in Commander 13.1
 - *Breaking*: throw on multiple calls to `.parse()` if `storeOptionsAsProperties: true` ([#2299])
 - TypeScript: include implicit `this` in parameters for action handler callback ([#2197])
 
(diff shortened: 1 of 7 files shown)


shadowspawn:
This PR is now a complete draft with tests and documentation. Feedback welcome.

abetomo (approved):

PR #2313, opened 2025-01-17
Fix example code in Readme_zh-CN.md

fwgood:
## Problem

There are differences between the Chinese and English readme, and the Chinese sample code cannot sit properly.

```
// En

program
  .option('--first')
  .option('-s, --separator <char>')
  .argument('<string>');

program.parse();

const options = program.opts();
const limit = options.first ? 1 : undefined;
console.log(program.args[0].split(options.separator, limit));
```

```
// Cn

program
  .option('--first')
  .option('-s, --separator <char>');

program.parse();

const options = program.opts();
const limit = options.first ? 1 : undefined;
console.log(program.args[0].split(options.separator, limit));
```

## Solution

Use English sample code in the Chinese version。

### DIFF
--- a/Readme_zh-CN.md
+++ b/Readme_zh-CN.md
@@ -77,7 +77,8 @@ const { program } = require('commander');
 
 program
   .option('--first')
-  .option('-s, --separator <char>');
+  .option('-s, --separator <char>')
+  .argument('<string>');
 
 program.parse();
 


shadowspawn (approved):
Thanks

shadowspawn:
(I'll prepare a release with PR #2312 soon and this can get promoted to master branch with that.)

abetomo (approved):

Issue #2338, opened 2025-03-13
Support negative number arguments


mmkal:
It would be good to allow passing in negative numbers via CLI - ideally without workarounds for _options_, but as far as I can tell, it doesn't work at all for _arguments_.

Example:

```js
import {program} from 'commander'

program.argument('<number>', 'The number to square').action(num => {
  console.log(Math.pow(Number(num), 2))
})

program.parse()
```

You can run `node square.js 2` but not `node square.js -2`. Since it's not an option, there's no equivalent to the `--foo=-2` suggested in https://github.com/tj/commander.js/issues/61. That issue has been closed for a few years but this comment suggests it's worth opening a new one https://github.com/tj/commander.js/issues/61#issuecomment-504636612.

My suggested fix:

If an argv value matching `^-\d(\.\d+)?$` is encountered, and doesn't match any option, don't throw `error: unknown option '-2'`. Just treat it like an non-option argv. I haven't looked into the code but hopefully that could be done by a single check somewhere. If feasible, this would also probably solve for the original complaint in #61 without needing end-users to know the `=` workaround.

shadowspawn:
The generic answer is use `--` to terminate option processing, like:
```
node square.js -- -2
```

-----

I am interested in revisiting negative numbers.

Related issues: #61 #413

POSIX has a section about numbers in 12.1.6:

- https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap12.html

Some research I have done on negative number treatment in other packages:
- https://github.com/pkgjs/parseargs/issues/77
- https://github.com/pkgjs/parseargs/pull/62#issuecomment-1034656341

shadowspawn:
This has been released in Commander v14.

Issue #413, opened 2015-06-19
Arguments of negative numbers


ciceropablo:
It makes sense?
```sh
$ node index.js 1 2 3
# [ '1', '2', '3' ]

$ node index.js 1 -2 3
# [ '1' ]
```

SomeKittens:
In this case you'd use the double hyphen `--` to end option parsing.

ciceropablo:
Ok!
Documentation about it (literal args)?

SomeKittens:
That's a general CLI thing not specific to commander

sjlongland:
Hi,

Just stumbled on this today… Would it not be better for the code to first *check* that there exists at least one purely numeric argument defined for the action or globally, before assuming that the `-` parameter is in fact an option?

In my case, it's a option to an "action" which is getting confused:
```javascript

program
    .command('readloadsurvey <num> [start] [end]')
    .description('read load survey <num> starting at record <start>')
    .option('--search-ts', 'Search by time-stamp')
    .action((num, start, end, options) => {
        /* … etc… */
    });
```

Here, there is no option that matches the regex `^-\d$`; therefore, an option like `-1` cannot be anything *but* a negative number.  In fact, there isn't even any "short form" arguments defined anywhere, so it'd be nice to be able to disable this, as the "feature" if possible.  I can see it being ambiguous if I had defined any options matching this form, but here, that is not the case so it'd be better to assume "this is a number" rather than "this is an option".

shadowspawn:
Commander does not currently treat numbers as special. It is a little appealing, but this has not come up very often and would be a special case in the quite complicated parsing, so my first reaction is unlikely.

If you are interested in gathering more feedback, please open a new issue.

Related: #61

PR #2339, opened 2025-03-15
Allow negative numbers as optional optional arguments and command arguments when unambiguous

shadowspawn:
## Problem

It is unexpected when negative numbers don't work as option-arguments or command arguments and require special treatment.

e.g.

`test --optional -3` => `test --optional=-3`
`test -12.4` => `test -- -12.4`

See: #61 #413 #2338 

## Solution

Allow negative numbers for optional option-arguments and for command arguments.

Like Python, if there are options with digit short flag then don't treat negative numbers as special since they overlap with some options. Keep the behaviour simple and conservative.

Description of what some other libraries do: https://github.com/pkgjs/parseargs/issues/77#issue-1164568986
- yargs has support for negative numbers
- Python allows negative numbers for positionals if there are not options that look like negative numbers (Python [Arguments containing -](https://docs.python.org/3/library/argparse.html#arguments-containing))

## ChangeLog

- add: support for negative numbers as option-arguments and command-arguments

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -333,9 +333,9 @@ $ pizza-options --cheese mozzarella
 add cheese type mozzarella
 ```
 
-Options with an optional option-argument are not greedy and will ignore arguments starting with a dash,
-except negative numbers which are accepted.
+Options with an optional option-argument are not greedy and will ignore arguments starting with a dash.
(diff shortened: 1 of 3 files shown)


shadowspawn:
Moved out of draft, ready for review.

abetomo (approved):

shadowspawn:
This has been released in Commander v14.

PR #2341, opened 2025-03-17
Unknown option - negative number suggestion

mmkal:
## Problem

See: https://github.com/tj/commander.js/issues/61 https://github.com/tj/commander.js/issues/413 https://github.com/tj/commander.js/issues/2338 (doesn't solve, just informs user how to workaround issue)

More basic alternative/complement to https://github.com/tj/commander.js/pull/2339

## Solution

When there's an "unknown option" that looks like a negative number, add a suggestion to the `unknown option` error message telling the user how they can get the value to be parsed as an argument or option value.

- Used a regex to test if the unknown option "looks like" a negative number
- Doesn't cover hex numbers like `-0xa` or anything, but I think that's ok
- Added a test for the suggestion message
- Also tested manually by linking this lib to a local project
- No docs added as it's a bit of an edge case. Better to educate the end user only if they hit this problem, since the majority probably won't

## ChangeLog

Add suggestion for negative number unknown option error message

### DIFF
--- a/lib/command.js
+++ b/lib/command.js
@@ -2090,6 +2090,9 @@ Expecting one of '${allowedValues.join("', '")}'`);
         command = command.parent;
       } while (command && !command._enablePositionalOptions);
       suggestion = suggestSimilar(flag, candidateFlags);
+    } else if (flag.match(/^-\d+(\.\d+)?$/) && this._showSuggestionAfterError) {
+      // looks like a negative number, offer the workarounds to allow it to be processed as an argument/option value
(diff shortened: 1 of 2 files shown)


shadowspawn:
We haven't been using the suggestions for other than typos of known options/commands, that is interesting. 

A related example, we did some similar messaging for missing option values in parseArgs: https://github.com/pkgjs/parseargs/blob/8038b6f67202d473f65cbea1541f029c87559c5e/index.js#L73-L89

shadowspawn:
On hold until decide on #2339

shadowspawn:
Will be releasing support for negative numbers: #2339 

Thanks for your contributions

mmkal:
Great, thanks!

PR #2348, opened 2025-03-24
Fix leading space in option helptext

mmkal:
## Problem

There's a leading space when an option has no description but does have a default value (or other "extra info"):

image

## Solution

Copy the existing logic for arguments to avoid a leading space:

image

Added a test which failed before the change to `lib/help.js`

image

## ChangeLog

Fix leading space in option helptext


shadowspawn (approved):
LGTM, thanks

shadowspawn:
This has been released in Commander v14.

Issue #2342, opened 2025-03-19
copyInheritedSettings and configureOutput have side-effects


mmkal:
Hi - noticed something that maybe isn't desired behaviour with the above two methods. Since `copyInheritedSettings` sets `this._outputConfiguration = sourceCommand._outputConfiguration` and `configureOutput` does `Object.assign(...)` - this means that one command can change another command's settings unexpectedly:

```ts
import {program, Command} from 'commander'

const one = new Command('one')
  .description('A very very very very very very very very very very very very very very very very very very very very very very very very long description')
  .action(() => console.log('hi i am one'))
  .configureOutput({
    writeOut: str => console.log(str + '🚀🚀🚀🚀'),
    getOutHelpWidth: () => 70,
  })

const two = new Command('two')
  .copyInheritedSettings(one)
  .description('Another very very very very very very very very very very very very very very very very very very very very very very very very long description')
  .action(() => console.log('hi i am two'))
  .configureOutput({
    getOutHelpWidth: () => 200,
  })

program
  .addCommand(one)
  .addCommand(two)

program.parse()
```

When you do `mycommand one --help` it uses 200 for the help width, not 70 as I'd expect. (Note: `mycommand two --help` prints the `🚀🚀🚀🚀` at the bottom, which _is_ correct/expected, IMO).

Suggested solution: use a spread operator in either/both of the methods.

shadowspawn:
Probably uncommon to configure the output at a leaf command, but that isn't the intended behaviour. `.configureOutput()` should merge to a new object instead of merging in place.

This will change the behaviour for people configuring output after setting up subcommands (not the intended pattern), which can be touched on in the release notes. (Note to self: check README too.)

shadowspawn:
This has been released in Commander v14.

PR #2350, opened 2025-03-26
Make .configureOutput() create copy of settings instead of in-place change

shadowspawn:
## Problem

Modifying command using `.configureOutput()` modifies other commands sharing same settings.

Fixes: #2342

## Solution

Make a copy of settings instead of modifying in place. 

## ChangeLog

- fixed: `.configureOutput()` now makes copy of settings instead of modifying in-place, fixing side-effects

### DIFF
--- a/lib/command.js
+++ b/lib/command.js
@@ -239,7 +239,7 @@ class Command extends EventEmitter {
   configureOutput(configuration) {
     if (configuration === undefined) return this._outputConfiguration;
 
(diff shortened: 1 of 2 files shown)


shadowspawn:
Could go in a minor version since a bug fix, but thinking hold until we do a release to drop node 18 since will break anyone who was inadvertently relying on previous behaviour.

abetomo (approved):

galaxycores (approved):

PR #2365, opened 2025-05-07
ci: Update node-version

abetomo:
Update to the currently supported version.
https://github.com/nodejs/Release

* Add 24.x
* Drop 18.x


shadowspawn (approved):
LGTM

I will prepare a release for 14.x which drops node 18 in next week or so.

PR #2384, opened 2025-06-27
List excess arguments in error message

stiff:
Sometimes a typo in arguments list can be parsed as an excess argument rather than option. When the options list is long, finding it can be tricky, so it would be helpful to see exactly what arguments are treated as excess.

## Solution

Real life example by mistakenly adding spaces before # in systemd service file:

```
error: too many arguments. Expected 0 arguments but got 2: #, #--address=0x72.
```

## ChangeLog

List excess arguments in error message

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -74,7 +74,7 @@ Now shows an error:
 
 ```console
 $ node example.js a b c
-error: too many arguments. Expected 0 arguments but got 3.
+error: too many arguments. Expected 0 arguments but got 3: a, b, c.
 ```
 
 You can declare the expected arguments. The help will then be more accurate too. Note that declaring
@@ -455,7 +455,7 @@ program.parse();
 ```
 
 ```sh
-$ node trivial.js 
+$ node trivial.js
 error: missing required argument 'file'
 ```
 
@@ -557,7 +557,7 @@ program.showHelpAfterError();
   - pad width calculation was not including help option and help command
   - pad width calculation was including hidden options and commands
 - improve backwards compatibility for custom command event listeners ([#1403])
-  
+
 ### Deleted
 
 - *Breaking:* `.passCommandToAction()` ([#1409])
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -74,7 +74,7 @@ Now shows an error:
 
 ```console
 $ node example.js a b c
-error: too many arguments. Expected 0 arguments but got 3.
+error: too many arguments. Expected 0 arguments but got 3: a, b, c.
```
shadowspawn, line 77: This is the description of the behaviour in Commander v13, which has not changed.

The PR should not be updating the CHANGELOG.

shadowspawn:
For listing the arguments, I wonder whether they should be quoted. Given the arguments could contain spaces and commas, I think quoting might be appropriate? (Note: we can't match exactly how they appear on the command-line as any original quotes have been stripped during processing by the shell.)

-----

I looked at how choices are handled, and we have done it both ways!

In the help we are fairly consistent and use `JSON.stringify`. Although that is partly because some values, like default, not being plain strings. And use same pattern so consistent in display. 

Given:
```javascript
program.addOption(new Option('-d, --drink <size>', 'drink size').choices(['small', 'medium', 'large']).default('medium'));
```

See quoted in the help:
```text
Options:
  -d, --drink <size>  drink size (choices: "small", "medium", "large", default: "medium")
```

But unquoted in the error:
```console
$ node index.mjs --drink huge
error: option '-d, --drink <size>' argument 'huge' is invalid. Allowed choices are small, medium, large.
```

CHANGELOG.md
stiff: Indeed, I was doing too quickly didn't look at the file name. I also realised that implementation is doing completely wrong thins since it compares argument name()s with actual values.

shadowspawn (changes requested):
I think this is a useful improvement. I would like the changes to CHANGELOG removed.

shadowspawn:
It is an interesting question whether change to error messages are technically a semver breaking change, but I know some projects test messages in unit tests or scrape the error message for custom handling. Unless/until we have a stated policy that messages may change I prefer to keep the changes in a semver major release.

stiff:
I don't think it's very urgent and big enough for major release on its own, but I think you are right it should be included in next major release.

shadowspawn:
(Note: waiting for revert on accidental changes to CHANGELOG.)

shadowspawn (approved):
LGTM, thanks @stiff

abetomo (approved):
:+1:

shadowspawn:

PR #2387, opened 2025-07-01
Add MissingArgumentError with argument name

jpozo20:
## Problem

When an exception occurs due to a missing argument, it would be good to have the argument name in the thrown exception so I can use it in the `exitOverride` callback. 
## Solution

Added a `MissingArgumentError` class that is thrown when a missing argument is trigger in the `.parse` method.
This class contains a property `argumentName` that can be used to customize the error message in the callback of `exitOverride`. 

I added 4 tests in `command.exitOverride.test.js`. 
Two of them test the error is indeed a `MissingArgumentError`. The other two test that the `argumentName` is in the error object. 
## ChangeLog

Add MissingArgumentError with argument name for exitOverride()

### DIFF
--- a/lib/error.js
+++ b/lib/error.js
@@ -36,7 +36,7 @@ class InvalidArgumentError extends CommanderError {
 }
 
 /**
- * InvalidArgumentError class
+ * MissingArgumentError class
  */
 class MissingArgumentError extends CommanderError {
   /**


jpozo20:
Maybe I didn't explain myself correctly. The `MissingArgumentError` is an _error class_ extending from `CommanderError`. I based it on the `InvalidArgumentError` class and added the property `argumentName`.

jpozo20:
I'm already using the package with the new class in the meantime. But when I tried to use class in the `exitOverride` callback, I got an exception saying the `MissingArgumentError` was not exported. 

Is there no way to generate the typings? Since I couldn't find anything related to it in the package.json nor the documentation, are they added manually?

shadowspawn:
> I'm already using the package with the new class in the meantime.

Is this in a public repo? I am interested in seeing the usage, but no worries if not visible.

jpozo20:
Eventually, it will be public and OSS but right now I'm working on the initial version of the CLI. 

The code where I will be using the new class will resemble something like this

```javascript
import { logError, logOutput } from './messages.js';
import { Command, CommanderError, InvalidArgumentError, MissingArgumentError } from 'commander';

const handleProjectsCommandError = (err: CommanderError, command: Command) => {
  if (err instanceof MissingArgumentError) {
    const message = `mycustommessage with missing argument: ${err.argumentName}`;
    logError(message);
  } else if (err instanceof InvalidArgumentError) {
    logError(`mycustommessage with invalid argument`);
  } else {
    logError(`An unexpected error occurred: ${err.message}`);
  }
  process.exit(0);
};

export const onCommanderError = (err: CommanderError, command: Command) => {
  switch (command.name()) {
    case 'projects':
      handleProjectsCommandError(err, command);
      break;
    default:
      logError(`An error occurred in command ${command.name()}: ${err.message}`);
  }
};

export const outputConfiguration = {
  writeOut: (str: string) => process.stdout.write(`[OUT] ${str}`),
  // hide output errors from commanderjs
  writeErr: (str: string) => {},
  outputError: (str: string, write: (str: string) => never) => {},
};

```

But now I realize I will need to refactor my code if I want that second parameter on the `onCommanderError` callback since those commands are subcommands and I didn't use `createCommand()` for them.

```javascript
projectsCommand
  .command('list')
  .alias('l')
  .description('List all projects in the database.')
  .option('-d, --details', 'Show detailed information about each project.', false)
  .action(async (options) => {
    await projectsHandler.listProjects(options.details);
  });

projectsCommand
  .command('remove')
  .alias('rm')
  .description('Remove a project from the database.')
  .argument('<name>', 'Name of the project to remove')
  .action(async (name) => {
    if (!name) {
      // right now handled in the onCommanderError of the projectsCommand
      // since this code isn't reached if the required argument is missing
      throw new Error('Project name is required');
    }

    await projectsHandler.removeProject(name);
  });
```

shadowspawn:
I am not sure the current form of the PR is right. It looks like you have a big customisation in flight, and so I am not doing anything until that comes together further. See if the proposed PR actually proves usable and useful as the customisation takes shape.

shadowspawn:
I am not sure this is the right path, and do not want to set a precedent by adding a custom field and class for one extra piece of information.

It isn't immediately apparent, but the reason for a custom `InvalidArgumentError` class is that it is a class also thrown by end-users, so I wanted it easy to construct.

If we do add further information, I am thinking it might be richer like including:
- `command`: Command that is throwing the error
- `argument`: Argument that is missing
- (`value`: string with problem value for `InvalidArgumentError`)

I'll leave the issue open for months to collection feedback and see if there is more interest and feedback.

Good luck with your project and custom Commander! Thanks for the sneak peep. 😄

sbgitZZ (commented):
h

sbgitZZ (commented):
h

PR #2395, opened 2025-07-05
refactor: use an object spread instead of Object.assign

noritaka1166:
## Solution
use an object spread instead of Object.assign to make it more concise and readable


shadowspawn (approved):
Ah, that was me in #2350. I am in the habit of using assign. I agree using spread is more idiomatic.

PR #2396, opened 2025-07-07
refactor:  use endsWith

noritaka1166:
## Solution
use endsWith for better readability

### DIFF
--- a/lib/argument.js
+++ b/lib/argument.js
@@ -33,7 +33,7 @@ class Argument {
         break;
     }
 
-    if (this._name.length > 3 && this._name.slice(-3) === '...') {
+    if (this._name.length > 3 && this._name.endsWith('...')) {
       this.variadic = true;
       this._name = this._name.slice(0, -3);
     }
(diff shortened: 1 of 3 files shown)


lib/argument.js
```diff
@@ -33,7 +33,7 @@ class Argument {
         break;
     }
 
-    if (this._name.length > 3 && this._name.slice(-3) === '...') {
+    if (this._name.length > 3 && this._name.endsWith('...')) {
```
shadowspawn, line 36: The slice was a bit unusual, and the length test is redundant too.

```suggestion
    if (this._name.endsWith('...')) {
```

shadowspawn:
I don't find the changes from testing `str[0]` against a character to `str.startsWith(chr)` are as compelling. I am pretty familiar with testing characters this way in this code and in other languages. But not strongly against the change.

Do you like this change @abetomo ? Otherwise I would prefer not to change to `.startsWith()` (at least not without more justification).

abetomo:
I too prefer `str[0]` if only the first character is tested.

shadowspawn (approved):

abetomo (approved):
👍

Issue #2397, opened 2025-07-11
Make order of negated options irrelevant


micheleb:
Hey there! 👋

Thanks for this great library. I've been using it for years, and it's really amazing. :)

It would be really nice if the order in which options and their negations are specified didn't matter, since it's an invisible requirement that easily breaks when you want your options to be alphabetically sorted.

I was able to fix the issue with the patch below. Would something like this be considered for inclusion in the library? I can create a PR in that case, of course.

Thanks!

```diff
diff --git a/node_modules/commander/lib/command.js b/node_modules/commander/lib/command.js
index efbb8f6..5c74f6a 100644
--- a/node_modules/commander/lib/command.js
+++ b/node_modules/commander/lib/command.js
@@ -659,15 +659,43 @@ Expecting one of '${allowedValues.join("', '")}'`);
     if (option.negate) {
       // --no-foo is special and defaults foo to true, unless a --foo option is already defined
       const positiveLongFlag = option.long.replace(/^--no-/, '--');
-      if (!this._findOption(positiveLongFlag)) {
+      const positiveOption = this._findOption(positiveLongFlag);
+      if (!positiveOption) {
         this.setOptionValueWithSource(
           name,
           option.defaultValue === undefined ? true : option.defaultValue,
           'default',
         );
+      } else {
+        // If positive option exists and was auto-defaulted to true due to this negative option being added later,
+        // remove that auto-default to prevent {foo: true} when neither --foo nor --no-foo is used
+        const positiveOptionName = positiveOption.attributeName();
+        if (this.getOptionValueSource(positiveOptionName) === 'default' && 
+            this.getOptionValue(positiveOptionName) === true &&
+            positiveOption.defaultValue === undefined) {
+          // Remove the auto-set default
+          delete this._optionValues[positiveOptionName];
+          delete this._optionValueSources[positiveOptionName];
+        }
       }
     } else if (option.defaultValue !== undefined) {
       this.setOptionValueWithSource(name, option.defaultValue, 'default');
+    } else {
+      // Check if this positive option has a corresponding negative option that was added first
+      // and auto-set a default value. If so, remove that auto-default.
+      const negativeLongFlag = option.long ? option.long.replace(/^--/, '--no-') : null;
+      const negativeOption = negativeLongFlag && this._findOption(negativeLongFlag);
+      if (negativeOption && negativeOption.negate) {
+        // If we already have a default value set for this option due to the negative option,
+        // and the negative option didn't have an explicit default, remove the auto-default
[...]

PR #2405, opened 2025-08-03
Make order not matter for adding combo positive/negative options

shadowspawn:
## Problem

Lone negated options default to true. (Historical.)
Positive then negative does not default to true. (By design.)
Negative then positive still defaults to true, for implementation reasons rather than by design.

See: #2397

## Solution

Move the setting of the default to when the options are known.

Alternative approach to #2398. That PR retroactively removes the "default true" when the positive option is added second.

## ChangeLog

Fix: only lone negated option defaults option value to true

### DIFF
--- a/lib/command.js
+++ b/lib/command.js
@@ -675,17 +675,7 @@ Expecting one of '${allowedValues.join("', '")}'`);
     const name = option.attributeName();
 
     // store default value
-    if (option.negate) {
-      // --no-foo is special and defaults foo to true, unless a --foo option is already defined
-      const positiveLongFlag = option.long.replace(/^--no-/, '--');
-      if (!this._findOption(positiveLongFlag)) {
-        this.setOptionValueWithSource(
-          name,
-          option.defaultValue === undefined ? true : option.defaultValue,
-          'default',
-        );
-      }
-    } else if (option.defaultValue !== undefined) {
+    if (option.defaultValue !== undefined) {
       this.setOptionValueWithSource(name, option.defaultValue, 'default');
     }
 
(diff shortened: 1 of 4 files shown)


shadowspawn:
Oh drat, started work from develop branch so includes extra commits. Will rework.

lib/command.js
```diff
@@ -1126,6 +1116,30 @@ Expecting one of '${allowedValues.join("', '")}'`);
...
+              option.attributeName(),
+              true,
+              'default',
+            );
+          }
+        });
+    }
+
+    // Save the state the first time, then restore the state before each subsequent parse.
     if (this._savedState === null) {
```
abetomo, line 1143: Since the conditon is the same, how about not separating them so it is easier to understand?

```suggestion
    if (this._savedState === null) {
      // Before parsing and before saving the state, do the special default of lone negated option to true.
      // Some paranoia about not overwriting a value since doing this lazily instead of when option added.
      this.options
        .filter(
          (option) =>
            option.negate &&
            option.defaultValue === undefined &&
            this.getOptionValue(option.attributeName()) === undefined,
        )
        .forEach((option) => {
          // lone --no-foo is special and defaults foo to true, unless a --foo option is already defined
          const positiveLongFlag = option.long.replace(/^--no-/, '--');
          if (!this._findOption(positiveLongFlag)) {
            this.setOptionValueWithSource(
              option.attributeName(),
              true,
              'default',
            );
          }
        });
      // Save the state the first time, then restore the state before each subsequent parse.
```

shadowspawn: I thought it might be useful having the simple save/restore together, but didn't really help. Refactored as suggested!

abetomo (approved):
:+1:

shadowspawn:

PR #2427, opened 2025-09-25
minor change in index.d.ts

danielrentz:
## Problem

No problem, just a subtlety in the typings. The deprecated versions of `help` and `outputHelp` should not include the parameter-less version to make it entirely clear that only the callback variant itself is deprecated.

## Solution

Remove optional marker from the deprecated signatures.

### DIFF
--- a/typings/index.d.ts
+++ b/typings/index.d.ts
@@ -1044,7 +1044,7 @@ export class Command {
    */
   outputHelp(context?: HelpContext): void;
   /** @deprecated since v7 */
-  outputHelp(cb?: (str: string) => string): void;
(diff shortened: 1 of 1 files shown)


shadowspawn (approved):
Clearer for readers. LGTM.

shadowspawn:
Would you like to make the same change in https://github.com/commander-js/extra-typings ?

abetomo (approved):
:+1:

danielrentz:
> Would you like to make the same change in https://github.com/commander-js/extra-typings ?

https://github.com/commander-js/extra-typings/pull/127

shadowspawn:
Related in v14.0.3

Issue #2423, opened 2025-09-09
Configurable option (camel-)casing


mmkal:
Hi! For trpc-cli, I'd like to support properties like `useMCPServer: z.string()` nicely. Right now, the only way to get commander to parse an option into `program.opts().useMCPServer` is to define an option `--use-m-c-p-server` which is horrible looking.

It'd be nice if I could do something like this:

```ts
const option = new Option('--use-mcp-server', {attributeName: 'useMCPServer'})
```

I'm trying a workaround in trpc-cli by sub-classing `Option` and overriding `attributeName()`, but it seems a bit brittle: https://github.com/mmkal/trpc-cli/pull/149/files#diff-a2a171449d862fe29692ce031981047d7ab755ae7f84c707aef80701b3ea0c80R287

mmkal:
Oh nice, didn't spot `createOption`. If subclassing like that is considered an ok way to do this, I'm fine with it. Feel free to close unless you want to keep open to track `setAttributeName`.

Thanks!

PR #2447, opened 2025-11-08
Implement useName() method for setting a custom attribute name

nielsnl68:
## Problem

In my setup i use a fixed configuration file that is loaded from an INI file. To make thinks easier for me i like the result of the arg parser have the same key names as this configuration class members so i could easily set the values without first modify the result.

## Solution

With this PR we can now set an alternative attribute but name.

This small PR adds a method `useName()` to have a different attribute name then the current option name and the `attributeName()` checks first if the `useName()` is used will then us the given attrName or use the old method to determine the attributes name.

### DIFF
--- a/lib/option.js
+++ b/lib/option.js
@@ -34,7 +34,7 @@ class Option {
     this.conflictsWith = [];
     this.implied = undefined;
     this.helpGroupHeading = undefined; // soft initialised when option added to command
-    this.attrName = undefined;
+    this.attributeName = undefined;
   }
 
   /**
@@ -198,13 +198,13 @@ class Option {
     /**
    * Allow to change the attribute name of this option.
    *
-   * @param {string} the used attribute name
+   * @param {string} attributeName is used for a alternated attribute name
    * @return {Option}
    */
 
-  useName(artName) {
-    if (typeof artName === 'string') {
-      this.attrName = artName;
+  setAttributeName(attributeName) {
+    if (typeof attributeName === 'string') {
+      this.attributeName = attributeName;
     }
     return this;
   }
@@ -224,15 +224,15 @@ class Option {
   /**
    * Return option name, in a camelcase format that can be used
    * as an object attribute key. 
-   * This method now uses the attrName when it is set otherwise 
+   * This method now uses alternativly the this.attributeName when it is set otherwise 
    * it will still use the option name as before
    *
    * @return {string}
    */
 
   attributeName() {
-    if (typeof this.attrName === 'string') {
-      return attrName;
+    if (typeof this.attributeName === 'string') {
+      return this.attributeName;
     } 
     if (this.negate) {
       return camelcase(this.name().replace(/^no-/, ''));
--- a/lib/option.js
+++ b/lib/option.js
@@ -34,6 +34,7 @@ class Option {
     this.conflictsWith = [];
     this.implied = undefined;
     this.helpGroupHeading = undefined; // soft initialised when option added to command
+    this.attrName = undefined;
   }
 
   /**
@@ -194,6 +195,19 @@ class Option {
     return this;
   }
 
+    /**
+   * Allow to change the attribute name of this option.
+   *
+   * @param {string} the used attribute name
+   * @return {Option}
+   */
+
+  useName(artName) {
+    if (typeof artName === 'string') {
+      this.attrName = artName;
+    }
+    return this;
+  }
   /**
    * Return option name.
    *
@@ -209,12 +223,17 @@ class Option {
 
   /**
    * Return option name, in a camelcase format that can be used
-   * as an object attribute key.
+   * as an object attribute key. 
+   * This method now uses the attrName when it is set otherwise 
+   * it will still use the option name as before
    *
    * @return {string}
    */
 
   attributeName() {
+    if (typeof this.attrName === 'string') {
+      return attrName;
+    } 
     if (this.negate) {
       return camelcase(this.name().replace(/^no-/, ''));
     }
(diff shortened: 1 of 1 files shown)


shadowspawn:
I don't think this is a common need, but it is the second time this has come up recently. See: #2423

shadowspawn:
Looking into existing code and patterns. 

For `-s, --some-option`:
- `option.name()` returns `some-option` and is only used for events. Calculated on fly.
- `option.attributeName()` returns `someOption` (#685).  Calculated on fly from `name()`.

`Command` class uses getter/setter pattern for some properties, but this pattern is not used in `Option` class.

https://github.com/tj/commander.js/blob/0692be58103e1ea8052d5d45ab11cc02e197eea5/lib/command.js#L2331-L2335

All `Option` data properties are public.

shadowspawn:
A small challenge with having public data properties on `Option` is that the obvious name for the data property and for the related chaining method are often the same. But you can not call both the data property and the chaining method the same thing!

nielsnl68:
> A small challenge with having public data properties on `Option` is that the obvious name for the data property and for the related chaining method are often the same. But you can not call both the data property and the chaining method the same thing!

Good point, not sure what to name the variable i was doubting to name it *`alternatedName`*. But it did not give a good explanation of what it is. Do you have a preference?

shadowspawn:
Just stick an underscore on the start for now to let the tests run. I'll come back to naming. (We work hard on the naming!)

nielsnl68:
> I don't think this is a common need, but it is the second time this has come up recently. See: #2423

I fully understand that. i just your component fine without it. And would not have created this PR, when there was more involved then this.

I just played a little bit with the Python ArgParse library and it did have something like this as well. And made my code a lot easier to setup.

This is currently what i use in my nodejs script to get the configuration:
```javascript
const config = {
    commport: 'none',
    baudrate: 115200,
    ini: "./NowTalk.ini",
    database: './nowTalk.sqlite',
    switchboardName: 'SwitchBoard',
    callname: "computer",
    dynamicExtIP: false,
    externelIP : "",
    allowGuests: true,
    webAddress: "*",//127.0.0.1
    webPort: 1215,
    allowNewDevice: true,
    badgeTimeout: 60,
    reconnectTimeout: 5000,
    connectTimeout: 1500
};

function defaultConfig(args) {
    let inifile = args.ini;
    var newConfig;

    if (inifile && fs.existsSync(inifile)) {
        newConfig = ini.parse(fs.readFileSync(inifile, 'utf-8'));
    } else {
        newConfig = {};
    }

    config.commport = args.port || newConfig.commport || config.commport;
    config.baudrate = makeNumber(args.baud || newConfig.baudrate || config.baudrate);
    ... etc.
}

defaultConfig({ ini: './nowTalk.ini' });

let portnames = []
const ports = await SerialPort.list();
for (const port of ports) {
    portnames.push(port.path);
}


program
    .version(JSON.stringify(pkg.version), '-v, --version', 'Show the current version.')
    .name(JSON.stringify(pkg.name))
    .usage('[options]')
    .description( JSON.stringify(pkg.description) + ' Pressing ctrl+c exits.')
    .addOption(new Option('-p, --port <commport>', 'Commport name of the serial port', config.commport).choices(portnames))
    .option('-b, --baud <baudrate>', 'Used baudrate', config.baudrate)
    // etc.
    .option('--ini <inifile>', 'Alternate ini filename')
    .option('-w --write', 'Write configuration settings to file')
    .parse(process.argv);

const args = program.opts();

defaultConfig(args);

if (args.write) {
    fs.writeFileSync(args.ini, ini.stringify(config));
}
```

And this was what could do in Python:
```python
config = {
    'commport': 'none',
    'baudrate': 921600,
    # etc. 
}

def loadConfig(inifile) :
  global config
  try:
    if (inifile):
      ini = configparser.ConfigParser(defaults=config, allow_unnamed_section=True)
      ini.read(inifile)
      if (ini.has_section(configparser.UNNAMED_SECTION)):
          config = ini[configparser.UNNAMED_SECTION]
  except FileNotFoundError:
    pass

loadConfig('./NowTalk.ini')

parser = argparse.ArgumentParser(description='NowTalk Switchboard Tool', epilog="Copyright 2025, LumenSoft Nederland. ")
parser.add_argument('-port', type=str, help='The COM port to use', choices=list, dest='commport')
parser.add_argument('-baud', type=int, help='Baud rate (921600)', nargs="?", default=config['baudrate'], dest='baudrate')
# etc.
parser.add_argument('--ini', type=str, help='Alternate ini filename')

args = parser.parse_args()
if (args.ini is not None):
    loadConfig(args.ini)
else:
    config = config | vars(args)
```

shadowspawn:
(Thanks for details on scenario. Interesting.)

shadowspawn:
Python argparse documentation for `dest`, per example: https://docs.python.org/3/library/argparse.html#dest

shadowspawn:
Yargs supports option  "alias". This gets used to specify short+long options, but also for alternative names.

https://yargs.js.org/docs/#api-reference-aliaskey-alias

shadowspawn:
I tried filling out the concept with unit tests in #2483.

shadowspawn:
I think the type inference changes which would be needed in [@commander-js/extra-typings](https://github.com/commander-js/extra-typings) are too extensive to add `.setAttributeName()` based on current demand.

However, having confirmed the simple implementation as proposed in this PR does work as hoped, I am open to an example file showing a custom `Option` object with `.setAttributeName()` as discussed in #2423.

Thanks @nielsnl68 for the suggestion and discussion.

Issue #2345, opened 2025-03-22
Migrate source code to ESM?


zanminkian:
Node is defaultly supporting `require('esm-module')`. ESM is the future.

Moreover, pure js project under ESM would be more easy to add type annotation to improve type-safety.

That means commander will not provide CJS distribution in npm package.

Feel free to close it if not planned.

shadowspawn:
I am interested in whether there are benefits in switching to esm-only.

> Moreover, pure js project under ESM would be more easy to add type annotation to improve type-safety.

I am not sure what you mean here. Can you give an example?

stefnotch:
@shadowspawn
There is this post which covers the most important points of why more and more authors are switching to esm only https://antfu.me/posts/move-on-to-esm-only

shadowspawn:
Thanks @stefnotch, I had seen that article and read it again.

shadowspawn:
Lots of notes.

I am excited about ESM becoming more practical to adopt for both leaf projects and libraries with node now supporting `require(esm)`, and reducing the expectations and extra work for libraries to support both CJS and ESM.

From the consumer point of view, Commander transparently supports ESM. There have not been many reports of problems with the CJS implementation, but there was one reported with an esm-only bundling tool (#1781). This might come up more in the future as more tooling moves to esm-only.

From the maintainer point of view, we ship a relatively low-overhead dual package implemented in CJS. There is a very small esm shim: https://github.com/tj/commander.js/blob/master/esm.mjs. We don't do a transpilation step for full dual native CJS/ESM.

The support for correct TypeScript definition support for dual CJS/ESM was a pain! (#1880) That was something I could have done without.

Node.js are reviewing their own documentation and guidance around packaging, and I'll be watching that space: https://github.com/nodejs/package-examples

At the moment I am interested in esm-only, but don't see enough benefits to make the move soon.

stefnotch:
Glad I could point you in a useful direction!

So I suppose the answer to the issue is: The migration will happen, but only once there's a pressing need, or if enough users politely request it.

shadowspawn:
My summary: the migration will happen when there are sufficient benefits, or is likely to happen anyway when there are less downsides.  (The maintainers like moving forward with modern Javascript.)

As a concrete example, if we went esm-only next month when node 18 goes EOL, any CJS users on an LTS version of node 20 (active LTS start 2023-10-24) lower than 20.19.0 would need to upgrade node to 20.19.0 to update Commander. I consider that a downside.

[...]

PR #2464, opened 2025-12-31
Switching to esm-only

shadowspawn:
## Problem

Commander is written in CommonJS with shims for ESM imports and types, and a complex exports in `package.json`. It would be nice to move to ESM now that node supports `require(esm)`.

We have a relatively low overhead CommonJS implementation with the ESM shim. This also avoids the dual-package hazard. However moving forward, there will be more situations where there is friction to support CommonJS implementation in ESM-first tooling (e.g. https://github.com/tj/commander.js/issues/2345#issuecomment-3442907147). 

With `require(esm)` now available in Node.js, and Deno and Bun having supported this for some time, there is solid runtime support for ESM-only dependencies. Hurrah! A major blocker removed.

Some tooling still requires extra work for ESM, with Jest being a common example. Also a familiar example, because we were using it in Commander ourselves! To be fair on Jest, it is stuck waiting for some API issues to be worked through in Node.js, and even upstream in V8. Some popular libraries went ESM-only years ago so the work required to support ESM-only is fairly well established.

Some ESM-only related links:
- https://e18e.dev/blog/migrating-the-ecosystem-to-esm.html
- https://antfu.me/posts/move-on-to-esm-only

Packages gone ESM-only:
- https://storybook.js.org/blog/storybook-is-going-esm-only/
- https://github.com/yargs/yargs/pull/2451

See: #2345

## Solution

Go ESM-only. Document migration tips for common scenarios. Note that Commander 14 will be supported until (approximately) May 2027 so users do not need to update now. (And support for ESM-only is likely to continue to improve in the meantime making an update easier.)

- switch implementation files from CommonJS to ESM
- drop the shim files
- delete deprecated `commander/esm.mjs` entry point
- simplify exports in `package.json`
- add Migration Tips in CHANGELOG as important part of migration

ToDo:
- [x] investigate whether FORCE_COLOR still needed where was added for breakage in Jest
- [x] change `package.json` type to `module`
- [x] convert tests to ESM
- [x] convert (most) example files to ESM
- [x] convert README examples to ESM
- [x] plan for impact on legacy setups, such as CommonJS projects using Jest (familiar example since that was us!)
- [ ] are the Jest `require(esm)` examples worthwhile??
  
If we proceed with this PR:
- [x] add pinned issue announcing plan
- [x] make prerelease available after this PR lands
- [ ] update commander dependency in jest examples to prerelease version

(Note: can try out Jest examples by doing `npm pack` in Commander directory and installing the resulting tar file into example folder, to overwrite the non-existent 15.0.0.)

## ChangeLog

Changed
- *Breaking:* Migrated Commander implementation from CommonJS to pure ESM (ECMAScript Modules). This is expected to be seamless for ESM consumers, but CommonJS consumers may hit issues with tooling requiring configuration for ESM-only dependencies.
- *Breaking:* Commander 15 requires Node.js v20 or higher. To be more precise, `^20.19.0 || >=22.12.0` for `require(esm)`.

Deleted
- *Breaking:* removed deprecated export of `commander/esm.mjs`


shadowspawn:
We put a lot of work into our TypeScript and TypeScript very widely used. Could get rid of the `typings` folder and put types in top level beside `index.js` as expected by default, and move the type tests into `tests/`.

Edit: Yeah, nah. Good enough as is.

shadowspawn:
I did hit some color breakages in tests when I removed `FORCE_COLOR`, so some of the color problems were not due to Jest.

The underlying issue was testing output from `console.log(array)` which does some coloring of array elements based on their type. I added a `JSON.stringify()` wrapper to log plain text.

examples/custom-command-class.js
```diff
@@ -20,7 +20,7 @@ function inspectCommand(command) {
 }
 
 const program = new CommandWithTrace('program')
-  .option('-v, ---verbose')
+  .option('-v, --verbose')
```
shadowspawn, line 23: For interest: I was running each of the example files after updating, and the current version of commander picked up the error in the long option.

lib/command.js
```diff
@@ -2077,8 +2077,8 @@ Expecting one of '${allowedValues.join("', '")}'`);
    * `Option` conflicts with another option.
    *
    * @param {Option} option
-   * @param {Option} conflictingOption
    * @private
+   * @param {Option} conflictingOption
    */
```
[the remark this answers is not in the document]
shadowspawn: Accidental change, reverted.

shadowspawn:
Lots of minor changes to convert implementation and examples and units tests from CommonJS to ESM!

shadowspawn:
As Copilot mentioned in summary, I used `createRequire` in a few tests files to allow existing mocks to keep working by pulling node packages in with `require` instead of `import`. I do not know how the require/import mocking works! The tests pass, and are a little flaky and added to bump up coverage, so not worried for now.

shadowspawn:
Well drat. Reality check. Nothing surprising, but I was being overly optimistic about impact.

I tried using this build of commander in an existing CommonJS project which uses TypeScript and Jest. No problem running program. So that part worked due to Node.js support for `require(esm)`!

But Jest in the consuming project blows up because sees an import in a CommonJS context due to the esm-only implementation of commander. Sigh. Consumers still have problems if they are using Jest, or other tooling that is not yet ESM friendly.

Possible options:
- We could stick with esm-only moving forward, but perhaps extend support for Commander v14 as the version to use if your project setup does not support esm-only, and perhaps provide example changes for Jest.
- We could transpile the esm to CommonJS for a dual build as an interim solution so CommonJS consumers can use latest Commander without upgrading their infrastructure .
- We could abandon conversion and postpone move to esm for years.

Time to reread those articles about moving to ESM and think some more. 😅

shadowspawn:
For my `jest`+`ts-config` setup in a project which depends on Commander there is sane support for handling selected esm-only dependencies, although it took some research to find the appropriate changes.

`jest.config.js` changes

```diff
+const esmModules = ["commander"];
+
 module.exports = {
   globals: {
     JEST_RUNNING: true,
   },
   testEnvironment: "node",
+  // Allow transpile of certain ESM modules from node_modules.
+  // 1. tell ts-jest to (also) handle .js and .mjs files
   transform: {
-    "^.+\\.tsx?$": "ts-jest",
+    "^.+\\.m?[tj]sx?$": [
+      "ts-jest",
+      {
+        useESM: false, // want CJS for Jest
+      },
+    ],
   },
+  // 2. allow transformation of specific node_modules
+  transformIgnorePatterns: [`node_modules/(?!(${esmModules})/)`],
```

shadowspawn:
(Updated original post with more background and explanation.)

examples/alias.cjs
```diff
@@ -28,9 +28,9 @@ program
 program.parse();
 
 // Try the following:
-//    node alias.js --help
-//    node alias.js exec script
-//    node alias.js ex script
-//    node alias.js print file
-//    node alias.js pr file
-//    node alias.js show file
+//    node alias.cjs --help
```
shadowspawn, line 31: Note: I wanted some CommonJS examples still. No particular reason it was alias.js.

examples/arguments-extra.js
```diff
@@ -2,22 +2,19 @@
 
 // This is used as an example in the README for extra argument features.
 
-const commander = require('commander');
-const program = new commander.Command();
+import { Command, Argument } from 'commander';
```
shadowspawn, line 5: Note: I could have kept the "default" (whole) import, but somewhat unusual these days. I switched to more common named import style.

shadowspawn:
Noting a change in approach for mocking imports in tests. I initially used `createRequire` to allow mocking some node library methods from ESM. Since we can freely mix and match CommonJS and ESM in tests, I switched the affected files back to CommonJS (`.cjs`) rather than needing the work-around. (Mocking of ESM is expected to come to `node:test` eventually.)

shadowspawn:
I have added examples of Jest  tests with changes for ESM-only Commander. I have mixed feelings about including specific migration examples, but some people are going to hit them, and we were using Jest, so trying it out.

shadowspawn:
While I was researching whether to go to ESM-only, I was reassured to see these two past articles from 2021 recommending dual support:
- https://antfu.me/posts/publish-esm-and-cjs
- https://dev.to/bcoe/esm-doesn-t-need-to-break-the-ecosystem-4p8b

And the same two authors now moving to esm-only:

- https://antfu.me/posts/move-on-to-esm-only
- https://github.com/yargs/yargs/pull/2451

abetomo (approved):

shadowspawn:

PR #2465, opened 2025-12-31
docs(README): Tweak formatting, punctuation for clarity

Zearin:
Simple copy-edits and small tweaks to the README.

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -530,7 +530,7 @@ You can specify (sub)commands using `.command()` or `.addCommand()`. There are t
 1. Using an action handler attached to the command; or,
 2. as a stand-alone executable file. (More detail about this later.)
 
-Subcommands may be nested. (See example: [nestedCommands.js](./examples/nestedCommands.js).)
+Subcommands may be nested. Example file: [nestedCommands.js](./examples/nestedCommands.js).
 
 In the first parameter to `.command()` you specify the command name. You may append the command-arguments after the command name, or specify them separately using `.argument()`. The arguments may be `<required>` or `[optional]`, and the last argument may also be `variadic...`.
 
--- a/Readme.md
+++ b/Readme.md
@@ -154,7 +154,7 @@ This is used in the examples in this README for brevity.
 const { program } = require('commander');
 ```
 
-For larger programs which may use commander in multiple ways, including unit testing, it is better to create a local Command object to use.
+For larger programs which may use commander in multiple ways, including unit testing, it is better to create a local `Command` object to use.
 
 ```js
 // CommonJS (.cjs)
@@ -176,7 +176,9 @@ const program = new Command();
 
 ## Options
 
-Options are defined with the `.option()` method, also serving as documentation for the options. Each option can have a short flag (single character) and a long name, separated by a comma or space or vertical bar ('|'). To allow a wider range of short-ish flags than just
+Options are defined with the `.option()` method, also serving as documentation for the options. Each option can have a short flag (single character) and a long name, separated by a comma, a space, or a vertical bar (`|`). 
+
+To allow a wider range of short-ish flags than just
 single characters, you may also have two long options. Examples:
 
 ```js
@@ -188,9 +190,9 @@ program
 
 The parsed options can be accessed by calling `.opts()` on a `Command` object, and are passed to the action handler.
 
-Multi-word options such as "--template-engine" are camel-cased, becoming `program.opts().templateEngine` etc.
+Multi-word options, such as `"--template-engine"`, are camel-cased, becoming `program.opts().templateEngine`, for example.
 
-An option and its option-argument can be separated by a space, or combined into the same argument. The option-argument can follow the short option directly or follow an `=` for a long option.
+An option and its option-argument can be separated by a space, or combined into the same argument. The option-argument can follow the short option directly, or follow an `=` for a long option.
 
 ```sh
 serve -p 80
@@ -245,7 +247,7 @@ pizza details:
 ```
 
 Multiple boolean short options may be combined following the dash, and may be followed by a single short option taking a value.
-For example `-d -s -p cheese` may be written as `-ds -p cheese` or even `-dsp cheese`.
+For example, `-d -s -p cheese` may be written as `-ds -p cheese` or even `-dsp cheese`.
 
 Options with an expected option-argument are greedy and will consume the following argument whatever the value.
 So `--id -xyz` reads `-xyz` as the option-argument.
@@ -276,8 +278,8 @@ cheese: stilton
 
 ### Other option types, negatable boolean and boolean|value
 
-You can define a boolean option long name with a leading `no-` to set the option value to false when used.
-Defined alone this also makes the option true by default.
+You can define a boolean option long name with a leading `no-` to set the option value to `false` when used.
+Defined alone, this also makes the option `true` by default.
 
 If you define `--foo` first, adding `--no-foo` does not change the default value from what it would
 otherwise be.
@@ -309,7 +311,7 @@ You ordered a pizza with no sauce and no cheese
 ```
 
 You can specify an option which may be used as a boolean option but may optionally take an option-argument
-(declared with square brackets like `--optional [value]`).
+(declared with square brackets, like `--optional [value]`).
 
 Example file: [options-boolean-or-value.js](./examples/options-boolean-or-value.js)
 
@@ -342,7 +344,9 @@ For information about possible ambiguous cases, see [options taking varying argu
 
 ### Required option
 
-You may specify a required (mandatory) option using `.requiredOption()`. The option must have a value after parsing, usually specified on the command line, or perhaps from a default value (say from environment). The method is otherwise the same as `.option()` in format, taking flags and description, and optional default value or custom processing.
+You may specify a required (mandatory) option using `.requiredOption()`. The option must have a value after parsing, usually specified on the command line, or perhaps from a default value (e.g., from environment). 
+
+The method is otherwise the same as `.option()` in format, taking flags and description, and optional default value or custom processing.
 
 Example file: [options-required.js](./examples/options-required.js)
 
@@ -363,7 +367,7 @@ error: required option '-c, --cheese <type>' not specified
 You may make an option variadic by appending `...` to the value placeholder when declaring the option. On the command line you
 can then specify multiple option-arguments, and the parsed option value will be an array. The extra arguments
 are read until the first argument starting with a dash. The special argument `--` stops option processing entirely. If a value
-is specified in the same argument as the option then no further values are read.
+is specified in the same argument as the option, then no further values are read.
 
 Example file: [options-variadic.js](./examples/options-variadic.js)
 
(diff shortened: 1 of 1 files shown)


shadowspawn:
So many changes! But lots of improvements, thanks. I'll query a few while taking the time to review.

Readme.md
```diff
@@ -176,7 +176,9 @@ const program = new Command();
 
 ## Options
 
-Options are defined with the `.option()` method, also serving as documentation for the options. Each option can have a short flag (single character) and a long name, separated by a comma or space or vertical bar ('|'). To allow a wider range of short-ish flags than just
+Options are defined with the `.option()` method, also serving as documentation for the options. Each option can have a short flag (single character) and a long name, separated by a comma, a space, or a vertical bar (`|`). 
+
```
shadowspawn: The blank line and `Examples:` may make it seem the example will be of short-ish flags. If you think the blank line is helpful, perhaps remove the `Examples:` to avoid that impression.

Readme.md
```diff
@@ -176,7 +176,9 @@ const program = new Command();
...
+To allow a wider range of short-ish flags than just
 single characters, you may also have two long options. Examples:
```
shadowspawn: As above, perhaps:
```suggestion
single characters, you may also have two long options.
```

Readme.md
```diff
@@ -188,9 +190,9 @@ program
 
 The parsed options can be accessed by calling `.opts()` on a `Command` object, and are passed to the action handler.
 
-Multi-word options such as "--template-engine" are camel-cased, becoming `program.opts().templateEngine` etc.
+Multi-word options, such as `"--template-engine"`, are camel-cased, becoming `program.opts().templateEngine`, for example.
```
shadowspawn: I think this is an old line with tweaks over the years! Still clunky. How about a bigger rewrite:

Multi-word options like `--template-engine` are normalized to camelCase option names, resulting in properties such as `program.opts().templateEngine`.

Readme.md
```diff
@@ -521,7 +525,12 @@ $ custom --list x,y,z
 
 ## Commands
 
-You can specify (sub)commands using `.command()` or `.addCommand()`. There are two ways these can be implemented: using an action handler attached to the command, or as a stand-alone executable file (described in more detail later). The subcommands may be nested ([example](./examples/nestedCommands.js)).
+You can specify (sub)commands using `.command()` or `.addCommand()`. There are two ways these can be implemented:
+
+1. Using an action handler attached to the command; or,
+2. as a stand-alone executable file. (More detail about this later.)
+
+Subcommands may be nested. (See example: [nestedCommands.js](./examples/nestedCommands.js).)
```
shadowspawn: I settled on `Example file:` elsewhere, and I think use that here too.

```suggestion
Subcommands may be nested. Example file: [nestedCommands.js](./examples/nestedCommands.js).
```

Readme.md
```diff
@@ -521,7 +525,12 @@ $ custom --list x,y,z
 
 ## Commands
 
-You can specify (sub)commands using `.command()` or `.addCommand()`. There are two ways these can be implemented: using an action handler attached to the command, or as a stand-alone executable file (described in more detail later). The subcommands may be nested ([example](./examples/nestedCommands.js)).
+You can specify (sub)commands using `.command()` or `.addCommand()`. There are two ways these can be implemented:
+
+1. Using an action handler attached to the command; or,
+2. as a stand-alone executable file. (More detail about this later.)
```
shadowspawn: I don't think this expansion is enough of an improvement over the concise statement.

Readme.md
```diff
@@ -553,19 +562,20 @@ program
 
 Configuration options can be passed with the call to `.command()` and `.addCommand()`. Specifying `hidden: true` will
 remove the command from the generated help output. Specifying `isDefault: true` will run the subcommand if no other
-subcommand is specified ([example](./examples/defaultCommand.js)).
+subcommand is specified. (See example: [defaultCommand.js](./examples/defaultCommand.js).)
```
shadowspawn:
```suggestion
subcommand is specified. (Example file: [defaultCommand.js](./examples/defaultCommand.js).)
```

Readme.md
```diff
@@ -553,19 +562,20 @@ program
...
 
-You can add alternative names for a command with `.alias()`. ([example](./examples/alias.js))
+You can add alternative names for a command with `.alias()`. (See example: [alias.js](./examples/alias.js).)
```
shadowspawn:
```suggestion
You can add alternative names for a command with `.alias()`. (Example file: [alias.js](./examples/alias.js).)
```

Readme.md
```diff
@@ -553,19 +562,20 @@ program
...
 For safety, `.addCommand()` does not automatically copy the inherited settings from the parent command. There is a helper routine `.copyInheritedSettings()` for copying the settings when they are wanted.
 
 ### Command-arguments
 
 For subcommands, you can specify the argument syntax in the call to `.command()` (as shown above). This
-is the only method usable for subcommands implemented using a stand-alone executable, but for other subcommands
-you can instead use the following method.
+is the only method usable for subcommands implemented using a stand-alone executable.
+
+For other subcommands, you can instead use the following method.
```
shadowspawn: The "For other subcommands" is in contrast to "implemented using a stand-alone executable" which gets lost with the line break. 

I think this needs a bigger rewrite. Want to have a go? The background is that `.argument()` is newer and somewhat preferred, but you can't use it with stand-alone executables (because there isn't any opportunity for chaining calls).

Readme.md
```diff
@@ -584,8 +594,10 @@ program
   });
 ```
 
- The last argument of a command can be variadic, and only the last argument.  To make an argument variadic you
- append `...` to the argument name. A variadic argument is passed to the action handler as an array. For example:
+The last argument of a command can be variadic, and _only_ the last argument.  To make an argument variadic, simply
+append `...` to the argument name.
+
+A variadic argument is passed to the action handler as an array. For example:
```
shadowspawn: I am neutral on the blank line. If you do like it, I suggest dropping the "for example" as example is for the whole thing.

```suggestion
A variadic argument is passed to the action handler as an array.
```

Readme.md
```diff
@@ -662,7 +674,7 @@ program
   });
 ```
 
-If you prefer, you can work with the command directly and skip declaring the parameters for the action handler. The `this` keyword is set to the running command and can be used from a function expression (but not from an arrow function).
+If you prefer, you can work with the command directly and skip declaring the parameters for the action handler. The `this` keyword is set to the running command, and can be used from a function expression (but not from an arrow function).
```
shadowspawn: I wanted one clause and not two. ChatGPT agreed with you that comma needed as written, but suggested a rewrite to allow no comma:

```suggestion
If you prefer, you can work with the command directly and skip declaring the parameters for the action handler. The `this` keyword is set to the running command and usable from a function expression (but not from an arrow function)..
```

Readme.md
Zearin: Alas, I’m out of time today. You’re welcome to resolve the rewrite as you like! 🙏🏽

Readme.md
Zearin: Suggest the change and I’ll give it a ✅

Zearin: never mind I’ve got a few minutes to do the change :)

Readme.md
Zearin: K.

Readme.md
```diff
@@ -1031,15 +1049,15 @@ program
 
 ### TypeScript
 
-extra-typings: There is an optional project to infer extra type information from the option and argument definitions.
+**extra-typings:** There is an optional project to infer extra type information from the option and argument definitions.
```
shadowspawn, line 1044: Ok, but I suspect my markdown linter is going to complain about using bold as headings. 😆

Readme.md
```diff
@@ -1054,15 +1072,15 @@ const { createCommand } = require('commander');
 const program = createCommand();
 ```
 
-`createCommand` is also a method of the Command object, and creates a new command rather than a subcommand. This gets used internally
+`createCommand()` is also a method of the `Command` object, and creates a new command rather than a subcommand. This gets used internally
 when creating subcommands using `.command()`, and you may override it to
-customise the new subcommand (example file [custom-command-class.js](./examples/custom-command-class.js)).
+customise the new subcommand. See example file: [custom-command-class.js](./examples/custom-command-class.js).
```
shadowspawn:
```suggestion
customise the new subcommand. Example file: [custom-command-class.js](./examples/custom-command-class.js).
```

Readme.md
```diff
@@ -1054,15 +1072,15 @@ const { createCommand } = require('commander');
...
 
 ### Node options such as `--harmony`
 
 You can enable `--harmony` option in two ways:
 
-- Use `#! /usr/bin/env node --harmony` in the subcommands scripts. (Note Windows does not support this pattern.)
+- Use `#! /usr/bin/env node --harmony` in the subcommands scripts. (**Note:** Windows does not support this pattern.)
```
shadowspawn: I don't like the bold here, too noisy.

```suggestion
- Use `#! /usr/bin/env node --harmony` in the subcommands scripts. (Note: Windows does not support this pattern.)
```

Readme.md
Zearin: ChatGPT? Meh.  I think I got a better  rewrite. Just committed.

Readme.md
```diff
@@ -1072,14 +1090,14 @@ An executable subcommand is launched as a separate child process.
...
+If you are using VSCode to debug executable subcommands you need to set the `"autoAttachChildProcesses": true` flag in your `launch.json` configuration.
 
 ### npm run-script
 
-By default, when you call your program using run-script, `npm` will parse any options on the command-line and they will not reach your program. Use
+By default, when you call your program using `run-script`, `npm` will parse any options on the command-line and they will not reach your program. Use
  `--` to stop the npm option parsing and pass through all the arguments.
 
- The synopsis for [npm run-script](https://docs.npmjs.com/cli/v9/commands/npm-run-script) explicitly shows the `--` for this reason:
+ The synopsis for [`npm run-script`](https://docs.npmjs.com/cli/v9/commands/npm-run-script) explicitly shows the `--` for this reason:
```
shadowspawn, line 1092: Side note: I had left out the code formatting in a couple of places you noticed as I thought it might make the link less clear, but looks fine.

Readme.md
Zearin: I squeezed in an attempted edit.

Readme.md
Zearin: I don’t think of it like a heading.  I think of it more like a [`dl` element](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/dl).  

Tell the linter who’s boss. 😈

Readme.md
```diff
@@ -1099,12 +1117,16 @@ program.error('Custom processing has failed', { exitCode: 2, code: 'my.custom.er
...
+By default, Commander calls `process.exit()` when it detects errors, or after displaying the help or version. You can override
 this behaviour and optionally supply a callback. The default override throws a `CommanderError`.
 
-The override callback is passed a `CommanderError` with properties `exitCode` number, `code` string, and `message`.
-Commander expects the callback to terminate the normal program flow, and will call `process.exit` if the callback returns.
-The normal display of error messages or version or help is not affected by the override which is called after the display.
+The override callback is passed a `CommanderError` with the properties 
+- `exitCode` (number), 
+- `code` (string), and 
+- `message` (string).
```
shadowspawn: I went compact on purpose, but agree list is clearer. I think not as a sentence though.

```suggestion
The override callback is passed a `CommanderError` with the properties:
- `exitCode`: number
- `code`: string 
- `message`: string
```

shadowspawn (approved):
Epic. Looks good to me.

abetomo (approved):
:+1:

shadowspawn:
Released in v14.0.3

PR #2473, opened 2026-01-19
Switch tests and examples to esm, leave core CommonJS

shadowspawn:
This is a rework of #2464 to leave the implementation as CommonJS, to avoid impact on consumers with CommonJS tooling that would need work to allow esm-only. Like Jest! Wait another year for tooling commonly used for CLI applications to improve esm-only support! (Front-end tooling is going esm-only and may need support for CommonJS implementation, but that does not affects Commander much.)

Keep most of the changes for tests and example files to esm. Renamed to `.mjs` to work in CommonJS package.

Will write up properly if this looks like the preferred approach.


shadowspawn:
I amleaning towards going ESM-only this year, after Node.js 20 goes EOL on 2026-04-30. I'll make a case for that in #2464 and close this experiment.

PR #2475, opened 2026-01-29
docs: avoid MINGW64/Git Bash slash interpretation

### LINKED ISSUE #2380: Warn users about Bash incompatibility with slash (/) arguments
### Summary

This documentation update adds a warning to help users avoid issues when using Commander CLI tools in **Git Bash / MINGW64** on Windows. In these environments, passing `/` as a flag value (e.g., `--separator "/"`) may result in unexpected behavior due to auto-expansion into `C:/Program Files/Git`.

---

---

Image

Image

taronsung:
## Summary

Add documentation warning about forward slash (`/`) argument expansion in Git Bash (MINGW64) on Windows.

Closes #2380

## Changes

- Added new section "Windows MINGW64 / Git Bash" to both English and Chinese READMEs
- Documents the MSYS2 path conversion behavior that affects slash arguments
- Provides practical workarounds for users encountering this issue

## Workarounds Documented

1. Use a different character instead of `/`
2. Double the slash: `--separator //`
3. Set `MSYS_NO_PATHCONV` environment variable
4. Quote with leading slash: `--separator '///'`

## References

- [MSYS2 Filesystem Paths Documentation](https://www.msys2.org/docs/filesystem-paths/#automatic-unix-windows-path-conversion)

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -91,10 +91,10 @@ console.log(program.args[0].split(options.separator, limit));
 ```
 
 ```console
-$ node split.js -s / --fits a/b/c
+$ node split.js -s - --fits a-b-c
 error: unknown option '--fits'
 (Did you mean --first?)
-$ node split.js -s / --first a/b/c
+$ node split.js -s - --first a-b-c
 [ 'a' ]
 ```
 
@@ -138,7 +138,7 @@ Options:
   -s, --separator <char>  separator character (default: ",")
   -h, --help              display help for command
 
-$ node string-util.js split --separator=/ a/b/c
+$ node string-util.js split --separator=- a-b-c
 [ 'a', 'b', 'c' ]
 ```
 
--- a/Readme_zh-CN.md
+++ b/Readme_zh-CN.md
@@ -88,10 +88,10 @@ console.log(program.args[0].split(options.separator, limit));
 ```
 
 ```console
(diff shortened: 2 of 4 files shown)


shadowspawn:
This is tidy, but there has only ever been one issue about `/` in MINGW64 and I feel documenting the behaviour is a problem for that shell rather than for Commander.

Since the problem came up because a Commander example uses `/` in a parameter (my fault!), I thought reworking the "split" example to use a different character would avoid the problem.

Is that something you would like to do? I think a dash (`-`) should be ok because Commander allows that as a parameter, but an underscore would be very safe too.

taronsung:
Thanks for the suggestion! I've updated the examples to use `-` instead of `/` as the separator character. This avoids the MINGW64 path expansion issue entirely rather than documenting the workaround.

Changes made:
- `examples/split.js`: Changed comments to use `-s - a-b-c`
- `examples/string-util.js`: Changed to `--separator=- a-b-c`  
- `Readme.md`: Updated corresponding example outputs
- `Readme_zh-CN.md`: Updated Chinese README with same changes

shadowspawn (approved):
LGTM
Thanks @taronsung

abetomo (approved):
👍

shadowspawn:

PR #2483, opened 2026-02-07
Add Option.setAttributeName

shadowspawn:
I wanted to fill out tests for #2447 to check working as expected.

## Problem

People sometimes want to override the default camelcase option attribute name.

See: #2423 #2447

## Solution

The implementation is easy because the code uses `.attributeName()` everywhere and does not make assumptions about the mapping from flags to attribute name. But needs lots of tests that the custom name gets used. And could be tricky in `@commander-js/extra-typings`.

To do:
- [ ] decide if adds enough value!
- [ ] add TypeScript
- [ ] document in README
- [ ] can handle custom attribute name in [@commander-js/extra-typings](https://github.com/commander-js/extra-typings)

### DIFF
--- a/lib/option.js
+++ b/lib/option.js
@@ -34,6 +34,7 @@ class Option {
     this.conflictsWith = [];
     this.implied = undefined;
     this.helpGroupHeading = undefined; // soft initialised when option added to command
+    this.customAttributeName = undefined; // override to default behaviour for .attributeName() of camelcase option.name()
   }
 
   /**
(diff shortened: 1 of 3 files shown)


shadowspawn:
I think the type inference changes which would be needed in [@commander-js/extra-typings](https://github.com/commander-js/extra-typings) are too extensive to add `.setAttributeName()` based on current demand.

However, having confirmed the simple implementation does work as hoped, I am open to an example file showing a custom `Option` object with `.setAttributeName()` as discussed in #2423.

PR #2484, opened 2026-02-20
Prepare for prerelease 15.0.0-0

shadowspawn:
Updating the changelog for Commander 15, and bumping the Commander version and minimum Node.js version. Aiming for May release after node 22 goes EOL.

I'll do a prerelease when approved, and put up pinned issues about prerelease and moving to esm-only.


abetomo (approved):
📦

PR #2485, opened 2026-02-21
Use simple match in test (to avoid warning about expensive regex)

shadowspawn:
## Problem

CodeQL reports:

> Polynomial regular expression used on uncontrolled data

for regex `assert(!/\W-h\W.*display help/.test(helpInformation));`

## Solution

This isn't a vulnerability since only used within test with controlled data, but want to fix warning/error anyway.

Just replace regex with some simple string matches.


abetomo (approved):

shadowspawn:

PR #2486, opened 2026-02-21
Use node:util stripVTControlCharacters instead of own code

shadowspawn:
## Problem

I wrote my own `stripColor`, but discovered that node has had [stripVTControlCharacters](https://nodejs.org/docs/latest/api/util.html#utilstripvtcontrolcharactersstr) from Node 16.11.0

## Solution

Use `stripVTControlCharacters`. Easy!

This will strip some special sequences that the custom code did not, so include in a semver major version.

## ChangeLog

- Changed: using `node:util` `stripVTControlCharacters()` for stripping control characters from help when needed


abetomo (approved):

shadowspawn:

PR #2496, opened 2026-03-07
Fix missing return after help() in _dispatchHelpCommand()

abetomo:
## Problem

`_dispatchHelpCommand()` does not return after calling `this.help()` or `subCommand.help()`.
When `exitOverride` is set with a non-throwing callback and `process.exit` is overridden,
execution falls through and calls `_dispatchSubcommand(undefined, ...)`,
which results in a `TypeError: Cannot read properties of undefined (reading '_prepareForParse')`.

## Solution

Added `return` statements after `this.help()` calls in `_dispatchHelpCommand()` so that execution stops after displaying help, regardless of how `exitOverride` is configured.
Added a test to verify the fix.

### DIFF
--- a/lib/command.js
+++ b/lib/command.js
@@ -1380,10 +1380,12 @@ Expecting one of '${allowedValues.join("', '")}'`);
   _dispatchHelpCommand(subcommandName) {
     if (!subcommandName) {
       this.help();
+      return;
     }
     const subCommand = this._findCommand(subcommandName);
     if (subCommand && !subCommand._executableHandler) {
       subCommand.help();
+      return;
     }
 
     // Fallback to parsing the help flag to invoke the help.
--- a/tests/command.helpCommand.test.js
+++ b/tests/command.helpCommand.test.js
@@ -177,4 +177,21 @@ describe('help command processed on correct command', () => {
       program.parse(['help', 'sub'], { from: 'user' });
     }).toThrow('sub help');
   });
+
+  test('when "program help" with non-throwing exitOverride then should not fall through to dispatch', () => {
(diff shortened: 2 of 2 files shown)


shadowspawn:
Can you show how to reproduce the error?

The intent is that anything which normally calls `process.exit()` can not continue through program flow. So we expect `.exitOverride()` to throw, but if it does not we call `process.exit()`  anyway.

In the `.help()` routine there is a call to `this._exit()` here:

https://github.com/tj/commander.js/blob/8247364da749736570161e95682b07fc2d72497b/lib/command.js#L2633

In `._exit()` there is a call to the `exitOverride` callback but if it returns, we call `process.exit()` anyway.

https://github.com/tj/commander.js/blob/8247364da749736570161e95682b07fc2d72497b/lib/command.js#L538

shadowspawn:
With this program:

```js

program.exitOverride(() => {
  console.log('exit override called');
});

program.parse();

console.log('after program');
```

I see this behaviour:
```console
% node index.mjs
after program
% node index.mjs --help
Usage: index [options]

Options:
  -h, --help  display help for command
exit override called
```

abetomo:
Ah, I see. I noticed a test failure while working on the tests and thought it was a bug, but it turns out it was caused by the mock setup.

shadowspawn:
If the mock setup mocked out `process.exit()` and did not throw, that could cause problems.

I used mocks a lot in early Commander work and have been intending to use more built-in support for overriding exit and logging since they got added. In the big test rework on  `release/15.x` branch,  I added `createTestCommand()` to use built-in support for clean tests in a more structured way.

Issue #1619, opened 2021-10-08
[feature request] first-party integration with fig-autocomplete


fedeci:
Hi there, I am currently working on some tools for [Fig autocomplete](https://github.com/withfig/autocomplete) to improve the automatic generation of Fig specs.
One of these is [commander-to-fig](https://github.com/fedeci/commander-to-fig), is there any way we can collaborate to embed this tool directly in commander and provide a better experience to users?

/cc @QuiiBz @brendanfalk @mschrage

shadowspawn:
Related autocompletion issues: #385 #424 #716 #907

shadowspawn:
It looks like Commander-to-fig is a tool that developers might run once if the spec needs further manual editing, or might run as a build step if the output does not require further editing? I am not sure what the developer would usually do with with the resulting spec file. It looks like the possibilities are submit to https://github.com/withfig/autocomplete as a PR (versions?), or publish for team consumption using `fig-teams deploy` (versions?).

I am not sure Commander-to-fig is a good fit for embedding in Commander. In package terms, I see Commander as a production dependency, and  Commander-to-fig as a devDependency. But I am not sure I have understood the developer workflow with Fig and specs.

An interesting approach would be to build the spec from the command help externally. This would have the big workflow advantage that it could be run externally without code modification of the target program. But it would be messier and more fragile to parse the help output than deconstruct the Command object in code.

(As alternatives to embedding, I have wondered about listing Commander add-on packages, or a Commander organisation. But withfig already has a potential location: https://github.com/withfig/autocomplete-tools)

I am interested in general in making it feasible or easier for third-party add-ons and integrations, including auto-complete. Fig auto-complete is potentially straight-forward compared with low-level auto-complete because it is doing most of the command-line parsing itself! Nice.

I would hope that the Commander-to-fig could describe the program without accessing private properties (https://github.com/fedeci/commander-to-fig/blob/401149868decf6fd93c63eaa71c112c554a4dc82/src/index.ts#L68)

A tip for deconstructing the program is use `Help` to get the `. visibleOptions()` and `.visibleCommands()`. This both adds the implicit help option and help command if present, and removes any hidden commands and options.

gutenye:
For anyone seeking a solution, I’ve developed [@gutenye/commander-completion-carapace](https://github.com/gutenye/commander-completion-carapace) using Carapace. It's fast, works well, and supports numerous shells like Bash, Zsh, Fish, Nushell, and others.

Issue #1292, opened 2020-07-03
Commander extensions & middlware


ad-m:
I see Commander as a full-fledged CLI framework.

In my Commander experience, the extension mechanism is missing to introduce the ability to pass specific information to the executed command, or to perform some additional operation before specifying the parameter to modify the context.

In this regard, it is worth noting the "toolbox" mechanism from Gluegun:

> Extensions are additional functionality that you can monkeypatch onto the `toolbox` object. They look something like this:
> 
> ```js
> // extensions/sayhello.js
> module.exports = toolbox => {
>   const { print } = toolbox
> 
>   toolbox.sayhello = () => {
>     print.info('Hello from an extension!')
>   }
> }
> ```
> 
> When you have this extension, you can access it in any command file, like this:
> 
> ```js
> // ...
> run: async toolbox => {
>   const { sayhello } = toolbox
> 
>   sayhello()
> 
>   // or
> 
>   toolbox.sayhello()
> }
> ```

In the same are there is plugins from `structured-cli`:

https://github.com/ggoodman/structured-cli/blob/94542bc0cdd15d93c0415bf637f4a103f78822e9/lib/index.js#L85-L110

Example plugin: 
https://github.com/hyperonecom/h1-cli/blob/master/bin/_plugins/outputFormat/index.js

For example, in the case of developing an API service, it can be used to verify that the user is correctly authenticated before executing the command. As well as for consistent output formatting.

shadowspawn:
Related to the `structured-cli` plugin/middleware and with links to some older issues: #1197

ad-m:
@shadowspawn , thank you for pointing that issue. I use different terms for that idea. Let's continue that discussion in #1197.

PR #2503, opened 2026-04-06
feat: add plugin support using `register`

danielpza:
## Problem

Add a composable plugin interface using the `Command.register(plugin)` api.

The `register` method is similar to how [fastify does plugins](https://fastify.dev/docs/latest/Reference/Plugins/): `fastify.register(plugin)`

This makes it easier to reuse code when creating multiple commands using method chaining.

```js
new Command()
  .option('-c, --compress')
  .command('sub-command')
  .register(prettyErrors)
  .register(setMeta)
  .register(logger)
  .action(() => {
    // noop
  })
  .parse();

// vs

const command = new Command() // forces binding the new command to a variable
  .option('-c, --compress')
  .command('sub-command')
  .action(() => {
    // noop
  })
  .parse();

prettyErrors(command);
setMeta(command);
logger(command);

// or worse
logger(
  setMeta(
    prettyErrors(
      new Command()
        .option('-c, --compress')
        .command('sub-command')
        .action(() => {
          // noop
        })
        .parse(),
    ),
  ),
);
```

## Solution

See the full example in https://github.com/danielpza/commander.js/blob/517749ff6094df484c701aaa3d5513e4e2eb1014/examples/plugins.js

## ChangeLog

Add plugin support using `register` api similar to fastify

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -1151,6 +1151,22 @@ program
   });
 ```
 
+### Composable Plugins using `register`
+
+Use `program.register(plugin)` (from `Command` class) to make it easier to change the behavior. `plugin` is a function that receives the `Command` object.
+
+Example:
+
+```js
+program
+  .register(prettyErrors)
+  .parse();
+
+function prettyErrors(command) {
+  command.configureOutput({/* snip */});
+}
+```
+
 ### Additional documentation
 
 There is more information available about:
--- a/examples/plugins.js
+++ b/examples/plugins.js
@@ -1,0 +1,53 @@
+#!/usr/bin/env node
+
+const { Command } = require('commander');
+
+new Command()
+  .option('-c, --compress')
+  .command('sub-command')
+  .register(prettyErrors)
+  .register(setMeta)
+  .register(logger)
+  .action(() => {
+    // noop
+  })
+  .parse();
+
+function prettyErrors(command) {
+  function errorColor(str) {
+    // Add ANSI escape codes to display text in red.
+    return `\x1b[31m${str}\x1b[0m`;
+  }
+
+  command.configureOutput({
+    // Visibly override write routines as example!
+    writeOut: (str) => process.stdout.write(`[OUT] ${str}`),
+    writeErr: (str) => process.stdout.write(`[ERR] ${str}`),
+    // Highlight errors in color.
+    outputError: (str, write) => write(errorColor(str)),
+  });
+}
+
+function setMeta(command) {
+  command.version('1.2.3');
+}
+
+function logger(command) {
+  command.option('--trace');
+  command.hook('preAction', (thisCommand, actionCommand) => {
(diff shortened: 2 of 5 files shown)


shadowspawn:
This adds a simple convenience method for chaining calls to mutate a `Command` using a passed function.

Thanks for the example code and fastify link. I didn't work out what the fastify  "scope" and "encapsulate" references in their docs cover, but their plugin wrapper adds some additional utility like giving plugin a name and checking host version and checking for required dependent plugins: https://fastify.dev/docs/latest/Reference/Plugins/

I don't want to overpromise and underdeliver by adding a "plugin" method which is a simple convenience callback. Before adding something I would like a clearer idea of what people want to do with plugins and how the interface Commander offers could assist with that and make it safer than a mutate-anything call, and optimistically whether this could allow an ecosystem of plugins as exists with various other packages.

-----

For interest, I have been collecting plugin links and Commander issues and have not previously posted this info. (The last item musing about approaches does match this PR approach.)

- plugins?? Future direction for expansion and customisation?
    - Commander issues
        - https://github.com/tj/commander.js/pull/907#issuecomment-505769990  
        - https://github.com/tj/commander.js/issues/1619  
        - https://github.com/tj/commander.js/issues/1197#issuecomment-653700165
        - https://github.com/tj/commander.js/issues/1292 
    - https://oclif.io/docs/plugins 
        - They can export 3 different types: commands, hooks, and other plugins.
        - hooks: init, rerun, postrun, command_not_found
    - https://github.com/infinitered/gluegun 
    - https://jmcdo29.github.io/nest-commander/docs/features/plugins/ 
    - middy: middleware
    - Cordova has lot of plugins
    - Koa has LOTS of middleware (>200?)
    - fastify: https://www.fastify.io/docs/latest/Write-Plugin/ 
    - https://yarnpkg.com/features/plugins#what-can-plugins-do
        - add resolvers, fetchers, linkers
        - add new commands
        - register to some events
    - https://github.com/getgauge/taiko/wiki/Taiko-Plugin 
    - [ ] https://github.com/unifiedjs/unified#plugin
    - [ ] plugin approach
        - [ ] plugin offers hooks and commands (say), restricts what plugin can do
        - [ ] plugin just runs code and modify command object directly. Allows say calling exitOverride, adding commands and/or options, et al

danielpza:
Hey, that's reasonable. 

Thanks for the detailed list of resources on this topic.

> Before adding something I would like a clearer idea of what people want to do with plugins and how the interface Commander offers could assist with that and make it safer than a mutate-anything call, and optimistically whether this could allow an ecosystem of plugins as exists with various other packages.

Should we create an issue/discussion to start gathering feedback from the community?

shadowspawn:
Yes, welcome to make an issue and look for interest. 👍

danielpza:
Created https://github.com/tj/commander.js/issues/2505#issue-4245588384

Issue #2325, opened 2025-02-06
preAction hook does not work with exitOverride if async


Kpoke:
If an async preAction hook is enabled, exitOverride does not work

example 1: without async

![Image](https://github.com/user-attachments/assets/560bfc1c-d9cb-4424-97f9-bca740934410)

result:
preAction
start action
custom error

example 2: with async

![Image](https://github.com/user-attachments/assets/85087023-0f77-46a8-8d05-c20000ab850d)

result: custom error is not called

preAction
start action
throw new Error();
        ^

Error

shadowspawn:
A tip: thanks for the example code, but it is more useful supplied as text so I can copy it to try reproducing, and so it is searchable in the future.

shadowspawn:
What command are you running on the command line? Perhaps both of:

```console
myProg start
myProg start --unknown
```

The `.exitOverride()` will change the behaviour of the second case but not the first.

shadowspawn:
I think the issue is your try/catch handling is not catching the async reject. If you have `await` available then I like using try/catch in conventional looking way::

```js
try {
  await program.parseAsync();
  console.log("parsed without error");
} catch (error) {
  console.log("caught error");
};

```

Or using callback style:
```js
program.parseAsync().then(() => {
  console.log("parsed without error");
}).catch((error) => {
  console.log("caught error");
});
```

```console
$ node index.mjs start
preAction
start action
caught error
$ node index.mjs start --unknown-option
error: unknown option '--unknown-option'
caught error
```

Kpoke:
You were correct; I used the callback style, and it works. Thank you very much, and thank you for the tip, I'll keep that in mind moving forward.

Issue #1898, opened 2023-06-29
Async action not executed fully


Firgeis:
The following code results in the action stopping in the second await:

```
program.command('action')
  .action(async () => {
    await Promise.resolve();
    console.log("1")
    await new Promise(() => setTimeout(() => {}, 1));
    console.log("2")
  });

program.parseAsync().then(() => {});
```

shadowspawn:
Thanks for the example, and then the simplified example. I'll have a look into it.

shadowspawn:
It might be a problem with your async and promises. I modified the `setTimeout` line and tried this:

```js
const { program } = require('commander')

program.command('action')
  .action(async () => {
    console.log("0")
    await Promise.resolve();
    console.log("1")
    await new Promise(resolve => setTimeout(resolve, 1000, 'dummy'));
    console.log("2")
  });

console.log("before parseAsync")
program.parseAsync().then(() => {});
console.log("after parseAsync")
```

And with node 18 see:
```
% node index.js action
before parseAsync
0
after parseAsync
1
2
```

Firgeis:
Trying to understand that is wrong here, they seem to run fine on their own. Also if they have an issue, is it possible to display it in any way? Why did the line below needed to be changed?
```
await new Promise(() => setTimeout(() => {}, 1));
```

shadowspawn:
It is a valid timer. But the code following the await call is never being called.

The `Promise` and `await` implementation require some help from the author. What `await` effectively does is wrap the following code up into a callback to call later. What `setTimeout` does is call a function later. You have a timeout, but are not calling the resolve callback from the Promise so the code after the await line never gets run.

I find it a bit magical, but _much_ better than writing nested callbacks by hand! `setTimeout` is commonly used in examples since it is easy to understand that code is running later: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/await#

shadowspawn:
> Also if they have an issue, is it possible to display it in any way?

By Commander? No, there is nothing for Commander to detect. You can take Commander out of the example and see the same behaviours.

```js
async function action() {
    console.log("0")
    await Promise.resolve();
    console.log("1")
    await new Promise(resolve => setTimeout(resolve, 1000, 'dummy'));
    console.log("2")
};

console.log("before action call")
action().then(() => {});
console.log("after action call")
```

shadowspawn:
> Why did the line below needed to be changed, seems like a valid timer to me?

[...]

Issue #1777, opened 2022-08-26
The parseAsync method doest not trigger the catch method if an error occurs.


yankos:
I am trying to log an error into a file with winstone module. Winston writes logs asynchronously, so the program must wait a while before exiting. I want to do this in the catch method - catch returns a promise that will be resolved when winstone completes. But the catch method, from the chain after the parseAsync method, is not executed. [Live example](https://replit.com/@yan_kos/Commander-Async-Issue#catch_issue.js)

```
...
program.parseAsync(process.argv)
    .then(
        () => {
            console.log("Done.");
        },
        errorHandler, // <-- here.
    )
    .catch(errorHandler) // <-- or here.
    .finally(() => {
        console.log("Exit.");
    });
...
```

I have:
```
~/Commander-Async-Issue$ node ./catch_issue.js test_as_var
error: unknown command 'test_as_var'
~/Commander-Async-Issue$ node ./catch_issue.js test
error: required option '--arg <arg[,...]>' not specified
~/Commander-Async-Issue$ 
```
Expecting:
```
~/Commander-Async-Issue$ node ./catch_issue.js test_as_var
error: unknown command 'test_as_var'
Error error: unknown command 'test_as_var'
errorHandler finished.
Exit.
~/Commander-Async-Issue$ node ./catch_issue.js test
error: required option '--arg <arg[,...]>' not specified
Error error: required option '--arg <arg[,...]>' not specified
errorHandler finished.
Exit.
```

Is this a bug or am I misunderstanding the documentation and i can't implement my idea?

shadowspawn:
(Thanks for the detail and the full example in the link. Including full examples in the issue itself is fine too, and the example code is very useful for reproducing and getting to the bottom of the problem.)

The default behaviour when Commander finds a parsing error is to display a message and call `process.exit()`, which happens before the `catch`.

To modify the exit behaviour and instead throw, use `.exitOverride()`.

https://github.com/tj/commander.js#override-exit-and-output-handling

(Note also, you are using `.addCommand()` in the linked example code and the exit behaviour is an inherited setting. Either copy the inherited settings like in #1778, or call `.exitOverride()` on the subcommand too.)

shadowspawn:
An answer was provided, and no further activity in a month. Closing this as resolved.

Feel free to open a new issue if it comes up again, with new information and renewed interest.

Issue #1144, opened 2020-01-07
Better error handling for parseAsync()


HNicolas:
Hello,
When using the new method `parseAsync()` it seems that it is not possible to override the default error handling behavior :
* calling a command that does not exist, `parseAsync()` does not throw an error or exit the process, the promise resolve
* calling a command with wrong parameters (missing required parameter) the process exit with code 1

It should be possible to override those default behaviors in order to catch an error like with the sync `parse()` method.

shadowspawn:
> It should be possible to override those default behaviors in order to catch an error like with the sync parse() method.

`parseAysnc()` calls `.parse()`, and you override the default behaviours the same way. I am guessing you have not done that, and would see the same error handling behaviours as you describe if you called `.parse()` too.

> calling a command that does not exist, parseAsync() does not throw an error or exit the process, the promise resolve

Commander sees a command that does not exist as an argument that you don't happen to handle. (Related collection of issues: #1088)

The current approach is to add a custom event listener and there is an example in README:
https://github.com/tj/commander.js#custom-event-listeners

> calling a command with wrong parameters (missing required parameter) the process exit with code 1

See `.exitOverride()`: https://github.com/tj/commander.js#override-exit-handling

HNicolas:
I didn't know that it was the correct behavior for when calling a command that does not exist.
But there is still a problem when calling `parseAsync()` with wrong parameters, I do call .exitOverride() but the process still exit without throwing an error or calling the provided callback if any.
Example of code :
```typescript
    const program = new Command();
    program.command("test <req>")
      .action(async () => {
        // do some async work
      });
    program.exitOverride()
      .parseAsync(process.argv)
      .catch(error => {
        // not reached on invalid command (missing req)
      });
```

shadowspawn:
To get the exitOverride to apply to the subcommand too, define it before adding the command. Then to catch the error I think you need to be in an async context when it is thrown.

```
const commander = require('commander');
const program = new commander.Command();

program
  .exitOverride()
  .command("test <req>")
  .action(async () => {
    // do some async work
  });

async function main() {
  return program.parseAsync(['node', 'my.js', 'test']); // Oops, req missing
}

main().then(() => {
  console.log('main.then');
})
.catch(error => {
  console.log('main.caught');
  console.log(error);
});
```

```
$ node index.js 
error: missing required argument 'req'
main.caught
CommanderError: error: missing required argument 'req'
...
```

HNicolas:
[...]

PR #2504, opened 2026-04-11
docs: add async error handling example

vincent067:
Hi there! 👋

First off, thanks for maintaining such an awesome library! Commander.js has been my go-to for CLI tools for years.

I noticed that while the README mentions using `parseAsync()` for async action handlers, there's no complete example showing how to properly handle errors in async contexts. This is a pretty common pain point for developers new to the library.

This PR adds an example that demonstrates:

- ✅ Proper error handling in async actions (without crashing the CLI)
- ✅ A simple retry mechanism with the `--retry` flag
- ✅ Batch processing with both "continue on error" and "fail fast" modes
- ✅ Correct exit codes for CI/automation compatibility
- ✅ Why `parseAsync()` is important (with `.catch()` handler)

Example usage:

```bash
# Retry failed requests
$ node async-error-handling.js fetch error --retry 3
Retry attempt 1/3...
Retry attempt 2/3...
Retry attempt 3/3...
Error: Failed to fetch data for ID: error

# Batch processing with partial failure
$ node async-error-handling.js batch 1 error 3
✓ 1
✗ error: Failed to fetch data for ID: error
✓ 3

--- Summary ---
Successful: 2
Failed: 1
```

The example is designed to be:
- **Practical**: Shows real-world patterns developers actually need
- **Educational**: Well-commented, explaining the "why" not just "how"
- **Runnable**: Fully functional code you can execute immediately

I'm open to any feedback or changes you'd prefer! 😊

Thanks for considering!

### DIFF
--- a/examples/async-error-handling.js
+++ b/examples/async-error-handling.js
@@ -1,0 +1,109 @@
+#!/usr/bin/env node
+
+// This example demonstrates proper error handling in async action handlers.
+// It shows how to handle errors gracefully without crashing the CLI.
+
+const { Command } = require('../');
+const program = new Command();
+
+// Simulate an async operation that might fail
+async function fetchData(id) {
+  return new Promise((resolve, reject) => {
+    setTimeout(() => {
+      if (!id || id === 'error') {
+        reject(new Error(`Failed to fetch data for ID: ${id}`));
+      } else {
+        resolve({ id, name: `Item ${id}`, timestamp: new Date().toISOString() });
+      }
+    }, 100);
+  });
+}
+
+program
+  .name('async-error-demo')
+  .description('Demonstrates async error handling patterns in Commander.js')
+  .version('1.0.0');
+
+program
+  .command('fetch <id>')
+  .description('Fetch data by ID (use "error" to simulate failure)')
+  .option('-r, --retry <count>', 'number of retries on failure', '0')
+  .action(async (id, options) => {
+    const maxRetries = parseInt(options.retry, 10);
+    let lastError;
+
+    for (let attempt = 0; attempt <= maxRetries; attempt++) {
+      try {
+        if (attempt > 0) {
+          console.log(`Retry attempt ${attempt}/${maxRetries}...`);
+        }
+        
+        const data = await fetchData(id);
+        console.log('Success:', data);
+        return; // Success - exit the action
+      } catch (error) {
+        lastError = error;
+        
+        if (attempt < maxRetries) {
+          // Wait before retrying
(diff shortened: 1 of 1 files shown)


shadowspawn:
> there's no complete example showing how to properly handle errors in async contexts. This is a pretty common pain point for developers new to the library.

Have you got any references to new developers having problems? 

Most of the problems people have reported in issues are:
- using asynchronous at all, rather than with Commander
- not using `.exitOverride()` when expecting to catch that as an error
- dangling promises in their own code and program not terminating

However, having an async error example is still a reasonable suggestion! I have different ideas about what to focus on to discuss.

Related past issues: #1144 #1777 #1898 #2325

vincent067:
Hi @shadowspawn, thanks so much for the thoughtful feedback! 🙏

To be completely honest, I don't have hard references or issue links to back up my claim about &quot;new developers having problems&quot; — that was more based on my own personal experience when I first started using Commander with async actions. I remember being a bit confused about whether to use <code>parse()</code> or <code>parseAsync()</code>, and my script would just silently hang or crash in ways I didn't expect.

Looking at the related issues you linked (#1144, #1777, #1898, #2325), I can see the actual pain points are a bit different from what I focused on here — especially the <code>.exitOverride()</code> situation and dangling promises causing the program to not terminate. That's really good insight, and honestly I hadn't fully appreciated how important the dangling promise issue is until now.

I'd love to revise this example to focus on the things that actually trip people up. Would you be open to me updating the PR to:

1. Emphasize <code>parseAsync()</code> vs <code>parse()</code> more clearly
2. Show how <code>.exitOverride()</code> interacts with async actions (if that's useful in an example)
3. Demonstrate proper promise handling so the program terminates cleanly
4. Remove or de-emphasize the retry/batch logic since that might be adding noise

I'm totally happy to pivot this in whatever direction you think would be most helpful for the examples directory. Let me know what you'd prefer to see! 😊

shadowspawn:
Yes please, a pivot. I do think the retry/batch is a bit noisy. Keep the "program" simple and focus on the asynchronous patterns.

I wonder whether should have two examples to show the two styles of outer code? One using try/await main()/catch (in order processing) and one using main().then().catch() (out of order processing). They felt quite different to me when I first used async. The command parser could be the same (copy/paste).

shadowspawn:
No activity in a month, closing. But can open a new PR or come back to this one in future. From issues I do see users struggle with async, but this PR was a bit heavy-weight and opinionated (e.g. retry is too much).

shadowspawn:

PR #2509, opened 2026-04-23
fix: skip help text wrapping when output is not a TTY

### LINKED ISSUE #2508: Bug: Option description shouldn't be wrapped if `process.stdout.columns` is `undefined`

mixelburg:
Fixes #2508

## Problem

When command output is piped (not a TTY), `process.stdout.columns` is `undefined`. Commander was defaulting to 80 columns for wrapping help text, causing option descriptions to be unnecessarily broken across lines.

## Fix

- In `prepareContext()`: remove the `?? 80` fallback so `helpWidth` stays `undefined` when no TTY width is available
- In `formatItem()`: when `helpWidth` is `undefined`, skip wrapping entirely (no line breaks in descriptions)
- Updated test to verify `boxWrap` is not called when `helpWidth` is `undefined`

## Behavior change

- **TTY (terminal)**: wrapping still works as before using actual terminal width
- **Piped output**: descriptions no longer wrapped — full text on each line
- **Custom `helpWidth`**: still respected as before

All 1367 tests pass.

### DIFF
--- a/lib/help.js
+++ b/lib/help.js
@@ -27,7 +27,7 @@ class Help {
    * @param {{ error?: boolean, helpWidth?: number, outputHasColors?: boolean }} contextOptions
    */
   prepareContext(contextOptions) {
-    this.helpWidth = this.helpWidth ?? contextOptions.helpWidth ?? 80;
+    this.helpWidth = this.helpWidth ?? contextOptions.helpWidth;
   }
 
   /**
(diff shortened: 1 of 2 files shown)


shadowspawn:
To be clear: is this issue entirely AI/LLM generated?

shadowspawn:
The `helpWidth` is used to wrap the command description at:

https://github.com/tj/commander.js/blob/8247364da749736570161e95682b07fc2d72497b/lib/help.js#L463

shadowspawn:
This PR is not complete, and no engagement since being opened. May have been AI submitted with no follow-through.

Issue #2487, opened 2026-02-21
Commander v15 going ESM in May 2026. Prerelease available now.


shadowspawn:
The big changes in Commander 15 are:
- switching from CommonJS implementation to ESM implementation
- requiring Node.js v22.12.0 or higher

A prerelease is available now. See the release notes for all the details: https://github.com/tj/commander.js/releases/tag/v15.0.0-0

The release of Commander 15.0.0 will be in May 2026, after Node.js 20 goes EOL.

joaopedrodcf:
With checking the release notes and noticed that commanded 15 kept jest https://github.com/tj/commander.js/tree/release/15.x/examples/using-esm-from-commonjs/jest-javascript do you guys plan on moving to vitest as it's ESM first, just curious 🙏

shadowspawn:
Commander switching to `node:test`: #2463

joaopedrodcf:
Ah thanks for sharing 🙏

shadowspawn:
Released in v15: https://github.com/tj/commander.js/releases/tag/v15.0.0

HyphensPeciwse33:
#2565

PR #2516, opened 2026-05-10
fix: avoid wrapping help for non-tty output

### LINKED ISSUE #2508: Bug: Option description shouldn't be wrapped if `process.stdout.columns` is `undefined`

innoprej:
## Summary

This changes Commander so help output avoids applying the default 80-column wrapping width when the configured output stream does not report terminal columns.

When help is being redirected to a pipe or file-like stream, `outputConfiguration.getOutHelpWidth()` now returns `Infinity` instead of falling back to 80 columns. That leaves descriptions unwrapped unless a width is explicitly configured.

Fixes #2508

## Validation

- `npm test -- command.configureOutput.test.js`
- `npm test`

## Disclosure

Prepared with AI/LLM assistance. I reviewed the patch before opening the PR and ran the validation listed above.

### DIFF
--- a/lib/command.js
+++ b/lib/command.js
@@ -63,9 +63,9 @@ class Command extends EventEmitter {
       writeErr: (str) => process.stderr.write(str),
       outputError: (str, write) => write(str),
       getOutHelpWidth: () =>
-        process.stdout.isTTY ? process.stdout.columns : undefined,
+        process.stdout.isTTY ? process.stdout.columns : Infinity,
       getErrHelpWidth: () =>
-        process.stderr.isTTY ? process.stderr.columns : undefined,
+        process.stderr.isTTY ? process.stderr.columns : Infinity,
       getOutHasColors: () =>
         useColor() ?? (process.stdout.isTTY && process.stdout.hasColors?.()),
       getErrHasColors: () =>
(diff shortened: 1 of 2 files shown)


shadowspawn:
To be clear: is this PR entirely AI/LLM generated?

shadowspawn:
Note: Commander 15 switches from Jest to `node:test` for the test runner.

See: #2487

innoprej:
Yes, this PR was generated with AI/LLM assistance, including the initial patch and PR description. I reviewed the change before opening the PR and ran the validation listed in the PR body, but I should have made the AI assistance explicit.

You are also right to flag the Commander 15 test runner change. This PR currently targets `master`, where the tests are still Jest-based, but I can adapt the tests to the `node:test`/`node:assert` style if you would prefer this change to be based on `release/15.x` or otherwise prepared for Commander 15.

I also noticed a wording mismatch in my PR description: the implementation returns `Infinity` for non-TTY help width, not `undefined`. I will update the PR description to match the actual implementation.

shadowspawn:
Normally we target Pull Requests at the `develop` branch, but currently the best branch to target is `release/15.x` which has the upcoming 15.x changes. Depending on progress on this PR and that release, we might land the PR on that branch or retarget to `develop` if the release gets merged into develop and/or released.

shadowspawn:
Do you want to have another go at this @innoprej ?

shadowspawn:
I did some research and decided not to change the existing behaviour.

See https://github.com/tj/commander.js/issues/2508#issuecomment-4814699048

PR #2519, opened 2026-05-22
Update details for 15.0.0 release

shadowspawn:
Preparing for 15.0.0 release. I put in a date of next weekend, but we could do it this weekend if you like @abetomo ?  Just when suits you, no rush.

For interest:
- the `15.0.0-0` version has had 4162 downloads from [npmjs](https://www.npmjs.com/package/commander?activeTab=versions), which is our most ever for a prerelease. 
- v13 of Commander went EOL 4 days ago.


abetomo (approved):
👍

PR #2521, opened 2026-05-26
Pin GitHub actions with hash

shadowspawn:
## Problem

1) Best practice to lower risk of supply chain attacks is to pin GitHub Actions to hash.

2) Can't manually run actions on branches (e.g. run CodeQL on `release/15.x`)

## Solution

1) Switch to hashes using `npx actions-up`. In theory, dependabot will update the cases and maintain the version number comment from now on, so this is a one-shot.

https://github.com/azat-io/actions-up

2) Add `workflow_dispatch` to events for actions.

https://docs.github.com/en/actions/how-tos/manage-workflow-runs/manually-run-a-workflow#configuring-a-workflow-to-run-manually

## References

https://nesbitt.io/2026/04/28/github-actions-is-the-weakest-link.html

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -8,6 +8,7 @@ on:
     branches: [ master, develop ]
   schedule:
     - cron: '27 5 * * 6'
+  workflow_dispatch:
 
 jobs:
   analyze:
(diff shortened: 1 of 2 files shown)


shadowspawn:
(Sorry about the multiple PR in short period! I have some time off after a busy time at work, and doing some open source maintenance. 😅 )

abetomo (approved):

PR #2523, opened 2026-05-29
Fix release dates in changelog

abetomo:


shadowspawn (approved):
Thanks! I was copying and pasting lots of dates.

PR #2531, opened 2026-06-17

### LINKED ISSUE #2530: Executable subcommand does not preserve end-of-options delimiter before option-like operands
### Description

Related PR: #2531

### Reproduction

Parent:

```js

new Command()
  .name('parent')
    executableFile: 'child.js',
  })
  .parse();
```

Child:

```js

new Command()
  .name('parent child')
  .argument('[input]')
  .parse();
```

Run:

```
```

### Expected

### Actual

### Notes

yminod:
## Summary

Fixes #2530

This fixes stand-alone executable subcommand dispatch so the first end-of-options delimiter `--` is preserved in the argv passed to the child process.

## Approach

The change keeps normal in-process parsing behavior unchanged. `--` is not added back as a normal operand for action subcommands.

Instead, `parseOptions()` records the position of the first operand-side end-of-options delimiter as internal metadata. `_parseCommand()` and `_dispatchSubcommand()` carry that index through subcommand dispatch. When the target is an executable subcommand, Commander restores `--` at the recorded position before calling `child_process.spawn()`.

This preserves the original command-line boundary for delegated CLIs without changing in-process command parsing.

## Behavior covered

- executable subcommands receive the `--` delimiter before option-like operands
- argument order around the delimiter is preserved
- a later `--` after the first delimiter remains an operand
- existing executable subcommand argument ordering without `--` is unchanged

## Validation

- `npm run test`
- `npm run check`

## Notes

Targeting `develop` per `CONTRIBUTING.md`. Happy to retarget if maintainers prefer another branch.

Prepared with AI/LLM assistance. I reviewed the change and ran the validation listed above.

### DIFF
--- a/lib/command.js
+++ b/lib/command.js
@@ -11,6 +11,8 @@ import { Help } from './help.js';
 import { Option, DualOptions } from './option.js';
 import { suggestSimilar } from './suggestSimilar.js';
 
+const OptionTerminatorSymbol = Symbol('optionTerminator');
+
 export class Command extends EventEmitter {
   /**
    * Initialize a new `Command`.
@@ -1361,7 +1363,7 @@ Expecting one of '${allowedValues.join("', '")}'`);
    * @private
    */
 
-  _dispatchSubcommand(commandName, operands, unknown) {
+  _dispatchSubcommand(commandName, operands, unknown, optionTerminatorIndex) {
     const subCommand = this._findCommand(commandName);
     if (!subCommand) this.help({ error: true });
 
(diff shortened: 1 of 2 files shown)


shadowspawn:
This is too heavy-weight for an issue that has not previously been reported. I will leave the issue open for at least a month to see if gathers any likes or comments.

Good human-readable description and detail. I appreciate the AI/LLM mention.

yminod:
Thanks. I understand that the compatibility/weight tradeoff needs more consideration.

PR #2533, opened 2026-06-17
fix: use Infinity instead of 80 for helpWidth when columns is undefined (#2508)

### LINKED ISSUE #2508: Bug: Option description shouldn't be wrapped if `process.stdout.columns` is `undefined`

rafaumeu:
Fixes #2508

When piping command output, `process.stdout.columns` is undefined, causing help text to be wrapped to 80 chars. Using Infinity prevents unwanted wrapping in pipe scenarios.

**Changes:**
- `prepareContext()`: 80 → Infinity
- `formatHelp()`: 80 → Infinity  
- `formatItem()`: 80 → Infinity
- Tests added in `command.infinite-width-piping.test.js`

### DIFF
--- a/lib/help.js
+++ b/lib/help.js
@@ -28,7 +28,7 @@ export class Help {
    * @param {{ error?: boolean, helpWidth?: number, outputHasColors?: boolean }} contextOptions
    */
   prepareContext(contextOptions) {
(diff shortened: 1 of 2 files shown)


shadowspawn:

shadowspawn:
Have an existing PR which I'll give first change to: #2516

Also appears heavily AI, and not a human responding.

PR #2537, opened 2026-06-20
docs: add Deno to install instructions

bartlomieju:
👋 Bartek from the Deno team here.

The install instructions list npm, but not Deno. Since Deno is a drop-in replacement for npm these days, I'd like to add a Deno line so Deno users have a copy-pasteable command, in parity with the others:

```sh
deno install commander
```

(As of Deno 2.8 there's no `npm:` prefix needed, it's the same command shape as `npm install commander`.)

Totally fine to close this if you'd rather keep the list short, no hard feelings. Happy to adjust wording, placement, or move it to a docs site instead.

_Disclosure: I work on Deno, so I have an interest here. The goal is parity with the package managers you already list, not promotion. This PR was prepared with AI assistance under human supervision: I review every change myself and personally respond to any comments or review feedback._

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -62,6 +62,7 @@ For information about terms used in this document see: [terminology](./docs/term
 
 ```sh
 npm install commander
+deno install commander
 ```
 
 ## Quick Start
--- a/Readme.md
+++ b/Readme.md
@@ -62,6 +62,11 @@ For information about terms used in this document see: [terminology](./docs/term
 
 ```sh
 npm install commander
+```
+
+Or using [Deno](https://deno.land):
+
+```sh
 deno install commander
 ```
 


Readme.md
```diff
@@ -62,6 +62,7 @@ For information about terms used in this document see: [terminology](./docs/term
 
 ```sh
 npm install commander
+deno install commander
```
abetomo, line 70:
````suggestion
```

Or using [Deno](https://deno.land):

```sh
deno install commander
````

Thanks!
Since `npm` and `deno` are alternatives, the single block makes it look like you run both.
How about splitting them?

Comment (bartlomieju) [reply Readme.md:70]: Good call, done in d0f4f41 — split into two separate blocks with an "Or using Deno" line between them. Thanks for the review!

**Comment by shadowspawn** [comment]
Thanks for being clear about affiliation and AI use.

I have some sympathy, and wondered about this myself: https://github.com/tj/commander.js/issues/1803#issuecomment-1327934790

> in parity with the others

Indeed, what about `bun` and `pnpm` and `yarn`? Looking at raw hits in issues here: deno (4), bun (6), pnpm (5), and yarn (17) is the most mentioned. 

I'll think some more about this.

**Comment by shadowspawn** [comment]
The place we do have multiple examples is the `program` section of the README: https://github.com/tj/commander.js#declaring-program-variable

**Comment by shadowspawn** [comment]
I do like clarifying that does work with Deno and Bun. This isn't as up-front as the installation instruction, but does add explicit mention for searches and reassurance that works (in https://github.com/tj/commander.js#declaring-program-variable). How about:

```diff
- // TypeScript (.ts)
+ // TypeScript (.ts), including Bun and Deno
import { Command } from 'commander';
const program = new Command();
```

bartlomieju:
@shadowspawn it's totally fine to close this PR - actually it shouldn't have been opened in the first place - in my prompt it slipped that PRs should be proposed only to repos that have more than just `npm install ...` mentioned in the readme. So your call, I'm happy to do it either way.

shadowspawn:
For future interest, some of the other Pull Requests are for projects with multiple alternatives, including some where presented together with introduction, and some as reveal-your-flavour:
- https://github.com/motdotla/dotenv/pull/1025/changes
- https://github.com/dcastil/tailwind-merge/pull/687/changes

We only have one installation method listed, and I don't want to go heavy on installation variations at this time, and if we did would want to be clear that alternatives (as @abetomo mentioned).

Since we ended up having this conversation...

@bartlomieju Is it a positive for Deno to be mentioned in the TypeScript `program` example, or is TypeScript support a given for Deno and not helpful or helpful enough to proceed?

bartlomieju:
> @bartlomieju Is it a positive for Deno to be mentioned in the TypeScript program example, or is TypeScript support a given for Deno and not helpful or helpful enough to proceed?

I'm not entirely sure which section your mentioning exactly, that said, it's mostly a given that TS works in Deno.

shadowspawn:
I'll close this, as somewhat missed the target.

PR #2545, opened 2026-07-02
Update dependencies

shadowspawn:
The `prettier` update introduced one small format change (dependabot PR failed tests) so do a manual update to pick up current updates and apply the format change.

Just ran:
```console
npm update
npm run fix:format
```


abetomo (approved):
👍

PR #2552, opened 2026-07-14
Unskip .ts subcommand lookup test now that node 20 is EOL

damienmeur:
## Problem

The `.ts` executable-subcommand lookup test was skipped in the ESM-only migration (#2464) because it failed on Node 20, which does not run `.ts` files.

## Solution

Node 20 is now EOL and was dropped from the test matrix in #2520, and all current matrix versions (22.x, 24.x, 26.x) run the plain-JS `.ts` fixture via default type stripping. This un-skips the test and removes the stale comment; it passes locally on Node 22.22.2, 24.15.0, and 26.3.0.

## ChangeLog

Test-only change, no CHANGELOG entry needed.

### DIFF
--- a/tests/command.executableSubcommand.lookup.test.js
+++ b/tests/command.executableSubcommand.lookup.test.js
@@ -90,8 +90,7 @@ describe('executable subcommand lookup ', () => {
     },
   );
 
-  // This test currently fails on Node 20 when run as ESM because Node does not execute .ts files as JavaScript by default.
-  test.skip('when subcommand suffix is .ts then lookup succeeds', async () => {
+  test('when subcommand suffix is .ts then lookup succeeds', async () => {
     // We support looking for ts files for ts-node in particular, but don't need to test ts-node itself.
     // The subcommand is both plain JavaScript code for this test.
     const binLinkTs = path.join(


shadowspawn:
This looks simple and tidy but...

damienmeur:
Fair question. The change itself is a one-liner (un-skip + drop a stale comment), so there wasn't much to generate — but yes, the PR text was AI-drafted and I used AI tooling to prepare it.

The backstory: I'm building a code-analysis tool (statistical, not an LLM), and commander has been one of the real-world repos I benchmark it against. It flagged this test as skipped with a stale reason. Commander's history has been genuinely useful to me, so fixing the things it surfaces felt like a fair way to give something back.

I did verify before filing: the un-skipped test passes 14/14 locally on Node 22.22.2, 24.15.0 and 26.3.0 — the current CI matrix — and Node 20 was dropped as EOL in #2520. Happy to close if you'd rather not take it.

shadowspawn:
I regularly ask this question because the human effort can be very mismatched with AI generated Pull Requests and a human reviewer. If heavy AI, the AI may be out of credit or disconnected by the time I ask questions if needed, so I limit my effort.

Thanks for the back story. Glad history proving useful. :-)

shadowspawn (approved):
LGTM, I had forgotten about that skipped test.

PR #2553, opened 2026-07-15

### LINKED ISSUE #2530: Executable subcommand does not preserve end-of-options delimiter before option-like operands

Hashim1999164:
## Summary

Fixes #2530

Executable subcommand dispatch now keeps the first end-of-options delimiter (`--`) in the argv passed to the child process.

## Approach

Normal in-process parsing is unchanged. `parseOptions()` records where the first operand-side `--` appeared. `_parseCommand()` and `_dispatchSubcommand()` carry that index through subcommand dispatch. For executable subcommands only, Commander splices `--` back at that position before `child_process.spawn()`.

## Test plan

- [x] `npm test` (2 unrelated `FORCE_COLOR` failures in this environment)
- [x] `node --test tests/command.executableSubcommand.lookup.test.js`
- [x] New cases: delimiter before option-like operand, operands around delimiter, repeated `--`

### DIFF
--- a/lib/command.js
+++ b/lib/command.js
@@ -11,6 +11,8 @@ import { Help } from './help.js';
 import { Option, DualOptions } from './option.js';
 import { suggestSimilar } from './suggestSimilar.js';
 
+const OptionTerminatorSymbol = Symbol('optionTerminator');
+
 export class Command extends EventEmitter {
   /**
    * Initialize a new `Command`.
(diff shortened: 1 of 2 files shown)


shadowspawn:

Is this Pull Request materially different than #2531 ?

Hashim1999164:
@shadowspawn  yes it is. :) . I believe the changes are same as #2531

HyphensPeciwse33:
@shadowspawn Lets Reopen

PR #2558, opened 2026-07-23
feat: allow custom env bag in parse options

### LINKED ISSUE #2549: Allow passing custom env instead of process.env

Errr0rr404:
## Summary
- Add `parseOptions.env` so `.env()` option values can be resolved from a caller-provided object instead of `process.env`.
- Useful for tests without mutating the real process environment.
- Updates TypeScript typings and adds tests.

Fixes #2549

## Test plan
- [x] `node --test tests/options.env.test.js` (41 passing)

### DIFF
--- a/lib/command.js
+++ b/lib/command.js
@@ -57,6 +57,8 @@ export class Command extends EventEmitter {
     this._showHelpAfterError = false;
     this._showSuggestionAfterError = true;
     this._savedState = null; // used in save/restoreStateBeforeParse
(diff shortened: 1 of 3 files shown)


Errr0rr404:
Closing this PR after re-reading the issue thread.

@shadowspawn already indicated there isn’t enough value in a custom env override vs the existing low-tech pattern of setting/deleting `process.env` in tests (and noted #1762). Opening a full feature PR against that guidance was a mistake on my part — sorry for the noise.

Thanks for keeping the issue open for interest; if the community shows demand later, a fresh PR can be considered then.

PR #2559, opened 2026-07-23

### LINKED ISSUE #2530: Executable subcommand does not preserve end-of-options delimiter before option-like operands

Hashim1999164:
## Summary
- Fixes #2530 by keeping the first end-of-options delimiter when spawning executable subcommands
- Records where `--` appeared during option parsing and splices it back into argv before `child_process.spawn()`
- Adds regression tests for delimiter preservation with option-like operands

## Test plan
- [x] `node --test tests/command.executableSubcommand.lookup.test.js`
- [x] Verified `pm echo -- --not-an-option` passes `--` through to the child process

### DIFF
--- a/lib/command.js
+++ b/lib/command.js
@@ -11,6 +11,8 @@ import { Help } from './help.js';
 import { Option, DualOptions } from './option.js';
 import { suggestSimilar } from './suggestSimilar.js';
 
+const OptionTerminatorSymbol = Symbol('optionTerminator');
+
 export class Command extends EventEmitter {
   /**
    * Initialize a new `Command`.
@@ -1361,7 +1363,7 @@ Expecting one of '${allowedValues.join("', '")}'`);
    * @private
    */
 
-  _dispatchSubcommand(commandName, operands, unknown) {
+  _dispatchSubcommand(commandName, operands, unknown, optionTerminatorIndex) {
     const subCommand = this._findCommand(commandName);
     if (!subCommand) this.help({ error: true });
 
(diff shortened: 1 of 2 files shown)


shadowspawn:
AI slop, one of 138 PR opened in July by this account.

And to add insult to injury, same PR from same account was opened a week prior and already closed. #2553

HyphensPeciwse33:
```js
import fs from 'fs';

program
  .name('file-manager')
  .description('A modified CLI utility for processing files')
  .version('2.0.0');

// Modified command structure with custom error validation
program.command('modify')
  .description('Modify file content by replacing target text')
  .argument('<file>', 'path to the target file')
  .requiredOption('-t, --target <string>', 'text string to find and replace')
  .option('-r, --replacement <string>', 'new text to insert instead', '')
  .option('-v, --verbose', 'display detailed execution logs')
  .option('-g, --global', 'replace all occurrences')
  .action((file, options) => {
    try {
      // Validate file exists
      if (!fs.existsSync(file)) {
        console.error(`Error: File not found: ${file}`);
        process.exit(1);
      }

      if (options.verbose) {
        console.log(`Reading file from path: ${file}`);
        console.log(`Replacing matches of "${options.target}" with "${options.replacement}"`);
      }

      // Read file content
      const content = fs.readFileSync(file, 'utf-8');

      // Perform replacement
      const modified = options.global
        ? content.replaceAll(options.target, options.replacement)
        : content.replace(options.target, options.replacement);

      // Write modified content back
      fs.writeFileSync(file, modified, 'utf-8');

      if (options.verbose) {
        console.log(`File content updated successfully`);
      }
      console.log(`Successfully modified: ${file}`);
    } catch (error) {
      console.error(`Error: ${error.message}`);
      process.exit(1);
    }
  });

program.parse();
```

PR #2560, opened 2026-07-24
Improve tests following lint advice from eslint-node-test

shadowspawn:
## Problem

I transferred a lot of tests to a new testing framework, which I learnt as I went. There is some room for improvement. 😊 

## Solution

Install https://github.com/sindresorhus/eslint-node-test and fix the lint errors.

Some of changes:
- two tests were duplicated
- LOTS of (sensible) errors to check error when using `assert.throws`, as otherwise even a syntax error would pass
- have just a single line of code in `assert.throws()`
- consistently use `assert.ok(expression)` instead of `assert(expression)`
- tidy some test names

### DIFF
--- a/examples/help-centered.mjs
+++ b/examples/help-centered.mjs
@@ -25,8 +25,6 @@ class MyCommand extends Command {
 
 const program = new MyCommand();
 
-program.configureHelp({ MyCommand });
-
 program
   .option('-s', 'short flag')
   .option('-f, --flag', 'short and long flag')
--- a/tests/argument.variadic.test.js
+++ b/tests/argument.variadic.test.js
@@ -4,7 +4,7 @@ import assert from 'node:assert/strict';
 
 // Testing variadic arguments. Testing all the action arguments, but could test just variadicArg.
 
-describe('Command variadic argument using .argument()', (t) => {
+describe('Command variadic argument using .argument()', () => {
   test('when no extra arguments specified for program then variadic arg is empty array', (t) => {
     const actionMock = t.mock.fn();
     const program = new commander.Command();
@@ -97,19 +97,6 @@ describe('Command variadic argument using .argument()', (t) => {
     let passedArg;
     program
       .addArgument(new commander.Argument('<value...>').choices(['one', 'two']))
-      .action((value) => {
-        passedArg = value;
-      });
-
-    program.parse(['one', 'two'], { from: 'user' });
-    assert.deepEqual(passedArg, ['one', 'two']);
-  });
-
-  test('when variadic has default array then specified value is used instead of default (not appended)', () => {
-    const program = new commander.Command();
-    let passedArg;
-    program
-      .addArgument(new commander.Argument('[value...]').default(['DEFAULT']))
       .action((value) => {
         passedArg = value;
       });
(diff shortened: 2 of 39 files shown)


examples/help-centered.mjs
```diff
@@ -25,8 +25,6 @@ class MyCommand extends Command {
 
 const program = new MyCommand();
 
-program.configureHelp({ MyCommand });
```
shadowspawn, line 28: This is a bogus call I spotted myself, not one reported by the linter.

shadowspawn:
I tried turning on all the rules to see what else looked good.

[require-top-level-describe](https://github.com/sindresorhus/eslint-node-test/blob/v0.3.0/docs/rules/require-top-level-describe.md) enforces the pattern I followed with a file-level describe for any files. It picked up a few cases where I had added tests after the describe block. For single test files, I manually disabled the check as nothing extra to add for context.

[require-hook](https://github.com/sindresorhus/eslint-node-test/blob/v0.3.0/docs/rules/require-hook.md) picked up the pattern I used for wrapping fixture loops inside a describe means the code is running at file load time rather than test time. It will be a noisy refactor, so I'll turn that on separately.

tests/argument.variadic.test.js
```diff
@@ -117,17 +117,4 @@ describe('Command variadic argument using .argument()', (t) => {
     program.parse(['one', 'two'], { from: 'user' });
     assert.deepEqual(passedArg, ['one', 'two']);
   });
-
```
shadowspawn, line 120: Duplicate of above test

tests/command.exitOverride.test.js
```diff
@@ -1,457 +1,435 @@
...
+import { createTestCommand } from './testHelpers.js'; // createTestCommand sets exitOverride()
 import { test, describe } from 'node:test';
 import assert from 'node:assert/strict';
 
 // Test details of the exitOverride errors.
-// The important checks are the exitCode and code which are intended to be stable for
-// semver minor versions. For now, also testing the error.message and that output occurred
-// to detect accidental changes in behaviour.
-
-function expectCommanderError(err, exitCode, code, message) {
```
shadowspawn, line 11: Removed `expectCommanderError` and put tests inline for clarity within tests.

tests/command.exitOverride.test.js
```diff
@@ -1,457 +1,435 @@
...
+        assert.equal(err.exitCode, 1);
+        assert.equal(err.code, 'commander.unknownOption');
+        assert.equal(err.message, "error: unknown option '-m'");
+        return true;
+      },
     );
   });
 
-  test('when specify unknown command then throw CommanderError', (t) => {
-    const stderrSpy = t.mock.method(process.stderr, 'write', () => {});
```
shadowspawn, line 43: Removed mocks to suppress/check output and used `createTestCommand()` for clean output instead

tests/options.variadic.test.js
```diff
@@ -180,12 +185,4 @@ describe('variadic options', () => {
 
     assert.deepEqual(program.opts().comma, ['CCC']);
   });
-
-  test('when option has default array then specified value is used instead of default (not appended)', () => {
```
shadowspawn, line 184: Duplicate of above test

shadowspawn:
Finished work, ready for review. No hurry. 

(I had intended to open as draft, but missed that sorry. Did at least put WIP in title initially!)

abetomo (approved):
👍

PR #2566, opened 2026-07-30
Group codeql updates as inter-dependent

shadowspawn:
## Problem

`github/codeql-action/init` and `github/codeql-action/analyze` need to be in sync, and dependabot PRs fail their checks when updated spearately:

- https://github.com/tj/commander.js/pull/2563
- https://github.com/tj/commander.js/pull/2564

## Solution

Add configuration to group the codeql updates together in a single PR.


abetomo (approved):
👍

PR #2573, opened 2026-08-01
fix: preserve -- for executable and in-process subcommands

### LINKED ISSUE #2530: Executable subcommand does not preserve end-of-options delimiter before option-like operands

aalhadxx:
## Problem

When using executable subcommands (or nested in-process subcommands), the `--` option terminator is consumed by the parent command's option parsing and never reaches the child subcommand.

This breaks POSIX-compliant argument handling where option-like operands after `--` should be treated as arguments.

### Reproduction

```js
// parent.js
new Command()
  .name('parent')
  .command('child', 'run child', { executableFile: 'child.js' })
  .parse();

// child.js
new Command()
  .name('parent child')
  .argument('[input]')
  .parse();
```

```bash
$ node parent.js child -- --not-an-option
# Expected: { input: '--not-an-option' }
# Actual: error: unknown option '--not-an-option'
```

## Solution

Track the position of `--` in `parseOptions()` via a new `optionTerminatorIndex` field in the return value. When dispatching to a subcommand, restore `--` at the correct position in the arguments:

- **Executable subcommands**: splice `--` back into the args passed to the child process
- **In-process subcommands**: pass `--` as part of the `unknown` args so the subcommand re-parses it correctly

This approach is lighter-weight than the Symbol-based fix proposed in #2531 and preserves backward compatibility for leaf commands.

## Changes

- `lib/command.js`: track `optionTerminatorIndex` in `parseOptions`, restore `--` in `_dispatchSubcommand`
- `tests/args.literal.test.js`: add tests for in-process and nested subcommands with `--`
- `tests/command.executableSubcommand.lookup.test.js`: add end-to-end tests for executable subcommands with `--`
- `tests/command.parseOptions.test.js`: update assertions to include `optionTerminatorIndex`

Fixes #2530

### DIFF
--- a/lib/command.js
+++ b/lib/command.js
@@ -1361,7 +1361,12 @@ Expecting one of '${allowedValues.join("', '")}'`);
    * @private
    */
 
-  _dispatchSubcommand(commandName, operands, unknown) {
+  _dispatchSubcommand(
+    commandName,
+    operands,
+    unknown,
+    optionTerminatorIndex,
+  ) {
     const subCommand = this._findCommand(commandName);
     if (!subCommand) this.help({ error: true });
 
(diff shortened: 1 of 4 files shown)


shadowspawn:

aalhadxx:
yeah I used claude to help explore the code and draft this, but I ran the tests and checked it works locally. saw #2531 got closed as too heavy so tried a lighter approach with just the index. let me know if anything needs changing

shadowspawn:
Still adding clutter to the parsing process and injecting argument back into parameters.

PR #2574, opened 2026-08-01
Use subtest not describe level loop

shadowspawn:
## Problem

Running fixture loops with embedded tests at top level of `describe` is running the tests at file load time instead of a test time. This is not causing any known problems at the moment, but could cause confusion in the future.

Issue picked up by `node-test/require-hook`:

- https://github.com/sindresorhus/eslint-node-test/blob/main/docs/rules/require-hook.md

## Solution

Wrap fixture loops in a test, and by convention use await for the nested test.

Couple of other tidies:
- use built-in support for conditional test/describe
- remove `node-test/require-top-level-describe` as we often use a top level `test` rather than `describe`, and don't want extra wrapper or to suppress the warning everywhere that pattern used.

The changes are a bit noisy, but a fairly pure refactor. (Turning off whitespace in the diff may help make this clearer.)


abetomo (approved):
👍

PR #2576, opened 2026-08-04
test: run .ts subcommand fixture as CommonJS so it works on all supported node versions

Jaybhade:
## Problem

`npm test` fails on node 22.12 – 22.17, which `package.json` `engines` (`>=22.12.0`) and the Commander 15 release notes both declare as supported:

```console
$ node --version
v22.12.0
$ npm test
not ok 11 - when subcommand suffix is .ts then lookup succeeds
  error: |-
    Command failed: node tests/fixtures-extensions/pm.js try-ts
    TypeError [ERR_UNKNOWN_FILE_EXTENSION]: Unknown file extension ".ts" for tests/fixtures-extensions/pm-try-ts.ts
# fail 1
```

The cause is a module-format detail rather than anything about the lookup itself. `tests/fixtures-extensions/` has no `package.json`, so since the ESM-only migration (#2464) it inherits the root `"type": "module"`. That sends `pm-try-ts.ts` to the ESM loader, which can only run it via TypeScript type stripping — and type stripping is enabled by default only from **node 22.18.0** (`process.features.typescript` is `false` on 22.12). Before the ESM migration the fixture went through the CommonJS loader, which runs an unknown extension as plain JavaScript, so no TypeScript support was needed.

This is the same fragility that had the test skipped in #2464. #2552 un-skipped it on the reasoning that every matrix entry runs the fixture via default type stripping, which is true — but the matrix entry `22.x` always resolves to the newest 22, so CI never exercises the lower end of the supported range that `engines` promises.

## Solution

Mark the fixture directory as CommonJS with a two-line `tests/fixtures-extensions/package.json`, and rename the parent fixture to `pm.mjs` so it stays ESM. The CommonJS loader runs `pm-try-ts.ts` as plain JavaScript, so the test no longer depends on TypeScript support at all and passes on every supported node version — removing the root cause instead of tracking node EOL dates.

The test's scope is unchanged: it still only checks that Commander *finds* `pm-try-ts.ts`, which is all the comment claims it sets out to do ("we support looking for ts files for ts-node in particular, but don't need to test ts-node itself"). Loading a child as ESM stays covered by the neighbouring `.mjs` test, and the parent fixture is still ESM.

Verified on node 22.12.0 (no type stripping available):

```console
$ node --test tests/command.executableSubcommand.lookup.test.js
# tests 14
# pass 14
# fail 0

$ npm test
# tests 1420
# pass 1420
# fail 0

$ npm run check   # type + lint + format, all clean
```

Also re-ran the lookup tests with `NODE_OPTIONS=--experimental-strip-types` to confirm the change is fine where type stripping *is* available (as on the current matrix): 14/14 pass.

Happy to swap this for a version-conditional `skip` on `process.features.typescript` if you would rather keep the fixture directory ESM — that is smaller, but it gives up the test on the lower half of the supported 22 range.

## ChangeLog

Test-only change, no CHANGELOG entry needed.


shadowspawn:
This PR is probably AI slop.

This particular test is not doing a genuine end-to-end tests of ts support, just that a `.ts` suffix is being looked for. Good enough.

PR #2581, opened 2026-08-10
Add `env` parse option to use instead of `process.env`

### LINKED ISSUE #2549: Allow passing custom env instead of process.env

justinjd00:
## Problem

Resolves #2549

There is currently no way to supply the environment used for options declared with `.env()` — Commander always reads `process.env`. When testing a CLI, that means mutating the real `process.env` (and remembering to clean up), or reaching for mocking libraries. The issue asks for an `env` override in `ParseOptions`:

```js
program.addOption(new Option('-p, --port <number>').env('PORT'));
program.parse([], { from: 'user', env: { PORT: '80' } });
// before this PR: program.opts() is {} — the env parse option is silently ignored
```

## Solution

Add an optional `env` property to the parse options for `.parse()` and `.parseAsync()`. When supplied, it is used instead of `process.env` when looking up options declared with `.env()`:

```js
program.parse([], { from: 'user', env: { PORT: '80' } });
program.opts(); // { port: '80' }, source 'env'
```

Details:

- The override is stored on the command in `_prepareUserArgs` and resolved in `_parseOptionsEnv` via a new private helper `_getParseEnv()`, which walks `_getCommandAndAncestors()` so action-handler subcommands see the `env` passed to the program's `parse()` call. Falls back to `process.env` when no override was supplied (no behaviour change for existing code).
- Passing `env` fully replaces `process.env` for option lookup (e.g. `env: {}` means no environment options are set), which makes tests deterministic.
- The override is per-parse: a subsequent `parse()` call without `env` reads `process.env` again.
- Out of scope, unchanged: color detection env vars (`NO_COLOR`/`FORCE_COLOR`/`CLICOLOR_FORCE`) and the environment of spawned stand-alone executable subcommands (the child process still inherits the real environment).
- Included: JSDoc, TypeScript typings (`ParseOptions.env?: Record<string, string | undefined>`, assignable from `process.env`), tsd tests, unit tests, and a short README note under `.parse() and .parseAsync()`.

Tested with `npm test` and `npm run check` (all passing). New tests in `tests/options.env.test.js` cover: custom env used for value and boolean options, `process.env` ignored when `env` is passed, cli still winning over env, subcommand inheritance, `parseAsync`, and reverting to `process.env` on a later `parse()` without the override.

## ChangeLog

- Added: `env` parse option for `.parse()` and `.parseAsync()` to specify environment variables to use instead of `process.env` for options with `.env()` ([#2549])

### DIFF
--- a/lib/command.js
+++ b/lib/command.js
@@ -57,6 +57,7 @@ export class Command extends EventEmitter {
     this._showHelpAfterError = false;
     this._showSuggestionAfterError = true;
     this._savedState = null; // used in save/restoreStateBeforeParse
+    this._parseEnv = undefined; // custom environment variables from parse option, see _getParseEnv()
 
     // see configureOutput() for docs
     this._outputConfiguration = {
@@ -994,6 +995,7 @@ Expecting one of '${allowedValues.join("', '")}'`);
       throw new Error('first parameter to parse must be array or undefined');
     }
     parseOptions = parseOptions || {};
+    this._parseEnv = parseOptions.env;
 
     // auto-detect argument conventions if nothing supplied
     if (argv === undefined && parseOptions.from === undefined) {
@@ -1075,6 +1077,7 @@ Expecting one of '${allowedValues.join("', '")}'`);
    * @param {string[]} [argv] - optional, defaults to process.argv
    * @param {object} [parseOptions] - optionally specify style of options with from: node/user/electron
    * @param {string} [parseOptions.from] - where the args are from: 'node', 'user', 'electron'
+   * @param {Object<string, string>} [parseOptions.env] - environment variables to use for options with env(), instead of process.env
    * @return {Command} `this` command for chaining
    */
 
@@ -1104,6 +1107,7 @@ Expecting one of '${allowedValues.join("', '")}'`);
    * @param {string[]} [argv]
    * @param {object} [parseOptions]
    * @param {string} parseOptions.from - where the args are from: 'node', 'user', 'electron'
+   * @param {Object<string, string>} [parseOptions.env] - environment variables to use for options with env(), instead of process.env
    * @return {Promise}
    */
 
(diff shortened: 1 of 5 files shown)


shadowspawn:

justinjd00:
> To be clear: is this Pull Request entirely AI/LLM generated?

Yes, to be upfront about it: the implementation was AI-generated (Claude), working under my direction. It picked the issue, wrote the change and the tests, and ran npm test and npm run check locally (all passing). I reviewed the summary and test results before submitting, but I want to be honest that the code itself is AI-written rather than hand-written by me.

I completely understand if the project prefers not to take AI-generated contributions, feel free to close this, no hard feelings. If you're open to keeping it, I'm happy to address any review feedback.

lib/command.js
```diff
@@ -2000,6 +2005,20 @@ Expecting one of '${allowedValues.join("', '")}'`);
...
   }
 
+  /**
+   * Get the environment variables to use for options with env, from the `env`
+   * parse option on this command or an ancestor, falling back to process.env.
+   *
+   * @return {Object}
+   * @private
+   */
+  _getParseEnv() {
```
shadowspawn, line 2015: This is a messy part, finding the environment variables relating to the current `.parse()` call. This will mostly work fine. In practice, normally will be on the program. 

However, technically possible to call `.parse()` directly on a subcommand too. So a call to parse on one level, then a call to parse on another level, may see stale state. Not a normal situation.

Would be more robust if cleared env in `.restoreStateBeforeParse()` so does not rely on being overwritten by subsequent call on _same_ command as the previous call.

shadowspawn:
Keeping the issue open to see if there is more interest in having this functionality available, and more use cases to check that a possible solution (like this PR) will actually meet the needs.

Closing this PR in the meantime. Not bad, but don't want to hold open a PR. Can reopen it if the approach meets future understanding of needs. 

Thanks for details on how PR was produced.

PR #2585, opened 2026-08-16
feat: add apply() helper to batch-configure commands

HosnainRafi:
## Summary

This PR adds an `apply()` method to the `Command` class, allowing users to batch-configure multiple commands with shared settings without breaking the call chain.

## Motivation

When building CLIs with many commands that share common options or hooks, it is tedious to repeat the same configuration across commands. This PR introduces `apply()` which accepts a function that receives the command instance, enabling reusable configurators:

```js
function addVerboseOption(cmd) {
  return cmd.option('-v, --verbose', 'verbose output');
}

program
  .apply(addVerboseOption)
  .apply(addOtherOptions)
  .action(() => { ... });
```

## Changes

- Added `apply(fn)` method to `Command` class in `lib/command.js`
- Added TypeScript type definition in `typings/index.d.ts`
- Added comprehensive tests in `tests/command.apply.test.js`

## Testing

- All 1378 existing tests pass
- New tests added for `apply()`: chaining, function invocation, configuration, multiple calls, and method chaining
- ESLint and Prettier checks pass

### DIFF
--- a/lib/command.js
+++ b/lib/command.js
@@ -500,6 +500,18 @@ Expecting one of '${allowedValues.join("', '")}'`);
   }
 
   /**
+   * Apply a function to this command, useful for batch configuration
+   * of multiple commands with shared settings.
+   *
+   * @param {Function} fn - function receiving this command, return value is discarded
+   * @return {Command} `this` command for chaining
+   */
+  apply(fn) {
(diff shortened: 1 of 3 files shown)


shadowspawn:
AI slop.

PR #2589, opened 2026-08-17
docs: use node --inspect in the Chinese debugging notes

JiuYue0820:
## Problem

The Chinese README documents the Node inspector flag as `node -inspect`. That is not a valid Node.js option. The English README already uses the correct `node --inspect`.

## Solution

Change `node -inspect` to `node --inspect` in `Readme_zh-CN.md` only.

## Verification

```
node -inspect -e "0"
# bad option: -inspect

node --inspect -e "0"
# Debugger listening on ws://127.0.0.1:...
```

### DIFF
--- a/Readme_zh-CN.md
+++ b/Readme_zh-CN.md
@@ -981,7 +981,7 @@ const program = createCommand();
 
 一个可执行的子命令会作为单独的子进程执行。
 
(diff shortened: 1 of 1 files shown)


shadowspawn:
This is AI slop, but a small tidy change.

shadowspawn (approved):
LGTM

PR #2594, opened 2026-08-30
feat: allow passing custom env in parse options

### LINKED ISSUE #2549: Allow passing custom env instead of process.env

vaibhavmashal:
## Problem

Resolves #2549

Currently, options configured with \.env()\ always read values directly from \process.env\. In testing scenarios or when wrapping Commander in custom application contexts, modifying \process.env\ globally requires manual setup and cleanup. Allowing a custom \env\ dictionary to be passed via \parseOptions\ provides a clean and predictable way to supply environment variables.

## Solution

- Added optional \env\ property to \ParseOptions\ for \.parse()\ and \.parseAsync()\.
- When \env\ is provided, options declared with \.env()\ read from the supplied object instead of \process.env\.
- Subcommands inherit \env\ configured on parent commands.
- State is properly saved and restored across multiple \.parse()\ calls.
- Updated TypeScript typings (\ParseOptions.env?: Record<string, string | undefined>\), \	sd\ tests, JSDoc documentation, and \Readme.md\ / \Readme_zh-CN.md\.
- Added unit tests in \	ests/options.env.test.js\ covering custom env lookup, priority over \process.env\, CLI precedence, subcommand inheritance, \parseAsync\, and state restoration.

## Testing

- \
pm test\ passed (100% tests pass).
- \
pm run check\ passed (\check:type\, \check:lint\, \check:format\).

## Suggested Changelog

- Added: \env\ property in \ParseOptions\ to specify custom environment variables for \.parse()\ and \.parseAsync()\ (#2549)

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -977,6 +977,12 @@ program.parse(process.argv); // assume argv[0] is app and argv[1] is script
 program.parse(['--port', '80'], { from: 'user' }); // just user supplied arguments, nothing special about argv[0]
 ```
 
+You can also pass `env` in the parse options to use custom environment variables instead of `process.env` (useful for testing):
+
+```js
+program.parse([], { from: 'user', env: { PORT: '80' } });
+```
+
 Use `.parseAsync()` instead of `.parse()` if any of your action handlers are async.
 
 ### Parsing Configuration
--- a/Readme_zh-CN.md
+++ b/Readme_zh-CN.md
@@ -918,6 +918,7 @@ program.on('option:verbose', function () {
 program.parse(process.argv); // 指明，按 node 约定
 program.parse(); // 默认，自动识别 electron
 program.parse(['-f', 'filename'], { from: 'user' });
+program.parse([], { from: 'user', env: { PORT: '80' } }); // 使用自定义环境变量替代 process.env
 ```
 
 ### 解析配置
--- a/lib/command.js
+++ b/lib/command.js
@@ -57,6 +57,7 @@ export class Command extends EventEmitter {
     this._showHelpAfterError = false;
     this._showSuggestionAfterError = true;
     this._savedState = null; // used in save/restoreStateBeforeParse
+    this._parseEnv = undefined;
 
     // see configureOutput() for docs
     this._outputConfiguration = {
@@ -986,6 +987,10 @@ Expecting one of '${allowedValues.join("', '")}'`);
    * Get user arguments from implied or explicit arguments.
    * Side-effects: set _scriptPath if args included script. Used for default program name, and subcommand searches.
    *
+   * @param {string[]} [argv]
+   * @param {object} [parseOptions]
+   * @param {string} [parseOptions.from] - where the args are from: 'node', 'user', 'electron'
+   * @param {Record<string, string | undefined>} [parseOptions.env] - custom environment variables to use instead of process.env
    * @private
    */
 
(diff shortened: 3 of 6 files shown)


shadowspawn:

official-burak (changes requested):
`from` is also optional on `ParseOptions` now, but `_prepareUserArgs` still defaults a missing `from` to `argv.slice(2)`. The new type test `program.parse(['--option'], { env: process.env })` type-checks, then treats `--option` as the script path and drops the user option. README already passes `from: 'user'`; the types and that test should match, otherwise a user argv array quietly loses its first two entries.

shadowspawn:
Oh hurrah, a probably AI generated review on a probably AI generated PR. A new low.

official-burak:
@shadowspawn I review from this account as myself. The `env: null` TypeError is reproducible on the branch. Happy to paste a tiny repro if that helps.

shadowspawn:
@official-burak Checking `env` for `null` seems very low value. Not a likely error for the author to make?

If you look around the rest of the interface, most parameters are not checked. There are some which are checked for type because there is an easy mistake to make, either reported in issues or encountered during maintenance.

official-burak:
@shadowspawn You are right, that null check is low value. Thanks for spelling out how the rest of the API treats this. I am dropping it.

vaibhavmashal:
Hi @shadowspawn,

No, I designed and implemented this change directly to address the feature request for custom environment variable support in \parse()\ and \parseAsync()\ options (similar to #2549).

Here is a breakdown of the solution and architecture:

1. **\ParseOptions\ Extension**: Added optional \env?: Record<string, string | undefined>\ to \ParseOptions\ so users can supply a mock or scoped dictionary of environment variables without mutating global \process.env\.
2. **Hierarchical Inheritance**: Implemented \_getParseEnv()\ to traverse \_getCommandAndAncestors()\. This ensures nested subcommands automatically inherit the custom environment passed to the root \program.parse()\.
3. **State Isolation**: \_parseEnv\ is reset in \_init()\ alongside other parse-specific state so successive parses on the same Command instance remain independent.
4. **Option Resolution**: \_parseOptionsEnv()\ resolves values against the active environment returned by \_getParseEnv()\ instead of hardcoding \process.env\.
5. **Type Test Cleanup**: Updated \	ypings/index.test-d.ts\ (commit \7e20f4f\) to pass standard full node argv arrays when \rom\ is omitted.

All 1,412 unit tests and type checks pass cleanly with \
pm test\ and \
pm run check:type:ts\.

vaibhavmashal:
Thanks for the reviews and feedback @official-burak @shadowspawn!

Regarding the \rom\ option and argv handling (@official-burak):
You're completely right. Since omitting \rom\ retains Commander's default behavior (\rom: 'node'\), passing a raw user argument array without node/script prefixes resulted in the first two arguments being dropped. We updated the typings test in commit \7e20f4f\ to use \['node', 'script.js', '--option']\ when \rom\ is omitted to accurately reflect the standard argv structure, while keeping \{ from: 'user' }\ for direct user option arrays.

Regarding the background and technical implementation (@shadowspawn):
The motivation behind this change is to enable clean environment variable isolation for commands without having to mutate global \process.env\. In testing suites, workers, and embedded CLI contexts where commands may run concurrently or require simulated environments, mutating \process.env\ introduces race conditions and test pollution. By accepting an optional \env\ object via \parse\ / \parseAsync\ options and plumbing it through option resolution (falling back to \process.env\ when unset), callers can pass scoped configurations deterministically while leaving the existing command behavior and defaults untouched.

All unit tests and type checks pass cleanly. Please let me know if any further adjustments would be helpful!

PR #2604, opened 2026-09-05
fix: do not use electron argv parsing with ELECTRON_RUN_AS_NODE

### LINKED ISSUE #2603: `parse()` gets `argv` wrong with `ELECTRON_RUN_AS_NODE`
## Summary

In an Electron project, I ran into an argument-parsing issue when using commanderjs in a child process with `ELECTRON_RUN_AS_NODE=1`.

The process still has `process.versions.electron`, so `program.parse()` picks Electron argument handling. But it is effectively being run as `Electron script.js [...arguments]`, just like `node script.js [...arguments]`: `script.js` is the script being launched, not a user argument. Since `process.defaultApp` is `undefined`, Commander parses from the script path and includes it as a user argument.

This happens when an Electron main process uses `child_process.spawn()` to run a JavaScript child with `ELECTRON_RUN_AS_NODE=1`.

## Reproduction

`main.js`, started as an Electron main process:

```js
const { app } = require('electron');
const { spawn } = require('node:child_process');
const path = require('node:path');

app.whenReady().then(() => {
  const child = spawn(process.execPath, [path.join(__dirname, 'child.js'), '--verbose'], {
    env: {
      ...process.env,
      ELECTRON_RUN_AS_NODE: '1',
    },
    stdio: 'inherit',
  });

  child.once('close', () => app.quit());
});
```

`child.js`:

```js

console.log({
  argv: process.argv,
  electron: process.versions.electron,
  defaultApp: process.defaultApp,
  type: process.type,
  runAsNode: process.env.ELECTRON_RUN_AS_NODE,
});

const program = new Command().argument('[input]').option('--verbose');

program.parse();

console.log({ args: program.args, options: program.opts() });
```

## Actual behavior

Here is the output from the child:

```js
{
  argv: [
    '/path/to/Electron.app/Contents/MacOS/Electron',
    '/path/to/child.js',
    '--verbose'
  ],
  electron: '43.4.1',
  defaultApp: undefined,
  type: undefined,
  runAsNode: '1'
}
{ args: ['/path/to/child.js'], options: { verbose: true } }
```

Because `process.versions.electron` is set, `program.parse()` selects the Electron convention. Since `process.defaultApp` is falsey, it uses `argv.slice(1)`, so it parses the script path as the optional `input` argument.

For comparison, if I remove `ELECTRON_RUN_AS_NODE`, the same executable starts as an Electron main process and Commander parses the arguments correctly:

```js
{
  argv: [
    '/path/to/Electron.app/Contents/MacOS/Electron',
    '/path/to/child.js',
    '--verbose'
  ],
  electron: '43.4.1',
  defaultApp: true,
  type: 'browser',
  runAsNode: undefined
}
{ args: [], options: { verbose: true } }
```

## Expected behavior

When `ELECTRON_RUN_AS_NODE=1` is set, I would expect `program.parse()` to use the Node argument convention and leave the script path out of user arguments.

## Proposed fix

One possible fix is to skip Electron auto-detection when the process is running in Node mode:

```js
if (process.versions?.electron && process.env.ELECTRON_RUN_AS_NODE === undefined) {
  parseOptions.from = 'electron';
}
```

When `ELECTRON_RUN_AS_NODE` is set, `parseOptions.from` stays unset. The existing default path then uses Node conventions and `argv.slice(2)`.

## Environment

- Commander: 15.0.0
- Electron: 43.4.1
- OS: macOS

nirmal-shaji:
## Summary

`Command.parse()` mis-parses `argv` when `ELECTRON_RUN_AS_NODE` is set (#2603).

When an Electron main process spawns a JS child with `ELECTRON_RUN_AS_NODE=1`, the child runs as plain Node — the argv layout is `Electron script.js [...args]`, exactly like `node script.js [...args]`, so `script.js` is the script path, not a user argument. But `process.versions.electron` is still truthy, so auto-detection picked `from: 'electron'`; and because `process.defaultApp` is `undefined`, the script path was kept as a user argument (`argv.slice(1)`).

## Fix

Only auto-select the `electron` convention when the process is actually running as an Electron app, i.e. when `ELECTRON_RUN_AS_NODE` is **not** set:

```js
if (process.versions?.electron && !process.env.ELECTRON_RUN_AS_NODE) {
  parseOptions.from = 'electron';
}
```

This is intentionally minimal and only touches auto-detection; passing `{ from: 'electron' }` explicitly is unchanged.

## On the fragility

I saw @shadowspawn's note that this heuristic is inherently fragile — we can't know at parse time whether the executable was launched with the env var or it was set afterwards. Agreed. This guard doesn't try to solve that general case; it aligns auto-detection with the documented meaning of `ELECTRON_RUN_AS_NODE` ("run as a normal Node.js process"), which covers the common case in the report (env var set as part of the `spawn`). It's a small, spec-consistent improvement over unconditionally trusting `process.versions.electron`, and explicit `from` remains available for anything more exotic.

## Test

Added a case to `tests/command.parse.test.js` alongside the existing electron auto-detection test: with `process.versions.electron` and `ELECTRON_RUN_AS_NODE` set and `argv = 'node script.js user'`, the script path is excluded and `program.args` is `['user']`. Confirmed it fails before the fix and passes after.

`npm test` (node --test) green — 1373 passing.

Fixes #2603

### DIFF
--- a/lib/command.js
+++ b/lib/command.js
@@ -997,7 +997,10 @@ Expecting one of '${allowedValues.join("', '")}'`);
 
     // auto-detect argument conventions if nothing supplied
     if (argv === undefined && parseOptions.from === undefined) {
-      if (process.versions?.electron) {
+      // ELECTRON_RUN_AS_NODE makes Electron behave as plain Node (script.js is
+      // argv[1]), even though process.versions.electron is still truthy. Only
+      // use Electron argument handling when it is running as an Electron app.
+      if (process.versions?.electron && !process.env.ELECTRON_RUN_AS_NODE) {
         parseOptions.from = 'electron';
       }
       // check node specific options for scenarios where user CLI args follow executable without scriptname
--- a/tests/command.parse.test.js
+++ b/tests/command.parse.test.js
@@ -33,6 +33,25 @@ describe('Command.parse()', () => {
       assert.deepEqual(program.args, ['user']);
     });
 
+    test('when no args and electron properties but ELECTRON_RUN_AS_NODE then use process.argv and app/script/args', () => {
+      const program = new commander.Command();
+      program.argument('[args...]');
+      const holdArgv = process.argv;
+      const holdRunAsNode = process.env.ELECTRON_RUN_AS_NODE;
+      process.versions.electron = '1.2.3';
+      process.env.ELECTRON_RUN_AS_NODE = '1';
+      process.argv = 'node script.js user'.split(' ');
+      program.parse();
+      delete process.versions.electron;
+      if (holdRunAsNode === undefined) {
+        delete process.env.ELECTRON_RUN_AS_NODE;
+      } else {
+        process.env.ELECTRON_RUN_AS_NODE = holdRunAsNode;
+      }
+      process.argv = holdArgv;
+      assert.deepEqual(program.args, ['user']);
+    });
+
     test('when args then app/script/args', () => {
       const program = new commander.Command();
       program.argument('[args...]');


nrps9909 (approved):
Reviewed `3ac5c5d01fa91090218b8c4ed10663616cd03825`.

Confirmed the reported script-path leak before the change and checked nine parsing cases on this head: unset/empty `ELECTRON_RUN_AS_NODE`, non-empty `1`/`0`/`false`, explicit `from: 'electron'`/`'node'`/`'user'`, and a supplied argv array. All matched the expected arguments. Using truthiness here also matches Electron's non-empty environment-value check in its [macOS entry point](https://github.com/electron/electron/blob/main/shell/app/electron_main_mac.cc); restricting the value to `'1'` would miss other enabled values.

`npm test`: 1,373 passed, one skipped. `npm run check`: types, lint and formatting passed (Node 26.8.1, macOS). I cleared inherited color variables for the full test run: the existing configureOutput tests otherwise fail with my `NO_COLOR=1`; that also reproduces independently on develop and is outside this change.

No blocking findings in this auto-detection change. The additional parsing probe mocks Electron process metadata, consistent with the existing test file; it is not a real Electron process launch.

shadowspawn:

shadowspawn:
This account opened multiple PR on the same day, the PR shows heavy use of AI, and does not follow the contributing guidelines.

And in addition, giving the person who reported the problem first choice for fixing the problem!

PR #2606, opened 2026-09-05
broaden the type to include number for option defaultValue

### LINKED ISSUE #2605: type declarations for the option function doesn't have a number type
The following example is from the documentation:
```ts
import { program, } from 'commander'

program
  .option('-p, --port <number>', 'port number', 80)

```

using TypeScript it shows this error:
```
 No overload matches this call.
 The last overload gave the following error.
 Argument of type 'number' is not assignable to parameter of type 'RegExp'.
 index.d.ts(660, 3): The last overload is declared here.
```

The declaration code for `option`  is:

```ts
 option(
    flags: string,
    description?: string,
    defaultValue?: string | boolean | string[],
  ): this;
  option<T>(
    flags: string,
    description: string,
    parseArg: (value: string, previous: T) => T,
    defaultValue?: T,
  ): this;
  /** @deprecated since v7, instead use choices or a custom function */
  option(
    flags: string,
    description: string,
    regexp: RegExp,
    defaultValue?: string | boolean | string[],
  ): this;
```
The first and the third declarations will show a type error if an integer is being passed as a default value because they are missing a number type.

I would like to work on this issue, and I will basically add a number type for the first and third declarations.

I'm not quite sure if not having a number type here is something intended and intentional, so please let me know why if that is the case.

SuperMo0:
closes #2605
## Problem
Type declarations for the option function don't include a number type

## Solution
adding the type `number` to 2 `option` declarations inside the index. d.ts

### DIFF
--- a/typings/index.d.ts
+++ b/typings/index.d.ts
@@ -647,7 +647,7 @@ export class Command {
   option(
     flags: string,
     description?: string,
-    defaultValue?: string | boolean | string[],
+    defaultValue?: string | boolean | string[] | number,
   ): this;
   option<T>(
     flags: string,
@@ -660,7 +660,7 @@ export class Command {
     flags: string,
     description: string,
     regexp: RegExp,
-    defaultValue?: string | boolean | string[],
+    defaultValue?: string | boolean | string[] | number,
   ): this;
 
   /**


shadowspawn:
See https://github.com/tj/commander.js/issues/2605#issuecomment-5554993156

PR #2608, opened 2026-09-05
test: isolate color environment in configureOutput tests

nrps9909:
Running the tests from a shell with `NO_COLOR=1` makes the two configureOutput force-color cases fail, even though Commander correctly gives `NO_COLOR` precedence. With `FORCE_COLOR=0`, the CLICOLOR_FORCE case also inherits a conflicting setting. The tests currently save and replace only the variable they are testing.

Clear all three color variables before each subtest and restore their original values once in `after` when the group finishes, so inherited settings cannot change the assertions and a failed assertion still restores the environment. This changes only test setup and cleanup.

Validation on develop `97411c6`:

- Reproduced two failures with `NO_COLOR=1 node --test tests/command.configureOutput.test.js` before the change.
- The focused file passes with no color variables, each of `NO_COLOR=1`, `FORCE_COLOR=0`, `FORCE_COLOR=1`, `CLICOLOR_FORCE=1`, and all three conflicting settings together.
- `npm test`: 1,420 passing with the original `NO_COLOR=1` environment.
- `npm run check`: types, lint and formatting pass.
- The focused file also passes on the minimum supported Node 22.12.0.

Suggested changelog summary: isolate configureOutput tests from inherited color environment variables.

Implemented and validated with Codex.

### DIFF
--- a/tests/command.configureOutput.test.js
+++ b/tests/command.configureOutput.test.js
@@ -342,18 +342,30 @@ describe('Command.configureOutput()', () => {
       ['FORCE_COLOR', true],
       ['CLICOLOR_FORCE', true],
     ];
+    const originalEnv = testCases.map(([envvar]) => [
+      envvar,
+      process.env[envvar],
+    ]);
+    t.beforeEach(() => {
+      for (const [envvar] of originalEnv) {
+        delete process.env[envvar];
+      }
+    });
+    t.afterEach(() => {
+      for (const [envvar, value] of originalEnv) {
+        if (value === undefined) delete process.env[envvar];
+        else process.env[envvar] = value;
+      }
+    });
     for (const [envvar, expected] of testCases) {
       await t.test(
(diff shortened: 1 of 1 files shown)


SuperMo0:
Why use `afterEach` instead of `after`?  The `beforeEach` is going to delete the keys anyway, so there is no need to reset them to the original value after every iteration.

nrps9909:
`afterEach` was more cleanup than this serial group needs. I've changed it to `t.after()` in `f57c87b`: `beforeEach` still clears all three controls before each case, and the original environment is restored once when the group finishes, including if a subtest fails. This also matches the cleanup scope used in `useColor.test.js`.

Rechecked all six environment configurations, the full 1,420-test suite, types/lint/format, and the 24-test focused file on Node 22.12.0. All pass.

shadowspawn:
Thank-you for the AI disclosure in the Pull Request.

I do not want to make the change at this time, and instead I am treating it as a support example for #2549

PR #2609, opened 2026-09-07
fix: only use electron argv parsing for electron main process

### LINKED ISSUE #2603: `parse()` gets `argv` wrong with `ELECTRON_RUN_AS_NODE`

yifanwww:
## Problem

Fixes #2603 

## Solution

The fix is pretty simple: add `process.type === 'browser'` to only turn on `from=electron` workflow in Electron main process. See https://github.com/tj/commander.js/issues/2603#issuecomment-5561794924 for more details


lib/command.js
```diff
@@ -997,7 +997,11 @@ Expecting one of '${allowedValues.join("', '")}'`);
 
     // auto-detect argument conventions if nothing supplied
     if (argv === undefined && parseOptions.from === undefined) {
-      if (process.versions?.electron) {
+      if (
+        process.versions?.electron &&
```
shadowspawn: I would prefer to have the `process.type` test in the `electron` handling. So for example, user can potentially specify `from: "electron"` and get the full logic.

shadowspawn:
I wondered about testing `process.ELECTRON_RUN_AS_NODE` too since it is explicitly named, but decided it is really an Electron detail and does not cover all the cases. After look at all the cases, I prefer the approach you have taken with `process.type`.

shadowspawn:
Per comment: https://github.com/tj/commander.js/issues/2603#issuecomment-5642813500

I think probably `type: "renderer"` is also missing the script argument (but have not tried reproducing).

shadowspawn:
I am going to have a play with the new renderer tests in the testbed, but the code changes are looking good.

Thanks for the extensive background work that went into this PR @yifanwww

shadowspawn (approved):
New logic looks good. Thanks for working through all the ways Electron can launch and launch processes, much appreciated!

Since users may have worked around the old behaviour, I am being cautious and treating this as a semver major release.

(I intend to refactor the tests a bit after they land.)

shadowspawn:
@yifanwww would you please rebase your changes onto branch `release/16.x`?

(No worries if you can't. I am happy to do it.)

shadowspawn:
(I marked Pull Request as Draft until rebase, so we don't accidentally squash `develop` branch changes into the merge.)

yifanwww:
sure no problem, rebased to `release/16.x`

PR #2610, opened 2026-09-13
docs: clarify precedence in optsWithGlobals

Hugohong258:
## Problem

The README does not explain which value wins when local and global option names overlap in `.optsWithGlobals()`.

## Solution

Document that global values take precedence. This matches the existing implementation and its same-name option test.

## Validation

Validation: `npm test` (1,404 passed, 4 skipped) and `npm run check` pass. Also checked a three-level command hierarchy with positional options, where the root value overrides both descendants.


shadowspawn:
This looks like a drive-by AI contribution. The account created 9 other Pull Requests on the same day.

Overlapping options is not a good pattern anyway.

PR #2617, opened 2026-09-20
docs: use the declared flag in optional-value examples

agammann:
## Problem

The "Options in Depth" example declares `-o, --other`, but the subsequent command uses `collect -a -v -l` and shows `any servings: true`. Running that command with the documented program exits with `error: unknown option '-a'`.

## Solution

Use `-o` and the corresponding `other servings` output in both the English and Chinese guides. Keep the nearby combined-flag illustration consistent with the declared options by changing `-avl` to `-ovl`.

Validation: extracted and executed each guide's JavaScript block with its documented arguments and asserted the exact output and parsed options. Both corrected examples pass; the original command fails. `npm test` passes (1,404 tests passed, 4 skipped, plus TypeScript checks), and `npm run check` passes (JavaScript/TypeScript checks, ESLint and Prettier).

## ChangeLog

Correct the optional-value flag examples to use the option declared by the sample program.


shadowspawn:
This is a drive-by AI contribution, but is a simple correction.

shadowspawn (approved):
LGTM

